Issue #4, opened 2022-11-07
pubspec.lock should be committed


michalsrutek:
To avoid problems with different dependencies, everybody should use the same versions. That is what `pubspec.lock` is for. It should be definitely committed.

PR #5, opened 2022-11-07
Commit pubspec.lock

michalsrutek:
This would solve https://github.com/lichess-org/mobile/issues/4


veloce:
Thank you, I'll commit it later.

PR #6, opened 2022-11-07
Use single app icon size

michalsrutek:
Since Xcode 14, it is enough to provide a single app icon size and Xcode can resize it accordingly. ([source](https://useyourloaf.com/blog/xcode-14-single-size-app-icon/)).


veloce:
Interesting. Thank you.

PR #15, opened 2022-11-19
dart format .

ornicar:
my vim kept reformatting on save (using the standard dart lang server LSP) so I checked out industry standards https://dart.dev/tools/dart-format

And when typing `dart format .` it reformats the current code as such


ornicar:
needs project-level config instead

PR #19, opened 2022-11-29
Play with computer

veloce:


lib/src/features/game/domain/game.dart
```diff
@@ -0,0 +1,43 @@
+import 'package:freezed_annotation/freezed_annotation.dart';
+import 'package:dartchess/dartchess.dart';
+
+part 'game.freezed.dart';
+part 'game.g.dart';
+
+@Freezed(toJson: false)
+class Game with _$Game {
+  factory Game({
+    required String id,
```
ornicar, line 10: One thing I'm doing a lot in lila these days is getting reads of `String`s, `Int`s, and other meaningless types. For type safety purpose, I much prefer handling a `GameId` type than a `String`, because then I know, and the compiler knows, what it is.

Is there a nice way to achieve that in dart, ideally without incurring runtime costs?

lib/src/features/game/domain/game.dart
```diff
@@ -0,0 +1,43 @@
...
+    required bool rated,
+    required Speed speed,
+    required String initialFen,
```
ornicar, line 13: a `Fen` type would be nice. In scalachess I even have several types based on the fen content: 

```scala
// r3k2r/p3n1pp/2q2p2/4n1B1/5Q2/5P2/PP3P1P/R4RK1 b kq - 6 20
opaque type Fen = String

// r3k2r/p3n1pp/2q2p2/4n1B1/5Q2/5P2/PP3P1P/R4RK1
opaque type BoardFen = String
```


https://github.com/lichess-org/scalachess/blob/master/src/main/scala/format/Fen.scala

lib/src/features/game/domain/game.dart
ornicar: Maybe this is what I'm looking for: https://github.com/dart-lang/language/blob/master/working/1426-extension-types/feature-specification-views.md

veloce: I suppose we can do it for very critical types (game id) and wait for dart to implement views for others (wrapping with a class comes with a perf penalty)

lib/src/features/game/domain/game.dart
veloce: this new `opaque type` feature of scala3 looks very cool indeed.

veloce: here we could do `typedef Fen = String` but that's not as much useful...

lib/src/features/game/domain/board_state.dart
```diff
@@ -0,0 +1,69 @@
...
+  @override
+  bool operator ==(Object other) {
+    return other is BoardState &&
+        other.whiteTime == whiteTime &&
+        other.blackTime == blackTime &&
+        other.status == status &&
+        other.uciMoves == uciMoves &&
+        other.sanMoves == sanMoves &&
+        other.position == position;
+  }
```
ornicar, line 64: do we need this? why can't it be `other.id == id`

ornicar, line 64: do we need this when we already have `hashCode`?

veloce: Looks like we don't indeed. But it's the standard way of overriding `==` I've seen it everywhere.

But I won't leave that code, I'll refactor it and use instead https://pub.dev/packages/equatable (or freezed where it makes sense).
In dartchess too.

Issue #25, opened 2023-01-09
Leaderboard screen


veloce:
Corresponding lichess page: https://lichess.org/player

Lichess API reference: https://lichess.org/api#tag/Users/operation/player

nav-28:
The screen would be added to the bottom nav for now, right?

veloce:
No I don't think we'll add it to the bottom nav. I have to figure out where to put it indeed :)

veloce:
See: https://github.com/lichess-org/mobile/commit/4eef5df186d53c49c62eb4da8692f035df3d1ba7

Now there's a home page. Put the leaderboard here (one line per perf, as in lichess.org home), and a link to the detailed leaderboard.

PR #29, opened 2023-01-11
Leaderboard Screen 

nav-28:
Add the leaderboard screen.
Completes #25 

- [x] Add models for the leaderboard
- [x] Fetch the data from the API
- [x] Build the UI


lib/src/features/leaderboard/ui/leaderboard_screen.dart
```diff
@@ -0,0 +1,216 @@
...
+      _buildLeaderboard(
+          context, leaderboard.horde, LichessIcons.horde, 'HORDE'),
+      _buildLeaderboard(
+          context, leaderboard.antichess, LichessIcons.antichess, 'ANTICHESS'),
+      _buildLeaderboard(context, leaderboard.racingKings,
+          LichessIcons.racing_kings, 'RACING KINGS'),
+    ];
+  }
+
+  Widget _buildLeaderboard(BuildContext context, List<LightUser> userList,
```
veloce, line 107: Prefer writing a private Stateless widget class instead.
Cf: https://stackoverflow.com/questions/53234825/what-is-the-difference-between-functions-and-classes-to-create-reusable-widgets for more info.
For a one-time private method it's ok to use a function, but for a reusable widget that produces a widget tree, it's better to take the habit to write a class (even if here it is private this widget is reused several times).

lib/src/features/leaderboard/ui/leaderboard_screen.dart
```diff
@@ -0,0 +1,216 @@
...
+          title: Text(title),
+        ),
+        const Divider(),
+        ..._leaderboardList(context, userList),
+        const SizedBox(height: 12),
+      ],
+    );
+  }
+
+  List<Widget> _leaderboardList(
```
veloce, line 123: I'd prefer not to have that function and write the code directly inside `_buildLeaderboard`

lib/src/features/leaderboard/ui/leaderboard_screen.dart
```diff
@@ -0,0 +1,216 @@
...
+
+class ListCard extends StatelessWidget {
+  const ListCard({required this.user, this.prefIcon, super.key});
+  final LightUser user;
+  final IconData? prefIcon;
+
+  @override
+  Widget build(BuildContext context) {
+    return ListTile(
+      leading: _onlineOrPatron(user),
```
veloce, line 141: better not using a function here imo

lib/src/features/leaderboard/ui/leaderboard_screen.dart
```diff
@@ -0,0 +1,216 @@
...
+                style: const TextStyle(
+                    color: LichessColors.brag, fontWeight: FontWeight.bold)),
+            const SizedBox(width: 5)
+          ],
+          Flexible(
+            child: Text(user.username, overflow: TextOverflow.ellipsis),
+          ),
+        ],
+      ),
+      trailing: _ratingAndProgress(user),
```
veloce, line 155: same, no function

lib/src/features/leaderboard/ui/leaderboard_widget.dart
```diff
@@ -0,0 +1,86 @@
...
+});
+
+class LeaderboardWidget extends ConsumerStatefulWidget {
+  const LeaderboardWidget({super.key});
+
+  @override
+  ConsumerState<LeaderboardWidget> createState() => _LeaderboardWidgetState();
+}
+
+class _LeaderboardWidgetState extends ConsumerState<LeaderboardWidget> {
```
veloce, line 28: there is no field in your state, so you don't need the `StatefulWidget` here and can use a `ConsumerWidget` instead.

lib/src/features/leaderboard/ui/leaderboard_widget.dart
```diff
@@ -0,0 +1,86 @@
...
+import 'package:lichess_mobile/src/common/lichess_icons.dart';
+
+import '../model/leaderboard.dart';
+import './leaderboard_screen.dart';
+
+final leaderListProvider = FutureProvider.autoDispose((ref) async {
+  final leaderRepo = ref.watch(leaderboardRepositoryProvider);
+  final either = await leaderRepo.getLeaderboard().run();
+  return either.match((error) => throw error, (data) {
+    ref.keepAlive();
```
veloce, line 16: don't use `ref.keepAlive()` as it will cache forever the data (sorry for the example in my code where I used it wrongly too...);

The leaderboard data don't change often so we may later add a more clever mechanism to cache the data but with a timeout when it invalidates itself. 
Perhaps later we'll also implement a pull to refresh mechanism on the home page. But let's not add unnecessary complexity for now.

lib/src/features/leaderboard/ui/leaderboard_widget.dart
```diff
@@ -0,0 +1,86 @@
...
+        },
+        error: (error, stackTrace) {
+          debugPrint(
+              'SEVERE: [LeaderboardWidget] could not lead leaderboard data; $error\n $stackTrace');
+          return const Text('could not lead leaderboard');
+        },
+        loading: () => const CenterLoadingIndicator());
+  }
+
+  List<Widget> _buildList(BuildContext context, Leaderboard leaderboard) {
```
veloce, line 66: don't use function where you don't need them

Issue #23, opened 2023-01-09
Perf stats screen


veloce:
Create the performance screen for each time control and variant. Those screens are accessed from the corresponding "perf" card in the profile screen.

See on lichess.org: 
- https://lichess.org/@/Shant7777777/perf/bullet
- https://lichess.org/@/veloce/perf/blitz

Note the difference between the first and second: the second correspond to the logged in user profile, the working is different.

See lichess API reference to obtain the data: https://lichess.org/api#tag/Users/operation/apiUserPerf

Dante-010:
Is it ok if I (try to) create everything but the graph?
I would also like to apologize in advance for any misinterpretations, mistakes, or troubles I may cause; this is my first time contributing to open source.

Dante-010:
Don't know if it's necessary to post this, but here's how the UI is coming along:
 
![ui](https://user-images.githubusercontent.com/66489857/213090597-09a338a2-237f-4295-a319-f096696141fa.jpg)

I'm mostly done with business logic, and I should finish the UI by tomorrow. Then I'll add some tests to finish this off.

If there is anything you wish me to change, please tell me.

veloce:
I'd probably display information differently, but that's too hard to describe here precisely what I want.

Your UI is clear but some informations are more important that other and here everything is display at the same level.
I'd try to use the `PlatformCard` widget for more important infos.

Here I don't see "Highest rating" and "Winning streak". Are they accessible on lichess API?

Dante-010:
Ok, I'll try to implement PlatformCard and change some of the layout. As you can see, designing UI is not something I can do well, so I tried making it as simple as possible in order to make it easy to change in the future.

I have already implemented "Highest rating" and "Winning streak", I simply didn't show them in this screenshot. They both are accessible through the API, and can be found inside "stat" in the JSON response from this link: https://lichess.org/api#tag/Users/operation/apiUserPerf.

Thank you for your response! I'll be getting back in touch as soon as I can.

Dante-010:
Here's the new UI layout (still a WIP):
![new_ui](https://user-images.githubusercontent.com/66489857/213631888-6caaf764-ff8f-4282-9c07-c0ad71364af3.jpg)

Don't worry if you see missing/wrong values, it's just that my development environment's database is not properly populated. I've already tested that these stats work. 

If you want to, once I'm done with it, I can send you screenshots for all the stats.

If you think this is OK, I'll go on to write some tests and, if everything is fine, make the pull request.

PR #33, opened 2023-01-23
Perf stats screen

Dante-010:
Completes #23.


lib/src/features/user/ui/perf_stats_screen.dart
```diff
@@ -0,0 +1,457 @@
...
+      },
+      loading: () => const CenterLoadingIndicator(),
+    );
+  }
+
+  String _getPercentageString(num numerator, num denominator) {
+    return '${((numerator / denominator) * 100).round().toString()}%';
+  }
+
+  PlatformCard _customPlatformCard(
```
veloce, line 232: It is preferable to write a private `class` (StatelessWidget) instead of a function to create a reusable widget.
cf: https://stackoverflow.com/questions/53234825/what-is-the-difference-between-functions-and-classes-to-create-reusable-widgets/53234826#53234826

lib/src/features/user/ui/perf_stats_screen.dart
```diff
@@ -0,0 +1,457 @@
...
+          else if (widgets != null)
+            ...widgets
+          else
+            Text('?', style: trueValueStyle)
+        ],
+      ),
+    ));
+  }
+
+  IntrinsicHeight _rowFromCards(List<PlatformCard> cards) {
```
veloce, line 272: Same remark: make a private `class` instead of function (the remark applies to all other cases)

lib/src/features/user/ui/profile_screen.dart
```diff
@@ -235,43 +238,52 @@ class PerfCards extends StatelessWidget {
...
-                                : LichessColors.red,
-                            size: 12,
-                          ),
-                          Text(userPerf.progression.abs().toString(),
-                              style: TextStyle(
-                                  color: userPerf.progression > 0
-                                      ? LichessColors.good
-                                      : LichessColors.red,
-                                  fontSize: 12)),
+              child: InkWell(
```
veloce, line 241: shouldn't the `InkWell` wrap the `PlatformCard` instead of being its child?

lib/src/utils/duration.dart
```diff
@@ -0,0 +1,44 @@
+import 'package:flutter/material.dart';
+
+import 'package:lichess_mobile/src/utils/l10n_context.dart';
+
+extension DurationExtensions on Duration {
+  String toDaysHoursMinutes(BuildContext context) {
+    /*
```
veloce, line 7: same remark about useless comment in the code that could be a GitHub comment.

lib/src/utils/duration.dart
```diff
@@ -0,0 +1,44 @@
+import 'package:flutter/material.dart';
+
+import 'package:lichess_mobile/src/utils/l10n_context.dart';
+
+extension DurationExtensions on Duration {
+  String toDaysHoursMinutes(BuildContext context) {
```
veloce, line 6: perhaps this function could have a unit test?

test/features/user/data/user_repository_test.dart
```diff
@@ -95,6 +96,298 @@ void main() {
     });
   });
 
+  group('UserRepository.getUserPerfStatsTask', () {
+    const testPerf = Perf.rapid;
+    final String perfApiString = testPerf.toString().split('.').last;
```
veloce, line 101: `perf.name`

lib/src/features/user/data/user_repository.dart
```diff
@@ -24,6 +25,21 @@ class UserRepository {
     });
   }
 
+  TaskEither<IOError, UserPerfStats> getUserPerfStatsTask(
+      String username, Perf perf) {
+    final String perfApiString = perf.toString().split('.').last;
```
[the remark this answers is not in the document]
Dante-010: I did not use `perf.name` due to the fact that, for example, `Perf.threeCheck.name` is 'Three-Check`, but the api needs to be called with `threeCheck` to work (https://lichess.org/api#tag/Users/operation/apiUserPerf)

lib/src/features/user/ui/perf_stats_screen.dart
```diff
@@ -0,0 +1,457 @@
...
+    return Scaffold(
+        appBar: AppBar(
+          titleSpacing: 0,
+          title: Row(
+            mainAxisAlignment: MainAxisAlignment.start,
+            children: [
+              Icon(perf.icon),
+              const SizedBox(width: 5),
+              Expanded(
+                child: FittedBox(
```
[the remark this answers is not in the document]
Dante-010: This `FittedBox` allows the text to be scaled down in case it is too large to fit in one line, instead of the word placed in two lines (thus keeping the UI more natural, in my opinion).

lib/src/features/user/ui/perf_stats_screen.dart
```diff
@@ -0,0 +1,457 @@
...
+                            '${data.percentile!.toStringAsFixed(2)}%',
+                            perf.name),
+                    style: TextStyle(
+                        fontSize: _defaultStatFontSize,
+                        color: textShade(context, _customOpacity)),
+                    textAlign: TextAlign.center)
+            ]),
+            // The number '12' here is not arbitrary, since the API returns the progression for the last 12 games (as far as I know).
+            _customPlatformCard(context,
+                context.l10n.progressOverLastXGames('12').replaceAll(':', ''),
```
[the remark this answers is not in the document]
Dante-010: I also did not find any value in the JSON response which corresponds to how many games this progression corresponds to.

lib/src/features/user/model/user.dart
```diff
@@ -134,3 +134,152 @@ class UserStatus with _$UserStatus {
...
+    required String username,
+    required Perf perf,
+  }) = _UserPerfStatsParameters;
+}
+
+@freezed
+class UserPerfStats with _$UserPerfStats {
+  const factory UserPerfStats(
+      {required double
+          rating, // <- These five parameters could be represented as a
```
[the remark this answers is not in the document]
Dante-010: I meant that these five statistics are the same ones you find in the `UserPerf` class, but you're right, the comment doesn't add anything.

lib/src/features/user/data/user_repository.dart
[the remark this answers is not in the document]
Dante-010: This was the easiest workaround I found, but I also thought of using `perf.name` and adding special cases for the names that do not match, what do you think?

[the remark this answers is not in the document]
veloce: my bad, this one is an extended enum.. I added the field `name` so it overrides the default behaviour (which is what we wanted here). This is weird from dart to allow this.
If you rename the enum fields `name` and `shortName` to `title` and `shortTitle` then you should be able to access the original `name` property which is the one we want.

lib/src/features/user/model/user.dart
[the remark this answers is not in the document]
veloce: oh I see. If they are the same as `UserPerf` then you can compose your `UserPerfStats` like that:
```dart
const factory UserPerfStats({
   required UserPerf userPerf,
   ...
})
```

but it's not mandatory

lib/src/features/user/ui/perf_stats_screen.dart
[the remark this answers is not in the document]
veloce: ok thanks! I've never used that widget, thus the asking

lib/src/features/user/ui/perf_stats_screen.dart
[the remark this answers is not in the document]
veloce: I think your right: https://lichess.org/@/penguingim1/perf/bullet 

every user perf page on lichess mention the last 12 games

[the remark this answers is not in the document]
veloce: why 12? no idea but we can safely follow the website ;)

lib/src/features/user/ui/profile_screen.dart
Dante-010: I did it like that at first, because, as you say, it makes more sense. The thing is when I do that the `InkWell`'s border is around the `PlatformCard`'s border. I'm sure there is a way to fix this, but when I swapped the hierarchy the problem fixed itself, so I thought it was simpler and more readable to leave it like that.

lib/src/utils/duration.dart
Dante-010: I wasn't aware of GitHub comments, sorry!

lib/src/utils/duration.dart
Dante-010: Of course! It looks like it can fail in lots of disastrous ways...

lib/src/features/user/model/user.dart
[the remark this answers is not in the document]
Dante-010: I could do it, but since the JSON is different, if you don't mind, I'll stick with what I've got.

lib/src/features/user/ui/profile_screen.dart
Dante-010: If you want, I can try to fix it, but I would leave it that way (not because of laziness, but instead due to code readability).

lib/src/features/user/ui/perf_stats_screen.dart
Dante-010: I was unaware of that, thanks for pointing it out! I'll write the classes the conventional way, but after reading through the Stack Overflow question, maybe you should consider adding the following package: https://pub.dev/packages/functional_widget ?
Maybe it'll help reduce some boilerplate code, but it's up to your own preference.

On a side note: are we (contributors) allowed to add new packages ourselves? (Just in case, I did not add one).

lib/src/utils/duration.dart
Dante-010: This method works for english, but I'm not 100% sure it will work for every language out there...
Although now that I think about it, this could be solved by checking `Intl.getCurrentLocale()`, and if there's a special case, just alter the function behaviour.

Should I add this as a comment in case someone in the future stumbles with this problem?

Dante-010:
There appears to be an error regarding the localization file, since not only I did not modify the file, but also the same error is happening with other pull requests. 

What's more, when I run `flutter gen-l10n`, even after running `flutter pub get` or even when I build the app, everything seems to run just fine!

lib/src/features/user/ui/perf_stats_screen.dart
```diff
@@ -0,0 +1,536 @@
...
+        fontSize: _defaultStatFontSize,
+        color: textShade(context, _customOpacity));
+
+    final List<Widget> streakWidgets = List.empty(growable: true);
+    final longestStreakStr = context.l10n.longestStreak('').replaceAll(':', '');
+    final currentStreakStr = context.l10n.currentStreak('').replaceAll(':', '');
+    // This variable may seem useless, why not check (streak == maxStreak)?
+    // Because this avoids the rare case where maxStreak == curStreak and the same title is shown twice.
+    bool inFirstLoop = true;
+
```
veloce, line 433: There's nothing wrong with the way you've written it, but for the sake of example here's a more functional approach, using the `Iterable.mapIndexed` method of flutter collection lib: https://api.flutter.dev/flutter/package-collection_collection/IterableExtension/mapIndexed.html

```dart
final List<Widget> = [maxStreak, curStreak].mapIndexed((index, streak) {
  final streakTitle = Text(index == 0 ? longestStreakStr : currentStreakStr);
  ...
}).toList(growable: false);
```
I general I prefer the functional style because:
1. you don't need to mutate a list
2. no need for the mutable `inFirstLoop` variable

Sometimes I also write in the imperative style, but when I can I'd always choose the more functional style.

lib/src/features/user/ui/profile_screen.dart
```diff
@@ -231,47 +234,65 @@ class PerfCards extends StatelessWidget {
...
-                                      ? LichessColors.good
-                                      : LichessColors.red,
-                                  fontSize: 12)),
+            child: InkWell(
+              splashFactory: isPerfWithoutStats
+                  ? NoSplash.splashFactory
+                  : InkSplash.splashFactory,
+              customBorder: kPlatformCardBorder,
+              onTap: isPerfWithoutStats
+                  ? () {}
```
veloce, line 248: It is better to put `null` instead of `() {}`. In the case of a button this is the way flutter will mark it as `disabled`. Not sure it applies to `InkWell` though.

test/utils/duration_test.dart
```diff
@@ -0,0 +1,73 @@
...
+import 'package:flutter_localizations/flutter_localizations.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:flutter/material.dart';
+
+import 'package:lichess_mobile/src/utils/duration.dart';
+
+void main() {
+  group('DurationExtensions.toDaysHoursMinutes()', () {
+    // The following tests are widget tests because the function being
+    // tested need access to the BuildContext, and this is a simple way to access it.
```
veloce, line 11: Not a big fan of doing like that. I see a couple of options:
- instead of passing `BuildContext` as arg to the function you can pass `AppLocalization` which can be easily mocked 
- instead of an extension you can use a service with `Riverpod` that depends on the `l10nProvider`: https://github.com/lichess-org/mobile/blob/main/lib/src/common/l10n.dart#L7 (that way this method will be accessible anywhere, as long there is a `ref` in the scope, no need for the `BuildContext`.

lib/src/features/user/model/user.dart
```diff
@@ -134,3 +134,179 @@ class UserStatus with _$UserStatus {
...
+
+  factory UserStreak.fromPick(RequiredPick pick) {
+    final path = pick.path;
+    if (path.length <= 1) {
+      throw ArgumentError('pick must have a path longer than one');
+    }
+
+    // Since we are passing 'cur' or 'max' as the last pick,
+    // we check the previous value in the path (the parent).
+    final type = path.reversed.elementAt(1);
```
Dante-010, line 259: Arguably this "problem" could be solved by creating a class that stores both the current streak and the longest streak, but I thought it would be better to leave it this way in order not to make these separated (although related) streaks coupled.
Even so, I can change this if you want to.

lib/src/features/user/model/user.dart
```diff
@@ -211,86 +211,77 @@ class UserPerfStats with _$UserPerfStats {
...
-  factory UserTimeStreak.fromJson(Map<String, dynamic> json) =>
-      UserTimeStreak.fromPick(pick(json).required());
-
-  factory UserTimeStreak.fromPick(RequiredPick pick) =>
-      UserTimeStreak.withPick(pick,
-          timePlayed: pick('v').asDurationFromSecondsOrThrow());
+  factory UserStreak.fromPick(RequiredPick pick) {
+    final path = pick.path;
+    if (path.length <= 1) {
+      throw ArgumentError('pick must have a path longer than one');
```
veloce, line 254: Here it is JSON decoding, so `ArgumentError` is not appropriate. Moreover in dart `Error` is not meant to be catched.
Cf: https://stackoverflow.com/questions/17315945/error-vs-exception-in-dart

It should be a `DataFormatException` or a `PickException` and the message should be more useful, like `'cannot decode $object as 'UserStreak''`

lib/src/features/user/model/user.dart
```diff
@@ -211,86 +211,77 @@ class UserPerfStats with _$UserPerfStats {
...
+      case 'win':
+      case 'loss':
+      case 'nb':
+        return UserStreak.gameStreak(
+            gamesPlayed: value.asIntOrThrow(),
+            isValueEmpty: isValueEmpty,
+            startGame: startGame,
+            endGame: endGame);
+      default:
+        throw UnsupportedError('Unsupported streak type $type');
```
veloce, line 282: idem, throw an Exception with a `cannot decode $object...` message

lib/src/features/user/ui/profile_screen.dart
Dante-010: You were right about this, I don't know how or why I thought this worked, but I tested it today and the border was not expecting as expected, sorry for my mistake.

PR #36, opened 2023-01-27
allow http on android devices/emulators

schlawg:
In concert with the following command (which must be run after the emulator/device has started): adb reverse tcp:9663 tcp:9663

Avoid 10.0.2.x networking as it will confuse CORS.  Use localhost:9663

If Chrome instacrashes, it is likely you need to disable vulkan in  emulator settings.  Will add docs for all this pending discussion of where they should go.

### DIFF
--- a/lib/src/features/auth/data/auth_repository.dart
+++ b/lib/src/features/auth/data/auth_repository.dart
@@ -56,6 +56,7 @@ class AuthRepository {
           await _appAuth.authorizeAndExchangeCode(AuthorizationTokenRequest(
         kLichessClientId,
         redirectUri,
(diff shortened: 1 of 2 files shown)


lib/src/features/auth/data/auth_repository.dart
```diff
@@ -56,6 +56,7 @@ class AuthRepository {
           await _appAuth.authorizeAndExchangeCode(AuthorizationTokenRequest(
         kLichessClientId,
         redirectUri,
+        allowInsecureConnections: true,
```
veloce, line 59: this should be `true` only in debug mode

PR #45, opened 2023-02-05
Add Live Streamers

### LINKED ISSUE #41: Improve TV with variants
TV screen

For this you have the lichess API: 

- https://lichess.org/api#tag/TV/operation/tvChannels
- https://lichess.org/api#tag/Games/operation/streamGame

It should replace the existing stream that only shows the main TV.

It should filter out unsupported variants (crazy house, racing kings, horde).

TV channel selector at the top should use a widget adapted to the platform, that can be something similar to what is implemented here:

https://github.com/flutter/samples/blob/main/platform_design/lib/widgets.dart#L289

### LINKED ISSUE #42: Live streamers
Streamers screen

Corresponding lichess API:

- https://lichess.org/api#tag/Users/operation/streamerLive

The widget should look similar to the already existing `LeaderboardWidget`. It is basically a `PlatformCard` with a `dense` `ListTileTheme` and a list of `PlatformListTile` inside. We may want to extract that into a custom widget here.

Tapping on a streamer item should open an external browser to the streaming platform (twitch, YouTube, ...).

nav-28:
Resolves #42


lib/src/ui/watch/watch_screen.dart
```diff
@@ -0,0 +1,200 @@
...
+        tvStream.when(
+          data: (position) {
+            final boardData = BoardData(
+              interactableSide: InteractableSide.none,
+              orientation: featuredGame?.orientation.cg ?? Side.white,
+              fen: position.fen,
+              lastMove: position.lastMove?.cg,
+            );
+            return Center(
+              child: ConstrainedBox(
```
veloce, line 130: Is it not the same as `SizedBox.square(350)`?

Perhaps we should also adapt to different screen sizes (including tablets), but that can be done later I guess.

lib/src/ui/watch/watch_screen.dart
```diff
@@ -0,0 +1,200 @@
...
+                constraints: BoxConstraints.tight(const Size.square(350)),
+                child: GameBoardLayout(
```
veloce, line 132: Here you should not use the `GameBoardLayout` as it is not meant to display board previews. 

We should make a dedicated `BoardPreview` widget instead as we'll certainly reuse it elsewhere.

lib/src/ui/watch/watch_screen.dart
```diff
@@ -0,0 +1,200 @@
...
+            return const EmptyBoard(errorMessage: 'Could not load TV Stream');
+          },
+          loading: () => const EmptyBoard(),
+        ),
+      ],
+    );
+  }
+}
+
+class EmptyBoard extends StatelessWidget {
```
veloce, line 157: If it is only used once, I'd rather not have a widget but the code directly in place. Here the size and board data have some duplication.

lib/src/ui/watch/watch_screen.dart
nav-28: added a board preview widget

lib/src/common/social_icons.dart
```diff
@@ -0,0 +1,36 @@
...
+///         Author:    Dave Gandy
+///         License:   SIL ()
+///         Homepage:  http://fortawesome.github.com/Font-Awesome/
+///
+import 'package:flutter/widgets.dart';
+
+class SocialIcons {
+  SocialIcons._();
+
+  static const _kFontFam = 'SocialIcons';
```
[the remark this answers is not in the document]
nav-28: I added the icons file directly, which seemed easier.

veloce:
Thanks for you great work, I think we're almost here. I tested and noticed a couple of bugs / things to change:

- the TV channel selector shows the enum instead of a label:
Capture d’écran 2023-02-21 à 08 30 57
- TV channel selector has no effect: if I change the tv channel the game doesn't change, and when we go back to the selector, the selected item is still the previous one
- on the watch tab (1st level), it's better not to have the sound (and no sound button); only keep the sound on the main TV screen
- instead of the 3 dots at the top right of the board let's make a tappable board to enter the main TV screen (the 3 dots makes sense for the list because items are tappable, but it doesn't for the board)

PR #48, opened 2023-02-10
Piece Set Picker

bclehmann:
Lets users choose the piece set that will be used app-wide:
image

I have two remaining questions:
- Should the default be merida or cburnett? The default in lila and lichobile is cburnett, but for now I left it as merida
- Should these be ordered alphabetically? In a custom order? Is the order from `PieceSet.values` correct?

Finally, I haven't meaningfully used flutter in a year and a half, so feel free to rip this PR apart, it might need it.


veloce:
Sorry closing in favour of https://github.com/lichess-org/mobile/pull/49 which was started at the same time.

PR #52, opened 2023-02-13
Add build android and upload artifacts workflow

PunchCakee:


PunchCakee:
Do note that doing `flutter build apk --release` requires signing the apk itself so it is not possible for me to make one

veloce:
I don't get it why you add another workflow since there is already a `build.yml`. 

You can add the upload step only for android to the existing workflow instead.

PunchCakee:
> I don't get it why you add another workflow since there is already a `build.yml`.
> 
> You can add the upload step only for android to the existing workflow instead.

The current workflow does not build an APK it builds an AAB which is completely usless for a developer to test it out on their phone. Also a lot of pepole want to try the mobile app and they really dont wanna build it all from scratch.

veloce:
I don't want to add another build since it is already long, just for people wanting to try the app.
When the app is ready we might have a public beta program. For now, there's not much to test.

PunchCakee:
> I don't want to add another build since it is already long, just for people wanting to try the app. When the app is ready we might have a public beta program. For now, there's not much to test.

No problem I'll just redirect them to my fork if they wanna try it

veloce:
Ok, closing it then.

Issue #50, opened 2023-02-10
Board theme picker


veloce:
Add a board theme picker in settings screen.

This feature depends on https://github.com/lichess-org/flutter-chessground/issues/6.

A board thumbnail should be passed to the `leading` parameter of the settings list tile. This thumbnail should be an image consisting of 6 board squares aligned (a row). So that we can display the 6 pieces type over the board thumbnail. Thumbnail pixel ratio should be the same as the original board image, so the asset size should be 768*128.

PR #55, opened 2023-02-13
Board Theme Picker

bclehmann:
#50 (should probably also close https://github.com/lichess-org/flutter-chessground/issues/6)
image


veloce:
That was quick! I think I'd leave the board sample empty (no pieces) for the board picker, and show it under the pieces in the pieces picker.

bclehmann:
Is this what you mean?
image
image

veloce:
I meant the first yes. I think it's nice to see the pieces over the board when you choose the pieces, but when you choose the board it's better to see it empty.

bclehmann:
I rebased to fix the merge conflicts, the only thing I'd note is that long pieceset names can be forced to break into two lines when the checkmark is present:
image
image

These screenshots are from an emulated Pixel 6 Pro which has a 19.5:9 aspect ratio. My Pixel 3 has a fatter aspect ratio of 18:9 and I still noticed it. I'm currently nursing my old phone back to health because it has a 16:9 aspect ratio.

I don't know if this is a concern, and I don't really know how many would notice it because I'm not really a mobile developer by trade, so I don't know if it's normal for phones to be this thin.

veloce:
There's a way to force it to one line... but I think also this will not fix on iOS so I may modify it a bit after my iOS tests. Like put everything in 2 lines.

PR #58, opened 2023-02-23
Fix empty space issue in rating section of profile screen.

Rica320:
when a player has not yet made a single game or obtained an appropriate kClueLessDeviation a empty space is show. That is horrible to the UI and considered a bad practise. The following changes were made...

Before             |  After
:-------------------------:|:-------------------------:
imagem |  imagem)

when tapping on the play button the player will be forward to the play screen

### DIFF
--- a/lib/src/ui/user/user_screen.dart
+++ b/lib/src/ui/user/user_screen.dart
@@ -205,7 +205,7 @@ class PerfCards extends StatelessWidget {
     }).toList(growable: false);
 
     if (userPerfs.isEmpty) {
-      return Container();
+      return const SizedBox.shrink();
     }
 
     return Padding(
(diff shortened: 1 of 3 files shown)


veloce:
Thank you for that fix. But we can't show that button. The "play with a friend" is not even implemented. Currently there is the play against computer but that feature was a demo it will be removed before first release.

So instead please just remove the empty space.

Rica320:
> Thank you for that fix. But we can't show that button. The "play with a friend" is not even implemented. Currently there is the play against computer but that feature was a demo it will be removed before first release.
> 
> So instead please just remove the empty space.

Hi There. I was reluctant to put the button for the exact same reason. I will just remove the empty space then. Maybe later we can rethink this.

lib/src/ui/user/user_screen.dart
```diff
@@ -204,6 +204,10 @@ class PerfCards extends StatelessWidget {
           p.ratingDeviation < kClueLessDeviation;
     }).toList(growable: false);
 
+    if (userPerfs.isEmpty) {
+      return Container();
```
veloce, line 208: Use `SizedBox.shrink()` instead.

Issue #24, opened 2023-01-09
User activity feed


veloce:
Add the user activity feed section on profile screen.

See it on lichess: https://lichess.org/@/veloce

API reference: https://lichess.org/api#tag/Users/operation/apiUserActivity

Design notes: the list should stay short, so we should show only the most important items, and keep one line per item. Example: don't show the followers in the "Gain X followers" item.

Ideally it should be place above the "recent games" section:

<img width="411" alt="Capture d’écran 2023-01-09 à 15 37 06" src="https://user-images.githubusercontent.com/423393/211333345-add85255-d066-44a0-8c10-bb6935c29344.png">

vrund297:
I am chess Player with around 1900 lichess & backend developer too.
can you please assign this issue to me?

vrund297:
@veloce

schlawg:
You can work on it without being assigned.

veloce:
Here you are @vrund297 . I missed your ping, and indeed you can work on it without being assigned but if you do so the best is to inform me on discord.

If I don't respond on GitHub, you can also reach on discord.

veloce:
Closed by https://github.com/lichess-org/mobile/pull/116

PR #60, opened 2023-02-25
User activity feed

ZTL-UwU:
#24


ZTL-UwU:
The activity feed is mostly done. However, the size of the list is not ideal. With so many items it generally longer than the screen height, blocking the **Recent Games** section.

640046d2ae446_preview

Should we put the **activity feed** section and the **recent games** section into two separate tabs like the web version.

64004832aad7e_preview

@veloce

veloce:
You should cap the length of that list to a predetermined value, like 10. The header should be tappable (with a three dots on the right, like the leaderboard on home page) and push a detailed screen with the full length list.

veloce:
@ZTL-UwU your PR is still a draft. I think this is a nice feature to have in the app, do you still want to work on it or should I take over?

veloce:
BTW you should not edit manually the `app_en.arb` file. See https://github.com/lichess-org/mobile/blob/main/CONTRIBUTING.md

ZTL-UwU:
@veloce Thank you! I am not very experienced in mobile developing and I found myself not able to cap the length of the list so the pr is kinda suspended. It would be great if you can take over.

PR #63, opened 2023-03-01
Add Setting-Menu for changing Sound-Theme 

youngunand:
Done: 

- Added sound assets
- Added Provider for Storing user Selection
- Built UI

To do:

- Maybe someone can provide an Icon, for the List-Tile of the new setting entry Sound (just used a random one now (Storm))
- I'm unsure how to implement the loading logic of the sounds in sound_service.dart (loadSounds), what could be the best approach?
- Maybe find a better place to store the enum SoundTheme than common


veloce:
589 files! that's way too much. Please add only the necessary files. Let's take this one step at a time.
For now the app uses only these sounds:
- capture
- move
- dong

So you can remove everything else. And we don't need both `.ogg` and `.mp3` files. Only keep the `.mp3` for now. We'll add the `.ogg` if and only if some devices don't support the `.mp3`.

Thank you!

veloce:
For the icon in the settings list entry I'd use the material design `Icons.volume_up` .

veloce:
Closing because https://github.com/lichess-org/mobile/pull/152 has been merged.

PR #117, opened 2023-04-23
Puzzle history

### LINKED ISSUE #79: Puzzle history
Offline puzzle history, using a new SQLite table.

We should store only the last n (2?) months of history. Deleting old items should be done at startup time, when opening the database, or with a cron job.

On main puzzle tab screen, only show the last 10 items. Tapping the "..." goes to the full history screen.

Tablet support needed here. See leaderboard screen for an example.

Puzzle history

nav-28:
Resolves #79


lib/src/model/puzzle/puzzle_history_storage.dart
```diff
@@ -0,0 +1,229 @@
...
+        'lastModified': DateFormat('yyyy-MM-dd').format(date),
+        'data': jsonEncode(puzzle.toJson()),
+      },
+      conflictAlgorithm: ConflictAlgorithm.replace,
+    );
+  }
+}
+
+@Freezed(fromJson: true, toJson: true)
+class PuzzleHistory with _$PuzzleHistory {
```
veloce, line 187: If I understand well your code, this class represents a single day of puzzle history. So the name is not good, it should be called `PuzzleHistoryDay` or sth like that.
And the `date` field should be called `day`.

I'm not sure this is the best way to represent the history actually, but I think to think about it.

lib/src/model/puzzle/puzzle_history_storage.dart
```diff
@@ -0,0 +1,229 @@
...
+    required PuzzleTheme angle,
+  }) = _PuzzleHistory;
+
+  factory PuzzleHistory.fromJson(Map<String, dynamic> json) =>
+      _$PuzzleHistoryFromJson(json);
+}
+
+// only used for parsing and saving PuzzleIdAndResult from database
+@Freezed(fromJson: true, toJson: true)
+class PuzzleHistoryData with _$PuzzleHistoryData {
```
veloce, line 200: should be private then

lib/src/model/puzzle/puzzle_history_storage.dart
```diff
@@ -0,0 +1,229 @@
...
+  const factory PuzzleHistoryData({
+    required IList<PuzzleIdAndResult> puzzles,
+  }) = _PuzzleHistoryData;
+
+  factory PuzzleHistoryData.fromJson(Map<String, dynamic> json) =>
+      _$PuzzleHistoryDataFromJson(json);
+}
+
+@Freezed(fromJson: true, toJson: true)
+class PuzzleIdAndResult with _$PuzzleIdAndResult {
```
veloce, line 210: I guess that one should be private too

lib/src/model/puzzle/puzzle_providers.dart
```diff
@@ -34,11 +35,16 @@ Future<PuzzleStreakResponse> streak(StreakRef ref) {
   return Result.release(repo.streak());
 }
 
-// TODO when history database is available should first try to fetch from there
 @Riverpod(keepAlive: true)
-Future<Puzzle> puzzle(PuzzleRef ref, PuzzleId id) {
```
veloce, line 39: the return type should stay `Future<Puzzle?>` imo; we're not certain the puzzle exist in the db;

lib/src/model/puzzle/puzzle_providers.dart
```diff
@@ -34,11 +35,16 @@ Future<PuzzleStreakResponse> streak(StreakRef ref) {
...
-  final repo = ref.watch(puzzleRepositoryProvider);
-  return Result.release(repo.fetch(id));
+Future<Puzzle?> puzzle(
+  PuzzleRef ref,
+  PuzzleId id,
+) {
+  // ignore: avoid_manual_providers_as_generated_provider_dependency
+  final repo = ref.watch(puzzleHistoryStorageProvider);
+  return repo.fetchPuzzle(
```
veloce, line 45: this provider can be used to get puzzles never played before on the app so it should first try to fetch locally in the db, then if not found fetch from lichess server

lib/src/model/puzzle/puzzle_providers.dart
```diff
@@ -59,3 +65,10 @@ Future<PuzzleDashboard> puzzleDashboard(PuzzleDashboardRef ref, int days) {
   final repo = ref.watch(puzzleRepositoryProvider);
   return Result.release(repo.puzzleDashboard(days));
 }
+
+@Riverpod(keepAlive: true)
+Future<IList<PuzzleHistory>?> puzzleHistory(PuzzleHistoryRef ref) {
```
veloce, line 70: as said above, this should be paginated.

lib/src/model/puzzle/puzzle_service.dart
```diff
@@ -99,12 +104,29 @@ class PuzzleService {
...
     required UserId? userId,
     required PuzzleSolution solution,
+    required Puzzle puzzle,
     PuzzleTheme angle = PuzzleTheme.mix,
   }) async {
     final data = await storage.fetch(
       userId: userId,
       angle: angle,
     );
+    final history = await historyStorage.fetch(
```
veloce, line 114: The fetch and concatenation should be handled by the storage, in a transaction. Here we should only do:
```dart
await historyStorage.save(
  userId: userId,
  angle: angle,
  puzzleResult: puzzleResult
);
```

lib/src/ui/puzzle/puzzle_dashboard_screen.dart
```diff
@@ -81,6 +82,7 @@ class _PuzzleDashboardScreenState extends ConsumerState<PuzzleDashboardScreen> {
   }
 
   Future<void> _refreshData() {
+    ref.invalidate(puzzleHistoryProvider);
```
veloce, line 85: See: https://github.com/lichess-org/mobile/blob/main/lib/src/ui/account/profile_screen.dart#L138

on how to refresh several providers at once.

lib/src/ui/puzzle/puzzle_history_widget.dart
```diff
@@ -0,0 +1,234 @@
...
+        debugPrint(
+          'SEVERE: [PuzzleHistoryWidget] could not load next Puzzle; $e\n$s',
+        );
+        return const Center(child: Text('Could not load Puzzle History'));
+      },
+      loading: () => const Center(child: CircularProgressIndicator()),
+    );
+  }
+
+  IList<PuzzleHistory> getFirst10(IList<PuzzleHistory> history) {
```
veloce, line 51: That selection should be done on DB level

lib/src/ui/puzzle/puzzle_session_widget.dart
```diff
@@ -112,7 +112,7 @@ class PuzzleSessionWidgetState extends ConsumerState<PuzzleSessionWidget> {
                                 final nextContext = PuzzleContext(
                                   userId: widget.initialPuzzleContext.userId,
                                   theme: widget.initialPuzzleContext.theme,
-                                  puzzle: puzzle,
+                                  puzzle: puzzle!,
```
veloce, line 115: why is that needed now?

veloce (commented):
I did not review everything yet, (I left the history widget) because the storage part need to be changed to support paginated queries.

lib/src/model/puzzle/puzzle_history_storage.dart
```diff
@@ -0,0 +1,229 @@
...
+    required PuzzleTheme angle,
+    required IList<PuzzleIdAndResult> data,
+    required Puzzle puzzle,
+  }) async {
+    await _db.insert(
+      _dbHistory,
+      {
+        'userId': userId?.value ?? _anonUserKey,
+        'angle': angle.name,
+        'solvedDate': DateFormat('yyyy-MM-dd').format(date),
```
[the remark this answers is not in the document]
nav-28: `DateTime.toIso8691Strin()` includes the time precision. So I am assuming we'd want to remove that. So I think that something like this should work
``` dart
final now = DateTime.now();
final dateOnly = DateTime(now.year, now.month, now.day);
dateOnly.toIso8601String();
```

nav-28:
closing in favour of #187

PR #144, opened 2023-05-04
Game material difference

ZTL-UwU:


ZTL-UwU:
Will add the icons later

veloce:
Hi thanks for this! A screenshot or video showing the feature would be nice.

ZTL-UwU:
@veloce

lib/src/model/game/game.dart
```diff
@@ -87,13 +90,98 @@ class ClockData with _$ClockData {
   }) = _ClockData;
 }
 
+@freezed
+class MaterialDiffSide with _$MaterialDiffSide {
+  const factory MaterialDiffSide({
+    required Map<Role, int> pieces,
```
veloce, line 96: You should use an `IMap` instead.

lib/src/model/game/game.dart
```diff
@@ -87,13 +90,98 @@ class ClockData with _$ClockData {
...
+class MaterialDiff with _$MaterialDiff {
+  const MaterialDiff._();
+
+  const factory MaterialDiff({
+    required MaterialDiffSide black,
+    required MaterialDiffSide white,
+  }) = _MaterialDiff;
+
+  factory MaterialDiff.fromBoard(Board board) {
+    final Map<Role, int> pieceScores = {
```
veloce, line 111: should be defined as a `const IMap<Role, int>` outside

lib/src/ui/game/archived_game_screen.dart
```diff
@@ -76,6 +81,41 @@ class ArchivedGameScreen extends ConsumerWidget {
   }
 }
 
+class _AppTitle extends ConsumerWidget {
```
veloce, line 84: This shouldn't be there, it's not part of the material difference feature, so please remove it.

I'd like to review features/improvements separately, and I'm not sure to accept this change.

lib/src/ui/user/user_activity.dart
veloce, line 1: I don't understand the change in the whole file. Again this is not part of the material diff feature.

Maybe you fixed a bug or you just improved the code, I have no idea. So this should be part of another PR where you explain what is this change. Thanks.

lib/src/widgets/player.dart
```diff
@@ -15,15 +19,35 @@ class BoardPlayer extends StatelessWidget {
...
   });
 
   final Player player;
   final Duration? clock;
   final bool? active;
+  final MaterialDiffSide? diff;
 
   @override
   Widget build(BuildContext context) {
+    final IList<Role> roleList = {
```
veloce, line 33: that should be a `const` defined outside. But you can actually use `Role.values` instead.

lib/src/widgets/player.dart
```diff
@@ -15,15 +19,35 @@ class BoardPlayer extends StatelessWidget {
...
+      Role.king,
+      Role.queen,
+      Role.rook,
+      Role.bishop,
+      Role.knight,
+      Role.pawn,
+    }.toIList();
+
+    final Map<Role, IconData> iconByRole = {
```
veloce, line 42: should be a private const defined outside

lib/src/widgets/player.dart
```diff
@@ -101,16 +152,37 @@ class PlayerRating extends StatelessWidget {
...
   final TextStyle? style;
 
   @override
   Widget build(BuildContext context) {
     final ratingStr =
         rating is double ? rating.toStringAsFixed(2) : rating.toString();
-    return Text(
-      '$ratingStr${provisional == true || deviation > kProvisionalDeviation ? '?' : ''}',
-      style: style,
+    return Row(
```
veloce, line 163: This has again nothing to do with material diff, please remove it.

veloce (commented):
Thanks for the screenshots.

Please add some unit tests to test the logic that compute the material diff and the score.

veloce:
Oops, I messed up with git and your pr was not cleanly merged with a commit. But it's merged.

Issue #51, opened 2023-02-10
Sound theme picker


veloce:
Add a new settings section to choose a sound theme (or silent).

Available sounds are listed here: https://github.com/lichess-org/lila/blob/master/modules/pref/src/main/SoundSet.scala

And the assets are here: https://github.com/lichess-org/lila/tree/master/public/sound

Kirito-Excalibur:
Can you please assign me this?

veloce:
done!

youngunand:
Could you assign me?
 I suppose this issue is still open.

brandonpl8:
Should the sound theme setting belong the boardPreferences or generalPreferences? An argument can be made for both depending on how the sound is used in the rest of the app.

veloce:
Closed by #152

PR #152, opened 2023-05-08
Sound Theme Picker

brandonpl8:
To close Sound Theme Picker #51


veloce:
Thank you @brandonpl8 ! 

I just handled the missing files since storm was merged, and improved the way to load files.

PR #158, opened 2023-05-12
Set puzzle themes text style

CloudyDino:
Makes sure the "Puzzle themes" text is bolded/styled like in the other buttons.

| Before | After |
| --- | --- |
| image | image |


veloce:
It was actually not bold on purpose, as I want to put more accent on "Healthy mix" which should be the main way to play puzzles.

But I agree it's a bit weird not having it bold like the others.

PR #159, opened 2023-05-12
Refactor _PuzzleButton and _ThemedPuzzleButton

CloudyDino:
Makes sure the "Puzzle themes" text is bolded/styled like in the other buttons. I also refactored the code to use `_PuzzleButton` and `_ThemedPuzzleButton` to make sure this issue doesn't happen again.

If we don't want the whole refactor, feel free to just accept #158

| Before | After |
| --- | --- |
| image | image |


veloce:
I'm not sure to understand the advantage of the refactor here.

veloce:
Thanks for this, but I won't take the refactoring.

PR #172, opened 2023-05-17
Fix recent game trailing result

ZTL-UwU:
from discord


ZTL-UwU:
**Game Aborted**  


**Draw**

PR #173, opened 2023-05-18
Reorder perf cards

### LINKED ISSUE #163: Reorder perf cards

ZTL-UwU:
closes #163


veloce:
Thanks for this but I don't want to let the user choose this as a preference. Settings should be only last resort, it's always better  to provide sane defaults than to let the user choose, for a maintenance perspective, and for the user XP.

So, how the card should be ordered? There are 2 options:
- by number of games played
- by latest game played

I think we should stick to option 1., by number of games played. We already have that information and its a good way to order.

ZTL-UwU:
> Thanks for this but I don't want to let the user choose this as a preference. Settings should be only last resort, it's always better  to provide sane defaults than to let the user choose, for a maintenance perspective, and for the user XP.
> 
> So, how the cards should be ordered? There are 2 options:
> - by number of games played
> - by latest game played
> 
> I think we should stick to option 1., by number of games played. We already have that information and its a good way to order.

I see. Makes a ton of sense. Order for other user's profile will be more reasonable as well.

veloce:
@ZTL-UwU please fix lint so I can merge this.

ZTL-UwU:
@veloce

PR #187, opened 2023-05-31
Puzzle History

### LINKED ISSUE #79: Puzzle history

nav-28:
Resolves #79

### DIFF
--- a/lib/src/db/database.dart
+++ b/lib/src/db/database.dart
@@ -5,11 +5,26 @@ import 'package:lichess_mobile/src/app_dependencies.dart';
 
 part 'database.g.dart';
 
+const puzzleHistoryTTL = Duration(days: 60);
+
 Future<Database> openDb(DatabaseFactory dbFactory, String path) async {
   return dbFactory.openDatabase(
     path,
     options: OpenDatabaseOptions(
       version: 1,
+      onOpen: (db) async {
+        final nDaysAgo = DateTime.now().subtract(puzzleHistoryTTL);
+        final tableExists = await db.rawQuery(
+          "SELECT name FROM sqlite_master WHERE type='table' AND name='puzzle'",
+        );
+        if (tableExists.isNotEmpty) {
+          await db.delete(
+            'puzzle',
+            where: 'lastModified < ?',
+            whereArgs: [nDaysAgo.toIso8601String()],
+          );
+        }
+      },
       onCreate: (db, version) async {
         await db.execute(
           '''
@@ -21,6 +36,15 @@ Future<Database> openDb(DatabaseFactory dbFactory, String path) async {
           )
           ''',
         );
+
+        await db.execute('''
+          CREATE TABLE puzzle(
+          puzzleId TEXT NOT NULL,
+          lastModified DATE NOT NULL,
+          data TEXT NOT NULL,
+          PRIMARY KEY (puzzleId)
+        )
+          ''');
       },
     ),
   );
--- a/lib/src/db/database.dart
+++ b/lib/src/db/database.dart
@@ -40,7 +40,7 @@ Future<Database> openDb(DatabaseFactory dbFactory, String path) async {
         await db.execute('''
           CREATE TABLE puzzle(
           puzzleId TEXT NOT NULL,
-          lastModified DATE NOT NULL,
+          lastModified TEXT NOT NULL,
           data TEXT NOT NULL,
           PRIMARY KEY (puzzleId)
         )
--- a/lib/src/model/puzzle/history_provider.dart
+++ b/lib/src/model/puzzle/history_provider.dart
@@ -1,0 +1,70 @@
+import 'dart:async';
+
+import 'package:async/async.dart';
+import 'package:freezed_annotation/freezed_annotation.dart';
+import 'package:riverpod_annotation/riverpod_annotation.dart';
+import 'package:fast_immutable_collections/fast_immutable_collections.dart';
+import 'package:lichess_mobile/src/utils/riverpod.dart';
+import 'package:lichess_mobile/src/model/puzzle/puzzle.dart';
+import 'package:lichess_mobile/src/model/puzzle/puzzle_repository.dart';
+
+part 'history_provider.g.dart';
+part 'history_provider.freezed.dart';
+
+const _maxPuzzles = 500;
+
+@riverpod
+Future<IList<PuzzleHistoryEntry>> puzzleRecentActivity(
+  PuzzleRecentActivityRef ref,
+) {
+  ref.cacheFor(const Duration(minutes: 10));
+  final repo = ref.watch(puzzleRepositoryProvider);
+  return Result.release(repo.puzzleActivity(10));
+}
+
+@riverpod
+class PuzzleActivity extends _$PuzzleActivity {
+  PuzzleRepository? _repo;
+  @override
+  Future<PuzzleActivityState> build() async {
+    ref.cacheFor(const Duration(minutes: 10));
+    _repo = ref.read(puzzleRepositoryProvider);
+    final value = await ref.read(puzzleRecentActivityProvider.future);
+    return PuzzleActivityState(list: value, isLoading: false, hasMore: true);
+  }
+
+  void getNext() {
+    final currentVal = state.requireValue;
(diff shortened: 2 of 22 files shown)


nav-28:
The provider code for history is a but hacky. Don't know how to improve it.

lib/src/ui/puzzle/puzzle_history_screen.dart
```diff
@@ -0,0 +1,295 @@
...
+        ),
+      ),
+    );
+  }
+}
+
+class PuzzleHistoryScreen extends StatelessWidget {
+  const PuzzleHistoryScreen(this.historyList);
+
+  final IList<HistoryPuzzle> historyList;
```
veloce, line 114: I don't think we should have this parameter. If the reason is to avoid fetching twice the data, we should use riverpod caching mechanism instead (if needed).

lib/src/model/puzzle/puzzle_providers.dart
```diff
@@ -73,3 +80,74 @@ Future<PuzzleDashboard> puzzleDashboard(
...
+    ref.onDispose(() => _streamSub?.cancel());
+
+    _repo = ref.watch(puzzleRepositoryProvider);
+    final stream =
+        _repo!.puzzleActivity(10, DateTime.now()).asBroadcastStream();
+    _streamSub = stream.listen(
+      (event) {
+        _list.add(event);
+      },
+      onDone: () {
```
veloce, line 104: so here the stream is not used as a stream, rather a future. So we could simplify a lot by removing completely this class and return directly a `Future<List<HistoryPuzzle>>` from the `PuzzleRepository` (see comment here)

lib/src/ui/puzzle/puzzle_history_screen.dart
```diff
@@ -0,0 +1,295 @@
...
+class _Body extends ConsumerStatefulWidget {
+  const _Body(this.historyList);
+  final IList<HistoryPuzzle> historyList;
+  @override
+  ConsumerState<_Body> createState() => _BodyState();
+}
+
+class _BodyState extends ConsumerState<_Body> {
+  final ScrollController _scrollController = ScrollController();
+  List<HistoryPuzzle> _historyList = [];
```
veloce, line 147: So yes it's weird here to have both a `provider` and `state`. The provider isn't actually used as a provider but rather as a repository. And the state is in the stateful widget. It should be one or another but not both.

lib/src/db/database.dart
```diff
@@ -21,6 +36,15 @@ Future<Database> openDb(DatabaseFactory dbFactory, String path) async {
           )
           ''',
         );
+
+        await db.execute('''
+          CREATE TABLE puzzle(
+          puzzleId TEXT NOT NULL,
+          lastModified DATE NOT NULL,
```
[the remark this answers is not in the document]
nav-28: My bad `sql` has Date type not `sqllite`. But somehow it still worked and never gave an error.

lib/src/model/puzzle/puzzle_providers.dart
nav-28: I guess I over engineered the solution while trying to fix some other problems  😆

lib/src/ui/puzzle/puzzle_history_screen.dart
```diff
@@ -0,0 +1,263 @@
...
+                (e) => _HistoryBoard(e, boardWidth),
+              )
+              .toList(),
+        );
+      },
+    );
+  }
+}
+
+class _HistoryBoard extends ConsumerWidget {
```
nav-28, line 199: This widget is used instead of `PuzzleHistoryBoards` to support lazy loading of `ListView.builder`

lib/src/ui/puzzle/puzzle_history_screen.dart
```diff
@@ -0,0 +1,263 @@
...
+  void dispose() {
+    _scrollController.removeListener(_scrollListener);
+    _scrollController.dispose();
+    super.dispose();
+  }
+
+  void _scrollListener() {
+    if (_scrollController.position.pixels ==
+            _scrollController.position.maxScrollExtent &&
+        _historyList.length < 500) {
```
veloce, line 141: 500 should be a declared const

lib/src/ui/puzzle/puzzle_history_screen.dart
```diff
@@ -0,0 +1,263 @@
...
+  final double boardWidth;
+
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {
+    final (fen, turn, lastMove) = puzzle.preview;
+    final isLoading = ref.watch(_puzzleLoadingProvider);
+    return SizedBox(
+      width: boardWidth,
+      height: boardWidth + MediaQuery.of(context).textScaleFactor * 14 + 16.5,
+      child: BoardPreview(
```
veloce, line 212: Use `BoardThumbnail` rather than `BoardPreview`

lib/src/ui/puzzle/puzzle_history_screen.dart
```diff
@@ -0,0 +1,263 @@
...
+class _Body extends ConsumerStatefulWidget {
+  const _Body(this.historyList);
+  final IList<PuzzleHistoryEntry> historyList;
+  @override
+  ConsumerState<_Body> createState() => _BodyState();
+}
+
+class _BodyState extends ConsumerState<_Body> {
+  final ScrollController _scrollController = ScrollController();
+  List<PuzzleHistoryEntry> _historyList = [];
```
veloce, line 120: While there is nothing wrong to use a widget state, I think we could improve this by using riverpod to hold the state of the history. (Using a riverpod `Notifier`).
This would allow us to cache the data for some time. Like if a user goes back to the history screen after exiting it, with the widget state we need to fetch all the data again from the server. With riverpod it can be cached.

Also the screen widget should not take the recent items list as a parameter, because that data can be stale.

veloce (commented):
Nice work, I think we're getting close now :) 

I'd like to be able to cache the history data with riverpod since it's expensive to fetch (cf comment). This is the only thing left before we can merge.

lib/src/model/puzzle/history_provider.dart
```diff
@@ -0,0 +1,70 @@
...
+    _repo = ref.read(puzzleRepositoryProvider);
+    final value = await ref.read(puzzleRecentActivityProvider.future);
+    return PuzzleActivityState(list: value, isLoading: false, hasMore: true);
+  }
+
+  void getNext() {
+    final currentVal = state.requireValue;
+    if (currentVal.list.length < _maxPuzzles) {
+      state = AsyncData(currentVal.copyWith(isLoading: true));
+      Result.release(
```
veloce, line 40: Error handling is missing here.
Instead of `Result.release` you can use `fold` from `result_extensions` package to handle error:
```dart
_repo!.puzzleActivity(50, before: currentVal.list.last.date).fold(
  (value) => ....,
  (error, stackTrack) => AsyncValue.error(error, stackTrace ?? StackTrace.current),
);
```

veloce (commented):
Nice! there's just one error handling missing; now I'll test it :)

lib/src/model/puzzle/history_provider.dart
veloce: I wrote that too fast, maybe we should ignore the error instead, otherwise the whole list will become an error.

nav-28: We could ignore the error or show a snackbar on error

nav-28: added the snackbar

veloce (approved):
Code looks good to me. 
Now I should test that on some real devices to see if the scrolling perf of the long list is ok.
Also I'll try to improve the feature a bit if I can. I'd like to display the puzzles by theme and date for instance; Will have a look to see if not too complicated to to.

veloce:
Continued in #211

PR #189, opened 2023-06-01
Rewind moves

### LINKED ISSUE #143: Fast rewind moves

nav-28:
close #143

I had to remove the tooltip for the bottom button as it won't work with it.

### DIFF
--- a/lib/src/ui/game/archived_game_screen.dart
+++ b/lib/src/ui/game/archived_game_screen.dart
@@ -198,31 +198,45 @@ class _BottomBar extends ConsumerWidget {
                   : null,
               icon: const Icon(LichessIcons.fast_backward, size: 20),
             ),
-            BottomBarIconButton(
-              key: const ValueKey('cursor-back'),
-              // TODO add translation
-              semanticsLabel: 'Backward',
-              onPressed: canGoBackward
-                  ? () {
-                      ref
-                          .read(gameCursorProvider(gameData.id).notifier)
-                          .cursorBackward();
-                    }
+            LongPressButton(
+              callback: canGoBackward
+                  ? () => ref
+                      .read(gameCursorProvider(gameData.id).notifier)
+                      .cursorBackward()
                   : null,
-              icon: const Icon(LichessIcons.step_backward, size: 20),
+              child: BottomBarIconButton(
+                key: const ValueKey('cursor-back'),
+                // TODO add translation
+                semanticsLabel: 'Backward',
+                hasTooltip: false,
+                onPressed: canGoBackward
+                    ? () => ref
+                        .read(gameCursorProvider(gameData.id).notifier)
+                        .cursorBackward()
+                    : null,
+                icon: const Icon(LichessIcons.step_backward, size: 20),
+              ),
             ),
-            BottomBarIconButton(
-              key: const ValueKey('cursor-forward'),
-              // TODO add translation
-              semanticsLabel: 'Forward',
-              onPressed: canGoForward
-                  ? () {
-                      ref
-                          .read(gameCursorProvider(gameData.id).notifier)
-                          .cursorForward();
-                    }
+            LongPressButton(
+              callback: canGoForward
+                  ? () => ref
+                      .read(gameCursorProvider(gameData.id).notifier)
+                      .cursorForward()
                   : null,
-              icon: const Icon(LichessIcons.step_forward, size: 20),
+              child: BottomBarIconButton(
+                key: const ValueKey('cursor-forward'),
+                // TODO add translation
+                semanticsLabel: 'Forward',
+                hasTooltip: false,
+                onPressed: canGoForward
+                    ? () {
+                        ref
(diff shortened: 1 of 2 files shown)


lib/src/widgets/buttons.dart
```diff
@@ -392,3 +396,85 @@ class _CardButtonState extends State<CardButton> {
     }
   }
 }
+
+/// Button to add onTap on long pressing
+///
+/// ### Note
+/// Widgets with `tooltip` don't handle onLongPress
+/// Child's onTap will have priority
+class LongPressButton extends StatefulWidget {
```
veloce, line 405: While I think the implementation is correct, the name of the class and the doc are not relevant.

I'd call this button `RepeatButton` because that's what he does, calling repeatedly a function on hold, triggered after a long press.

The doc comment should explain that and the fact the button handles the `onTap` too.

It should mention too that the visual of the button is delegated to the child, this one is just a repeat wrapper.

lib/src/widgets/buttons.dart
```diff
@@ -392,3 +396,85 @@ class _CardButtonState extends State<CardButton> {
...
+  const LongPressButton({
+    required this.callback,
+    required this.child,
+    this.longDelay = const Duration(milliseconds: 600),
```
veloce, line 409: If I got it well the 3 three delays are there to produce an acceleration effect. So instead of `longDelay`, `middleDelay` and `minDelay` names, which are not very clear I think it'd be better to use a list of duration param:
```dart
/// Delays between callbacks at the beginning. Leave default to get an acceleration effect.
final List<Duration> triggerDelays;

/// Delay between callbacks on hold.
final Duration holdDelay
```

lib/src/widgets/buttons.dart
```diff
@@ -392,3 +396,85 @@ class _CardButtonState extends State<CardButton> {
...
+class _LongPressButtonState extends State<LongPressButton> {
+  bool _isPressed = false;
+  Timer? _timer;
+  List<Duration> _delayTime = [];
+
+  @override
+  void initState() {
+    super.initState();
+    _delayTime = [widget.longDelay, widget.middleDelay, widget.minDelay];
+  }
```
veloce, line 445: the `onDispose()` is missing, where `_timer` should be canceled.

lib/src/ui/game/archived_game_screen.dart
```diff
@@ -198,31 +198,45 @@ class _BottomBar extends ConsumerWidget {
...
+                      .read(gameCursorProvider(gameData.id).notifier)
+                      .cursorBackward()
                   : null,
-              icon: const Icon(LichessIcons.step_backward, size: 20),
+              child: BottomBarIconButton(
+                key: const ValueKey('cursor-back'),
+                // TODO add translation
+                semanticsLabel: 'Backward',
+                showTooltip: false,
+                onPressed: canGoBackward
```
veloce, line 212: the `onPressed` callback should be a no op to explicitly show the real callback is the one of the parent widget.

lib/src/widgets/buttons.dart
nav-28: Although making it into a list is great but we don't know the length of the list. Maybe adding an asset such that the length of the list is less than 3 would be ideal.

lib/src/ui/game/archived_game_screen.dart
nav-28: I don't think it can be a no op as the `onTap` is delegate to the child widget so it will do nothing on tap. It can also not be null as it won't be highlighted. I have changed the name of the callback in the `RepeatButton` to be `longPressCallback` and added a bool `handleOnTap` to be more clear. Added doc as well.

lib/src/widgets/buttons.dart
veloce: oh yeah, good catch, but is a variable length really problematic?

or instead of a list a 3-uple (Duration, Duration, Duration) is possible too.

lib/src/ui/game/archived_game_screen.dart
veloce: My understanding was that `HitTestBehavior.opaque` prevented all touch events to reach the child, and the `RepeatButton` would handle the `onTap` itself.

lib/src/widgets/buttons.dart
nav-28: Not really but a long list might be so capping it is a good idea. Also it can be empty if we don't want the acceleration effects. 

I already added an assert, i think that should be enough

lib/src/ui/game/archived_game_screen.dart
nav-28: I think the documentation is a bit weird. Here is an explanation https://github.com/flutter/flutter/issues/18450#issuecomment-397372078

veloce: ok then I suppose we can declare the callback as a private method to avoid duplicate the code

PR #200, opened 2023-06-08
Next puzzle immediately

### LINKED ISSUE #125: Add "Jump to next puzzle immediately" setting
Screenshot_2023-04-27-08-56-32-261-edit_com android chrome
I'd like to suggest a feature that already exists on the Lichess website:

"Jump to the next puzzle immediately"

It's a very simple but very useful setting: if it's enabled, the user gets a new puzzle immediately after solving one correctly. It makes solving puzzles a much faster and smoother experience.

I don't have to knowledge to do it, nor do I know how difficult it is to implement, but in my opinion a lot of users would find it useful. Let me know what you think, if it's possible and if there is someone who could/would implement it. Thanks!

brandonpl8:
Resolves #125. Adds a puzzle settings screen.


brandonpl8:
Setting_screen
Puzzle_setting_screen

veloce:
Closing in favour of #412 as this PR is stale.

PR #241, opened 2023-07-20
Update puzzle screen setting

nav-28:
Adds the setting button to the puzzle screen, streak, and storm to be similar to the game screen.


veloce:
Not sure we should add this to the puzzle screens, I'll have to think about it.

veloce:
Let's not do that for now.

PR #321, opened 2023-09-22
Add chess-opening database and fetch opening in analysis

veloce:


nav-28:
Why not directly create opening files in dart similar to https://github.com/lichess-org/scalachess/blob/master/src/main/scala/opening/OpeningDbPartA.scala instead of created a db?

veloce:
Because I thought it was simpler like that, and we have more power to query with sqlite than with dart classes.

nav-28:
> Because I thought it was simpler like that, and we have more power to query with sqlite than with dart classes.

Right, I didn't think about the impact to memory it might have as well.

Issue #322, opened 2023-09-23
Integrate FVM (Flutter Version Management) for Enhanced Team Collaboration and Compatibility


mucahitbircan:
We propose implementing FVM (Flutter Version Management) to streamline Flutter SDK version management within our project. This will ensure compatibility, stability, and consistency across the team, fostering efficient collaboration and addressing compatibility concerns.

PR #323, opened 2023-09-23
integrate: fvm (flutter version management)

mucahitbircan:
#322

### DIFF
--- a/.fvm/fvm_config.json
+++ b/.fvm/fvm_config.json
@@ -1,0 +1,4 @@
+{
+  "flutterSdkVersion": "stable",
+  "flavors": {}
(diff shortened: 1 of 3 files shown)


mucahitbircan:
> Hi! I didn't know this tool. I think it's good to have it integrated for those who want to use it so thanks for this PR.
> 
> On the other hand I don't want to force people using it (I will not use it myself for now) so let's not replace all the commands in the README. Just include a sub section under `## Setup`, like:
> 
> ```
> ### FVM
> 
> If you want to use FVM, do 
> ...
> ```

Hello, Thanks for your suggestion

Issue #242, opened 2023-07-20
Friends popup


veloce:
Implement friends popup.

![Home friends](https://github.com/lichess-org/mobile/assets/423393/a3e25d25-df31-429a-b739-39ca973c09e4)

This is a full screen popup accessible from the home tab button in the top right.

Online friends are fetched with a websocket API.

Suggestions need a server endpoint that does not exist yet, so it can be done later.

Tapping "More" goes to another popup with all friends. If no friends are online, a "Friends" button should be displayed under the "0 friend online" title.

This endpoint can be used to fetch all friends for the 2nd popup: https://lichess.org/api#tag/Relations/operation/apiUserFollowing

In order to unfollow someone, we'd swipe the tile to the left to uncover the unfollow action, as shown in the wireframe.

More actions will be available later, when available, like "challenge", "PM", etc.

Follow and unfollow endpoints are also documented.

BeardedAmbivert:
I would like to implement this if someone hasn't started working on this. Thanks!

papo1011:
Hi everyone! I'm working on this issue and I'm making good progress.
I started to open draft pr #337 and maybe you want to give some feedback :)

veloce:
Sure, will do asap!

NewSupKaizo55:
Is there going to be a "Challenge friend" button like in the old app? Is a new separate PR for the challenge user function needed?

PR #337, opened 2023-10-10
Feature: friends popup

papo1011:
Friends popup #242 

- [x] Relation Repository (getFollowing, follow, unfollow)
- [x] Test Relation Repository
- [x] Following screen
- [x] Following online socket topic (following_onlines, following_enters, following_leaves)
- [x] Following online screen


lib/src/model/relation/relation_ctrl.dart
```diff
@@ -0,0 +1,87 @@
...
+
+  AuthSocket get _socket => ref.read(authSocketProvider);
+}
+
+@freezed
+class RelationCtrlState with _$RelationCtrlState {
+  const RelationCtrlState._();
+
+  const factory RelationCtrlState({
+    required List<dynamic> followingOnlines,
```
veloce, line 85: this cannot by `dynamic` here and not a `List`, it should be an `IList<LightUser>`

Riverpod state must be immutable, see why here: https://github.com/lichess-org/mobile/wiki/Coding-style#immutability

lib/src/model/relation/relation_ctrl.dart
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  void getFollowingOnlines() {
+    _socket.send('following_onlines', null);
+  }
+
+  void _handleSocketTopic(SocketEvent event) {
+    switch (event.topic) {
+      case 'following_onlines':
+        final data = event.data as List<dynamic>;
```
veloce, line 41: Don't cast as dynamic but as `List<String>` and then parse the string to make a `LightUser`.
Something like that but in dart:
```js
function parseFriend(friend: string) {
  const splitted = friend.split(' ')
  const name = splitted.length > 1 ? splitted[1] : splitted[0]
  const title = splitted.length > 1 ? splitted[0] : undefined
  return [name, title]
}

The id is then just `name.toLowerCase`

File: lib/src/model/relation/relation_ctrl.dart [hunk 3]
```diff
@@ -0,0 +1,87 @@
...
+        if (state is AsyncData) {
```
veloce, line 42: instead of checking the type and cast, it's best to make a single check at the top of the `_handleSocketTopic`:
```dart
void _handleSocketTopic(SocketEvent event) {
  if (!state.hasValue) return;
 
  ...
  // then you can write:
  state = AsyncValue.data(state.requireValue.copyWith(followingOnlines: newData));
  ...
}
```

lib/src/model/relation/relation_repository_providers.dart
```diff
@@ -0,0 +1,25 @@
...
+@Riverpod(keepAlive: true)
+RelationRepository relationRepository(RelationRepositoryRef ref) {
+  final apiClient = ref.watch(authClientProvider);
+  return RelationRepository(
+    logger: Logger('RelationRepository'),
+    apiClient: apiClient,
+  );
+}
+
+@riverpod
```
veloce, line 20: we should cache this with `@Riverpod(keepAlive: true)`

lib/src/view/relation/following_screen.dart
```diff
@@ -0,0 +1,214 @@
...
+        middle: const Text("Following"),
+      ),
+      child: const _Body(),
+    );
+  }
+
+  Widget _buildAndroid(BuildContext context) {
+    return Scaffold(
+      appBar: AppBar(
+        title: Text(context.l10n.leaderboard),
```
veloce, line 39: wrong title

lib/src/view/relation/following_screen.dart
```diff
@@ -0,0 +1,214 @@
...
+
+  @override
+  Widget build(BuildContext context) {
+    return PlatformWidget(androidBuilder: _buildAndroid, iosBuilder: _buildIos);
+  }
+
+  Widget _buildIos(BuildContext context) {
+    return CupertinoPageScaffold(
+      navigationBar: CupertinoNavigationBar(
+        previousPageTitle: context.l10n.friends,
```
veloce, line 29: normally you don't have to put the `previousPageTitle` and `middle` params, flutter handle these automatically

lib/src/view/relation/following_screen.dart
```diff
@@ -0,0 +1,214 @@
...
+        middle: const Text("Following"),
```
veloce, line 30: there is a translation for `following`

lib/src/view/relation/following_screen.dart
```diff
@@ -0,0 +1,214 @@
...
+                                        if (res.isError) {
+                                          setState(() {
+                                            followingUsers = oldState;
+                                          });
+                                        }
+                                      },
+                                    ),
+                                  );
+                                },
+                                backgroundColor: const Color(0xFFFE4A49),
```
veloce, line 105: We have a `LichessColors.red` for that

lib/src/view/relation/relation_screen.dart
```diff
@@ -0,0 +1,132 @@
...
+class _OnlineFriendsWidget extends ConsumerWidget {
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {
+    final relationState = ref.watch(relationCtrlProvider);
+
+    return relationState.when(
+      data: (data) {
+        return ListSection(
+          header: Text(
+            '${data.followingOnlines.length} online friends', // TODO: we need good translations for this
```
veloce, line 71: you have the `nbFriendsOnline` which takes an int as parameter

lib/src/view/relation/relation_screen.dart
```diff
@@ -0,0 +1,132 @@
...
+          ),
+        ),
+      ),
+    );
+  }
+
+  void _handleTap(BuildContext context) {
+    pushPlatformRoute(
+      context,
+      title: "Following",
```
veloce, line 128: translation

lib/src/view/relation/relation_screen.dart
```diff
@@ -0,0 +1,132 @@
...
+                  onPressed: () => _handleTap(context),
+                  child: Text(
+                    context.l10n.more,
+                  ),
+                ),
+          children: [
+            if (data.followingOnlines.isEmpty)
+              PlatformListTile(
+                title: const Text(
+                  "Following",
```
veloce, line 85: translation

veloce (commented):
Many thanks for your work, it already looks really good!

lib/src/model/relation/relation_repository_providers.dart
papo1011: with @Riverpod(keepAlive: true) I have this behavior, steps to reproduce:

- go to following screen
- unfollow user
- go back to friends screen
- go to following screen

it serves old results, not updated with unfollow request at step 2
because cached with @Riverpod(keepAlive: true)

veloce: yes I see... the unfollow action should also invalidate the cache with `ref.invalidate`

papo1011: Now I've added @Riverpod(keepAlive: true) and ref.invalidate, but currently, the UI list updates only after the unfollow request is completed

papo1011: Now I've added `@Riverpod(keepAlive: true)` and `ref.invalidate` :)

veloce:
@papo1011 I'm going to review the PR. Could you fix the conflict please?

lib/src/view/relation/following_screen.dart
```diff
@@ -0,0 +1,372 @@
...
+  Widget _buildAndroid(BuildContext context) {
+    return Scaffold(
+      appBar: AppBar(
+        title: Text(context.l10n.following),
+      ),
+      body: const _Body(),
+    );
+  }
+
+  void _invalidateFollowingCache() {
```
veloce, line 92: looks like a bit complex logic here: we should probably not cache the data at all to avoid such code

veloce: the added benefit of caching is not worth the complexity imo

lib/src/view/relation/following_screen.dart
```diff
@@ -0,0 +1,372 @@
...
+}
+
+class _Body extends ConsumerWidget {
+  const _Body();
+
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {
+    final following = ref.watch(followingProvider);
+    final relationState = ref.watch(relationCtrlProvider);
+    final followingOnlines = relationState.when(
```
veloce, line 109: To avoid having 2 `.when()`, which creates branching complexity, it is usually better to combine the async data upstream in a single provider.
As done here for instance: https://github.com/lichess-org/mobile/blob/38fa6bc675f0c580eeced938ff7166e07a101ed1/lib/src/view/puzzle/puzzle_themes_screen.dart#L26

lib/src/view/relation/following_screen.dart
```diff
@@ -0,0 +1,372 @@
...
+                                        final res = await ref
+                                            .read(relationRepositoryProvider)
+                                            .unfollow(user.username);
+                                        if (res.isError) {
+                                          setState(() {
+                                            followingUsers = oldState;
+                                          });
+                                        } else {
+                                          ref
+                                              .read(
```
veloce, line 170: Same as other comment, the whole caching logic seems too complex here. Normally just invalidating the `followingProvider` should be enough to trigger a screen refresh.

Perhaps you could take the account preferences notifier as an example: https://github.com/lichess-org/mobile/blob/38fa6bc675f0c580eeced938ff7166e07a101ed1/lib/src/model/account/account_preferences.dart#L55

It works the same way: the initial data is fetched async from the server, and methods that update the state trigger an `invalidateSelf` to force fetching new data: https://github.com/lichess-org/mobile/blob/38fa6bc675f0c580eeced938ff7166e07a101ed1/lib/src/model/account/account_preferences.dart#L90

veloce (commented):
Except for this caching mechanism that need to be removed it looks really good! thanks!

lib/src/view/relation/following_screen.dart
papo1011: Yes, I understand.
But if we add await `Future.delayed(const Duration(seconds: 1));` before line 89 in the account_preferences, simulating a slow POST request, the delay will affect the toggle component, causing it to wait for 1 second before switching. The same behavior would occur when removing a user tile on the following screen.

veloce: Yes it's not perfect, I agree, but in the vast majority of cases it will work smoothly. 

UX can be improved I guess: set an opacity, or any other way to indicate something is loading. It will have the big advantage to be simple, compared to complex caching.

veloce:
Too bad the diff includes already merged commits... dunno how to fix that.

papo1011:
> Too bad the diff includes already merged commits... dunno how to fix that.

Here [https://github.com/lichess-org/mobile/compare/main...papo1011:mobile:feature/friends-popup-242](https://github.com/lichess-org/mobile/compare/main...papo1011:mobile:feature/friends-popup-242) the three dots diff shows the difference between the two branches, displaying the correct commits.

Issue #279, opened 2023-08-13
In game chat


veloce:
![Game chat](https://github.com/lichess-org/mobile/assets/423393/927f6eef-90f1-45d8-a442-18badf800527)

TODO:

- [ ] show chat in archived game screen
- [ ] consider a settings to disable chat by default

julien4215:
#339

kkorus:
I would like to consider giving option to disable in game chat for every game so it would be applied for each live game, thanks!

nav-28:
@veloce this can be closed now right?

veloce:
I'm waiting to add chat in the archived game screen before closing the issue. And I needed some server update before doing it.

PR #339, opened 2023-10-13
In-game chat feature

julien4215:
In-game chat feature #279 
- [x] Create the in-game chat screen
- [x] Add message icon to the game screen
- [x] Add server message (like "Draw offer")
- [x] Improve the color of bubbles messages
- [x] Create iOS builder
- [x] Edit game controller


lib/src/model/game/chat_controller.dart
```diff
@@ -0,0 +1,118 @@
...
+
+  int? _socketEventVersion;
+
+  AuthSocket get _socket => ref.read(authSocketProvider);
+
+  @override
+  Future<ChatState> build() {
+    _socketEventVersion = 0;
+    final stream = _socket.stream;
+    stream.listen(_handleSocketTopic);
```
veloce, line 25: you should assign the `_socketSubscription` here

lib/src/model/game/chat_controller.dart
```diff
@@ -0,0 +1,118 @@
...
+
+    ref.onDispose(() {
+      _socketSubscription?.cancel();
+    });
+
+    return stream.firstWhere((e) => e.topic == 'full').then(
```
veloce, line 31: in order to be generic, the chat controller shouldn't listen to the `full` event which is only related to "game" socket.
It should be easy to replace this by a setter:
```
void setMessages(IList<Message> messages) {
   state = ChatState(
     messages: messages,
     unreadMessages: 0,
   );
}
```

the `build` method does not return a Future anymore, it will even simplify things

lib/src/model/game/chat_controller.dart
```diff
@@ -0,0 +1,118 @@
...
+    final curState = state.requireValue;
+    state = AsyncValue.data(curState.copyWith(unreadMessages: 0));
+  }
+
+  void _handleSocketTopic(SocketEvent event) {
+    if (!state.hasValue) return;
+
+    switch (event.topic) {
+      // Full game data, received after a (re)connection to game socket
+      case 'full':
```
veloce, line 63: same remark, don't listen to `full` but use `setMessages`

lib/src/view/game/game_body.dart
```diff
@@ -78,6 +80,7 @@ class GameBody extends ConsumerWidget {
   @override
   Widget build(BuildContext context, WidgetRef ref) {
     final ctrlProvider = gameControllerProvider(id);
+    ref.watch(chatControllerProvider);
```
veloce, line 83: I assume you needed an eager initialisation of the provider to ensure it could fetch the `full` event, this should not be needed anymore if we refactor as suggested above comment.

lib/src/view/game/game_body.dart
```diff
@@ -523,9 +526,59 @@ class _GameBottomBar extends ConsumerWidget {
...
+                    pushPlatformRoute(
+                      context,
+                      builder: (BuildContext context) {
+                        return MessageScreen(game: gameState.game);
+                      },
+                    );
+                  },
+                  icon: CupertinoIcons.chat_bubble,
+                ),
+                ref.watch(chatControllerProvider).when(
```
veloce, line 544: I prefer to have all `ref.watch` declarations at the beginning of the `build()` method so you know at first glance what widget listen to what.

lib/src/view/game/message_screen.dart
```diff
@@ -0,0 +1,292 @@
...
+import 'package:lichess_mobile/src/navigation.dart';
+import 'package:lichess_mobile/src/styles/lichess_colors.dart';
+import 'package:lichess_mobile/src/styles/styles.dart';
+import 'package:lichess_mobile/src/utils/l10n_context.dart';
+import 'package:lichess_mobile/src/widgets/buttons.dart';
+import 'package:lichess_mobile/src/widgets/feedback.dart';
+import 'package:lichess_mobile/src/widgets/platform.dart';
+
+class MessageScreen extends ConsumerStatefulWidget {
+  final PlayableGame game;
```
veloce, line 17: this could be refactored to a more generic screen by removing the `PlayableGame` param and replace it with:

```dart
required Widget title,
User? me,
```

should be all what is needed here

lib/src/view/game/message_screen.dart
```diff
@@ -0,0 +1,292 @@
...
+      ),
+      margin: const EdgeInsets.all(10.0),
+      padding: Styles.horizontalBodyPadding,
+      child: SafeArea(
+        top: false,
+        child: Row(
+          mainAxisAlignment: MainAxisAlignment.spaceAround,
+          children: [
+            Expanded(
+              child: TextField(
```
veloce, line 260: There's an `AdaptiveTextField` which chooses the UI according to platform.

veloce (commented):
This is good! thank you for your work, I think it's almost done as the comments I've made only require easy refactoring, I think :)

lib/src/model/game/game_controller.dart
```diff
@@ -431,6 +438,10 @@ class GameController extends _$GameController {
         _socketEventVersion = fullEvent.socketEventVersion;
         _lastMoveTime = null;
 
+        ref
```
veloce, line 441: here it should be `ref.read`, as per riverpod documentation. We only want to watch in widget `build()` methods or in other providers because we want them to react to changes.

lib/src/view/game/message_screen.dart
```diff
@@ -231,13 +237,15 @@ class _ChatBottomBarState extends ConsumerState<_ChatBottomBar> {
...
           cupertinoDecoration: BoxDecoration(
-            border: Border.all(color: Theme.of(context).colorScheme.outline),
+            border: Border.all(
+              color: CupertinoColors.separator.resolveFrom(context),
+            ),
             borderRadius: borderRadius,
           ),
           controller: _textController,
-          keyboardType: TextInputType.text,
+          keyboardType: TextInputType.multiline,
```
julien4215, line 246: I don't think the user should be able to make new lines as lila (or the website frontend) dosen't allow it.

lib/src/view/game/game_body.dart
```diff
@@ -549,22 +551,25 @@ class _GameBottomBar extends ConsumerWidget {
...
+                          size: 20.0,
                         ),
-                        Text(
-                          chatState.unreadMessages.toString(),
-                          textAlign: TextAlign.center,
-                          style: const TextStyle(
-                            color: Colors.white,
-                            fontSize: 8.0,
-                            fontWeight: FontWeight.w500,
+                        SizedBox(
```
julien4215, line 563: I don't think this SizedBox is useful

lib/src/view/game/message_screen.dart
veloce: yes indeed; in that case `maxLines` should be 1 too

julien4215: I am not sure `maxLines` should be 1. On the one hand, if `maxLines` is greater than one, the user can fully see his message before sending it. On the other hand, it can be confusing for the user. But the return key is not avaible (at least on Android) for `TextInputType.text` so that should help the user to understand that his message is only one line.

Also, it is the website frontend which applies a treatment to text (I think it reduces white spaces to 1 and remove new lines).

veloce: yes it is a compromise and we should take the approach that is the less confusing for the user; still not sure what is the best actually;

on the website it will remove the new lines, but on mobile it will show them; why not let the user send multiline text on mobile after all?

julien4215: > on the website it will remove the new lines, but on mobile it will show them if we want to; why not let the user send multiline text on mobile after all?

Because if a user sends a message with newlines, it will weird for the receiver on the website as newlines will be removed.
I guess the best solution would be to remove the treatment of text on the website but they surely did this for a reason.

veloce: I'm pretty sure it's only to keep things simple because you hit <enter> to send the message.
Anyway, even though I'd like to make the best out of the mobile app, we're not a chat application and having a forced 1 line text is already good for a start. 
I'll ask the server team what they think about multiline text on the website.

PR #402, opened 2023-12-21
Refine default number of cores for smoother animations

HaonRekcef:
More conservative value for the default number of cores used in the analysis board. In my opinion the improved animations and the reduced lag when entering moves and opening settings outweigh the perfomance gain of the chess engine. 
On my Pixel 6a the lag was non-neglible when using maxEngineCores and I presume on older devices it might be even worse. 
In my opinion the default value could be even lower.


veloce:
Engine eval start is debounced so it is supposed not to hinder animations. I don't have problem with my devices at least.

HaonRekcef:
Indeed this seems to be an issue of the debug build, in the release version it is fine

PR #412, opened 2023-12-29
New feature: Jump to next puzzle immediately

kpsroka:
https://github.com/lichess-org/mobile/assets/24893014/04fa1a7c-99cd-4f9d-894b-e802ef840dea

https://github.com/lichess-org/mobile/assets/24893014/7af2f017-09b3-4312-b968-e52d4e2cb448

@veloce Let me know if this is the right approach. I'll then go and implement tests and android-specific version.


veloce:
Looks good to me. Settings screen is not the same as the game one. It should be eventually consistent but let's keep it like you did for now because I'm not satisfied with what I did for game or analysis screen.
I'll create a separate issue to make settings screen consistent across app.

What would be specific on android here?

lib/src/model/puzzle/puzzle_preferences.dart
```diff
@@ -37,8 +44,19 @@ class PuzzlePrefState with _$PuzzlePrefState {
   const factory PuzzlePrefState({
     required UserId? id,
     required PuzzleDifficulty difficulty,
+
+    /// If `true`, will show next puzzle after successful completion. This has
+    /// no effect on puzzle streaks, which always show next puzzle. Defaults to
+    /// `false`.
+    required bool autoNext,
```
veloce, line 51: Since the field is `required` the `fromJson` method would fail with previously saved data from the current version of the app. 
Making it non required would prevent any issue.

kpsroka: Removed `required`, added `@Default(false)` as required by freezed.

kpsroka:
> What would be specific on android here?

Added non-cupertino scaffold. I need to check on Android that this works OK, and will mark it as ready for review.

veloce:
@kpsroka is this PR ready?

kpsroka:
> @kpsroka is this PR ready?

I need to verify this works properly, I'll do that over the next 2 days.

kpsroka:
> @kpsroka is this PR ready?

Ready now.

veloce (commented):
Nice job! Just a couple of improvements and we're good.

PR #419, opened 2023-12-30
Home search

### LINKED ISSUE #246: Home search
Home search

Search interface to globally search on lichess. Similar to the search function of the GitHub mobile application.

For now it will only search players, but as the application will evolve more searching possibilities will be added, like teams, tournaments, studies, etc.

Tapping the search bar on the Home Screen will transition to the search screen with focus already on search bar. Ideally the transition is made with a [Hero animation](https://docs.flutter.dev/ui/animations/hero-animations).

Returned players from the search are displayed with their online status (circle or patron icon).

The autocomplete endpoint is documented here:
https://lichess.org/api#tag/Users/operation/apiPlayerAutocomplete

nav-28:
Closes #246


veloce:
I tested it and it is working well.

Before merging it, we'd need the missing header (as shown in the issue): "Players with '<searched terms>'". We need to show it because the plan is to make a global search later, when the app evolves. Eventually we should be able to search for studies, tournaments, teams, etc.

PR #420, opened 2023-12-31
Add issue template

nav-28:


.github/ISSUE_TEMPLATE/other.md
veloce, line 1: What is the use of this file?

nav-28: I copied all the files from Lila. It can be used as a template for feature requests but I don't know how to organise it so it's just a blank template for now.

veloce: Ok, I'm merging it, we'll improve later when needed.

PR #446, opened 2024-01-06
Improve game summary layout

### LINKED ISSUE #436: Improve game summary tablet layout
When we have enough space, players stats should be disposed side by side instead of stacked like in this screenshot:

Screenshot_20240104_000628

ZTL-UwU:
Closes #436 
<p>

Screenshot_2024-01-06-21-08-51-003_org lichess mo

</p>
</details> 

It will show the sumaries side by side if the screen width is over 350.


nav-28:
Look good. One improvement I would suggest is that we should make it into a table with the numbers on each side. 

| User1  |              | User2  | 
|-----|------------------------|----|
| 1   | Inaccuracy             | 4   | 
| 21  | Mistakes               | 11  | 
| 11  | Blunders               | 10  | 
| 163 | Average Centipawn Loss | 163 | 
| 42% | Accuracy | 52%|


I believe this can be done with flutter's `Table` widget

ZTL-UwU:
I've made it into a table and it looks really good. I've also added a side indicator. @nav-28 

<p>


Screenshot_2024-01-07-14-13-03-127_org lichess mo (2)
Screenshot_2024-01-07-14-12-29-467_org lichess mo (1)


</p>
</details>

lib/src/view/analysis/analysis_screen.dart
```diff
@@ -835,27 +834,123 @@ class ServerAnalysisSummary extends ConsumerWidget {
...
+                    (
+                      serverAnalysis.white.inaccuracies.toString(),
+                      'Mistakes',
+                      serverAnalysis.black.inaccuracies.toString()
+                    ),
+                    (
+                      serverAnalysis.white.inaccuracies.toString(),
+                      'Blunders',
+                      serverAnalysis.black.inaccuracies.toString()
+                    ),
```
ZTL-UwU, line 876: We need new translations here, though I have no idea how to add them.

HaonRekcef: I think we already have them, they should be:

```
Text(context.l10n.nbInaccuracies(data.inaccuracies))
Text(context.l10n.nbMistakes(data.mistakes))
Text(context.l10n.nbBlunders(data.blunders))
```

nav-28: This is fine for now.  It can be added later.

veloce: One trick could be to use the translation with the number and then do `translatedString.replaceAll('\d\', '').trim()`

veloce:
It is certainly a better layout on a phone, but can you make sure it is also good on a tablet (that it doesn't leave too many empty space)?

ZTL-UwU:
> It is certainly a better layout on a phone, but can you make sure it is also good on a tablet (that it doesn't leave too many empty space)?

I've set the max width to 500. I guess this looks good enough? @veloce 

Screenshot_1704954980

veloce:
In terms of spacing it looks fine. 

Text wrap of the color dot before the player name is not looking good. If would be better if always above and centred. It could even be replaced by a chess king of the right color (like in puzzle screen).

ZTL-UwU:
Screenshot_1704965110

ZTL-UwU:
> looks really good thank you for your work, we from the discord have a few suggestions: The result and the ratings should be displayed somewhere. Average at the top. Inaccuracies, Mistakes, Blunders in middle. ( with the equal top-bottom padding to the Accuracy and Avg. Centipawn losses ) Average centipawn losses at the bottom. The center the strings to make it look more balanced. an example image (ignore the profile pic): image

Thanks for the suggestions! I've made them happen except for the results and ratings. They allready exist in its parent page (which is the archieved game screen) so i dont think they should be in analysis screen. More over, the data are not accessible without huge changes to the code.

<p>


<br>


</p>
</details>

HaonRekcef:
I looks really good, thanks for the work!
The avg centi pawn loss should be centered too.

ZTL-UwU:
> I looks really good, thanks for the work! The second line of avg centi pawn loss should be centered too.

Fixed now

lib/src/view/analysis/analysis_screen.dart
```diff
@@ -834,27 +832,132 @@ class ServerAnalysisSummary extends ConsumerWidget {
...
           AcplChart(options),
-          if (serverAnalysis != null) ...[
-            const SizedBox(height: 16.0),
-            _PlayerStats(Side.white, serverAnalysis.white, pgnHeaders),
-            _PlayerStats(Side.black, serverAnalysis.black, pgnHeaders),
-          ],
+          if (serverAnalysis != null)
+            Center(
+              child: SizedBox(
+                width: math.min(MediaQuery.of(context).size.width, 500),
```
veloce, line 844:
```suggestion
                width: math.min(MediaQuery.sizeOf(context).width, 500),
```

lib/src/view/analysis/analysis_screen.dart
```diff
@@ -834,27 +832,132 @@ class ServerAnalysisSummary extends ConsumerWidget {
...
+                    children: [
+                      TableRow(
+                        decoration: const BoxDecoration(
+                          border: Border(
+                            bottom: BorderSide(color: Colors.grey),
+                          ),
+                        ),
+                        children: [
+                          _SummaryPlayerName(Side.white, pgnHeaders),
+                          const SizedBox(width: 0),
```
veloce, line 863:
```suggestion
                          const SizedBox.shrink(),
```

veloce (commented):
Thanks for your work. I suggest just a couple of minor fixes.

PR #449, opened 2024-01-08
Profile Bio Enhancement

nav-28:
Some user have links in their bio and we should make them clickable. 

<p>

screen
</p>
</details> 

Usernames will open a user screen if the user exists. Other urls will launch a window.


veloce:
Thanks for this. I'm not very fond of the underline style though. It would be better to leave just the blue color without the underline.

lib/src/view/user/user_profile.dart
```diff
@@ -47,10 +55,38 @@ class UserProfile extends StatelessWidget {
...
-            Text(
-              user.profile!.bio!,
+            Linkify(
+              onOpen: (link) async {
+                if (link.originText.startsWith('@')) {
+                  final userId =
+                      UserId.fromUserName(link.originText.substring(1));
+                  try {
+                    final user =
+                        await ref.read(UserProvider(id: userId).future);
```
veloce, line 65: weird syntax here: a provider shouldn't start with a majuscule

lib/src/view/user/user_profile.dart
```diff
@@ -47,10 +55,38 @@ class UserProfile extends StatelessWidget {
...
+                }
+              },
+              linkifiers: const [
+                UrlLinkifier(),
+                EmailLinkifier(),
+                UserTagLinkifier(),
+              ],
+              text: user.profile!.bio!,
               maxLines: bioMaxLines,
-              overflow: TextOverflow.ellipsis,
```
veloce, line 53: why removing it? should be kept imo.

lib/src/view/user/user_screen.dart
```diff
@@ -13,9 +13,18 @@ import 'user_activity.dart';
 import 'user_profile.dart';
 
 class UserScreen extends ConsumerWidget {
-  const UserScreen({required this.user, super.key});
+  const UserScreen._({this.lightUser, this.user, super.key});
```
veloce, line 16: since one of them is required (but not both) we should have an `assert` to enforce it.

pubspec.yaml
```diff
@@ -41,6 +42,7 @@ dependencies:
   http: ^1.1.0
   intl: ^0.18.0
   json_annotation: ^4.7.0
+  linkify: ^5.0.0
```
veloce, line 45: Why do we need this package since `flutter_linkify` is already handling the job alone it seems?

nav-28: This is cause `flutter_linkify` is depended on this library but it's not showing `UserTagLinkifier` which is needed to link user tags.
[repo link](https://github.com/Cretezy/flutter_linkify/blob/b43e490da124721d003f5f95f000372548db27e3/lib/flutter_linkify.dart#L5)

One way to solve that would be to create our own `Linkify` widget by copying it from `flutter_linkify`

lib/src/view/user/user_profile.dart
```diff
@@ -47,11 +56,45 @@ class UserProfile extends StatelessWidget {
               child: userFullName,
             ),
           if (user.profile?.bio != null)
-            Text(
-              user.profile!.bio!,
+            Linkify(
+              onOpen: (link) async {
+                if (link.originText.startsWith('@')) {
+                  final userId =
```
veloce, line 62: It is better to immediately go to the user screen with the id, and load the user while in the screen.

lib/src/view/user/user_screen.dart
```diff
@@ -13,9 +13,23 @@ import 'user_activity.dart';
...
-  const UserScreen({required this.user, super.key});
+  const UserScreen._({this.lightUser, this.user, super.key})
+      : assert(
+          (lightUser != null || user != null) &&
+              !(lightUser != null && user != null),
+          'Either lightUser or User should be provided but not both',
+        );
 
-  final LightUser user;
+  const factory UserScreen.fromUser({required User user, Key? key}) =
```
veloce, line 23: I don't understand this change actually. 

Why adding more complexity? We should always load the user from ID. 

It looks like this change was made to handle the case where the user is not found in a profile link. But it is better to handle this case in the async error handler and show a "User not found" here instead.

veloce (changes requested):
The changes of the UserScreen should be reverted before merging it. 
- go back to only `LightUser` param
- handle user not found in the error handler
- don't try to fetch user before going to the screen in `UserProfile`

lib/src/view/user/user_profile.dart
```diff
@@ -47,11 +54,35 @@ class UserProfile extends StatelessWidget {
...
+                      UserId.fromUserName(link.originText.substring(1));
+                  pushPlatformRoute(
+                    context,
+                    builder: (ctx) => UserScreen(
+                      user: LightUser(id: userId, name: userId.value),
```
veloce, line 65: I'd do instead:

```dart
final username = link.originText.substring(1);
final user = LightUser(id: UserId.fromUserName(username), name: username);
pushPlatformRoute(
  ...
```

veloce (approved):

PR #476, opened 2024-01-12
Tv enhancement closes #325

### LINKED ISSUE #325: Add "Flip board" to Lichess TV
I don't want to flood the issues tab with my feature requests, just a little enhancement that popped into my mind:

Sometimes it's good to see a game from another perspective.

I can see it's a bit more work as the clocks and player names has to be switched when the "Flip board" button is pushed. This was done very smoothly in old Lichobile. 

I think the layout of the bottom icons could be copied from the old app as well (there are no navigation buttons on the TV screen yet):
from left to right - flip board, one move backward, one move forward. 

Let me know if it's a reasonable idea or not. Thanks as always!

HaonRekcef:
Adds flip board, one move backward and one move forward button to the tv screen, closes #325 
 
I am new to this project and flutter and not sure if I did things correctly. As far as I could test the buttons work as intended, however my code to check if it is possible to go back or forward is probably is not clean. I am not sure how I should rewrite it.

Also I do not like the flip icon, I liked the lichobile flip Icon more.
<details>
<summary>Screenshot</summary>

Screenshot_20240112-164913

</details>

### DIFF
--- a/lib/src/model/tv/tv_controller.dart
+++ b/lib/src/model/tv/tv_controller.dart
@@ -100,6 +100,22 @@ class TvController extends _$TvController {
     state = AsyncValue.data(newState);
   }
 
+  bool canGoBack() {
+    if (state.hasValue) {
+      final curState = state.requireValue;
+      return curState.stepCursor > 0;
+    }
+    return false;
+  }
+
+  bool canGoForward() {
+    if (state.hasValue) {
+      final curState = state.requireValue;
+      return curState.stepCursor < curState.game.steps.length - 1;
+    }
+    return false;
+  }
+
   void toggleBoard() {
     if (state.hasValue) {
       final curState = state.requireValue;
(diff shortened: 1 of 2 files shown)


lib/src/model/tv/tv_controller.dart
```diff
@@ -100,6 +100,70 @@ class TvController extends _$TvController {
     state = AsyncValue.data(newState);
   }
 
+  bool canGoBack() {
```
veloce, line 103: This code is fine and perfectly readable in my opinion, even if, I agree the `if (state.hasValue` followed by `state.requireValue` is a bit verbose and redundant.

If you're looking for alternate way, I think you could also write it like:
```dart
bool canGoBack() => state.map(
  data: (d) => d.stepCursor > 0,
  error: (_) => false,
  loading: (_) => false,
);
```

or even
```dart
bool canGoBack() => state.mapOrNull(data: (d) => d.stepCursor > 0) ?? false
```

or maybe there is even a better way? The `AsyncValue` api is powerful but also complex.

HaonRekcef: yeah I like it better now

veloce (approved):

PR #482, opened 2024-01-17
Add color coding to analysis tree view

### LINKED ISSUE #479: Add color coding to analysis moves/text
Everywhere on Lichess (and in other chess websites/software too) the text of the evaluated moves are color coded, just like the NAG icons. 

This helps finding the key moments of the game a lot. I mean, having the analysis text in a separate line is great and I love that the current change in the eval number is shown, but I'm sure a lot of people are used to having this very useful feature.

In my opinion even the eval text (for example "Blunder. ... was best") could be colored as well, making the important information even more visible and easier to notice.

Just in case I'll attach a screenshot from old Lichobile. Thanks again as always!

ZTL-UwU:
closes #479 

<p>

</p>
</details>


NewSupKaizo55:
It looks great! Now that I'm looking at the details trying to understand what changed, I noticed that in old Lichobile the moves are in bold and the move numbers are not. Do you think it'd be worth it to add that as well?

ZTL-UwU:
I've made all the main line moves bold, and it does look much better and it gives more clarity to the actual move in the game. @NewSupKaizo55 

<p>



</p>
</details>

NewSupKaizo55:
Looks really good! Thanks for your work!

lib/src/view/analysis/tree_view.dart
```diff
@@ -240,14 +241,26 @@ class _InlineTreeViewState extends ConsumerState<AnalysisTreeView> {
...
+  final defaultColor = defaultTargetPlatform == TargetPlatform.android
       ? Theme.of(context).textTheme.bodyLarge?.color?.withOpacity(opacity)
       : CupertinoTheme.of(context)
           .textTheme
           .textStyle
           .color
           ?.withOpacity(opacity);
+
+  return nag != null
+      ? [
```
veloce, line 254: I'd extract that in a `const`.

veloce (approved):

veloce:
Thanks!

PR #483, opened 2024-01-17
Clickable AcplChart 

### LINKED ISSUE #471: Improve navigation in game analysis
One should be able to click on the `AcplChart` to jump to the move.

HaonRekcef:
Fixes #471

I've modified the graph such that it is clickable. It solely depends on the x-value of the touch input. Maybe there is more sophisticated method that involves the distance on the y -axis that works better. 
However I observed that the website implementation also considers only the x-value.

Also I changed the graph type to be non-curved. The current curved representation sometimes leads to the maximum not being on a move itself.

To illustrate the difference, please refer to the following videos:
<details>
<summary>isCurved: true</summary>

https://github.com/lichess-org/mobile/assets/78898963/de35e001-213c-4b22-b543-cf5fd30b8361
</details>
<details>
<summary>isCurved: false</summary>

https://github.com/lichess-org/mobile/assets/78898963/0fc200ca-3e25-40f1-ad83-c1a9acd154b6
</details>

While implementing these changes, I noticed a small "bug." If one promotes a line other than the game to the mainline, the graph still jumps around when pressing it. I'm uncertain about what the intended behavior in this scenario should be. On the website the same "bug" exists. Probably we can just ignore that.

### DIFF
--- a/lib/src/model/analysis/analysis_controller.dart
+++ b/lib/src/model/analysis/analysis_controller.dart
@@ -184,6 +184,31 @@ class AnalysisController extends _$AnalysisController {
     );
   }
 
+  void jumpToNthNode(int n) {
+    UciPath path = _root.mainlinePath;
+    while (!path.penultimate.isEmpty) {
+      path = path.penultimate;
+    }
+    Node? node = _root.nodeAt(path);
+    int count = 0;
+
+    while (node != null && count < n) {
+      if (node.children.isNotEmpty) {
+        path = path + node.children.first.id;
+        node = _root.nodeAt(path);
+        count++;
+      } else {
+        break;
+      }
+    }
+
+    if (node != null) {
+      _setPath(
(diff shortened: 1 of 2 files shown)


ZTL-UwU:
For longer games the curve might be fine. Should we set a threshold between the curve and no curve or just keep everything not curved?

HaonRekcef:
I also like the curved one more, it looks  bit more forgiving.

<details>
<summary>curved</summary>

curved
</details>
<details>
<summary>not curved</summary>

sharp
</details>

But I do not like that it always goes up before going down, look at the end of the game.

veloce:
If we make it clickable it has to be non curved I guess.

For the graph position indicator, it is easy to hide it if we're not on the mainline, but yeah it's a detail.

veloce (approved):
Thanks!

PR #523, opened 2024-02-03
Replace defaultTargetPlatform with Theme.of(context).platform

### LINKED ISSUE #470: Replace `defaultTargetPlatform` with `Theme.of(context).platform`
According to the official doc: https://api.flutter.dev/flutter/foundation/defaultTargetPlatform.html

`Theme.of(context).platform` is the preferred choice to get the current platform inside a Widget.

- [ ] Replace all `defaultTargetPlatform` where it is possible in the code
- [ ] make sure the tests are passing

BAW2501:
Closes #470 
Note that in  lib/src/utils/layout.dart i changed 
`final appBarHeight = Theme.of(context).platform == TargetPlatform.iOS ? 44.0 : 56.0;`
to inside `double estimateRemainingHeightLeftBoard(BuildContext context)`
to use the `context `for `Theme.of(context).platform`

### DIFF
--- a/lib/src/utils/immersive_mode.dart
+++ b/lib/src/utils/immersive_mode.dart
@@ -1,7 +1,6 @@
 import 'package:device_info_plus/device_info_plus.dart';
-import 'package:flutter/foundation.dart';
+import 'package:flutter/material.dart';
 import 'package:flutter/services.dart';
-import 'package:flutter/widgets.dart';
 import 'package:wakelock_plus/wakelock_plus.dart';
 
 final _deviceInfoPlugin = DeviceInfoPlugin();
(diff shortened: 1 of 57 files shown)


lib/src/utils/immersive_mode.dart
```diff
@@ -43,7 +42,7 @@ mixin ImmersiveMode<T extends StatefulWidget> on State<T>
   }
 
   Future<void> _setImmersiveMode() async {
-    if (defaultTargetPlatform == TargetPlatform.android) {
+    if (Theme.of(context).platform == TargetPlatform.android) {
```
veloce, line 45: In that context I'd not use the `Theme.of(context).platform` because we're about to make a system call which is only in android. This is not for styling purposes.
Here the docs says is ok to even use `dart:io.Platform` but I guess we can also use `defaultTargetPlatform`.

BAW2501: nice catch! will revert in latest commit

veloce (approved):

PR #524, opened 2024-02-04
Add clock tool

KarnerTh:
I really appreciate the new app and all the work so far, so I would like to contribute to that. I find the clock tool feature from the old app quite useful, so I decided to migrate it to the new app. I tried to reuse as much as possible from the existing codebase and use the same code styles etc. - but I am open for feedback if anything needs to change.

I would love to see the feature in the new app and I hope that other people also find this helpful :) 

Features: 
- select time and increment
- pause/resume
- reset clock
- show move count

## Open todos/questions

- [x] Test on iOS (I do not have the hardware to test it by myself)
- [x] The clock screen does currently not keep the screen active without user interaction (which could be a problem on long moves when the device has a short screen on setting) - there are plugins to achieve that (e.g. https://pub.dev/packages/wakelock) but I did not want to add any dependencies without asking first. (can be added later if that's desired)

## Screens
<details><summary>Clock screen</summary>
<p>

screen

</p>
</details>

<details><summary>Select time options</summary>
<p>

time

</p>
</details> 

<details><summary>Time is up</summary>
<p>

finished

</p>
</details>


nav-28:
The app already uses a wake lock. Look at the immersive mode file. You can use that. https://github.com/lichess-org/mobile/blob/76420d63d2763c964139f04997fc3709bebfef02/lib/src/utils/immersive_mode.dart#L5

KarnerTh:
> The app already uses a wake lock. Look at the immersive mode file. You can use that.
> 
> https://github.com/lichess-org/mobile/blob/76420d63d2763c964139f04997fc3709bebfef02/lib/src/utils/immersive_mode.dart#L5

Thanks for the hint, I added immersive mode to the clock screen 👍🏼

veloce:
Looks nice indeed! I agree with all the above comments from @HaonRekcef 
- [ ] selecting a time control shouldn't change the selected one on Home Screen
- [ ] disable buttons while clock is active

Also since it's a new app, we should change the clock color. We could use another lichess color (defined in `LichessColor.dart`).

veloce:
I also think the move counter is weirdly placed here. Could we not just put it just below the clock instead?

We can even prepend the string: "Moves played:" before, which is translated.

KarnerTh:
Thanks for all the feedback 🙌🏼 

- [x] Special time increment case with 0+1 gets handled as 3s +1
- [x] Converted time control component to a "dumb component" and handle selection in parent to not change setting in home screen
- [x] Buttons are now disabled when game is running
- [x] Changed the clock color to lichess green

@veloce I am afraid I do not get your point with the move counter - a "move" is completed once both White and Black have played one turn so the same value has to be visible to both players. What do you mean with "below the clock"? Should every player see the move count below his own clock? That would duplicate the information and take up screen space 🤔

veloce:
I don't really like the move counter being placed here.

But why is that counter useful in the first place? I thought it was only for some special clock setup which is not supported at the moment, isn't it?

KarnerTh:
To be honest, I migrated the move counter feature and design 1:1 from the old app, because I thought it could be useful to keep track for the 50 move draw rule in longer games.

If that's not desired, we can simply remove the feature.

veloce:
I think we can keep the counter but with a better design.

Let's add a move counter, but per player. Each time the clock is tapped the counter increments (independently for each player).

Display the move number with a small font at the bottom right corner of each clock button.

KarnerTh:
Comments are all resolved now

<details><summary>Settings screen without ultrabullet</summary>
<p>

time

</p>
</details> 

<details><summary>New clock screen with turn count per player</summary>
<p>

game

</p>
</details>

veloce:
Looks very good! Thanks @KarnerTh , I will test it on iOS asap.

I think we're missing one last important thing though: the click sound on tap. Let's not bother with the sound themes and use the same sound for all themes. We can reuse the sound of lichobile.

KarnerTh:
Added the clock sound from lichobile on every tap.

Tested only on Android, because I am still not able to run it on iOS - so help is appreciated :)

PR #528, opened 2024-02-05
improve anon error messages

HaonRekcef:
Improves the error messages when an anon user requests an analysis or wants to chat.


veloce (approved):

veloce:
Thanks.

PR #531, opened 2024-02-06
localize numbers

### LINKED ISSUE #504: localize numbers
like in POtD "Played 104537 times" should be `104,537` in english or `104 537` in french.

HaonRekcef:
closes #504
I added localizations to integers that are likely to be larger than 1000 (number of games played and also the number of times a puzzle has been played), except ratings. It could be easily extended to more numbers. 
Tested for German, French, and English.

### DIFF
--- a/lib/src/utils/string.dart
+++ b/lib/src/utils/string.dart
@@ -1,5 +1,6 @@
 import 'dart:convert';
 import 'dart:math';
+import 'package:intl/intl.dart';
 
 final _random = Random.secure();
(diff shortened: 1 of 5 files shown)


ornicar:
:+1: ratings and move numbers must not be localized. Every other number that represents a quantity is a candidate.

HaonRekcef:
So do you recommend localizing numbers in the activity feed, for example, even if the number is likely to be small?

PR #538, opened 2024-02-07
display information on closed accounts

HaonRekcef:
Displays a message on the user screen indicating that the account has been closed, rather than showing a "could not load recent games" error.


veloce (approved):
Thanks!

Issue #462, opened 2024-01-08
Improve offline mode display


veloce:
While in offline mode (offline status obtained with `connectivityProvider`):

In puzzle tab:
- remove error messages for dashboard and history that could not load

In play tab:
- disable friends button

In the bottom navigation bar:
- disable the "watch" button
- show a toast: "Not available in offline mode" if the watch button is tapped

georgkrause:
I reviewed the issue and I think everything is solved already:

* remove error messages for dashboard and history that could not load -> Cannot reproduce, don't know where this was changed
* disable friends button -> #645 
* disable watch button -> #1028
* show a toast: -> #645

The only thing that might be left is a visual indicator, that the watch button is inactive in the button bar.

robertterra-wmtrading:
mengo

blastoncrush:
Hello, is it still relevant to implement this last feature ?

isarvy:
Hey, It's my first bugfix, will be taking this one.

dufferzafar:
I'm looking to contribute to Lichess. Just wanted to check whether this is done or actually still open?

PR #546, opened 2024-02-08
offline mode improvements

brandonguigo:
Closes: #462 

# Recap
In puzzle tab:
- [x] ~~_remove error messages for dashboard and history that could not load_~~ : cannot reproduce the error message, even offline, the history and dashboards are displayed

In play tab:

- [x] disable friends button

In the bottom navigation bar:
- [x] disable the "watch" button
- [x] show a toast: "Not available in offline mode" if the watch button is tapped

### DIFF
--- a/lib/src/navigation.dart
+++ b/lib/src/navigation.dart
@@ -161,7 +161,11 @@ class BottomNavScaffold extends ConsumerWidget {
   /// scrollable to the top.
   /// Otherwise, switch to the tapped tab.
   void _onItemTapped(
-      BuildContext context, WidgetRef ref, int index, bool isOnline) {
+    BuildContext context,
+    WidgetRef ref,
+    int index,
+    bool isOnline,
+  ) {
     final curTab = ref.read(currentBottomTabProvider);
     final tappedTab = BottomTab.values[index];
 
--- a/lib/src/navigation.dart
+++ b/lib/src/navigation.dart
@@ -3,12 +3,14 @@ import 'package:flutter/material.dart';
 import 'package:flutter_gen/gen_l10n/l10n.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:lichess_mobile/src/styles/puzzle_icons.dart';
+import 'package:lichess_mobile/src/utils/connectivity.dart';
 import 'package:lichess_mobile/src/utils/l10n.dart';
 import 'package:lichess_mobile/src/utils/l10n_context.dart';
 import 'package:lichess_mobile/src/view/home/home_tab_screen.dart';
 import 'package:lichess_mobile/src/view/puzzle/puzzle_tab_screen.dart';
 import 'package:lichess_mobile/src/view/tools/tools_tab_screen.dart';
 import 'package:lichess_mobile/src/view/watch/watch_tab_screen.dart';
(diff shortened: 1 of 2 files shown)


lib/src/navigation.dart
```diff
@@ -106,6 +108,12 @@ class BottomNavScaffold extends ConsumerWidget {
     final currentTab = ref.watch(currentBottomTabProvider);
     final tabs = ref.watch(tabsProvider);
 
+    bool isOnline = true;
+
+    ref.listen(connectivityChangesProvider, (_, connectivity) {
```
veloce, line 113: You should not use `ref.listen` here but fetch connectivity value directly in `_onItemTapped`.

lib/src/navigation.dart
```diff
@@ -151,9 +160,24 @@ class BottomNavScaffold extends ConsumerWidget {
...
+  void _onItemTapped(
+    BuildContext context,
+    WidgetRef ref,
+    int index,
+    bool isOnline,
+  ) {
     final curTab = ref.read(currentBottomTabProvider);
     final tappedTab = BottomTab.values[index];
+
+    if (!isOnline && tappedTab == BottomTab.watch) {
```
veloce, line 172: This logic would be better placed in the `else` of 
```dart
if (tappedTab == curTab) {
...
} else {
 // put it here because we want it to apply only if the tapped tab is not the current tab
}
```

lib/src/view/home/home_tab_screen.dart
```diff
@@ -749,20 +749,33 @@ class _GamePreview<T> extends StatelessWidget {
 }
 
 class _RelationButton extends ConsumerWidget {
-  const _RelationButton();
+  const _RelationButton(this.wasOnline);
```
veloce, line 752: The `wasOnline` is weird here. We want to disable the button if currently not online, the `wasOnline` serves another purpose.

lib/src/view/home/home_tab_screen.dart
```diff
@@ -749,20 +749,33 @@ class _GamePreview<T> extends StatelessWidget {
...
+
+  final bool wasOnline;
 
   @override
   Widget build(BuildContext context, WidgetRef ref) {
     return AppBarIconButton(
-      icon: const Icon(Icons.people),
+      icon: Icon(
```
veloce, line 759: You shouldn't change the button color. In flutter you disable a button by passing a `null` callback to the `onPressed` parameter. Flutter will disable and grey out the button automatically then.

lib/src/view/home/home_tab_screen.dart
```diff
@@ -749,20 +749,33 @@ class _GamePreview<T> extends StatelessWidget {
...
       semanticsLabel: context.l10n.friends,
       onPressed: () {
-        pushPlatformRoute(
-          context,
-          title: context.l10n.friends,
-          builder: (_) => const RelationScreen(),
-          fullscreenDialog: true,
-        );
+        if (!wasOnline) {
+          showPlatformSnackbar(
```
veloce, line 766: Not necessary to show a snackbar, just disabling the button is enough.

brandonguigo:
@veloce thanks for the review, i will do the changes when i have a bit of time

veloce:
Closing as #645 was merged.

Issue #550, opened 2024-02-12
When you do the same castling in different ways, it behaves like a different variation


okmsbun:
### What is the problem?

When the king is placed in different places during the same castling, the game seems to continue from a different variation. In one case, the forward button is disabled.

I encountered this error while trying to solve issue [this](https://github.com/lichess-org/mobile/issues/543).

A similar [issue](https://github.com/lichess-org/mobile/issues/182) has been opened before, but the problem still persists.

| e8a8 | e8c8 |
|---------|---------|
| ![e8a8](https://github.com/lichess-org/mobile/assets/49310621/8420aaaf-a20d-4825-9cf2-3c06c82810dc) | ![e8c8](https://github.com/lichess-org/mobile/assets/49310621/7863821c-b428-4cfc-b322-627de2ec9b49) |

### Steps to reproduce the bug

1. Start analysis for any castled game.
2. Do the same castling in two different ways. example: 'e8a8' and 'e8c8'.

### What did you expect to happen?

The button used to move to the next move on the analysis screen must be active in two different castling ways.

### What happened instead?

In one of them, it was acted as if a different move was played and another variation was moved on.

### Device Info

Pixel 4a

### Operating System Version

Android 13 API33

Issue #543, opened 2024-02-08
Annotation symbols displayed on wrong square if castling


HaonRekcef:
### What is the problem?

Annotation symbols displayed on wrong square if castling was the last move

### Steps to reproduce the bug

1. Find a game where castling is an inaccuracy, mistake or blunder
2. Go to analysis

### What did you expect to happen?

The shape should be displayed on the square the king moved to

### What happened instead?

The shape was displayed on the square the rook moved from

### Device Info

Pixel 6a

### Operating System Version

Android 14

### Additional information

<details>
<summary>Screenshot</summary>

![Screenshot_20240207-233949](https://github.com/lichess-org/mobile/assets/78898963/cd4c2090-a0af-45f0-8e91-8788866572dd)
</details>

okmsbun:
Yes I tested it and it seems wrong. It's not like Lischess.org. In fact, in different forms of castling, it is sometimes not shown at all. In the video below, when I do the same castling in different ways, 'annotation' is shown (incorrectly) in one and not shown at all in the other.

Here for that 
https://github.com/lichess-org/mobile/blob/3c6d1105b3ec2a2d46efa529eb9f8fb23cf2f0b7/lib/src/view/analysis/analysis_screen.dart#L317

than
https://github.com/lichess-org/mobile/blob/3c6d1105b3ec2a2d46efa529eb9f8fb23cf2f0b7/lib/src/model/common/chess.dart#L42
and checking whether castling is done or not, it seems like it can be solved.


https://github.com/lichess-org/mobile/assets/49310621/58af79d6-92b6-4ccf-afe1-51ad3fd5c468

okmsbun:
When I leave the king to c8, the 'currentNode.nags' list ([]) [here](https://github.com/lichess-org/mobile/blob/3c6d1105b3ec2a2d46efa529eb9f8fb23cf2f0b7/lib/src/view/analysis/analysis_screen.dart#L283) appears empty. This list should not be empty. The list should be the same as the list ([2]) that comes when the king is left on a8.

Game = https://lichess.org/407v1JxQ/black#29

PR #553, opened 2024-02-14
Alternate castling was forced into normal castling

okmsbun:
I did something like this to solve [this](https://github.com/lichess-org/mobile/issues/550) problem. If there was a castling move containing the values of the map here, I forced it to change it with its key. Before solving [this](https://github.com/lichess-org/mobile/issues/543) problem, this needs to be solved first.

On the game analysis screen, I reduced eight castling alternatives to four. In all moves, there will be a movement just like the movement in the keys. Is this logic correct? If true, is this function the place of application?


veloce:
I think the logic is correct, also `Node` looks like a good place to put this. 

Since the alternative castling notation is always tricky it would be good to also add a new test to the `Node` tests to ensure all types of castling moves produces the same branch `id`.

veloce (approved):
Perfect! Thank you!

PR #565, opened 2024-02-20
Android system colors

### LINKED ISSUE #558: [Android] Add Monochrome Icon
Please consider adding a [monochrome/Material You/Themable](https://developer.android.com/develop/ui/views/launch/icon_design_adaptive) icon to the Android app.

This does not replace the standard icon and is really important for consistency on users with it enabled (on modern Android phones).

See the comparison between the default monochrome mode on my Samsung OneUI/Android 14 phone and the two lichess apps.

### LINKED ISSUE #560: [Android] Option (if not default) For Material You Theme
Please consider adding a theming option to color the app with the user's Material You palette (derived by the OS from the wallpaper color palette). This already seems to be partially implemented, but instead of using system colors, it uses the board colors.

Please either add a toggle in appearance settings or make Material You the default.

Also consider adding a Material You tint to the board theme as an option or default.

Lichess by default:

Other Material You/3 apps (My color palette):

veloce:
Closes #560 
Closes #558 

- Adds a new settings entry to choose palette (only on android 12+):
- Adds a new board theme settings entry (on android 12+):
- Improve settings icons
- Improve android launcher icons, add the monochrome variant


veloce:
I'm not sure how to make the best looking "system color" board.

The approach I used is pretty simple: https://github.com/lichess-org/mobile/blob/82bbe4afd4682ca514c4d570929199e902fb5348/lib/src/model/settings/board_preferences.dart#L208

It uses dimmed version of primary and secondary colors. There is maybe a way to improve that?

veloce:
This is the best I could come up with. 

The system color board looks nice with all the different wallpaper I tried, in my opinion. 

For now I don't think I'll make it the default though. I'm a bit reluctant because of the strong identify of the lichess brown board, and also because with the system colors the last move and selected square are not as much visible as with the stock boards.

The option to switch to system colors is quite visible in the settings screen so people who like that feature would find it easily.

Screenshot_1708529453
Screenshot_1708529423
Screenshot_1708529408
Screenshot_1708529276
Screenshot_1708529200
Screenshot_1708529186

LSeelig:
I like it alot. Might be open to future tweaks of the square highlight, though I don't know how I would make it better yet.

LSeelig:
> This is the best I could come up with. 
> 
> The system color board looks nice with all the different wallpaper I tried, in my opinion. 
> 
> For now I don't think I'll make it the default. I'm a bit reluctant because of the strong identify of the lichess brown board, and also because with the system colors the last move and selected square are not as much visible as with the stock boards.
> 
> The option to switch to system colors is quite visible in the settings screen so people who like that feature would find it easily.
> 
> Screenshot_1708529453
> Screenshot_1708529423
> Screenshot_1708529408
> Screenshot_1708529276
> Screenshot_1708529200
> Screenshot_1708529186
> 
> 

I dug into the current themes. This theme very well mimics my system color pallette. Square highlights are in shades of green.

<summary>








</summary>

LSeelig:
I dug into the current themes. This purple theme is fairly similar my system color pallette. Square highlights are in shades of green, despite the purple squares. Could you get better contrast if you weren't trying to theme the highlights?

<summary>








</summary>

veloce:
I have to theme the highlights, because I have to ensure the highlights are visible in any case. If I keep the default green highlights they won't be visible with green themes.

LSeelig:
Understood. I think your end result is excellent and agree w.r.t. user choice.

PR #567, opened 2024-02-22
Fixed incorrectly positioned annotation when performing an alternative castling move on the analysis screen.

okmsbun:
Closes https://github.com/lichess-org/mobile/issues/543

Fixed incorrectly positioned annotation when performing an alternative castling move on the analysis screen.

| before | after | lichess.org |
| Screenshot_1708587191 | Screenshot_1708587219 | Screenshot 2024-02-22 at 10 36 54 |


veloce (approved):
Thanks for taking care of this!

Issue #454, opened 2024-01-08
Full game list of a user


veloce:
For now we're only displaying the last 10 recent games.

A "More" button should open another screen with a lazy loading, infinite list of games.

On top of the list, an horizontal list of filters should be displayed (like in the notifications screen of GitHub app). Filters would include: "rated", "perf type", "bookmarks", etc. (list non exhaustive).

Use this endpoint: https://lichess.org/api#tag/Games/operation/apiGamesUser

(All filters are not yet available in the endpoint)

List items should be slidable to uncover actions:
- bookmark/unbookmark

Long press should open the context menu.

TODO: wireframe

ijm8710:
Piece color and game type played (variant) should also probably be filtering options too

ijm8710:
I agree with your first part Julien, but I still think bookmark as a game-list filter should be added eventually and is still not present.

Especially since the detailed view that archived games provides gives you more to sift through if you have a lot of bookmarks to scroll thru than the access point from your profile

PR #572, opened 2024-02-24
User full game list

ZTL-UwU:
### TODO

- [ ] Full game list (#454)
  - [x] Infinite scroll
  - [x] Lazy loading
  - [x] Context menu
  - [ ] Bookmark / Unbookmark
    - [ ] Add slidable
    - [ ] Implement bookmark action
  - [ ] Filtering

### DIFF
--- a/lib/src/model/game/game_repository.dart
+++ b/lib/src/model/game/game_repository.dart
@@ -8,7 +8,7 @@ import 'package:lichess_mobile/src/model/game/playable_game.dart';
 class GameRepository {
   const GameRepository(this.client);
 
-  final http.Client client;
+  final LichessClient client;
 
   Future<ArchivedGame> getGame(GameId id) {
     return client.readJson(
--- a/lib/src/model/game/game_repository.dart
+++ b/lib/src/model/game/game_repository.dart
@@ -1,3 +1,4 @@
+import 'package:dartchess/dartchess.dart';
 import 'package:fast_immutable_collections/fast_immutable_collections.dart';
 import 'package:http/http.dart' as http;
 import 'package:lichess_mobile/src/model/common/http.dart';
@@ -34,26 +35,39 @@ class GameRepository {
     }
   }
 
-  Future<IList<LightArchivedGame>> getUserGames(
+  Future<IList<LightArchivedGameWithPov>> getUserGames(
     UserId userId, {
     int? max = 20,
     DateTime? until,
   }) {
-    return client.readNdJsonList(
-      Uri(
-        path: '/api/games/user/$userId',
-        queryParameters: {
-          'max': max.toString(),
-          if (until != null) 'until': until.millisecondsSinceEpoch.toString(),
-          'moves': 'false',
-          'lastFen': 'true',
-          'accuracy': 'true',
-          'opening': 'true',
-        },
-      ),
-      headers: {'Accept': 'application/x-ndjson'},
-      mapper: LightArchivedGame.fromServerJson,
-    );
+    return client
+        .readNdJsonList(
+          Uri(
+            path: '/api/games/user/$userId',
+            queryParameters: {
+              'max': max.toString(),
+              if (until != null)
+                'until': until.millisecondsSinceEpoch.toString(),
+              'moves': 'false',
+              'lastFen': 'true',
+              'accuracy': 'true',
+              'opening': 'true',
+            },
+          ),
+          headers: {'Accept': 'application/x-ndjson'},
+          mapper: LightArchivedGame.fromServerJson,
+        )
+        .then(
+          (value) => value
+              .map(
+                (e) => (
+                  game: e,
+                  // we know here user is not null for at least one of the players
+                  pov: e.white.user?.id == userId ? Side.white : Side.black,
+                ),
+              )
+              .toIList(),
+        );
   }
 
   /// Returns the games of the current user, given a list of ids.
--- a/lib/src/model/game/game_repository.dart
+++ b/lib/src/model/game/game_repository.dart
@@ -37,7 +37,7 @@ class GameRepository {
 
   Future<IList<LightArchivedGameWithPov>> getUserGames(
     UserId userId, {
-    int? max = 20,
+    int max = 20,
     DateTime? until,
   }) {
     return client
--- a/lib/src/view/game/game_list_tile.dart
+++ b/lib/src/view/game/game_list_tile.dart
@@ -414,12 +414,12 @@ class _ContextMenu extends ConsumerWidget {
 class ExtendedGameListTile extends StatelessWidget {
   const ExtendedGameListTile({required this.item, this.userId});
 
-  final (LightArchivedGame, Side) item;
+  final LightArchivedGameWithPov item;
   final UserId? userId;
 
   @override
   Widget build(BuildContext context) {
-    final (game, youAre) = item;
+    final (game: game, pov: youAre) = item;
     final me = youAre == Side.white ? game.white : game.black;
     final opponent = youAre == Side.white ? game.black : game.white;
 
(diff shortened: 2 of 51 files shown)


ZTL-UwU:
Currently we are downloading all game histories at once, which takes a very long time to load on startup if the user has a large number of games. We will need a pagination api to get a few games everytime.

veloce:
Hello! Pagination is indeed mandatory, even for a first version.

ZTL-UwU:
Pagination is now implemented! 🚀

ijm8710:
Hello, accuracy should prob be fetched as well similar to the recent game list, thoughts?

ZTL-UwU:
> Hello @ZTL-UwU, accuracy should prob be fetched as well similar to the recent game list, thoughts?

The accuracy should be fetched, yes. Problem is that the server API only return a `"analysed": true` instead of the full game analysis like the one used recent game list. I believe that getting the accuracy would require a change in the lila server.

veloce:
@ZTL-UwU I think we should merge this PR with its current implementation (so without filters and bookmark action).

The rest can be done later. Could you resolve the conflicts please, then I'll review asap. Thanks!

lib/src/model/game/game_repository.dart
```diff
@@ -42,6 +42,17 @@ class GameRepository {
     );
   }
 
+  Future<FullGamePaginator> getFullGames(
+    UserId userId,
+    int page,
+  ) {
+    return client.readJson(
+      Uri.parse('$kLichessHost/@/$userId/all?page=$page'),
```
veloce, line 50: This endpoint is not opened to OAuth. I'm not sure we should use it since it's an old one not part of the API, and it would be rate limited.

I think https://lichess.org/api#tag/Games/operation/apiGamesUser has everything we need.

cc @ornicar , what do you think about that?

lib/src/view/user/full_games_screen.dart
```diff
@@ -0,0 +1,277 @@
...
+import 'package:lichess_mobile/src/model/game/game_repository.dart';
+import 'package:lichess_mobile/src/model/user/user.dart';
+import 'package:lichess_mobile/src/styles/styles.dart';
+import 'package:lichess_mobile/src/view/game/game_list_tile.dart';
+import 'package:lichess_mobile/src/widgets/platform.dart';
+import 'package:riverpod_annotation/riverpod_annotation.dart';
+
+part 'full_games_screen.g.dart';
+
+/// Create a Screen with Top 10 players for each Lichess Variant
```
veloce, line 19: The comment is wrong here.

lib/src/view/user/full_games_screen.dart
```diff
@@ -0,0 +1,277 @@
...
+
+  @override
+  Widget build(BuildContext context) {
+    return PlatformWidget(androidBuilder: _buildAndroid, iosBuilder: _buildIos);
+  }
+
+  Widget _buildIos(BuildContext context) {
+    return CupertinoPageScaffold(
+      navigationBar: const CupertinoNavigationBar(
+        previousPageTitle: 'Home',
```
veloce, line 32: Ideally you should not use this parameter, it's better with the default.

lib/src/view/user/full_games_screen.dart
```diff
@@ -0,0 +1,277 @@
...
+
+      setState(() {
+        games.addAll(nextPage?.games ?? []);
+      });
+      isLoading = false;
+    }
+  }
+}
+
+class _SlideMenu extends StatefulWidget {
```
veloce, line 194: I'm not against a custom implementation, but is there a reason you choose it instead of `flutter_slidable` package which is already used in the project?

ZTL-UwU: Oh I just didn't know it was already in the project 🤣

lib/src/model/game/game_repository.dart
ZTL-UwU: I did consider using this API, but it doesn't fit with `LightGame` which means adding custom implementation many things like the game tiles. Where as the old API just works. It would be nice if the new API can have the same response format as recent games.

veloce: I don't understand @ZTL-UwU 
This API does perfectly fit with `LightArchivedGame`, it is already used in the code, here: https://github.com/lichess-org/mobile/blob/d8b63a1d0afde6b54a57f31cba4279065598c9c3/lib/src/model/game/game_repository.dart#L41

ZTL-UwU: @veloce Sorry, I've got myself confused, it was written quite some time ago. What really happened was that the API doesn't have pagination, there is only a `max` query so I had to get all the games at once.

↓ This was the comment in this PR previously.
> Currently we are downloading all game histories at once, which takes a very long time to load on startup if the user has a large number of games. We will need a pagination API to get a few games everytime.

The API documentation says _We recommend streaming the response_. But I don't know how it will work with an infinite scroll so I used the old API instead in 78ff233c0c5c6471b91592512b90a6dcc04a72f1.

veloce: Actually you can paginate using the new endpoint, using `max` and `until`.

You already have an example in the app with the puzzle activity API which works the same way:
- https://lichess.org/api#tag/Puzzles/operation/apiPuzzleActivity
- https://github.com/lichess-org/mobile/blob/e56f52fd8751bb7ca2d8f404cba5b07ce5675c64/lib/src/model/puzzle/puzzle_activity.dart#L63

veloce:
Thanks for making the change @ZTL-UwU , I just realised you updated the PR. Will take take of it now.

android/gradle.properties
```diff
@@ -3,4 +3,4 @@ android.useAndroidX=true
 android.enableJetifier=true
 android.defaults.buildfeatures.buildconfig=true
 android.nonTransitiveRClass=false
-android.nonFinalResIds=false
+android.nonFinalResIds=false
```
veloce, line 6: this change should be removed

lib/src/model/game/game_repository.dart
```diff
@@ -51,6 +51,28 @@ class GameRepository {
     );
   }
 
+  Future<IList<LightArchivedGame>> getFullGames(
```
veloce, line 54: This method should now replace the `getRecentGame` because both almost do the same thing.
So let's remove `getRecentGames`, and have a default `max` parameter at 20.

lib/src/view/game/game_list_tile.dart
```diff
@@ -402,3 +409,101 @@ class _ContextMenu extends ConsumerWidget {
     );
   }
 }
+
+class ExtendedGameListTile extends StatelessWidget {
```
veloce, line 413: We don't really understand what this widget does just by reading its name. Since it's a public class it should have a doc comment explaining what it is.
And if possible use a better name.

veloce: What does it add to the existing `GameTile` and why is it needed? not sure to understand.

veloce (commented):
It is starting to look good now!

As you may have noticed, now the recent games section include offline games too. Pagination support is already done in the `GameStorage` class, so it should be pretty straightforward now to include the offline game in full game list when having no connectivity.
Using same logic as in https://github.com/lichess-org/mobile/blob/60530a1feccff195c698e88ff0dab9c39ac6492e/lib/src/view/user/recent_games.dart#L57

veloce:
Thanks for your work. I pushed some refactoring, I'll continue to work a bit on it but I think we're close to merging this :)

veloce:
By the way, each time I push to a remote repository branch, it messes up the diff. I don't know what I'm doing wrong...
I used this command to push the code to your repo:
```
git push git@github.com:ZTL-UwU/mobile.git ZTL-UwU-full-game-list:full-game-list
```

veloce:
🎉

ijm8710:
@ZTL-UwU how much effort would it be to have a similar game list view+more button for even more game history but on the perf cards themselves?

I know you were considering working on bookmarks but curious since you basically have built this already if applying to perf cards would be a quicker hit

Hoping you and Veloce agree it makes sense to have them there

ZTL-UwU:
@ijm8710 I thinks there is a merged PR by for this already #830

PR #587, opened 2024-03-09
Remove .remainder(60) from opponentLeftCounter

### LINKED ISSUE #586: Fix timer countdown when opponent leaves
### What is the problem?

When the timer that counts down when your opponent leaves the game is greater than 60 seconds, the app displays the seconds.remainder(60) as the number of seconds remaining until you can either 'Claim Victory' or 'Call Draw', instead of the correct total seconds value.

1. Create a 10+0 rapid game.
2. Have your opponent leave.
3. Observe the countdown timer go from T...3..2..1..0..59..

Countdown timer will subtract 1 each time step until it reaches 0.

### What happened instead?

Countdown timer subtracts 1 each time step, then jumps to 59 from 0.

### Device Info

iPhone 15 Pro

### Operating System Version

iOS 17.3.1

### Additional information

_No response_

kweithers:
Closes #586.

### DIFF
--- a/lib/src/view/game/game_body.dart
+++ b/lib/src/view/game/game_body.dart
@@ -993,7 +993,7 @@ class _ClaimWinCountdown extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
-    final secs = duration.inSeconds.remainder(60);
+    final secs = duration.inSeconds;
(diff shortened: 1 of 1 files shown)


veloce:
Thanks!

PR #596, opened 2024-03-11
fix alternative castling

### LINKED ISSUE #593: Wrong move gets played in the analysis board
### What is the problem?

I was analysing the following position: 

screenshot

[Variant "From Position"]
[FEN "6k1/6P1/8/p2p2KQ/4r3/5R2/PP6/4q3 b - - 0 46"]

46... Qg1+

And I wanted to play the move Qg1+, but Qh1 was played, same problem when I try Qc1+, Qa1 happens.

1. Copy this into the analysis board: [Variant "From Position"]
[FEN "6k1/6P1/8/p2p2KQ/4r3/5R2/PP6/4q3 b - - 0 46"]

46... Qg1+

2. Try to play Qc1 or Qg1

Qg1 gets played

### What happened instead?

Qh1 was played

### Device Info

Pixel 6a

### Operating System Version

Android 14

### Additional information

- I tested it in a game and I was able to play Qg1 in the specific position.
- Qg1 also does not executed when pressing the engine line, Qh1 happens.

HaonRekcef:
fixes #593 #601 #602

### DIFF
--- a/lib/src/model/common/node.dart
+++ b/lib/src/model/common/node.dart
@@ -192,9 +192,14 @@ abstract class Node {
     bool prepend = false,
   }) {
     final pos = nodeAt(path).position;
-    final (newPos, newSan) = pos.makeSan(convertAltCastlingMove(move) ?? move);
+    final isKingMove =
+        pos.board.roleAt(parseSquare(move.uci.substring(0, 2)) ?? 0) ==
+            Role.king;
+    final convertedMove =
+        isKingMove ? convertAltCastlingMove(move) ?? move : move;
+    final (newPos, newSan) = pos.makeSan(convertedMove);
     final newNode = Branch(
-      sanMove: SanMove(newSan, convertAltCastlingMove(move) ?? move),
+      sanMove: SanMove(newSan, convertedMove),
       position: newPos,
     );
     return addNodeAt(path, newNode, prepend: prepend);
--- a/lib/src/model/common/node.dart
+++ b/lib/src/model/common/node.dart
@@ -193,8 +193,7 @@ abstract class Node {
   }) {
     final pos = nodeAt(path).position;
     final isKingMove =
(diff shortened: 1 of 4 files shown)


lib/src/model/common/node.dart
```diff
@@ -192,9 +192,14 @@ abstract class Node {
     bool prepend = false,
   }) {
     final pos = nodeAt(path).position;
-    final (newPos, newSan) = pos.makeSan(convertAltCastlingMove(move) ?? move);
+    final isKingMove =
+        pos.board.roleAt(parseSquare(move.uci.substring(0, 2)) ?? 0) ==
```
veloce, line 196: You could use `move.from` instead of `move.uci`.

veloce (commented):
It would be nice to add a test case to prevent further regression. Thanks!

lib/src/model/common/node.dart
HaonRekcef: Ok. The from property is only available on instances of NormalMove, so I check if its a NormalMove first. Castling will always be a normalmove anyways.

HaonRekcef:
I added a testcase, which fails with the old code and works with my bug fix. I tested the fix in the analysis screen and for the puzzles, in both cases it seems to fix the wrong behaviour. So #601 and #602 can be closed as well.
The checks for this pull request fail because some unrelated code features are deprecated in the newest flutter beta build.

veloce:
Damn this bug is annoying. Unfortunately I won't be able to deal with it before 1 month when I return from my vacation. 

If you want to, you can fix the flutter beta issues in another PR. We'd then be able to make a hot fix release for at least android users.

veloce:
Hi @HaonRekcef 

If you commit the version change 7.0.5 to the pubspec.yaml file in this PR I can release the fix for all android users. Better than nothing I think 🙂

HaonRekcef:
I agree, I committed it, @veloce

veloce:
Thank you!

PR #598, opened 2024-03-11
wrap around tos violation

HaonRekcef:
fixes bug reported on discord that for certain languages not the complete text was displayed on accounts that violated tos.


HaonRekcef:
Yes, I tested it and it works well.

PR #603, opened 2024-03-15
Fix deprecated warnings for some colorSchemes properties

HaonRekcef:
I fixed the deprecated warnings that appear when running `flutter analyze`. The `background` and `onBackground` were easily replaced with `surface` and `onSurface` without any visible changes. The suggested replacement for `surfaceVariant` did not work; the best I could come up with is `tertiary.withOpacity(0.15)`, and it looks similar to before now.
It changes the color of the tiles with the timecontrols.

I tested it with light, dark, and system colors and different themes, and it looks pretty good.

| New                                            | Old                                            |
|------------------------------------------------|------------------------------------------------|
| Image 1 | Image 3 |
| Image 2 | Image 4 |


veloce:
Thanks!

PR #605, opened 2024-03-19
fixes the notation if the analysis starts with black to move

HaonRekcef:
The move number was not displayed if it's Black's turn to move in the initial position of the analysis.
<details>
<summary>Before</summary>

before
</details>

<details>
<summary>After</summary>

after

</details>


veloce (approved):

veloce:
Thanks!

PR #608, opened 2024-03-20
Fix Storm Dashboard for Other Users

HaonRekcef:
This PR resolves the issue of displaying other users storm highscores. At the moment users would either be directed to their own highscore screen or encounter an error if not logged in. 
Now it should correctly display the highscores of other users. Also, I have included the user's name in the screen title, as it's now possible to view highscores of other users.
<details>
<summary>Preview</summary>

image
</details>
I can not test it for iOS but I copied the code from perf_stats_screen.

One question remaining is whether we really should implement caching for the storm dashboard. Currently, if a user views their history, plays, beats their daily highscore and then revisits their history, the updated highscore isn't displayed.

### DIFF
--- a/lib/src/model/puzzle/puzzle_providers.dart
+++ b/lib/src/model/puzzle/puzzle_providers.dart
@@ -101,11 +101,12 @@ Future<IList<PuzzleHistoryEntry>?> puzzleRecentActivity(
 }
 
 @riverpod
-Future<StormDashboard?> stormDashboard(StormDashboardRef ref) async {
-  final session = ref.watch(authSessionProvider);
-  if (session == null) return null;
+Future<StormDashboard?> stormDashboard(
+  StormDashboardRef ref, {
+  required UserId id,
+}) async {
   return ref.withClientCacheFor(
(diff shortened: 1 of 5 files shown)


veloce:
So you're right we should remove caching to ensure it's always up to date.

I'm not sure about player name in title though. Because sometimes it is too long and cannot be displayed entirely. We could maybe show player's name as an intermediate level title in body, only for other players screens.

HaonRekcef:
True, sometimes the name gets cut off. But this also happens for the other stats screen, probably we should fix this on all of them?
Should I remove the caching for all users, one option would also be to just disable it for the logged in user.

HaonRekcef:
Okay, I removed the caching. However, I haven't found a good way to position the username somewhere other than in the title bar. Designing isn't exactly my strong suit... 🙃

veloce:
Yeah I see that is not obvious to make it look good. 

Perhaps we can just not display the player's name for now. This screen has to be accessed from the player profile so you'll always know what you're looking at.

HaonRekcef:
I removed the names from the screens, so no more names that are cut off.

veloce (approved):

Issue #595, opened 2024-03-11
Puzzle Streak: enable engine analysis after wrong move


stephanhuebsch:
After revealing the solution to a (regular) puzzle, it is possible to analyze the board using the engine. 

It would be nice if this feature also existed after getting a puzzle wrong in puzzle streak. Right now, it's only possible to share the puzzle, open it in a browser and using the browser version of Lichess - or continue with a new streak.

HaonRekcef:
What is your opinion, Veloce, should an evaluation bar appear, or should the user have the option to analyze the last puzzle of the puzzle streak?
Also, for normal puzzles, shouldn't there be the option to analyze them with the analysis board instead of just the evaluation?
I would try to implement this.

veloce:
Yes we should be able to analyse the last puzzle in streak. 

And also with regular puzzles. I guess we should keep the bar here but also have a button to launch the analysis screen.

HaonRekcef:
this can be closed due to #610

PR #610, opened 2024-03-24
puzzle(-streak) analysis in analysisboard

HaonRekcef:
Implements #595: Allows the user to analyze the last puzzle (the one they failed) of the streak or normal puzzles on the analysis board. For streak, there is a button in the bottom bar. For normal puzzles I merged it together with the share puzzle to a menu, because we should not have two separate buttons for analysis (evaluation bar and analysis board) in the same button bar.

When the user presses to go to the analysis board the initial position corresponds to the starting position of the puzzle and the moves of the solution are there.

### DIFF
--- a/lib/src/model/puzzle/puzzle_controller.dart
+++ b/lib/src/model/puzzle/puzzle_controller.dart
@@ -6,6 +6,7 @@ import 'package:dartchess/dartchess.dart';
 import 'package:fast_immutable_collections/fast_immutable_collections.dart';
 import 'package:freezed_annotation/freezed_annotation.dart';
 import 'package:http/http.dart' as http;
+import 'package:lichess_mobile/src/constants.dart';
 import 'package:lichess_mobile/src/model/common/chess.dart';
 import 'package:lichess_mobile/src/model/common/http.dart';
 import 'package:lichess_mobile/src/model/common/node.dart';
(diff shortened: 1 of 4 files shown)


veloce:
Nice! 

I wonder if we should use the whole game PGN instead. As in the lichess website.

Or maybe keep it like that, but also add a button in the menu to go the the game screen of that puzzle (jumping directly to the puzzle move). What do you think?

HaonRekcef:
The puzzle API does not send the complete game, but only until the moment where the tactic happened. So for simplicity and the offline mode we should probably keep the analysis in the analysis board like this.
I agree, I can try to add a button in the menu that loads the game where the puzzle is from.

veloce:
It does load the full game tree. But I agree let's keep the analysis as you did it. And with the game button I'll be nice!

HaonRekcef:
I added the linking to the game in the menu of the puzzle. It goes to the archived game screen of that game. 
As far as I could see there is no translation for this and I called it Puzzle Source Game, maybe there is a better name for this. 

Also unrelated but I improved the ordering in the puzzle themes a bit.

veloce:
I think there is a "From game X" translation we can use.

Doesn't the ordering follow the website's one?

HaonRekcef:
Yeah, I saw that string but was not sure how to use it. "From game player1 vs player 2" will be too long.
The other option would be "From game #gameid".

No, on the website it's different as well in the meantime.

veloce:
I'd use From game #id yes.

HaonRekcef:
done

PR #613, opened 2024-03-29
Display text vertically for game phases

julien4215:


lib/src/view/analysis/analysis_screen.dart
HaonRekcef, line 1: lines 850 and 851 can be removed

julien4215: Good catch, I totally forgot this.

veloce:
Thanks!

I'd remove vertical padding, and have only 1 for horizontal.

Also it would be better with a slight opacity on the text color, so the label doesn't hinder the view of the graph.

veloce (approved):

PR #621, opened 2024-04-08
Fix dependencies and deprecated warning

HaonRekcef:
Fixes dependencies and resolves the deprecated warning which are the reason why the workflow of the pr #620 failed


HaonRekcef:
... still fails. not sure what the issue is, on my local machine with the same flutter beta version flutter analyze does not give an error, I will try to investigate when I find some time in the next days

veloce:
Looks like an issue with l10n_gen. Perhaps some things changed with the new beta version related to that.

PR #631, opened 2024-04-12
Navigate the puzzle with back and forward buttons

HaonRekcef:
Implements a request from discord to navigate the puzzles. It is possible to go back and see the previous moves, including the introductory move.
I am not entirely happy with the buttons, at the moment the navigating buttons do not have a tooltip while the show solution and the difficulty selector have it. I figured this is the best way though, as there is no translation for previous.


HaonRekcef:
I had some test failing because of the accessibility requirements, so I fixed it by putting them in a semantics container, which fixed it, but I have not seen this in the rest of the codebase. What is the recommend way to fix this?

veloce:
Closing it as not planned for the moment.

PR #643, opened 2024-04-20
Disable premove while navigating history

### LINKED ISSUE #641: Premove should be cancelled when player goes to a previous move using backward arrow / cursor
### Description
Currently the premove is not being cancelled when the player goes to one of the previous moves by using the backward arrow / cursor.
In the older Lichess v1 app the premove was being cancelled in this scenario which is a good user experience and should be replicated in the v2 app.

#### App Version
v0.7.8

#### Device / OS Info
Pixel 4A (Android 14)

veloce:
Closes #641


dbamit291:
@veloce this is a good value add, thanks for the quick fix.

PR #645, opened 2024-04-21
Offline mode behavior

Silipok:
Issue #462
In play tab:

- [x] disable friends button

In the bottom navigation bar:

- [x] disable the "watch" button
- [x]  show a toast: "Not available in offline mode" if the watch button is tapped


Silipok:
@veloce If it's possible review pls.

lib/src/navigation.dart
```diff
@@ -110,6 +117,8 @@ class BottomNavScaffold extends ConsumerWidget {
   @override
   Widget build(BuildContext context, WidgetRef ref) {
     final currentTab = ref.watch(currentBottomTabProvider);
+    final connectivity = ref.watch(connectivityChangesProvider);
+    final isOnline = connectivity.value?.isOnline ?? false;
```
veloce, line 121: You should not assume the value is false when in loading state, I'd prefer to have `true` by default.

Also the connectivity is only needed for the watch tab, so instead of having the `ref.watch` here, it would be better to wrap the `NavigationDestination` with a `Consumer` widget so only that one rebuilds.

lib/src/navigation.dart
```diff
@@ -98,6 +100,11 @@ final watchScrollController = ScrollController(debugLabel: 'WatchScroll');
 final RouteObserver<PageRoute<void>> rootNavPageRouteObserver =
     RouteObserver<PageRoute<void>>();
 
+final cupertinoTabControllerProvider =
```
veloce, line 103: Not really useful to declare a state provider here, you could just use a private `final` variable.

lib/src/view/home/home_tab_screen.dart
```diff
@@ -602,16 +602,21 @@ class _PlayerScreenButton extends ConsumerWidget {
 
   @override
   Widget build(BuildContext context, WidgetRef ref) {
+    final connectivity = ref.watch(connectivityChangesProvider);
+    final isOnline = connectivity.value?.isOnline ?? false;
```
veloce, line 606: You should not access `value` directly but use `AsyncValue`'s `when()` method: https://pub.dev/documentation/riverpod/latest/riverpod/AsyncValue-class.html

veloce (commented):
Thanks for this. It needs some changes regarding the use of riverpod API.

lib/src/view/home/home_tab_screen.dart
Silipok: If a consumer of an [AsyncValue](https://pub.dev/documentation/riverpod/latest/riverpod/AsyncValue-class.html) does not care about the loading/error state, consider using [value](https://pub.dev/documentation/riverpod/latest/riverpod/AsyncValue/value.html)/valueOrNull to read the state:

Do we really need error and load value here?

lib/src/navigation.dart
Silipok: It's impossible  cause i need this value in `onDestinationSelected` cb, all what can i do, it wrap `NavigationBar` with Consumer, but I coudn’t do it with `CupertinoTabBar`.

veloce:
Just added a couple of tweak on top of it. Thanks!

PR #656, opened 2024-05-01
Fix puzzle streak

### LINKED ISSUE #646: Puzzle Streak: New puzzle streak button doesn't always work after you fail your first puzzle
### What is the problem?

Report from Lichess Forum:
When one plays puzzle streak, after a wrong move one can restart the streak by pressing the play button (right pointing triangle), but if you make a mistake at the first puzzle, the triangle does not work. You can press it multiple times but the puzzle streak does not restart. Sometimes it restarts after couple of presses.

1: Go to puzzle streak
2: Make an incorrect first move
3: Press the New Puzzle Streak button ( blue triangle)

You should be taken to a new puzzle streak session

### What happened instead?

No action happens

### Device Info

Pixel 6

### Operating System Version

Android 14

### Additional information

It happens pretty consistently, although not guaranteed.

julien4215:
Closes #646

### DIFF
--- a/lib/src/view/puzzle/streak_screen.dart
+++ b/lib/src/view/puzzle/streak_screen.dart
@@ -363,7 +363,13 @@ class _BottomBar extends ConsumerWidget {
                 Expanded(
                   child: BottomBarButton(
                     onTap: ref.read(streakProvider).isLoading == false
-                        ? () => ref.invalidate(streakProvider)
+                        ? () {
+                            ref.invalidate(streakProvider);
+                            // Since the API sends the same streak during an
+                            // interval of time, the provider will receive the
+                            // same arguments and so will not be recomputed
+                            ref.invalidate(ctrlProvider);
+                          }
                         : null,
                     highlighted: true,
                     label: context.l10n.puzzleNewStreak,


HaonRekcef:
A similar bug seems to exist for puzzle storm. Start a new run, make a move and immediately try to start another run, sometimes the button doesnt work. Can it be fixed the same way?

veloce:
I don't understand how it can fix the bug, since the streak provider is watched and if invalidated should simply build another puzzle controller.

Do you reproduce consistently predictably the issue?

julien4215:
The problem is that it doesn't build another puzzle controller when the provider takes the same values as parameters.

To reproduce the issue, you need to launch a puzzle streak and quickly make a wrong move to launch a new one. The API sends the same set of puzzles during a certain period of time (something like 30 seconds). If you launch a new streak before the API changes its set of puzzle, you will reproduce the issue.

julien4215:
> A similar bug seems to exist for puzzle storm. Start a new run, make a move and immediately try to start another run, sometimes the button doesnt work. Can it be fixed the same way?

I tried to fix it the same way but it didn't work

veloce:
> The problem is that it doesn't build another puzzle controller when the provider takes the same values as parameters.
> 
> 
> 
> To reproduce the issue, you need to launch a puzzle streak and quickly make a wrong move to launch a new one. The API sends the same set of puzzles during a certain period of time (something like 30 seconds). If you launch a new streak before the API changes its set of puzzle, you will reproduce the issue.

I should have read the code comment indeed 😄. Tanks!

PR #657, opened 2024-05-02
resolves #654

### LINKED ISSUE #654: "Has x seconds for his/her move" switches side depending on the side that currently plays
### What is the problem?

When starting a game and the opponent does not start playing instantly there appears a warning below their name.
The warning states that the player has x seconds until the game is aborted.

If you are white and go back a move (so to move 0 where no pieces have moved) the warning switches players and is now below my name.

1. New Game
2. Be White
3. Play first move
4. Opponent does not play first move
5. Opponent gets warning "has x seconds to move"
6. "go back" one move
7. warning appears below my name

warning stays below his/her name

### What happened instead?

warning switched to my name

### Device Info

OnePlus 8 Pro

### Operating System Version

Android 14

### Additional information

_No response_

Mauritz8:
resolves #654

### DIFF
--- a/lib/src/view/game/game_body.dart
+++ b/lib/src/view/game/game_body.dart
@@ -119,7 +119,6 @@ class GameBody extends ConsumerWidget {
     return gameStateAsync.when(
       data: (gameState) {
         final position = gameState.game.positionAt(gameState.stepCursor);
-        final sideToMove = position.turn;
         final youAre = gameState.game.youAre ?? Side.white;
         final archivedBlackClock =
             gameState.game.archivedBlackClockAt(gameState.stepCursor);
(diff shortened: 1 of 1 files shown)


lib/src/view/game/game_body.dart
```diff
@@ -119,7 +119,7 @@ class GameBody extends ConsumerWidget {
     return gameStateAsync.when(
       data: (gameState) {
         final position = gameState.game.positionAt(gameState.stepCursor);
-        final sideToMove = position.turn;
+        final sideToMove = gameState.game.sideToMove;
```
veloce, line 122: The variable name adds confusion. 

I'd rather not declare it and pass `gameState.game.sideToMove` directly where it's needed.

veloce (commented):
Good catch, thanks.

lib/src/view/game/game_body.dart
Mauritz8: Now I have fixed it.

veloce (approved):

PR #660, opened 2024-05-05
Fix puzzle storm

### LINKED ISSUE #633: Puzzle Storm - Can't move after starting new run
### What is the problem?

After starting a new run in Puzzle Storm, sometimes it's not possible to move for some reason. 

**Edit:** It seems like the bug only occurs when you start a new run exactly when the 10 seconds time penalty is given out. So if you tap the New run button while the numbers are still colored red and are ticking faster, the bug will appear.

~~I can't always reproduce it, but I managed to do it several times:~~

1. Open Puzzle Storm
2. Start a new run while it's your move **right when the 10 seconds are taken away from your time**
3. Try to move any of your pieces by tapping them and tapping a destination square

The selected piece gets moved to the destination square. 

### What happened instead?

The piece gets deselected when the destination square is tapped. It's not possible to move any pieces. 

### Device Info

Redmi Note 10 5G

### Operating System Version

MIUI 14 / Android 13

### Additional information

~~As mentioned above, it's not always possible to reproduce this bug.~~
https://github.com/lichess-org/mobile/assets/118524250/3b78dfce-b975-4087-be04-791d9dab3cba

julien4215:
Close #633

### DIFF
--- a/lib/src/model/puzzle/storm_controller.dart
+++ b/lib/src/model/puzzle/storm_controller.dart
@@ -36,12 +36,14 @@ class StormController extends _$StormController {
 
   @override
   StormState build(IList<LitePuzzle> puzzles) {
+    final pov = Chess.fromSetup(Setup.parseFen(puzzles.first.fen));
+    final clock = StormClock();
+
     ref.onDispose(() {
       _firstMoveTimer?.cancel();
-      state.clock.dispose();
+      clock.dispose();
     });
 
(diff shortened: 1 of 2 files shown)


veloce (approved):
Thanks!

veloce:
Oh, I didn't see it was a draft.

julien4215:
Closed as there is already #661

PR #679, opened 2024-05-13
docs: Add instructions for using Lichess dev server in README

iamnabink:
Updated README with instructions for using Lichess dev server and added a note regarding URL scheme changes in the recent code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,22 +36,12 @@ Instructions to install both are found in [the lila wiki](https://github.com/lic
 
 The mobile application is configured by default to target `http://127.0.0.1:9663` and `ws://127.0.0.1:9664`, so keep these when installing lila.
 
-### Using Lichess Own Dev Server
+### Using Lichess dev server
 
-If you don't want to set up your local Lila server, you can use [lichess dev](https://lichess.dev/) server to run this app.
-Set up the following values as environment variables:
+To use the [Lichess dev](https://lichess.dev/) to run this app, run the following command to set up the Lichess host URLs in the app.
 
 ```
-LICHESS_HOST=lichess.dev
-LICHESS_WS_HOST=socket.lichess.dev
-```
-
-To do so just use `--dart-define` argument during `flutter run` or `flutter build` command.
-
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -36,6 +36,27 @@ Instructions to install both are found in [the lila wiki](https://github.com/lic
 
 The mobile application is configured by default to target `http://127.0.0.1:9663` and `ws://127.0.0.1:9664`, so keep these when installing lila.
 
+### Using Lichess Own Dev Server
```
veloce, line 39: I'd simply put `Using Lichess dev server`

README.md
```diff
@@ -36,6 +36,27 @@ Instructions to install both are found in [the lila wiki](https://github.com/lic
...
+
+If you don't want to set up your local Lila server, you can use [lichess dev](https://lichess.dev/) server to run this app.
+Set up the following values as environment variables:
+
+```
```
veloce, line 44: This code block is a bit confusing as it looks like you'd need to define shell env vars, which is actually not working. It would be better to replace it with the actual full `flutter run --dart-define=....` command.

veloce (commented):
Thanks! Have only a couple of comments, then we can merge it.

README.md
iamnabink: Got it. How about this?

### Using Lichess dev server

To use the [Lichess dev](https://lichess.dev/) server to run this app, set up the following values as environment variables in cloned lichess flutter app.

```
flutter run --dart-define=LICHESS_HOST=lichess.dev --dart-define=LICHESS_WS_HOST=socket.lichess.dev
```

iamnabink: Got it. How about this?

### Using Lichess dev server

To use the [Lichess dev](https://lichess.dev/) to run this app, run the following command to set up the Lichess host URLs in the app.

```
flutter run --dart-define=LICHESS_HOST=lichess.dev --dart-define=LICHESS_WS_HOST=socket.lichess.dev
```

veloce (approved):

PR #680, opened 2024-05-13
remove error msgs in offline mode

raphaelladinig:
Issue #462 
 - [x] remove error messages for dashboard and history that could not load
 ---
I'm not 100% sure if that's what you wanted, please suggest improvements.

### DIFF
--- a/lib/src/view/puzzle/dashboard_screen.dart
+++ b/lib/src/view/puzzle/dashboard_screen.dart
@@ -131,7 +131,7 @@ class PuzzleDashboardWidget extends ConsumerWidget {
           'SEVERE: [PuzzleDashboardWidget] could not load puzzle dashboard; $e\n$s',
         );
 
-        if (!isOnline) {
+        if (isOnline) {
           return Padding(
             padding: Styles.bodySectionPadding,
             child: Column(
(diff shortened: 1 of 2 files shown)


lib/src/view/puzzle/dashboard_screen.dart
```diff
@@ -123,25 +124,33 @@ class PuzzleDashboardWidget extends ConsumerWidget {
...
-              ),
-              if (e is ClientException && e.message.contains('404'))
-                Text(context.l10n.puzzleNoPuzzlesToShow)
-              else
-                const Text('Could not load dashboard.'),
-            ],
-          ),
-        );
+
+        if (isOnline) {
```
veloce, line 134: This change doesn't make sense anymore I find, since now the dashboard is in its own screen.

I'd leave it as it was.

lib/src/view/puzzle/puzzle_tab_screen.dart
```diff
@@ -347,13 +347,21 @@ class PuzzleHistoryWidget extends ConsumerWidget {
...
+
         debugPrint(
           'SEVERE: [PuzzleHistoryWidget] could not load puzzle history',
         );
-        return Padding(
-          padding: Styles.bodySectionPadding,
-          child: const Text('Could not load Puzzle history.'),
-        );
+
+        if (isOnline) {
```
veloce, line 357: Here is makes sense, since the whole puzzle tab reacts to connectivity. 👍

veloce (commented):
Sorry for the long delay before review.

You did what was intended in the issue, but I think the change in the dashboard screen is no more relevant, so I'd revert just that one.

Thanks!

Issue #628, opened 2024-04-11
Allow time handicap in clock tool


veloce:
Currently the clock tool only allows to select the same time for each player.

Allow to customise the initial time per side, so a player can have a handicap.

Mauritz8:
I'll have a go at this one.

PR #684, opened 2024-05-15
Update custom time control in clock tool

sandeep-adireddi:
1. There's a new modal with sliders that let you change the time and Increment for each player's clock in clock Tool
2. Clock controller is updated to have unique clock options for the top and bottom player clocks.

https://github.com/lichess-org/mobile/issues/628

### DIFF
--- a/lib/src/view/play/custom_time_control_modal.dart
+++ b/lib/src/view/play/custom_time_control_modal.dart
@@ -54,9 +54,13 @@ class CustomTimeControlModal extends ConsumerWidget {
                 onPressed: () {
                   controller.updateOptions(
                     TimeIncrement(
-                        topClockTime * secondsPerMinute, topClockIncrement,),
-                    TimeIncrement(bottomClockTime * secondsPerMinute,
-                        bottomClockIncrement,),
+                      topClockTime * secondsPerMinute,
+                      topClockIncrement,
+                    ),
+                    TimeIncrement(
+                      bottomClockTime * secondsPerMinute,
+                      bottomClockIncrement,
+                    ),
                   );
                   Navigator.pop(context);
                   onSet();
--- a/lib/src/view/play/custom_time_control_modal.dart
+++ b/lib/src/view/play/custom_time_control_modal.dart
@@ -1,0 +1,143 @@
+import 'package:flutter/material.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+import 'package:lichess_mobile/src/model/clock/clock_controller.dart';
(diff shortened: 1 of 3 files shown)


veloce:
Thanks for making this PR. 

Can you post a video showing how the feature works?

Also I don't think we should be able to change the increment per player, but only the initial time, as it was specified in the issue.

sandeep-adireddi:
https://github.com/lichess-org/mobile/assets/16081707/165e3edd-9546-491d-a993-d02b996d9910

Hey, The Users can keep the increment slider on both clocks,If they need same increment.The current code also adds option to select different increments.

veloce:
Thanks for the video.
Looks like the sliders are not the best solution when it is difficult to choose a precise time with them.

How about making these improvements:
- replace the sliders with a time picker ([CupertinoTimePicker](https://api.flutter.dev/flutter/cupertino/CupertinoTimerPicker-class.html) and ([showTimePicker](https://api.flutter.dev/flutter/material/showTimePicker.html) for iOS and android respectively)
- when a custom time is selected show it in the clock (smaller font below the time with this format `<tuning icon> <time> + <increment>`

sandeep-adireddi:
@veloce 

Uploading Screen Recording 2024-06-12 at 1.11.33 AM-1.mov…

veloce:
Thanks for the new recording!

Can you please:
- send another video where we see the keyboard
- pre-fill the time in custom value input with the currently selected quick settings (which should be 10+0 by default).

lib/src/view/play/time_control_modal.dart
```diff
@@ -33,9 +35,48 @@ class TimeControlModal extends ConsumerWidget {
         child: ListView(
           shrinkWrap: true,
           children: [
-            Text(
-              context.l10n.timeControl,
-              style: Styles.title,
+            Row(
```
veloce, line 38: This file should not be modified because it the one used by the create game screen. We do not want the clock to interfere with it.

You can reuse the code, simply copy/paste it to the `view/clock` folder. Code duplication is not an issue here.

lib/src/view/play/custom_time_control_modal.dart
veloce, line 1: This file belongs in `view/clock` instead.

ijm8710:
@veloce image

Were you thinking more like this? And do you think this picker makes senses for custom game creation as well or do you prefer slider there?

veloce:
Closing it since it has been done here: https://github.com/lichess-org/mobile/commit/84f2cd9e9b3f7fb52bb0af505cde2304d32c1794

Sorry if it wasn't merged finally and thanks for the PR anyway.

Issue #687, opened 2024-05-15
Can no longer play anonymous games


casbarof:
### What is the problem?

As of 0.8.9, it’s no longer possible to play anonymous games. As the website still supports anonymous games, I assume this is a bug.

### Steps to reproduce the bug

Open the app. Cannot play a game without logging in.

### What did you expect to happen?

Can play anonymous games.

### What happened instead?

 Can’t play anonymous games.

### Device Info

iPad Pro M2

### Operating System Version

iPadOS 17.5

veloce:
Can you explain why is it not possible anymore? Do you see an error or something?

ijm8710:
I think they're not aware that you can click the plus icon in top right

casbarof:
“Home” only has two links: “Sign in” and “About lichess”. The controls to start a game only appear once logged in.

ijm8710:
Top right there is a plus sign

casbarof:
@ijm8710 There is no plus sign on my screen.

ijm8710:
> @ijm8710 There is no plus sign on my screen.

Would help to share picture of your home then

casbarof:
@ijm8710 Are you aware that this repo is for the new beta app, and that we’re talking about 0.8.9 specifically?

ijm8710:
> @ijm8710 Are you aware that this repo is for the new beta app, and that we’re talking about 0.8.9 specifically?

Yes I'm aware. But you keep not sharing the picture of the home screen. If Veloce needs to debug why the plus sign is not appearing I'm sure he can look into it but frankly I'm still slightly skeptical you don't have it

casbarof:
Only you asked for it, and you don’t seem to be a maintainer. I don’t need a plus sign to start a game after signing in. Why would I need a plus sign when I’m not signed in? Doesn’t make sense.

casbarof:
![IMG_0069](https://github.com/lichess-org/mobile/assets/169491148/f7a470f7-ea97-4185-bdcd-4a8c079684ae)

ijm8710:
> Only you asked for a screenshot, and you don’t seem to be a maintainer. I don’t need a plus sign to start a game after signing in. Why would I need a plus sign when I’m not signed in? Doesn’t make sense.

Veloce asked if you had any error messages which you didn't provide any.
So first instinct was that maybe you didn't realize that where the new mechanism for the create game is.
And actually, you do need the plus sign to start a game now when logged in.
[...]

PR #688, opened 2024-05-16
fix #2323: Adjust visibility of plus button on iPad Pro [HomeTabScreen]

iamnabink:
Hi @veloce,

I just reproduced this issue #687 on the iPad Pro 6th generation. It looks like the plus button is not displaying on the iPad mainly because of this check `(home_tab_screen.dart)`:

```
if (isHandset)
    ...[
        const SizedBox(width: 6.0),
        AppBarIconButton()
    ]
```
The visibility is controlled by `isHandset()`. I want to clarify whether this is an intended handset check. If we remove that condition, the button will be visible on any device, which should resolve the issue.

I'll go ahead and make a PR. If that's fine with you, please merge it.


veloce:
Fixed it differently, thanks!

Issue #717, opened 2024-05-26
Engine suggestion continues to be displayed after move was made


casbarof:
### What is the problem?

As shown in the screenshot, after mate by promotion to rook, the engine suggestion (promotion to queen) continued to be displayed in game analysis.

![image](https://github.com/lichess-org/mobile/assets/169491148/86ecb2fc-955f-417c-9392-be18359c229a)

### Steps to reproduce the bug

See screenshot

### What did you expect to happen?

Engine suggestion is no longer displayed after a move has been made

### What happened instead?

Engine suggestion continues to be displayed

### Device Info

iPad Pro M2

### Operating System Version

iOS 17.5

veloce:
Fixed by #724

PR #724, opened 2024-05-27
Clear engine suggestion arrows after reaching checkmate or stalemate

timmcca-be:
# Summary

This ensures that we emit an evaluation from `UCIProtocol` once checkmate or stalemate has been reached, addressing https://github.com/lichess-org/mobile/issues/717.

Stockfish sends the message `info depth 0 score mate 0` when checkmate is reached and `info depth 0 score cp 0` when stalemate is reached. Before this change, these messages were being filtered out (by an explicit check for checkmate and because `nodes` and `time` were missing for stalemate), so no evaluation was emitted. This led to the analysis board retaining a stale list of "best moves" and suggesting nonsense.

An alternative approach would be to allow for the stale "best moves" list, and just hide the arrows if the game is over. I prefer the approach taken in this PR, however, because it respects Stockfish as the source of truth for the suggestions.

This assumes that there are no cases in which it is important to filter out a message from Stockfish because it did not include `nodes` or `time`. If such cases do exist, please let me know.

# Repro steps

- Go to Tools > Analysis board
- Click Start
- Play Fool's Mate (1. f3 e6 2. g4 Qh4#)

# Screenshots

### DIFF
--- a/lib/src/model/common/eval.dart
+++ b/lib/src/model/common/eval.dart
@@ -72,7 +72,10 @@ class ClientEval with _$ClientEval implements Eval {
   }
 
   IList<Move?> get bestMoves {
-    return pvs.map((e) => Move.fromUci(e.moves.first)).toIList();
+    return pvs
+        .where((e) => e.moves.isNotEmpty)
+        .map((e) => Move.fromUci(e.moves.first))
+        .toIList();
   }
 
   @override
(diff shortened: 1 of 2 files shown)


veloce:
Thanks for this PR!

> This assumes that there are no cases in which it is important to filter out a message from Stockfish because it did not include nodes or time. If such cases do exist, please let me know.

I'm not able to answer this question. Not sure Stockfish output is a 100% consistent across versions also.

I'd say your fix is more logic, so let's give it a go?

veloce (approved):

PR #726, opened 2024-05-29
Broadcast feature, first iteration

julien4215:
[broadcast_dark.webm](https://github.com/lichess-org/mobile/assets/120588494/dbc0751a-68b0-42c0-8c23-dd8e4a75adbc)


veloce:
Looks very good! Thank you. I think we should see the big images even in watch tab root, without having to tap to see more.

What do you plan to add on top of this? I think we should limit the set of features for this PR, that'll be easier to review. 

I can review the current state (feature wise) and the rest would be done in other PRs.

lib/src/view/broadcast/broadcast_screen.dart
```diff
@@ -0,0 +1,136 @@
...
+      iosBuilder: _iosBuilder,
+    );
+  }
+
+  Widget _androidBuilder(
+    BuildContext context,
+  ) {
+    return Scaffold(
+      appBar: AppBar(
+        title: const Text('Broadcast'),
```
veloce, line 33: We should translate this. I'll write the documentation that explain how to add a new i18n module, so you can do it yourself and verify my doc is good ;)

lib/src/widgets/layout_board.dart
```diff
@@ -0,0 +1,43 @@
+import 'package:flutter/widgets.dart';
+import 'package:flutter_layout_grid/flutter_layout_grid.dart';
+import 'package:lichess_mobile/src/widgets/board_thumbnail.dart';
+
+class LayoutBoard extends StatelessWidget {
```
veloce, line 5: I'd name it `BoardGrid` it's more explicit.

lib/src/view/broadcast/broadcast_screen.dart
veloce: Here is the doc: https://github.com/lichess-org/mobile/blob/main/docs/internationalisation.md

julien4215:
I'd like to use the radio tower icon as on the website but I didn't find it in the Lichess icons of the mobile app and I don't know how to add it

veloce:
The radio tower in the website correspond to the study feature, and broadcast are actually implemented as a study. So I'm not sure but this may be the reason why there is that icon in the website.

We won't use it for broadcast in the app, so don't bother with it.

I think this PR should not evolve anymore than just the list of broadcasts and the list of boards for now. It's already big enough to review.

The feature will not be ready yet to be released but we can merge the code when it's ready and hide it, so you can continue to work in another PR when this one is done.

julien4215:
I wanted to use the radio tower to replicate this default image for broadcast like here
image
I thought it would make sense to keep the same default image (a linear gradient with the radio tower icon).

julien4215:
> Looks very good! Thank you. I think we should see the big images even in watch tab root, without having to tap to see more.
> 
> What do you plan to add on top of this? I think we should limit the set of features for this PR, that'll be easier to review.
> 
> I can review the current state (feature wise) and the rest would be done in other PRs.

I prefer to keep the list section because it is more coherent with the streamer and live games sections but I can change it if you think it would be better with the big images.

I still don't really know how to proceed after this. I think the biggest challenge is to find an easy way to navigate between all the pages. I think there will be at least 3 pages for a tournament, maybe more.

Also. in the grid board the positions are not updated, there is only one call which is made when the page is opened where I get the current fen for each. Do you want this to be implemented before merging the PR or should it be made in a different one ?

ijm8710:
image

My initial suggestion is that once a broadcast is open, it'll have horizontally slidable panels similar to the referenced image (but can also be tapped)

Panels eventually being:
- overview
- boards
- leaderboard 

The leaderboard would be the tab that down the line could have player links clickable to expand to a deep dive screen to see round by round vertical list performance  for said player

Understood that this design would slightly deviate from the rest of the app but I think regardless that's going to have to happen

lib/src/view/broadcast/broadcast_screen.dart
julien4215: Thanks for the documentation. The only issue I had is that I didn't have nodejs nor the modules used by the script so I had to installed them. Maybe you could add in the documentation the version of nodejs that this script works with.

scripts/update-arb-from-crowdin.mjs
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env node
+#!/home/julien/Downloads/node-v20.14.0-linux-x64/bin/node
```
veloce, line 1: Oops! the `/usr/bin/env node` shebang doesn't work for you? Still it's the most common. But yours will certainly not work on other machines ;)

julien4215: Good catch ! I need to be more careful with my commits, I am going to commit sensitive information one of these days.

I installed nodejs on my machine to run this script but I don't want to keep it so I didn't bother adding it to the path variable.

veloce:
Nice!

> Should I add an hidden option to activate the broadcast feature like you activate the developer mode on Android by tapping several times the build number ?

No need to make sth that advanced. But an option that hide the feature by default and you can activate using a --dart-define would be nice.

julien4215:
> Nice!
> 
> > Should I add an hidden option to activate the broadcast feature like you activate the developer mode on Android by tapping several times the build number ?
> 
> No need to make sth that advanced. But an option that hide the feature by default and you can activate using a --dart-define would be nice.

The idea was to allow also the users to activate the broadcast feature if they want too

ZTL-UwU:
Shouldn't it be on by default?

veloce:
Thank you for putting this together! I pushed a new commit with the radio tower asset to `main`, you can now rebase your branch if you want to use it.

julien4215:
> Thank you for putting this together! I pushed a new commit with the radio tower asset to `main`, you can now rebase your branch if you want to use it.

Thanks for the icon.

What should about the visibility of this feature ? (I don't know if you have seen the discussion above before the list of commit)

lib/src/model/broadcast/broadcast_repository.dart
```diff
@@ -0,0 +1,89 @@
...
+      Uri(path: '/api/broadcast'),
+      headers: {'Accept': 'application/x-ndjson'},
+      mapper: _makeBroadcastFromJson,
+      compare: (Broadcast a, Broadcast b) => b.priority.compareTo(a.priority),
+    );
+  }
+
+  Future<IList<BroadcastGameSnapshot>> getRound(String broadcastRoundId) {
+    return client.readJson(
+      Uri(path: 'api/broadcast/-/-/$broadcastRoundId'),
```
veloce, line 25: Add a comment explaining why these path param are replaced by `-`.

lib/src/model/broadcast/broadcast_repository.dart
```diff
@@ -0,0 +1,89 @@
...
+  BroadcastRepository(this.client);
+
+  final LichessClient client;
+
+  Future<IList<Broadcast>> getBroadcasts() {
+    return client.readNdJsonList(
+      Uri(path: '/api/broadcast'),
+      headers: {'Accept': 'application/x-ndjson'},
+      mapper: _makeBroadcastFromJson,
+      compare: (Broadcast a, Broadcast b) => b.priority.compareTo(a.priority),
```
veloce, line 19: I believe the sorting could be done server side.

lib/src/model/broadcast/broadcast_repository.dart
```diff
@@ -0,0 +1,89 @@
...
+  return Broadcast(
+    tour: Tour(
+      name: pick('tour', 'name').asStringOrThrow(),
+      description: pick('tour', 'description').asStringOrThrow(),
+      imageUrl: pick('tour', 'image').asStringOrNull(),
+      tier: pick('tour', 'tier').asIntOrThrow(),
+    ),
+    rounds: pick('rounds')
+        .asListOrEmpty(_roundFromPick)
+        .sorted((Round a, Round b) => a.startsAt.compareTo(b.startsAt))
```
veloce, line 45: Should be sorted server side as well.

lib/src/model/broadcast/broadcast.dart
```diff
@@ -0,0 +1,91 @@
...
+}
+
+@freezed
+class BroadcastPlayer with _$BroadcastPlayer {
+  const factory BroadcastPlayer({
+    required String name,
+    required String? title,
+    required int? rating,
+    required Duration? clock,
+    required String? fed,
```
veloce, line 76: We prefer full names in the app, so `federation`.

lib/src/view/broadcast/game_screen.dart
```diff
@@ -0,0 +1,218 @@
...
+import 'package:lichess_mobile/src/utils/chessground_compat.dart';
+import 'package:lichess_mobile/src/utils/duration.dart';
+import 'package:lichess_mobile/src/utils/l10n_context.dart';
+import 'package:lichess_mobile/src/utils/lichess_assets.dart';
+import 'package:lichess_mobile/src/widgets/board_thumbnail.dart';
+import 'package:lichess_mobile/src/widgets/grid_board.dart';
+import 'package:lichess_mobile/src/widgets/platform.dart';
+import 'package:lichess_mobile/src/widgets/shimmer.dart';
+
+class GameScreen extends ConsumerWidget {
```
veloce, line 20: It would be better named `BroadcastGameScreen` to avoid any ambiguity.

lib/src/view/broadcast/game_screen.dart
```diff
@@ -0,0 +1,218 @@
...
+  final String roundId;
+
+  const _Body(this.roundId);
+
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {
+    final games = ref.watch(roundProvider(roundId));
+    return games.when(
+      data: (games) => BroadcastPreview(games: games),
+      loading: () => const Shimmer(
```
veloce, line 66: 👍

lib/src/model/broadcast/broadcast_repository.dart
julien4215: I asked to backend developers if they could make the sort on server side

lib/src/model/broadcast/broadcast.dart
```diff
@@ -0,0 +1,91 @@
...
+  }
+}
+
+@freezed
+class Tour with _$Tour {
+  const factory Tour({
+    required String name,
+    required String description,
+    required String? imageUrl,
+    required int tier,
```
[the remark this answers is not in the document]
julien4215: Since the tournaments are now sorted, I don't need anymore the tier field

lib/src/view/watch/watch_tab_screen.dart
```diff
@@ -176,11 +187,66 @@ class _WatchScreenState extends ConsumerState<WatchTabScreen> {
...
+  static const int numberOfItems = 5;
+
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {
+    final broadcastList = ref.watch(broadcastsProvider);
+
+    return broadcastList.when(
+      data: (data) {
+        return ListSection(
+          header: const Text('Tournament broadcasts'),
```
veloce, line 208: Missing translation.

veloce:
Thanks again for this work @julien4215 ! Code is LGTM. Now I'm gonna test it, can you please update your branch (rebase to last main commit)?

lib/src/view/watch/watch_tab_screen.dart
julien4215: There is no translation for 'Tournament broadcasts'. Should I just use 'Broadcasts' instead ?

julien4215:
Right now the sorting is still done on the client side. ornicar created a new endpoint that includes the sorting https://lichess.org/api/broadcast/top but it also quite different from the previous endpoint as it adds pagination and group for tournaments.

veloce:
I think we're gonna want to use this new endpoint and add pagination support. 

Is it a lot of work for you to change that? Do you prefer me to test this PR as is and do the change in another one?

julien4215:
I would prefer to do the change in a different PR because it will require me a lot work to use the new endpoint.

veloce:
Couple of notes from my live test:

- there are differences between what I see on the website and the app.
- first there is a grouping issue:
Capture d’écran 2024-06-19 à 10 38 23

Here we're seeing several "Asian young championship" entries, where we should only see one.

- there is no separation between live broadcast, upcoming broadcast and past broadcasts, like on web
- I see a "Norway chess 2024 | Women" entry on the app and not on website

If you tap on a broadcast, it brings you to the games screen, but the title is wrong:

Capture d’écran 2024-06-19 à 10 43 33

Here it shows "Diffusions", we want to see the current broadcast title instead.

julien4215:
The grouping issue requires the new endpoint https://lichess.org/api/broadcast/top that ornicar created.

veloce:
Let's fix the title issue and I'm merging it then.

PR #752, opened 2024-06-07
fix: twice as slow text update frequency

### LINKED ISSUE #738: Frequency of text update
https://github.com/lichess-org/mobile/assets/155185001/92ce9e24-662d-4430-bda5-013534bf2746

Please, fix frequency update

Mauritz8:
close #738. I'm not sure what the sweet spot is for the duration, and it is quite hard to test in a realistic setting, but making it half as fast seems resonable.

### DIFF
--- a/lib/src/view/game/game_loading_board.dart
+++ b/lib/src/view/game/game_loading_board.dart
@@ -271,7 +271,7 @@ class _AnimatedLobbyNumberState extends State<_AnimatedLobbyNumber> {
         end: value,
       ),
       curve: Curves.linear,
-      duration: const Duration(seconds: 3),
+      duration: const Duration(seconds: 6),
       builder: (context, int value, _) {
         return Text(widget.labelBuilder(value), style: _lobbyNumbersStyle);
       },


lib/src/view/game/game_loading_board.dart
```diff
@@ -271,7 +271,7 @@ class _AnimatedLobbyNumberState extends State<_AnimatedLobbyNumber> {
         end: value,
       ),
       curve: Curves.linear,
-      duration: const Duration(seconds: 3),
+      duration: const Duration(seconds: 6),
```
veloce, line 274: I'm really not sure it fixes the issue. Were you able to reproduce it?

Mauritz8: I can see that the counter changes less frequently when I open/close many tabs with lichess at the same time. I don't know if it's possible to simulate the sheer amount of people opening and closing lichess at infrequent intervals, like what the app would observe in prod. Still, it shouldn't make a difference because there are never multiple transitions happening at the same time, since the widget is rebuilt every time the state changes.

veloce:
Closed by 4cd0e8cbabd971f1dd5e95acb21901821d25b73a

PR #767, opened 2024-06-10
adjust volume

### LINKED ISSUE #749: Sound isn't loud enough
Sound could be louder. No need for a sound control (a single value is fine), but the sound isn't loud enough.

Mauritz8:
close #749 
Screenshot from 2024-06-10 15-05-13
image


veloce:
Thanks for this. I'm not a fan of the current plugin API where `setVolume` is done each time before `play`; I'll try to change that first.

veloce:
If you want to have a go at it, it is here: https://github.com/lichess-org/flutter-soundpool

On lichess branch.

lib/src/model/common/service/sound_service.dart
```diff
@@ -93,6 +94,7 @@ class SoundService {
               _currentStream!.$2 == Sound.capture)) {
         await _pool.stop(_currentStream!.$1);
       }
+      await _pool.setVolume(soundId: soundId, volume: volume);
```
veloce, line 97: Play should have the lowest latency possible since we're playing short sounds very fast during game or replay.

So volume should not be set here.

Instead keep a reference to `masterVolume` in `SoundService` and have a `setVolume` method which loops over all loaded sound to change their volume one by one.

If needed at some point we could even tune the volume of each sound individually.

lib/src/model/settings/general_preferences.dart
```diff
@@ -87,6 +92,7 @@ class GeneralPrefsState with _$GeneralPrefsState {
     themeMode: ThemeMode.system,
     isSoundEnabled: true,
     soundTheme: SoundTheme.standard,
+    volume: 0.5,
```
veloce, line 95: Have you tested that the 0.5 default? How does it compare with the current `null` default?

lib/src/view/settings/settings_screen.dart
```diff
@@ -456,3 +493,55 @@ class _Body extends ConsumerWidget {
...
+  @override
+  Widget build(BuildContext context) {
+    return AlertDialog(
+      title: Text(widget.title),
+      content: StatefulBuilder(
+        builder: (context, setState) {
+          return Column(
+            mainAxisSize: MainAxisSize.min,
+            children: [
+              NonLinearSlider(
```
veloce, line 526: I think you could use flutter's `Slider.adaptive` instead since that one is linear.

lib/src/model/settings/general_preferences.dart
Mauritz8: I can't tell them apart, so it should be about the same.

lib/src/view/settings/settings_screen.dart
Mauritz8: Fixed

lib/src/model/common/service/sound_service.dart
Mauritz8: I've updated the implementation now, hopefully it's what you were looking for.

veloce:
I found a problem in the `soundpool` package. I've started to work on a new sound package that will replace it.

Mauritz8:
Will it fix the iOS simulator issue, or is that another matter? I haven't been able to work on it, since I don't have an apple device.

veloce:
Yes it will fix it. It’s not just a simulator issue but a general issue on iOS.

Issue #763, opened 2024-06-09
Add piece animation speed setting


EmmetSchuler:
Although the app already has a toggle for animating the piece movements, there is not yet an option to change how fast or slow they move. This functionality does appear on the website.

![Screenshot_20240609-102731.png](https://github.com/lichess-org/mobile/assets/108140722/f2a911f0-8712-416b-bb7d-eb079aef1903)

veloce:
Is it really useful to have different speed?

EmmetSchuler:
In my opinion, definitely! When people tap to move pieces, especially in faster time controls they'd like it to respond quicker. And some people also prefer slower and more calm movements when playing in general. I think this is pretty important for the overall enjoyment of many different types of players.

EmmetSchuler:
It might also be good to set default animation speeds for each time control. Seems like on the website default values for bullet are instant movement, 3+0 is fast movement and 5+3 is around regular movement.
In all honesty I don't really understand how the movement speed setting works. From what I've tested it seems to always be instant movement in bullet, even when I set the animation speed to "slow". Maybe specific time controls force a different speed, and only in classical, rapid, and slower blitz can you change the speeds?

veloce:
I didn't know that instant movement was the default in bullet. 

> Maybe specific time controls force a different speed, and only in classical, rapid, and slower blitz can you change the speeds?

I think you're right. It has to be fixed in bullet, otherwise there would be an option per time control, and we don't want to go that far.

veloce:
Actually I checked the server code to understand, and it's computed like that:

https://github.com/lichess-org/lila/blob/9a826e85e7fb67fdb76c464b9ed5f04dac1e60f6/modules/round/src/main/JsonView.scala#L337-L341

EmmetSchuler:
Don't forget about puzzle storm, as well. Seems like that defaults to instant move speed too

PR #773, opened 2024-06-11
fix: disable piece animation in bullet time controls

### LINKED ISSUE #757: Disable piece animation by default in bullet time controls
### What is the problem?

In bullet games I would expect the piece movement to be as fast as in the web and legacy mobile app. Instead it's slow as in longer time control games, which is very confusing.

Play bullet. Make moves.

Pieces should move very fast to the destination squares.

### What happened instead?

Pieces move very slowly.

### Device Info

Irrelevant

### Operating System Version

Android

### Additional information

This is from the beta app installed from Android's app store, a couple of days ago. I won't be playing bullet with this app again until this is resolved, so if it was fixed since then I don't know.

Mauritz8:
close #757

### DIFF
--- a/lib/src/view/game/game_body.dart
+++ b/lib/src/view/game/game_body.dart
@@ -210,6 +210,12 @@ class GameBody extends ConsumerWidget {
         final topPlayer = topPlayerIsBlack ? black : white;
         final bottomPlayer = topPlayerIsBlack ? white : black;
 
+        final animationDuration =
+            gameState.game.meta.speed == Speed.ultraBullet ||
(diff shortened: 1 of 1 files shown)


veloce:
I think this one will be a little more complicated. I'd like it to follow server preference while still working offline.

Mauritz8:
To my understanding, these changes will not alter the offline mode behavior. If that needs to be changed, then that seems like a different task altogether.

veloce (approved):

veloce:
Agreed this is another task. Until we deal with #763 this is better than nothing. Thanks!

PR #774, opened 2024-06-11
only show next correct move when viewing puzzle solution

### LINKED ISSUE #761: Puzzles: show next correct move
I quite liked the way that the old app only gave you the next correct move in a puzzle when you tapped "show solution". That enabled me to think about the next move in the sequence until I had the puzzle fully solved, even though it was counted as wrong.
Would it be feasible to add a "show next move" button in addition to the "show solution" that is there, currently?

Mauritz8:
close #761 
This alters the "show solution button" to only show the next move in the solution, which is in line with both the website and the old app.

### DIFF
--- a/lib/src/model/puzzle/puzzle_controller.dart
+++ b/lib/src/model/puzzle/puzzle_controller.dart
@@ -181,7 +181,9 @@ class PuzzleController extends _$PuzzleController {
       mode: PuzzleMode.view,
     );
 
-    _goToNextNode();
+    Timer(const Duration(milliseconds: 800), () {
+      _goToNextNode();
+    });
   }
 
   void skipMove() {
--- a/lib/src/model/puzzle/puzzle_controller.dart
+++ b/lib/src/model/puzzle/puzzle_controller.dart
@@ -106,6 +106,7 @@ class PuzzleController extends _$PuzzleController {
       resultSent: false,
       isChangingDifficulty: false,
       isLocalEvalEnabled: false,
+      viewedSolutionRecently: false,
       streak: streak,
       nextPuzzleStreakFetchError: false,
       nextPuzzleStreakFetchIsRetrying: false,
@@ -179,10 +180,17 @@ class PuzzleController extends _$PuzzleController {
 
     state = state.copyWith(
       mode: PuzzleMode.view,
+      viewedSolutionRecently: true,
     );
 
     Timer(const Duration(milliseconds: 800), () {
       _goToNextNode();
+    });
+
+    Timer(const Duration(seconds: 3), () {
+      state = state.copyWith(
+        viewedSolutionRecently: false,
+      );
     });
   }
 
@@ -553,6 +561,7 @@ class PuzzleState with _$PuzzleState {
     required bool isLocalEvalEnabled,
     required bool resultSent,
     required bool isChangingDifficulty,
+    required bool viewedSolutionRecently,
     PuzzleContext? nextContext,
     PuzzleStreak? streak,
     // if the automatic attempt to fetch the next puzzle in the streak fails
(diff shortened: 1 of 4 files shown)


juan-campuzano (approved):

lib/src/model/puzzle/puzzle_controller.dart
```diff
@@ -181,14 +181,7 @@ class PuzzleController extends _$PuzzleController {
...
     );
 
-    _viewSolutionTimer =
-        Timer.periodic(const Duration(milliseconds: 800), (timer) {
-      if (state.canGoNext) {
-        _goToNextNode();
-      } else {
-        timer.cancel();
-      }
-    });
```
veloce, line 191: I think a slight delay before jumping to next move is still desirable.

veloce (commented):
I think we want this but we could improve it a little more.

We could for instance have the forward button glow or blink for a short period (3s) to visually indicate that one can go forward.

There is already an example of blinking button in game screen when a draw offer is received.

Mauritz8:
> There is already an example of blinking button in game screen when a draw offer is received.

For me the draw offer button is just colored differently, it doesn't blink or anything like that. 
Is there another example of a blinking button?
image

veloce:
It doesn't blink? My bad then :)

Any visual indication would be good actually, blinking is not mandatory. Perhaps a simple background color inversion for a duration of 3s would be the best?

veloce:
If you have the time to post a screen recording showing the feature and button visual indicator in action that would be great. Thanks!

Mauritz8:
I tried both blinking and color inversion, and I think blinking works a lot better. And I set the duration to 5 seconds instead of 3 because it felt a bit too short.
[Screencast from 2024-07-05 12:20:21.webm](https://github.com/lichess-org/mobile/assets/67279312/7718e7d1-8d57-4864-a95c-6550646034cb)

lib/src/widgets/bottom_bar_button.dart
```diff
@@ -64,10 +66,16 @@ class BottomBarButton extends StatelessWidget {
                   mainAxisAlignment: MainAxisAlignment.center,
                   crossAxisAlignment: CrossAxisAlignment.center,
                   children: [
-                    Icon(
-                      icon,
-                      color: highlighted ? primary : null,
-                    ),
+                    if (blink) ...[
```
veloce, line 69: As there is a single child, I don't think the `...[` is necessary.

lib/src/widgets/bottom_bar_button.dart
```diff
@@ -64,10 +66,16 @@ class BottomBarButton extends StatelessWidget {
...
+                      _BlinkIcon(
+                        icon: icon,
+                        color: highlighted
+                            ? primary
+                            : Theme.of(context).iconTheme.color ?? Colors.black,
+                      ),
+                    ] else ...[
```
veloce, line 76: Same: unecessary `...[`

test/view/puzzle/puzzle_screen_test.dart
```diff
@@ -415,11 +415,11 @@ void main() {
         await tester.tap(find.byIcon(Icons.help));
 
         // wait for solution replay animation to finish
-        await tester.pump(const Duration(seconds: 3));
+        await tester.pump(const Duration(seconds: 1));
         await tester.pumpAndSettle();
 
         expect(find.byKey(const Key('h4-blackRook')), findsOneWidget);
-        expect(find.byKey(const Key('h8-whiteQueen')), findsNothing);
+        expect(find.byKey(const Key('h8-whiteQueen')), findsOneWidget);
```
veloce, line 422: We could add another check to show the "jump to next move" button is not disabled.

Mauritz8:
I addressed your comments. I'm a bit uncertain about the test, it might not be done in the most idiomatic way.

veloce (approved):

veloce:
Thanks!

PR #789, opened 2024-06-27
Put the resign button in bottom bar for easy access

veloce:


dennisjaheruddin:
Great to see these small changes made to improve usability. 
If I can note 1 suggestion of a related change which has been mentioned by me and others a few times but keeps getting lost in the storm:
On mobile (I use web but the same holds for the app) it would be great to have an option to put the clock on the left hand side.
Most players are right handed and in bullet/blitz your own finger makes it hard to see your clock.

Issue #759, opened 2024-06-08
Move Notation preference setting not followed in new app


LukeyPmoney:
### What is the problem?

In the new app, the list of moves is always in letter notation when playing a game, but the analysis board (both the separate tool or within a game) always shows moves in figurine notation. Also, this preference setting can only be updated on the website, not in the app.

### Steps to reproduce the bug

On the website, click/tap profile icon -> preferences -> display. Change the Move Notation setting to the opposite of what it was and save.
Next, start a game in the lichess app and make a non-pawn move. Click analysis icon or switch to analysis board tool and make a non-pawn move.

### What did you expect to happen?

Both in an active game and on the analysis board, the move notation preference setting should be respected. Also, this preference should be editable within the app instead of only on website.

### What happened instead?

In-game in the new lichess app: game moves notation is in letter form regardless of move notation setting. Switch to analysis board, either the separate tool or within a game: game moves and moves within analysis lines are always in figurine notation, regardless of move notation preference setting.

### Device Info

Samsung Galaxy S7

### Operating System Version

Android 8.0.0

eliogilvy:
@veloce which api call handles this? I'm looking at api/account/preferences and not seeing anything... I see it in
[Scala file](https://github.com/lichess-org/lila/blob/master/modules/pref/src/main/Pref.scala) but am not seeing the field when talking to the api/account/preferences.

awan1:
Verified on Android 14, Pixel 8

PR #792, opened 2024-06-27
feat: support disabling clock emergency sound

### LINKED ISSUE #785: Feature: clock warning sounds
As a user playing a timed real-time game, when my remaining time decreases below some threshold (described below), I would like to hear a warning sound.

As a user of the app, I would like to be able to configure whether the app makes this warning sound or not from the settings.

The threshold can be fixed (e.g. 30sec, 10sec, 5sec) or depend on the game format played. Ideally this behavior should match the web app.

tom-anders:
Closes #785

Needs https://github.com/lichess-org/lila/pull/15613

### DIFF
--- a/lib/src/widgets/countdown_clock.dart
+++ b/lib/src/widgets/countdown_clock.dart
@@ -17,7 +17,10 @@ class CountdownClock extends ConsumerStatefulWidget {
 
   /// If [timeLeft] is less than [emergencyThreshold], the clock will change
   /// its background color to [ClockStyle.emergencyBackgroundColor] activeBackgroundColor
+  /// If [emergencySoundEnabled] is `true`, the clock will also play a sound.
   final Duration? emergencyThreshold;
+
+  final bool emergencySoundEnabled;
 
   /// If [active] is `true`, the clock starts counting down.
   final bool active;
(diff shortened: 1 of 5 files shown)


tom-anders:
(fixed formatting)

lib/src/widgets/countdown_clock.dart
```diff
@@ -55,6 +59,7 @@ const _emergencyDelay = Duration(seconds: 20);
 class _CountdownClockState extends ConsumerState<CountdownClock> {
   Timer? _timer;
   Duration timeLeft = Duration.zero;
+  bool _emergencySoundEnabled = true;
```
veloce, line 62: You don't need that local property since it doesn't change. Use directly `widget.emergencySoundEnabled` instead.

veloce (commented):
Thanks! Will wait for the deploy of server change before merging this.

tom-anders:
Thanks for the quick feedback! Addressed all your comments.

@veloce if you don't mind, I might have a look at https://github.com/lichess-org/mobile/issues/759 next, it seems to require similar changes (probably also a similar server change as well

veloce (approved):
lgtm

PR #798, opened 2024-06-28
fix text being cut off in stormNewRun and stormEndRun translations

tom-anders:
In English, these strings are something like "End run (hotkey: ...)" and "New run (hotkey: ...)". The previous logic discarded the text in parenthesis by simply keeping the first two words.

However, in some languages we may have more than two words (e.g. in German it's "Beende den Run"). With this change, we keep everything until the '(' character, which should be more robust.

If the string for some reason does not contain '(' at all, split('(') will return a list containing the string itself, so that will work too.


tom-anders:
@veloce I haven't included any before/after screenshots, I think the change is trivial enough for that (and seems to also be covered by a unit test).

Issue #797, opened 2024-06-28
Add possibility to challenge someone from a position


veloce:
In the challenge setup screen, add the "From Position" entry to the "Variant" selector.

If the "From position" is chosen, display a readonly text field used to past a FEN. Once a valid FEN is loaded show the position as mini board (orientation from the selected side).

PR #804, opened 2024-06-29
Challenge from a Position with Fen

### LINKED ISSUE #803: Able to challenge with 0+0 leading to unrecoverable error
### What is the problem?

Users are able to challenge to a 0+0 game, leading to unrecoverable error.

1. Go to the challenge screen
2. Set timecontrol to 0+0

Users are not allowed to press the challenge button

### What happened instead?

I was able to challenge, leading to an error. 

### Device Info

-

### Operating System Version

-

### Additional information

_No response_

HaonRekcef:
implements #797
closes #803
Adds the possibility to challenge another player with a fen:
Behavior with an invalid fen: 

https://github.com/lichess-org/mobile/assets/78898963/a156e151-48af-4638-bc00-7548ea156520

Behavior with a valid fen:

https://github.com/lichess-org/mobile/assets/78898963/57f585cc-be6b-4182-b3d7-d968cbf16af4

Some remarks:
- From position games can not be rated, so the button is disabled
- The last position used to challenge is not stored (in my opinion this is not necessary)
- Assumes that the fen checker is the same as employed on the serverside, otherwise it might lead to strange behavior.
- Was not tested on iOS

### DIFF
--- a/lib/src/view/play/challenge_screen.dart
+++ b/lib/src/view/play/challenge_screen.dart
@@ -1,8 +1,12 @@
 import 'dart:async';
 
+import 'package:chessground/chessground.dart' as cg;
+import 'package:dartchess/dartchess.dart';
 import 'package:flutter/cupertino.dart';
 import 'package:flutter/material.dart';
+import 'package:flutter/services.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
+import 'package:lichess_mobile/src/constants.dart';
 import 'package:lichess_mobile/src/model/account/account_repository.dart';
 import 'package:lichess_mobile/src/model/challenge/challenge.dart';
 import 'package:lichess_mobile/src/model/challenge/challenge_preferences.dart';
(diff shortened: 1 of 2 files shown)


veloce:
Thanks for taking the time to create this! Multiple screen recording and remarks as you do are always very appreciated. I'll review it asap but for now I can say it looks very good.

lib/src/view/play/challenge_screen.dart
```diff
@@ -56,6 +63,25 @@ class _ChallengeBody extends ConsumerStatefulWidget {
 
 class _ChallengeBodyState extends ConsumerState<_ChallengeBody> {
   Future<void>? _pendingCreateGame;
+  final _controller = TextEditingController();
+
+  String? textInput;
```
veloce, line 68: `textInput` is too vague, prefer, for a local property, a long named describing exactly what it is representing.

lib/src/view/play/challenge_screen.dart
```diff
@@ -332,15 +382,23 @@ class _ChallengeBodyState extends ConsumerState<_ChallengeBody> {
                     child: FatButton(
                       semanticsLabel: context.l10n.challengeChallengeToPlay,
                       onPressed: timeControl == ChallengeTimeControlType.clock
-                          ? isValidTimeControl
+                          ? isValidTimeControl &&
+                                  ((textInput != '' && textInput != null) ||
```
veloce, line 386: I think you can use `textInput.isNotEmpty` instead of comparing with empty string.

PR #806, opened 2024-06-29
Use the new broadcast/top endpoint

julien4215:
[recording.webm](https://github.com/lichess-org/mobile/assets/120588494/129dc412-3c45-4ce3-be07-369da1d47b16)


veloce:
I'll review asap, in the meantime can you re-run the script to solve the conflicts? thanks!

lib/src/model/broadcast/broadcast.dart
```diff
@@ -4,46 +4,45 @@ import 'package:freezed_annotation/freezed_annotation.dart';
...
+    required String? group,
   }) = _Broadcast;
 
-  BroadcastRound? get curentRound =>
-      rounds.where((r) => r.status == RoundStatus.live).firstOrNull ??
-      rounds.where((r) => r.status == RoundStatus.finished).lastOrNull;
+  factory Broadcast.loading() => Broadcast(
+        tour: const BroadcastTournament(name: '', imageUrl: null),
+        lastRound: BroadcastRound(
+          id: '',
```
veloce, line 29: I missed that in the last PR, but we should avoid `String` for ids and use an extension type instead. See `common/id.dart` for example.

lib/src/model/broadcast/broadcast.dart
```diff
@@ -4,46 +4,45 @@ import 'package:freezed_annotation/freezed_annotation.dart';
...
     required BroadcastTournament tour,
-    required IList<BroadcastRound> rounds,
+    required BroadcastRound lastRound,
+    required String? group,
   }) = _Broadcast;
 
-  BroadcastRound? get curentRound =>
-      rounds.where((r) => r.status == RoundStatus.live).firstOrNull ??
-      rounds.where((r) => r.status == RoundStatus.finished).lastOrNull;
+  factory Broadcast.loading() => Broadcast(
```
veloce, line 26: Instead of that `Broadcast.loading()` that has empty name and id, is it possible to have a nullable `broadcast` field where the broadcast is not yet present?
I believe it is only used in `BroadcastPicture` for now, so it should be possible.

veloce (commented):
Nice job! only have a couple of comment for now;

Can you send a screen recording showing the new list and pagination? thanks.

julien4215:
https://github.com/lichess-org/mobile/assets/120588494/36533e9c-d2c7-4e52-88bf-b9a2aca087c1

julien4215:
https://github.com/lichess-org/mobile/assets/120588494/3871661f-c1f1-4cf5-93c5-77633e5d63ea

lib/src/model/broadcast/broadcast.dart
julien4215: Should I create a BroadcastRoundId ?

lib/src/model/broadcast/broadcast.dart
julien4215: Where should the `broadcast` field be nullable ? In the `BroadcastPicture` ?

veloce: Yes, in order to not use the `Broadcast.loading()`

julien4215: But why not keep it ? It is easier to create the shimmer like this.

I can also do it with defining a nullable broadcast field in the `BroadcastPicture` but I am not sure that this will make the code cleaner.

lib/src/view/broadcast/broadcast_tournament_screen.dart
```diff
@@ -47,58 +47,156 @@ class BroadcastTournamentScreen extends StatelessWidget {
...
   }
 }
 
 class BroadcastPicture extends StatelessWidget {
   final Broadcast broadcast;
 
-  const BroadcastPicture({super.key, required this.broadcast});
+  const BroadcastPicture({required this.broadcast});
+
+  BroadcastPicture.loading() : broadcast = Broadcast.loading();
```
veloce, line 188: It could use a const I think.

lib/src/model/broadcast/broadcast.dart
veloce: Ok let's use a dummy `Broadcast` in the `BroadcastPicture` if it's easier, but don't define the factory constructor here; we don't want it to be used elsewhere.
Just construct the dummy one directly in the view (and you can use a `const`).

lib/src/model/broadcast/broadcast_providers.dart
```diff
@@ -2,20 +2,50 @@ import 'package:fast_immutable_collections/fast_immutable_collections.dart';
...
-      .withClient((client) => BroadcastRepository(client).getBroadcasts());
+class BroadcastPage extends _$BroadcastPage {
+  @override
+  int build() {
+    return 1;
+  }
+
+  void next() {
+    if (state < 20) {
+      // 20 is the last page
```
veloce, line 19: You can precise the comment I think. Like: "server will return max 20 pages".

veloce: Or better use the doc comment `///` on top of the `BroadcastPage` class.

lib/src/model/broadcast/broadcast_providers.dart
```diff
@@ -2,20 +2,50 @@ import 'package:fast_immutable_collections/fast_immutable_collections.dart';
...
+    final page = ref.watch(broadcastPageProvider);
+
+    final broadcastResponse = await ref.withClient(
+      (client) => BroadcastRepository(client).getBroadcasts(page: page),
+    );
+
+    if (page == 1) {
+      return broadcastResponse;
+    } else {
+      return state.requireValue.copyWith(
```
veloce, line 38: There is a risk that this provider is initialised with a page that is not 1 and then it will throw an exception.

julien4215: I don't think it can happen because I watch the `broadcastListProvider` before the `broadcastPageProvider`.

Do you have an example of how it could happen?

veloce: Don't know, it just feels dangerous. It may be ok now, but could be not ok in the future, or with some edge case. We have to think about maintenance.

Also it is the first time I see this pattern where the previous state is used in the `build` method, which is not what we normally do.

The way it works is clever but I think it can be simplified. `BroadcastPage` and `BroadcastList` can be merged together. The fewer providers the better. The problem with your solution (on top of what I mentioned earlier) is that it is not obvious that state is accumulated from an `int` provider. And the page provider is a bit confusing since we don't really need to remember the page we're at, but the last downloaded page. It can be part of the `BroadcastList` state, by returning a tuple for instance. And we don't need to use our own logic for the `page` since the server handles it. Use the `currentPage` and `nextPage` fields in the json response.

lib/src/model/broadcast/broadcast.dart
julien4215: I can't use a `const` because DateTime doesn't have a const constructor

lib/src/view/broadcast/broadcast_tournament_screen.dart
julien4215: I can't because the value I assign to broadcast is not a const value due to the presence of Datetime in `Broadcast`

veloce:
> I added a delay of 5 seconds to the request in order to have time to see the loading behavior when reaching next page

5 seconds seems a lot. Can you try with 2 seconds?

julien4215:
> > I added a delay of 5 seconds to the request in order to have time to see the loading behavior when reaching next page
> 
> 5 seconds seems a lot. Can you try with 2 seconds?

I added it only for the video so we can see better what happens with the shimmer. Do you want me to record again with 2 seconds instead of 5 ?

ijm8710:
Hey guys, web has a row or two of smaller detail text below each of the larger tournament title.

Seems like they're still experimenting what to show on the tournament posters but these have included:
    - tournaments' location
    - tournament time control
    - tournament dates
    - list of players
    - current round #/time until round starts

There is also a red icon that shows # of viewers

image

Wpuld it make sense to have a list vs grid switcher so that you can alternate between 2 vs 1 title per row. With just 1 expanded title per row, at least some of that underlying detail can be shown. At least some that detail is important context whereby I think I'd personally keep it at 1 title per row if given the option.

image
image

And secondly, do you think the viewer count in red is worth showing in the app as well?

lib/src/model/broadcast/broadcast_providers.dart
julien4215: I remember why I split it in two providers. I thought that the current page should not be part of the state because the page number is not going to be used elsewhere than in the provider.

But I agree that it probably makes the code harder to understand. I'll create a `BroadcastListState` that will hold the the page number and the list of tournament broadcast so there will be only one provider.

lib/src/model/broadcast/broadcast.dart
```diff
@@ -1,56 +1,47 @@
 import 'package:dartchess/dartchess.dart';
 import 'package:fast_immutable_collections/fast_immutable_collections.dart';
 import 'package:freezed_annotation/freezed_annotation.dart';
+import 'package:lichess_mobile/src/model/common/id.dart';
 
 part 'broadcast.freezed.dart';
 
+@freezed
+class BroadcastTournamentsListState with _$BroadcastTournamentsListState {
```
veloce, line 9: I think we should start to use dart records instead of `@freezed` classes whenever possible. Also I would not use the `State` suffix because it is also the server response.

```dart
typedef BroadcastsList = ({
  IList<Broadcast> active,
  IList<Broadcast> upcoming,
  IList<Broadcast> past,
  int nextPage,
});
```

As long as there is no custom serialisation, or other logic, let's use a dart record. This can be done also for `BroadcastTournament`, `BroadcastRound`, `BroadcastGameSnapshot`, `BroadcastPlayer`.

lib/src/model/broadcast/broadcast_providers.dart
```diff
@@ -2,20 +2,48 @@ import 'package:fast_immutable_collections/fast_immutable_collections.dart';
...
 
 part 'broadcast_providers.g.dart';
 
+/// This provider is used to get a list of tournaments broadcasts.
+/// It implements pagination with a maximum number of pages of 20.
 @riverpod
-Future<IList<Broadcast>> broadcasts(BroadcastsRef ref) async {
-  return ref
-      .withClient((client) => BroadcastRepository(client).getBroadcasts());
+class BroadcastTournamentsList extends _$BroadcastTournamentsList {
```
veloce, line 13: I'd name it `BroadcastsPaginator`, it is more precise and then you can name the provider state `BroadcastsList` for instance.

lib/src/model/broadcast/broadcast_repository.dart
```diff
@@ -31,40 +35,42 @@ class BroadcastRepository {
...
+  Map<String, dynamic> json,
+) {
+  return BroadcastTournamentsListState(
+    active: pick(json, 'active').asListOrThrow(_broadcastFromPick).toIList(),
+    upcoming:
+        pick(json, 'upcoming').asListOrThrow(_broadcastFromPick).toIList(),
+    past: pick(json, 'past', 'currentPageResults')
+        .asListOrThrow(_broadcastFromPick)
         .toIList(),
+    nextPage: pick(json, 'past', 'nextPage').asIntOrThrow(),
```
veloce, line 48: I wouldn't assume the `nextPage` is always non null. Let's make it nullable.

lib/src/model/broadcast/broadcast.dart
julien4215: I can't use anymore `copyWith` with records. I use `copyWith` in the `BroadcastPaginator` to complete the broadcast list when I receive a new page.

veloce: It's a tradeoff in terms of syntax and code readability. You can construct a new record and only update some fields:
```dart
state = ({
  updatedField: newValue,
  remains: record.remains,
})
```

In that case for instance it is perfectly readable.

I'd say in some cases `@freezed` will be more convenient thanks to the deep copy mechanism, so it is not a golden rule. But in general records will do fine I think.

lib/src/model/broadcast/broadcast_repository.dart
```diff
@@ -31,40 +35,42 @@ class BroadcastRepository {
...
+  Map<String, dynamic> json,
+) {
+  return (
+    active: pick(json, 'active').asListOrThrow(_broadcastFromPick).toIList(),
+    upcoming:
+        pick(json, 'upcoming').asListOrThrow(_broadcastFromPick).toIList(),
+    past: pick(json, 'past', 'currentPageResults')
+        .asListOrThrow(_broadcastFromPick)
         .toIList(),
+    nextPage: pick(json, 'past', 'nextPage').asIntOrThrow(),
```
veloce, line 48: Should be `asIntOrNull()`

veloce (commented):
We cannot open any tournament anymore, I think it is missing a server scope config.

Capture d’écran 2024-07-15 à 11 56 44

lib/src/model/broadcast/broadcast_repository.dart
julien4215: oops I forgot to change that

veloce:
> > We cannot open any tournament anymore, I think it is missing a server scope config.
> 
> I don't have this problem on my side

It works as anonymous, but if I log in, it stops to work. Does it also work for you when you're logged in?

julien4215:
> It works as anonymous, but if I log in, it stops to work. Does it also work for you when you're logged in?

How can I try the broadcast feature on a local lila to be able to log in ? I think there is nothing to populate the database with broadcast.

PR #807, opened 2024-06-30
Use l10n.realName instead of l10n.name in profile

tom-anders:
Needed to update translations to get this new key


veloce:
Thanks! I've triggered another crowding build so you can run the `./script/update-arb-from-crowdin.mjs` again to get the latest translations.

tom-anders:
> Thanks! I've triggered another crowding build so you can run the `./script/update-arb-from-crowdin.mjs` again to get the latest translations.

Done :+1:

PR #808, opened 2024-06-30
Fix plural handling for zero quantity

tom-anders:
The removed code in the `mjs` file assumed that in all languages
`0` is followed by singular but that's not true e.g. for German and English.

I tested with French after removing the workaround and
updating localizations: `intl.Intl.pluralLogic` seems to handle
the plural logic correctly, choosing singular after 0 for French,
but plural for German and English.

Probably at the time that this code was added, `intl.Intl.pluralLogic`
was not used yet...?


veloce:
I don't know about `pluralLogic`, I think it is actually due to me not understanding how pluralisation work in other languages.

Thanks for catching this.

scripts/update-arb-from-crowdin.mjs
```diff
@@ -266,17 +266,17 @@ function transformTranslations(data, locale, module, makeTemplate = false) {
...
-    if (!plural.item.some((item) => item.$.quantity === 'zero')) {
-      const oneItem = plural.item.find((item) => item.$.quantity === 'one')
-      if (oneItem) {
-        plural.item.unshift({
-          $: { quantity: 'zero' },
-          _: oneItem._
-        })
-      }
-    }
+    //// add a zero quantity item if it doesn't exist
```
veloce, line 269: You can delete the block instead of commenting it out.

tom-anders: whoops, fixed

tom-anders:
> I don't know about `pluralLogic`, I think it is actually due to me not understanding how pluralisation work in other languages.
> 
> Thanks for catching this.

Me neither until yesterday :D

veloce:
@tom-anders can you re-run the script to solve the conflicts? thanks

tom-anders:
> @tom-anders can you re-run the script to solve the conflicts? thanks

Done :+1:

veloce:
Thanks!

PR #809, opened 2024-07-01
Stream broadcast pgn

julien4215:
Still experimenting with the streaming endpoint. I need to find a way to close the request.


veloce:
Is websocket not the way to go? Have you discuss with @ornicar about it? Sorry if I haven't followed this closely. I just want to make sure we follow the proper path from the beginning.
In the rest of the app we're not using any streaming endpoint but websocket instead. I'm not sure why it would be different here, but I can be wrong of course.

julien4215:
I have already discussed this with you and @ornicar but I agree that using the streaming pgn endpoint of the API doesn't feel right knowing that to get cloud analysis evals I will need to use the websocket.

veloce:
What is preventing you from using the websocket at the moment?

julien4215:
ornicar told me to use the http streaming for moves. Also I don't think the websocket allows you to get the full pgn of the games.

julien4215:
Close as it was decided to use the websocket instead of the http streaming

PR #827, opened 2024-07-03
feat: display db size in MB and add button to delete local db

mhnowak:


tom-anders:
> Also I assume the strings here should be translated, I couldn't find any appropiate translation for this feature. If someone could guide me on how to add one or is it fine to leave this untranslated for now.

You can add the translations to `translations/source/mobile.xml`.

After running `scripts/update-arb-from-crowdin.mjs && flutter gen-l10n` they should then be usable in dart (note that mobile translations use a `mobile` prefix, so `foo` in `mobile.xml` will be `context.l10n.mobileFoo` in dart)

mhnowak:
Thanks! I've added the string to the translations.

veloce:
Sorry, I'm just seeing this too late: we should not add a translation immediately in a new feature, as it is now explained in the documentation.

Please remove the translated string and all translation changes, we'll deal with that later. Thanks.

mhnowak:
> Sorry, I'm just seeing this too late: we should not add a translation immediately in a new feature, as it is now explained in the documentation.
> 
> Please remove the translated string and all translation changes, we'll deal with that later. Thanks.

Done.

veloce:
> the other one would be to somehow trigger invalidate each time we re-enter the screen. The second one could work nicely, just let me know if that's an issue for you and if you'd like me to work on that.

Yes indeed you can do that. There is already an example in the watch tab to refresh the current players on TV each time the tab is re-entered:
https://github.com/lichess-org/mobile/blob/c8219b53712d552f714b1763fe7dab06278ad254/lib/src/view/watch/watch_tab_screen.dart#L76-L80

mhnowak:
> > the other one would be to somehow trigger invalidate each time we re-enter the screen. The second one could work nicely, just let me know if that's an issue for you and if you'd like me to work on that.
> 
> Yes indeed you can do that. There is already an example in the watch tab to refresh the current players on TV each time the tab is re-entered:
> 
> https://github.com/lichess-org/mobile/blob/c8219b53712d552f714b1763fe7dab06278ad254/lib/src/view/watch/watch_tab_screen.dart#L76-L80

Made a commit yesterday, implementing the refresh. Think all should be completed now.

veloce:
Code is LGTM.

Now I wonder about whether the ability to delete the whole database makes sense? (I know I created the issue...).

Things to consider:
- several tables have records per user, should we delete for all users or the current user?
- what about cache deletion? would there be a confusion between cache and sql database?

So I'm still not sure about all this. But at least we have now the code to show the database size. We can show it at the bottom of the settings, and maybe figure out later the opportunity to have a delete button.

mhnowak:
> Code is LGTM.
> 
> Now I wonder about whether the ability to delete the whole database makes sense? (I know I created the issue...).
> 
> Things to consider:
> 
> * several tables have records per user, should we delete for all users or the current user?
> * what about cache deletion? would there be a confusion between cache and sql database?
> 
> So I'm still not sure about all this. But at least we have now the code to show the database size. We can show it at the bottom of the settings, and maybe figure out later the opportunity to have a delete button.

I understand your concerns, let me share my thoughts:

Having same db for cache and puzzles is not a problem within deletion, we're clearing each table manually instead of dropping the whole db at once (this is because if you wanted to drop the whole db at once you'd need to initialize it once again before usage, but also because it gives you more control of what you're deleting). So it wouldn't be a problem to delete only cache for requests or only puzzles. The only issue I see is that the user can see the MB for both cache and puzzles and when he'd like to clear them it won't drop to 0 (or close to 0), so it might give the User an idea of that the feature is not working correctly.

From my experience having that sort of a button is similar to going to your settings on the phone and seeing how much storage is the app taking and you also have the ability to clear local files/caches from there. The only use case I'd see for this is if the user doesn't want to lose all the data, but wants to clear some storage (since maybe if someone played a lot of puzzles he might have X MBs of storage taken by it), but he doesn't want to clear his user preferences, settings, login, etc. (I might be wrong here, but I assume the clear button in the system settings would clear).

Regarding cache and clearing cache in the app.. What are you using cache for exactly? If it's only to slightly increase the loading of some requests (similar to what GraphQL does), I agree it's nice, but it's not like the app is unusable without it.. some concerns could be offline mode - like maybe playing puzzles, or some bots in offline and you need to download and cache them.. but I think you're storing them separately to mentioned cache.. From my experience cache is only an additional thing in the app that improves the UX, but isn't necessary to run the app, and some of cache based frameworks like GraphQL fetch the data each time you make a request anyways, and the cache is there only to avoid loading indicators. 

Sorry for long text, let me know if any of that was useful. I feel like the best would be to give an user an option to delete either or both cache and puzzles, but honestly it requires much more effort to do it and I feel like there are more important features to be delivered, at least at this stage.

veloce:
Sorry I don't have the time right now to figure out what to do with the delete button.

But what you did is already interesting because it shows the size of the DB. How about changing the feature a bit, only to display the size but removing the ability to delete it?

mhnowak:
Will work on that during next week.

mhnowak:
Changed the PR title, it should only include displaying the db size in settings screen. 

Btw. If there are any other tasks you might suggest me take care of, let me know. I can review them and let you know if I have to complete them. 🙏

veloce:
> Btw. If there are any other tasks you might suggest me take care of, let me know. I can review them and let you know if I have to complete them. 🙏

You can pick anything you like. Don't hesitate to ask first on discord or GitHub if you're not sure. Also I'll take the time to create more issue properly described so you can wait for them if you like (no ETA).

veloce (approved):

PR #828, opened 2024-07-03
feat: scale arrows in analysis based on winning chances

tom-anders:
# Notes

- Not sure if we still need the different opacities now...? The old app does not have them, but I think it looks nice. The Web UI seems to use a different opacity for the best move, and the same opacity for all other moves
- Scaling logic could probably be adjusted in the feature based on user feedback, but I think the result looks pretty close to the old app and the web UI already.

# Screenshots

### DIFF
--- a/lib/src/model/common/eval.dart
+++ b/lib/src/model/common/eval.dart
@@ -1,5 +1,6 @@
 import 'dart:math' as math;
 
+import 'package:collection/collection.dart';
 import 'package:dartchess/dartchess.dart';
 import 'package:fast_immutable_collections/fast_immutable_collections.dart';
 import 'package:freezed_annotation/freezed_annotation.dart';
@@ -47,6 +48,16 @@ class ExternalEval with _$ExternalEval implements Eval {
   }
 }
 
+double _toWhiteWinningChances(int? cp, int? mate) {
+  if (mate != null) {
+    return mateWinningChances(mate);
+  } else if (cp != null) {
+    return cpWinningChances(cp);
+  } else {
+    return 0;
+  }
+}
+
 /// The eval from the client's own engine, typically stockfish.
 @freezed
 class ClientEval with _$ClientEval implements Eval {
@@ -71,10 +82,11 @@ class ClientEval with _$ClientEval implements Eval {
     return Move.fromUci(uci);
   }
 
-  IList<Move?> get bestMoves {
+  IList<MoveWithWinningChances> get bestMoves {
     return pvs
         .where((e) => e.moves.isNotEmpty)
-        .map((e) => Move.fromUci(e.moves.first))
+        .map((e) => e.firstMoveWithWinningChances(position.turn))
+        .whereNotNull()
         .toIList();
   }
 
@@ -89,13 +101,7 @@ class ClientEval with _$ClientEval implements Eval {
   double winningChances(Side side) => _toPov(side, _whiteWinningChances);
 
   double get _whiteWinningChances {
-    if (mate != null) {
-      return mateWinningChances(mate!);
-    } else if (cp != null) {
-      return cpWinningChances(cp!);
-    } else {
-      return 0;
-    }
+    return _toWhiteWinningChances(cp, mate);
   }
 }
 
@@ -136,7 +142,20 @@ class PvData with _$PvData {
     }
     return res;
   }
+
+  MoveWithWinningChances? firstMoveWithWinningChances(Side sideToMove) {
+    final uciMove = (moves.isNotEmpty) ? Move.fromUci(moves.first) : null;
+    return (uciMove != null)
+        ? (
+            move: uciMove,
+            winningChances:
+                _toPov(sideToMove, _toWhiteWinningChances(cp, mate)),
+          )
+        : null;
+  }
 }
+
+typedef MoveWithWinningChances = ({Move move, double winningChances});
 
 double cpToPawns(int cp) => cp / 100;
 
--- a/lib/src/view/analysis/analysis_screen.dart
+++ b/lib/src/view/analysis/analysis_screen.dart
@@ -377,6 +377,65 @@ class _Board extends ConsumerStatefulWidget {
 class _BoardState extends ConsumerState<_Board> {
   ISet<cg.Shape> userShapes = ISet();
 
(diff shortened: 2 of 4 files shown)


ijm8710:
Call me crazy but I think opacity matters more than thickness.

The super thin elongated secondary   Move arrows look out of place imo

tom-anders:
Again, I think it's important to be consistent with the Web UI.

Also, note that this is an extreme example, where only one move holds a draw, while the other two are loosing. I think it's nice that this is now immediately noticeable from the arrows, even if the variation display is disabled. 

Here's another example, where the move 2 and 3 are only slightly worse:

Screenshot_20240704-070854

veloce:
@tom-anders can you solve the conflicts please?

I also think the opacity is not needed with the scaled arrows. Now that they are scaled by winning chance it is even confusing and not relevant anymore.
We could have used an opacity scale by winning chance too, but since the web ui uses thickness it is best to stick with it indeed.

tom-anders:
> @tom-anders can you solve the conflicts please?

Done! 

> I also think the opacity is not needed with the scaled arrows. Now that they are scaled by winning chance it is even confusing and not relevant anymore. We could have used an opacity scale by winning chance too, but since the web ui uses thickness it is best to stick with it indeed.

Okay, the latest version removes the opacity scaling entirely now. However, to be even more consistent with the Web UI, the best move now has a higher (0.4) opacity than all other moves (0.2):

Untitled

Lichess Web UI for comparison:

Screenshot_2024-07-08_17-31-03

Do we also want to align the exact color/opacity values here?

veloce:
Yes let's keep the opacity difference between best move and others. Looks like there are some tests failing, can you check? thanks.

tom-anders:
Only formatting issues, my bad. Need to get into the habit of running `dart format` before pushing

veloce:
Oh sorry I haven't answered to the exact question. We can use the color/opacity difference from web UI between best move and others. It looks better like that I agree, and people accustomed to the web UI will feel more at home.

tom-anders:
Right, I’ll check the server code and update the PR

tom-anders:
@veloce by the way (sorry more or less off-topic), what’s the reasoning for using a different default piece set in the app compared to the Web UI? Don’t we want users to feel at home in that regard as well?

veloce:
We don't want the app to look the same as the website. This was the opportunity to use a more modern, nicer piece set.

We'd like to eventually cover all the features of the website, but being 1:1 with the website is a non goal. The app will be different.

lib/src/model/common/eval.dart
```diff
@@ -71,10 +82,11 @@ class ClientEval with _$ClientEval implements Eval {
     return Move.fromUci(uci);
   }
 
-  IList<Move?> get bestMoves {
+  IList<MoveWithWinningChanges> bestMoves() {
```
veloce, line 85: Why remove the getter here and use a method instead?

tom-anders: That happened by accident, fixed now.

tom-anders:
Colors are now aligned with the Web UI (thanks to quick help on discord):

Untitled

lib/src/model/common/eval.dart
```diff
@@ -136,6 +142,26 @@ class PvData with _$PvData {
...
+            move: uciMove,
+            winningChances:
+                _toPov(sideToMove, _toWhiteWinningChanges(cp, mate)),
+          )
+        : null;
+  }
+}
+
+@freezed
+class MoveWithWinningChanges with _$MoveWithWinningChanges {
```
veloce, line 159: Typo: should be `MoveWithWinningChances` with a C

Also I think here we could use a dart record instead of a new class:
```dart
typedef MoveWithWinningChances = (Move, double);
```

lib/src/model/common/eval.dart
```diff
@@ -47,6 +48,16 @@ class ExternalEval with _$ExternalEval implements Eval {
   }
 }
 
+double _toWhiteWinningChanges(int? cp, int? mate) {
```
veloce, line 51: typo

lib/src/view/analysis/analysis_screen.dart
```diff
@@ -377,6 +377,63 @@ class _Board extends ConsumerStatefulWidget {
...
+      // Force minimum scale if...
+      // 1) The best move is significantly better than this move
+      // 2) The signs differ, e.g. the best move is winning while this move draws (or even looses)
+      if (winningDiffComparedToBestMove > 0.3 ||
+          bestMove.winningChances.sign != moves[index].winningChances.sign) {
+        return minScale;
+      }
+      return math.max(
+        minScale,
+        1.4 - 5.0 * winningDiffComparedToBestMove,
```
veloce, line 397: I recently pushed a change to the `flutter-chessground` package: the `scale` is now constrained like: `0 < scale <= 1.0`.

So we want to clamp the value to these limits.

Also the 1.4 and 5 numbers come a bit out of nowhere; a comment or having names for them would be welcome.

lib/src/model/common/eval.dart
tom-anders: Fixed the typo, sorry about that.

Regarding the record: IMO this makes the code less readable, since in `_computeBestMoveShapes` you'd either write something like `moves[index].$2` instead of `moves[index].winningChances`, or you'd need to use destructuring (`var (move, winningChanges) = moves[index]` (but then, why use a record in the first place instead of a class?)

lib/src/view/analysis/analysis_screen.dart
tom-anders: Good point, these values were determined by the "play around and see what looks good" method, I'll give them better names (and also adapt to the changed scale limits)

lib/src/model/common/eval.dart
veloce: You can also defined a record with named fields: https://dart.dev/language/records#record-fields
```dart
typedef MoveWithWinningChances = ({ Move move, double winningChances });
```

Records are a language construct so I'd use them instead of `@freezed` classes whenever possible (when you need an immutable data structure without logic). Freezed classes bring overhead and more process time with `build_runner`.

tom-anders: Ahh thanks for the clarification, didn't learn about that feature yet. Will change it accordingly

veloce (approved):

PR #830, opened 2024-07-04
View recent games by time control/variant

Mauritz8:
I think being able to see past games in a certain time control/variant would be a really useful feature.

2024_07_04_0ra_Kleki
image


Mauritz8:
What do you think of the GamePerf enum I added, is it needed? It forces only valid values of the Perf enum to be passed as a param to the api call (puzzles and from position obviously not valid). But in the performance stats screen, the current logic already stops that from happening.

Mauritz8:
It does not support the current logged in user if offline, since you can't access the profile nor the performance stats screen if you're offline. Though, if those screens will be available offline in the future it might be worth supporting it anyways, what do you think?

veloce:
> What do you think of the GamePerf enum I added, is it needed? It forces only valid values of the Perf enum to be passed as a param to the api call (puzzles and from position not valid). But in the performance stats screen, the current logic already takes care of that.

I don't think we should add another `GamePerf` enum. You can just use `Perf` and add some `assert` where needed to make sure it is not used with unwanted perf types.

veloce:
This partially cover https://github.com/lichess-org/mobile/issues/454. I agree this feature itself is interesting, but the latter is the most interesting one.

Mauritz8:
Now that I see it, I agree that #454 would be really nice to have. Can I work on it, or should we wait until all the filters are available in the server api? Perhaps, I could just add a filter for perftype and integrate it with the current changes in this pr.

veloce:
Yes you can keep the feature in this PR, it's a nice one. If you are willing to implement partially #454 at the same time with the existing server filters that'd be great!

I'll take care of the missing server side API, and we can track the progress in #454

veloce:
I preferred to merge it right away. Feel free to open another PR for #454

ijm8710:
@Mauritz8 thank you for this. Quick check, do these games from
Perf screen also show accuracy when engine analysis is run and the force hold options that you have like on the game list on Home Screen?

Mauritz8:
Yes, this just links to the game history screen, so the behavior is the same as before.

PR #832, opened 2024-07-04
feat: hide move list in zen mode

tom-anders:
This is consistent with both the old app and the Web UI


veloce:
Could you solve the conflicts @tom-anders ? Thanks.

tom-anders:
Done :+1:

tom-anders:
Ah looks like I broke the build now, will take a look

PR #835, opened 2024-07-06
Extend translation comments based on crowdin feedback

tom-anders:
This adresses most of the feedback I've seen on crowdin about the new mobile-only texts


veloce (approved):
👍

PR #853, opened 2024-07-12
docs: add lila-docker instructions to settings_dev_env.md

tom-anders:
lila-docker runs on 8080 by default, so the port forwarding commands are a bit different


veloce (approved):
Thanks!

Issue #753, opened 2024-06-07
Missing German translations


AlphaYui:
### What is the problem?

At various points in the app the German translation is missing. As the app language cannot be changed, I am unable to test if this applies to other languages as well or is specific to German.

#### Screenshot 1:
1. Search for a new game
2. The text "Waiting for opponent to join" on the waiting screen isn't translated.

#### Screenshot 2:
1. Open a finished game from the home page
2. Open the hamburger menu in the bottom left
3. The menu elements "Show result" and "Share game URL" aren't translated.

#### Screenshot 3:
1. Open a game from the home page
2. Open the settings menu in the top right
3. The menu points "Haptic feedback" and "Blindfold" aren't translated.

#### Screenshot 4:
1. Click the new game button on the home page
2. Click the "Other time control" button
3. The tab "Join a game" in the top bar isn't translated.

#### Screenshot 5:
1. Open the puzzles page
2. The description of Themes and Puzzle Storm aren't translated.

#### Screenshot 6:
1. Open Puzzle Storm
2. Tap "About Storm" in the bottom left.
3. The long description text in the popup isn't translated.

#### Screenshot 7:
1. Go to Tools -> Analysis board
2. "Leave empty for initial position" is not translated.

#### Screenshot 8:
1. Go to Settings -> Game behavior
2. The options "Haptic feedback", "Immersive mode" and its description aren't translated.

#### Screenshot 9:
1. Open the home page.
2. The greeting text at the very top isn't translated.

### What did you expect to happen?

I expected the text to be translated into German.

### What happened instead?

Some parts of the text are in English while others are correctly translated into German.

### Device Info

Samsung Galaxy A70

### Operating System Version

Android 11

### Additional information

Screenshot 1: ![Screenshot_1](https://github.com/lichess-org/mobile/assets/5251203/aeadff68-1c51-4e74-a2be-1d3ba438e996)
Screenshot 2: ![Screenshot_2](https://github.com/lichess-org/mobile/assets/5251203/cc49c4bb-02e3-4065-9beb-bed5a6892544)
Screenshot 3: ![Screenshot_3](https://github.com/lichess-org/mobile/assets/5251203/cba67b75-9e6b-43d4-a896-47b1a9223b51)
Screenshot 4: ![Screenshot_4](https://github.com/lichess-org/mobile/assets/5251203/4965f68f-dcaf-490f-b6bb-5b716d1268e7)
[...]

PR #857, opened 2024-07-13
Add remaining missing mobile translations

tom-anders:
This solves #753

I didn't add any translation for the Puzzle Storm description, since the string is copied from lila (which does not have a translation yet either)

### DIFF
--- a/lib/src/utils/l10n.dart
+++ b/lib/src/utils/l10n.dart
@@ -52,3 +52,45 @@ class _LocaleObserver extends WidgetsBindingObserver {
     _didChangeLocales(locales);
   }
 }
+
+/// Takes a localization function ([l10nWithPlaceholder]) (from lib/l10n/l10n.dart) with a single placeholder
+/// and returns a Text widget containing the localized string with the placeholder replaced by the given [widget].
+///
+/// The [textStyle] parameter can be used to style the displayed text parts of the localized string.
+///
+/// For example:
+///
+/// ```dart
+/// String translationFun(String name) {
+///   return 'Hello, ${name}!';
+/// }
+///
+/// // returns a text widget containing MyFancyWidget('Magnus') in place of the placeholder
+/// replaceL10nPlaceholderWithWidget(translationFun, MyFancyWidget('Magnus'));
+/// ```
+Text replaceL10nPlaceholderWithWidget<T extends Widget>(
+  String Function(String) l10nWithPlaceholder,
+  T widget, {
+  TextStyle? textStyle,
+}) {
+  final localizedStringWithPlaceholder = l10nWithPlaceholder('%s');
+
+  final parts = localizedStringWithPlaceholder.split('%s');
+
+  // Localized string did not actually contain the placeholder
+  if (parts[0] == localizedStringWithPlaceholder) {
+    return Text(localizedStringWithPlaceholder);
+  }
+
+  return Text.rich(
+    TextSpan(
+      children: <InlineSpan>[
+        if (parts[0].isNotEmpty) TextSpan(text: parts[0], style: textStyle),
+        if (parts[0] != localizedStringWithPlaceholder)
+          WidgetSpan(child: widget),
+        if (parts.length > 1 && parts[1].isNotEmpty)
+          TextSpan(text: parts[1], style: textStyle),
+      ],
+    ),
+  );
+}
--- a/lib/src/view/home/home_tab_screen.dart
+++ b/lib/src/view/home/home_tab_screen.dart
@@ -13,6 +13,7 @@ import 'package:lichess_mobile/src/navigation.dart';
 import 'package:lichess_mobile/src/styles/styles.dart';
 import 'package:lichess_mobile/src/utils/chessground_compat.dart';
 import 'package:lichess_mobile/src/utils/connectivity.dart';
+import 'package:lichess_mobile/src/utils/l10n.dart';
 import 'package:lichess_mobile/src/utils/l10n_context.dart';
 import 'package:lichess_mobile/src/utils/navigation.dart';
 import 'package:lichess_mobile/src/utils/screen.dart';
@@ -507,8 +508,6 @@ class _HelloWidget extends ConsumerWidget {
 
     final user = accountUser ?? session?.user;
 
-    final greetingParts = context.l10n.mobileGreeting.split('%s');
-
     return Padding(
       padding:
           Styles.horizontalBodyPadding.add(Styles.sectionBottomPadding).add(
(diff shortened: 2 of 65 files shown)


lib/src/view/analysis/analysis_position_choice_screen.dart
```diff
@@ -85,7 +85,7 @@ class _BodyState extends State<_Body> {
               child: AdaptiveTextField(
                 maxLines: 500,
                 placeholder:
-                    '${context.l10n.pasteTheFenStringHere}\n\n${context.l10n.pasteThePgnStringHere}\n\nLeave empty for initial position',
+                    '${context.l10n.pasteTheFenStringHere}\n\n${context.l10n.pasteThePgnStringHere}\n\n${context.l10n.mobileLoadPositionLeaveEmpty}',
```
veloce, line 88: I'd say let's not translate that one for now. I don't think this screen will stay like this. And it would be better with a dedicated message.

veloce (commented):
Thanks.

There is also the `Join a game` from the custom game creation screen that is missing.

veloce:
@tom-anders can you solve the conflicts please?

lib/src/view/home/home_tab_screen.dart
```diff
@@ -527,7 +527,7 @@ class _HelloWidget extends ConsumerWidget {
             ),
             const SizedBox(width: 5.0),
             Text(
-              'Hello${user != null ? ', ' : ''}',
+              '${context.l10n.mobileHello}${user != null ? ', ' : ''}',
```
tom-anders, line 530: @veloce This should probably be fine for most languages, but I'm sure there are some languages where "Hello, %s" does not make any sense, we might want to revisit this at some point

tom-anders:
> @tom-anders can you solve the conflicts please?

Done :+1:

veloce (approved):

lib/src/view/home/home_tab_screen.dart
veloce: @tom-anders that is because we use the `UserFullNameWidget`.

Now that I think of it we should make it right from the start and have a parameterised message such as `helloX` : `Hello, <name>`.

We can adapt the HelloWidget later to show the flair for instance.

tom-anders: Ok, so the string is now 'Hello %s' and I added a special case in the translation generation, making sure that the %s is **not** converted to a function parameter. This way, we can instead split the (localized) string at '%s' and put the UserNameWidget between the two resulting parts.

lib/src/view/home/home_tab_screen.dart
```diff
@@ -528,11 +530,12 @@ class _HelloWidget extends ConsumerWidget {
               color: context.lichessColors.brag,
             ),
             const SizedBox(width: 5.0),
-            Text(
-              'Hello${user != null ? ', ' : ''}',
-              style: style,
-            ),
-            if (user != null) UserFullNameWidget(user: user, style: style),
+            if (user != null && greetingParts.length == 2) ...[
```
veloce, line 533: Not sure about this. Is the `greetingParts.length` always 2?

tom-anders: Yes, according to the flutter docs:

`A non-empty match at the start or end of the string, or after another match, is not treated specially, and will introduce empty substrings in the result`

So we'll always get 2 here, as long as '%s' is somewhere in the string, including start and end. (If not, it's a translation issue, the comment says to always include the %s)

veloce: I think it's a common use case (inclusion of a widget instead of string in a translation) so it should be factored out in a utility function.

And this function should have unit tests.

tom-anders: Ah yes good point, will do 👍

tom-anders: Ok, added a utility function

```
IList<Widget> replaceL10nPlaceholderWithWidget<T extends Widget>(
  String Function(String) l10nWithReplacement,
  T widget, {
  TextStyle? textStyle,
})
```
To avoid empty text widgets, this now also catches the case where the placeholder is at start the start or end and only adds one instead of two Text Widgets in that case

lib/src/utils/l10n.dart
```diff
@@ -52,3 +53,50 @@ class _LocaleObserver extends WidgetsBindingObserver {
...
+/// ```dart
+/// String translationFun(String name) {
+///   return 'Hello, ${name}!';
+/// }
+///
+/// // returns [Text('Hello, '), MyFancyWidget('Magnus'), Text('!')]
+/// replaceL10nPlaceholderWithWidget(translationFun, MyFancyWidget('Magnus'));
+/// ```
+///
+/// @param l10nWithReplacement The localization function corresponding to a localized string with a placeholder.
```
veloce, line 71: Dart doc comment are not like Java's. Instead of `@param` you need to put the triple slash above the function parameter declaration (line 83 in that case).

lib/src/utils/l10n.dart
```diff
@@ -52,3 +53,50 @@ class _LocaleObserver extends WidgetsBindingObserver {
...
+/// @param testStyle The style to apply to the Text widgets.
+///
+/// If the placeholder is not present in the localized string,
+/// will return a single Text widget with the localized string.
+/// If the placeholder is at the beginning or end of the localized string,
+/// will return a list with a single Text widget and the widget.
+/// Otherwise, will return a list with three Text widgets:
+/// the first with the text before the placeholder, the second with the widget,
+/// and the third with the text after the placeholder.
+IList<Widget> replaceL10nPlaceholderWithWidget<T extends Widget>(
```
veloce, line 82: Instead of an `List<Widget>`, the idiomatic way to handle this in flutter is the [Text.rich](https://api.flutter.dev/flutter/widgets/Text/Text.rich.html) constructor.

See [WidgetSpan](https://api.flutter.dev/flutter/widgets/WidgetSpan-class.html) to embed a widget inline inside text.

veloce (commented):
One remaining change about the signature of the utility function and we're good I think :)

lib/src/utils/l10n.dart
tom-anders: Whoops, I actually checked dart docs (https://dart.dev/effective-dart/documentation#do-use-prose-to-explain-parameters-return-values-and-exceptions), but overlooked that using `@param` is marked as `bad` :smile: 

The guidelines say to explain the parameters in the description of the method (via square brackets), so I'll do that instead

lib/src/utils/l10n.dart
tom-anders: thanks for the hint!

veloce (approved):

PR #858, opened 2024-07-13
Game Settings: Add toggle to disable Auto Queen if set to Always

incogg:


incogg:
i think because im on windows all the line endings changed?
sorry for the huge diff.

PR #863, opened 2024-07-14
Use the broadcast websocket to get real time moves

julien4215:


lib/src/model/broadcast/broadcast_controller.dart
```diff
@@ -0,0 +1,126 @@
...
+import 'package:lichess_mobile/src/model/common/http.dart';
+import 'package:lichess_mobile/src/model/common/id.dart';
+import 'package:lichess_mobile/src/model/common/socket.dart';
+import 'package:lichess_mobile/src/utils/json.dart';
+import 'package:riverpod_annotation/riverpod_annotation.dart';
+
+part 'broadcast_controller.g.dart';
+
+@riverpod
+class BroadcastController extends _$BroadcastController {
```
veloce, line 18: Looks like this controller handles all the games of a broadcast round. It would be more clear if it was named `BroadcastRoundController`.

lib/src/model/broadcast/broadcast_controller.dart
```diff
@@ -0,0 +1,126 @@
...
+      case 'chapters':
+        _handleChaptersEvent(event);
+      // Sent when clocks are updated from the broadcast
+      case 'clock':
+        _handleClockEvent(event);
+    }
+  }
+
+  void _handleAddNodeEvent(SocketEvent event) {
+    final relayPath = pick(event.data, 'p', 'relayPath').asUciPathOrThrow();
```
veloce, line 60: Explain with comments what is the difference between `relayPath` and `path`. Since I cannot figure it out while reviewing, I guess it will be the same for anyone reading this code.

lib/src/model/broadcast/broadcast_controller.dart
```diff
@@ -0,0 +1,126 @@
...
+    final path = pick(event.data, 'd', 'p', 'path').asUciPathOrThrow();
+    final id = pick(event.data, 'd', 'n', 'id').asUciCharPairOrThrow();
```
veloce, line 62: `id` is too vague here; I suppose here it is a node id, so let's name it `nodeId`.

lib/src/model/broadcast/broadcast.dart
```diff
@@ -46,15 +46,18 @@ class BroadcastRound with _$BroadcastRound {
   }) = _BroadcastRound;
 }
 
+typedef BroadcastMapGames = IMap<BroadcastGameId, BroadcastGameSnapshot>;
```
veloce, line 49: I'd name that `BroadcastRoundGames`.

lib/src/model/broadcast/broadcast.dart
```diff
@@ -46,15 +46,18 @@ class BroadcastRound with _$BroadcastRound {
...
 class BroadcastGameSnapshot with _$BroadcastGameSnapshot {
   const BroadcastGameSnapshot._();
 
   const factory BroadcastGameSnapshot({
-    required IList<BroadcastPlayer> players,
+    required IMap<Side, BroadcastPlayer> players,
     required String fen,
     required Move? lastMove,
     required String status,
+    required Duration? thinkTime,
```
veloce, line 60: What is the use of `thinkTime`?

julien4215: `thinkTime` is the amount of time that the player whose turn it is has been thinking since his last move

julien4215:
It looks like Github workflow merges my branch before running tests otherwise I don't understand why it is failing

lib/src/model/broadcast/broadcast.dart
veloce: > thinkTime is the amount of time that the player whose turn it is has been thinking since his last move.

Would be good to put that explanation as a field comment.

veloce (approved):
The broadcast screen will need tests but they can be done later since it's not finished.

PR #867, opened 2024-07-16
Broadcast - Use GridView.builder instead of ListView.builder

julien4215:


lib/src/view/broadcast/broadcast_screen.dart
```diff
@@ -84,104 +85,86 @@ class _Body extends ConsumerWidget {
...
+
+    const textHeight = 19.0;
+
+    const headerAndFooterPadding = EdgeInsets.symmetric(vertical: 5);
+
+    const boardSpacing = 10.0;
+
+    final numberOfBoardsByRow = isTabletOrLarger(context) ? 4 : 2;
+
+    final screenWidth = MediaQuery.of(context).size.width;
```
veloce, line 105: Use `MediaQuery.sizeOf(context)` instead.

lib/src/view/broadcast/broadcast_screen.dart
```diff
@@ -84,104 +85,86 @@ class _Body extends ConsumerWidget {
...
   Widget build(BuildContext context) {
-    const numberLoadingBoardRows = 6;
+    const numberLoadingBoards = 12;
+
+    const textHeight = 19.0;
+
+    const headerAndFooterPadding = EdgeInsets.symmetric(vertical: 5);
+
+    const boardSpacing = 10.0;
+
```
veloce, line 102: Avoid all the empty lines between declarations.

lib/src/view/broadcast/broadcast_screen.dart
```diff
@@ -84,104 +85,86 @@ class _Body extends ConsumerWidget {
...
-        itemCount: list.isEmpty ? numberLoadingBoardRows : list.length,
-        itemBuilder: (context, index) {
-          final itemPadding = EdgeInsets.only(
-            left: _kGridPadding,
-            top: _kGridPadding / 2,
-            bottom: index == list.length - 1 ? _kGridPadding : 0,
-          );
+      child: Padding(
+        padding: Styles.bodyPadding,
+        child: GridView.builder(
```
veloce, line 126: The number of elements is not that large, I wonder if the `GridView.count` would be more appropriate? The code would be a little easier to read I think.

julien4215:
> Looks like you based this PR on the websocket branch. You should base it on main branch so I can only review the changes related to the `GridView`. Thanks.

Ah I thought you would review this after the websocket PR. You can view the changes related only to the `GridView` by looking only at the last commit but if you want to merge this one before the websocket, I'll rebase it on main.

lib/src/view/broadcast/broadcast_screen.dart
julien4215: I don't know how it would affect performances. According to the `GridView.builder` documentation `the builder is called only for those children that are actually visible`. From what I understand the `GridView.count` will create all widgets at once but it will only render them when they are visible on the screen whereas the `GridView.builder` will create only the widgets that are needed.

Does the number of elements that is not that large applies to the puzzles too ?

veloce:
Yes please base it on `main`. It really helps the review process to only create PR based on main.

lib/src/view/broadcast/broadcast_screen.dart
veloce: No the puzzle history screen is paginated and thus potentially very large.

But the games number in a broadcast round are usually quite limited, like usually less than 50 and I doubt there are such big tournaments that this number can be more than like 100 games.

julien4215: Ok I'll use the `GridView.count`

julien4215: Actually I don't think it will make the code more readable because I don't have access anymore to `mainAxisExtent` so I would need to use `childAspectRatio` and also I would need to use `List.generate` to generate grid widgets.

PR #873, opened 2024-07-18
fix incorrect move sounds during analysis

### LINKED ISSUE #870: Move sounds are wrong when you take back a move on anlaysis board
### What is the problem?

Move sounds are wrong when you tak back a move on anlaysis board

1. Play a taking move like Nxd4
2. Play a normal move like Nf6
3. Take the normal move back and when you take back the move you hear piece capture sound instead of normal piece move sound 

You should hear piece move sound

### What happened instead?

You hear piece capture sound 

### Device Info

Samsung Galaxy M51

### Operating System Version

Android 12

### Additional information

_No response_

Lonec-L:
fixes #870 

sounds should now be played correctly when user goes to previous or next move on analysis board. The issue was that only associated move of current node (after the requested change) was considered when determining which sound to play. Now the associated move of current node is used when user moves forward through the tree and associated move of previous node (before the requested change) is used when backtracking.

This also fixes move sound not playing when reverting first move of the game. Now the normal move sound is played whenever current node is not a Branch. I am not sure about this, so I am looking for some feedback if this is undesired behavior.

### DIFF
--- a/lib/src/model/analysis/analysis_controller.dart
+++ b/lib/src/model/analysis/analysis_controller.dart
@@ -383,9 +383,9 @@ class AnalysisController extends _$AnalysisController {
         ? _root.view
         : state.root;
 
+    final isForward = path.size > state.currentPath.size;
     if (currentNode is Branch) {
       if (!replaying) {
-        final isForward = path.size > state.currentPath.size;
         if (isForward) {
           final isCheck = currentNode.sanMove.isCheck;
           if (currentNode.sanMove.isCapture) {
@@ -396,7 +396,7 @@ class AnalysisController extends _$AnalysisController {
             ref.read(moveFeedbackServiceProvider).moveFeedback(check: isCheck);
           }
         }
-      } else {
+      } else if (isForward) {
         final soundService = ref.read(soundServiceProvider);
         if (currentNode.sanMove.isCapture) {
           soundService.play(Sound.capture);


tom-anders:
I got curious what lichess.org and the old app do: They both play actually **no sound** at all when reverting a move in the analysis board.

Doesn't mean the new app has to do the same, I just thought I'd mention this.

Lonec-L:
To be honest, I did not pay attention to that. I just wanted to start contributing to the new app in some way, and this seemed like a good first issue to solve.

ArleyFelipe (approved):

veloce:
Thanks for your contribution @Lonec-L . I'm still not sure we should merge it because:
1. it is not a bug but was intended
2. it is the same behavior right now as in the website and the old app

tom-anders:
> Thanks for your contribution @Lonec-L . I'm still not sure we should merge it because:
> 
> 1. it is not a bug but was intended
> 2. it is the same behavior right now as in the website and the old app

I think you misunderstood the issue here:

1) lichess.org and the old app do **not** play a sound when reverting a move in the analysis board.
2) the new app **does** play a sound
3) the new app sometimes plays the **wrong** sound
4) this app fixes that

So this PR improves things, but it's still inconsistent with the old app and lichess.org.

veloce:
Oh my bad, sorry. I really thought that the new app would not play a sound when reverting a move.

tom-anders:
> Oh my bad, sorry. I really thought that the new app would not play a sound when reverting a move.

Yeah, so IMO we should be consistent with the old app and Web UI, by **not** playing a sound at all when reverting a move (instead of fixing the bug). Whats your opinion on this?

Lonec-L:
I have reverted previous changes and reopened this PR, now with sounds completely removed when reverting moves during analysis, matching the behavior of old app and lichess website.

veloce:
Ok, thank you for that.

I'm still not sure whether no sound when replaying backwards is a good idea. It is like lichess website for sure, but we are allowed to deviate from the website for things like that. I'll think about it.

PR #877, opened 2024-07-21
Filter games

Mauritz8:
Partially implement #454


Mauritz8:
I've implemented a filter for perf:

[Screencast from 2024-07-21 14:50:36.webm](https://github.com/user-attachments/assets/f8de5bc5-96ef-4c9d-8f6c-335182f757f8)

ZTL-UwU:
It looks awesome! One samll suggestion tho, is it possible to use chips for filtering like the recommended design in material 3? (https://m3.material.io/components/chips/overview)

ijm8710:
I agree with @ZTL-UwU. Another reason for this is because there will ultimately probably be other filters behind variant such as result type and piece color

Mauritz8:
Thanks! Using chips seems like a great idea, I'll try to make it work.

Mauritz8:
I think this is a significant improvement, let me know what you think.

[Screencast from 2024-07-25 17:41:35.webm](https://github.com/user-attachments/assets/2b1997b1-b1ca-40b0-bea3-b97bc7de69de)

veloce:
Thanks!

Can you try to:
- make the chip smaller
- put the chip in the app bar, or make it look like it is in the app bar; (I guess it is not that hard to do, but I'm not sure)

ijm8710:
For a first version this is great but would yall agree formatting the chips more similarly organized to the link shared here?

That way you can add a few more filters such as win/loss and white/black

I think this organization looks a bit more streamlined

image

veloce:
Yeah good idea. We could also separate the "Time control" chips from the variant chips. Using a line separator.

Mauritz8:
Now it looks like the picture, big improvement again I think. I also implemented your suggestions veloce. Though I'm not exactly sure what you mean by making it smaller, was shortening the text to just 'variant' enough? Also, I'm not able to test on ios, so I'm not sure whether the app bar will look good on it.

[Screencast from 2024-07-26 07:50:33.webm](https://github.com/user-attachments/assets/47a4fa6b-d707-42b4-9aef-43d1945eb5c6)

ZTL-UwU:
I think a filter icon button on the right of the appbar will be more concise.

Also, I think putting the chips into a bottom sheet or a separate screen instead of a dialog is better as it gives more space for the chips to spread.

Mauritz8:
If I understand correctly, you want a single button in the app bar, that opens all the filters?
That would be a different design than what's detailed in #454. To quote: 
> On top of the list, an horizontal list of filters should be displayed (like in the notifications screen of GitHub app). Filters would include: "rated", "perf type", "bookmarks", etc. (list non exhaustive).

I would prefer using a popup, rather than a separate screen. There's less context switching, and you can easily exit by pressing outside the popup. Overall I think it's a better experience. Also, I don't think space is an issue, It easily fits, and this is one of the filters with the most chips. Filters like color and rated will only have 2 chips, so a separate screen seems excessive, and a bottom sheet would be very far down since there would be so little content on it.

ijm8710:
I do think I had similar comments to @ZTL-UwU (even before I read his). 

Was also thinking separate screen or from bottom may work better for this

Smaller chips with headers. Just design wise closer to the piston so we can have more sections 
You can even have little icons in front of the section names to make it slightly prettier (like some pawn or something before variant type)

Mauritz8:
So, I created a new branch locally and implemented the design that you both want. I think it looks nice, and I'm willing to go for this design instead of the proposed one. @veloce what do you think?

[Screencast from 2024-07-27 14:41:50.webm](https://github.com/user-attachments/assets/9ebdb000-7653-4cd1-a362-a4c3b0662155)

ijm8710:
Awesome @Mauritz8 i thinks it's starting to look a lot better.

Few more things that I think would get it to finish line


- I feel order of variants should be either based on the same order of your perf cards (which I believe is based on total games played) or based on just most popular variant side wise. Small chance this is already implemented but if I had to guess, ultra bullet fits neither of those criteria
- I like that we have filter now!is it still possible to have the filter but still have the designation you had before to show what the filter is on? Perhaps can change the title of the header screen to say like bullet or something especially if it's just one filter selected
- would you be able to add a section for result: won/loss/draw
- I think icons before the section names would be the final polish to make them easy to discern quickly? Like perhaps the rabbit, bullet or turtle before the variant section. A black or white pawn before the side section. And the green plus icon for the result section?


Thanks for trialing out our ideas! It's getting really good!

veloce:
> So, I created a new branch locally and implemented the design that you both want. I think it looks nice, and I'm willing to go for this design instead of the proposed one. @veloce what do you think?
> 

Looks good to me. I think on top of the "filter" icon we need a chip indicating the number of selected filters. Same appearance as the "unread messages" chip on the chat icon on game screen.

Mauritz8:
Thanks @ijm8710, I've made adjustments now in regards to your feedback:

* Variants now work the same as the perf cards on the home screen. That is, they are ordered based on total games played, and ones with no games played are not shown.
* Chip indicating how many filters are in use, like veloce suggested.
* Icon before filter name

> would you be able to add a section for result: won/loss/draw

I don't think this is possible since games are only fetched a few at a time, so there's no way of even knowing the total number of games when using filters.

This is how it looks:

[Screencast from 2024-07-28 11:26:40.webm](https://github.com/user-attachments/assets/74580383-4e8b-4485-8c24-13995f02187d)

ijm8710:
Thanks @Mauritz8 think it looks perfect! You rock 


One question for you:
- if only one filter is chosen, any chance to have the title reflect filter chosen (ie Ivan's classical games or even "classical games")


And two questions for @veloce 
- mauritz mentioned game result is not supported in api. Would this be something worth adding to api if possible? I feel being able to query a win vs loss will have use-case here and possibly elsewhere
- on web, there are many more filters. Do you think it may make sense to have a link on the filter screen "advanced filters" for even more advanced filter access?
Dream scenario is a user would be able to open an overlay web screen of the filters, choose what they want and they generate in a native screen within the app. But perhaps this is overboard.

Mauritz8:
> if only one filter is chosen, any chance to have the title reflect filter chosen as well as the result count "83 classical games"...."83 games" if multi filters chosen (this would then match the format used when opening specific games from the variant screens)

Games are only fetched a few at a time, so it's unfortunately not possible to count the total when using filters. We could change the title when only a single filter is used, but I'm a bit worried that there might be too little horizontal space in the app bar. 

> these same filters can be run from another user profile too, correct?

Yes!

ijm8710:
Okay, only reason I thought otherwise is because your other pr if you pull all classical games from the variant card it had the count. No worries then! Thanks again for your work on this

Mauritz8:
Yes it's technically possible to get the number of games if the filter is only made up of variants, but it would require an additional api request for each variant selected, which doesn't seem great. I think it could be done in a better way, and in my opinon it's not that useful unless it works no matter how complex the filter is. If it should be implemented, I think the number of results should probably be a part of the api response instead. The advanced search on the website shows the number of results, so it should be possible to implement.

veloce:
> mauritz mentioned game result is not supported in api. Would this be something worth adding to api if possible? I feel being able to query a win vs loss will have use-case here and possibly elsewhere

Yes it is planned to add more filters later.

ijm8710:
@Mauritz8 one other request, not sure if you agree.

Before filtering on game type in your pr it says Ivan's games

Currently in beta it'll say total # games. I don't think that total count is visible anywhere in app so can we have it still say total count before filtering?

lib/src/model/game/game_history.dart
```diff
@@ -33,15 +33,19 @@ const _nbPerPage = 20;
 /// stored locally are fetched instead.
 @riverpod
 Future<IList<LightArchivedGameWithPov>> myRecentGames(
-  MyRecentGamesRef ref,
-) async {
+  MyRecentGamesRef ref, {
+  GameFilterState filter = const GameFilterState(),
```
veloce, line 37: Recent games should not be filtered (whether it is my recent games or any user recent games) so this parameter should be removed.

lib/src/model/game/game_filter.dart
```diff
@@ -0,0 +1,36 @@
...
+  GameFilterState build({GameFilterState? filter}) {
+    return filter ?? const GameFilterState();
+  }
+
+  void setFilter(GameFilterState filter) => state = state.copyWith(
+        perfs: filter.perfs,
+        side: filter.side,
+      );
+
+  int countFiltersInUse() {
```
veloce, line 22: Provider should not define public getters or expose state through public methods.

The `count` info is already available in the state. A `count` getter can be defined in the `GameFilterState` class directly if needed.

lib/src/view/user/game_history_screen.dart
```diff
@@ -1,65 +1,130 @@
...
-      androidBuilder: _buildAndroid,
-      iosBuilder: _buildIos,
+    final session = ref.read(authSessionProvider);
+    final username = user?.name ?? session?.user.name;
+    final title = Text(
+      username != null
+          ? '$username ${context.l10n.games.toLowerCase()}'
+          : context.l10n.games,
     );
+    final filtersInUse = ref
```
veloce, line 37: Cf. other comment. `ref.read` should not be used in `build()` methods; to count the filters you should get the filter state with `ref.watch`.

lib/src/view/user/game_history_screen.dart
```diff
@@ -183,3 +259,203 @@ class _BodyState extends ConsumerState<_Body> {
...
+              filter = filter.copyWith(
+                perfs: selected
+                    ? filter.perfs.add(value)
+                    : filter.perfs.remove(value),
+              );
+            },
+          ),
+        );
+
+    return Container(
```
veloce, line 338: A `Container` seems useless here. You could use directly a `Padding`.

lib/src/view/user/game_history_screen.dart
```diff
@@ -183,3 +259,203 @@ class _BodyState extends ConsumerState<_Body> {
...
+      child: DraggableScrollableSheet(
+        initialChildSize: .7,
+        expand: false,
+        snap: true,
+        snapSizes: const [.7],
+        builder: (context, scrollController) => ListView(
+          controller: scrollController,
+          children: [
+            if (userId != null)
+              ref.watch(userProvider(id: userId)).when(
```
veloce, line 349: It is better to put `ref.watch` at the beginning of `build()` methods so when we read the code we know immediately what are the watched dependencies of the widget.

lib/src/model/game/game_history.dart
Mauritz8: The game history provider uses the recent games providers in its implementation. We could create new providers that do the same thing, except they watch the game filter provider instead of passing it as a param. I don't really see the point though.

lib/src/view/user/game_history_screen.dart
Mauritz8: The `userId` might be null though, so we can't listen to it unconditionally.

Mauritz8:
> @Mauritz8 one other request, not sure if you agree.
> 
> Before filtering on game type in your pr it says Ivan's games
> 
> Currently in beta it'll say total # games. I don't think that total count is visible anywhere else in app so can we have it still say total count before filtering?

I'll add it back in.

lib/src/model/game/game_history.dart
veloce: I'd rather make the request again than have an extra complexity and unnecessary filter in the recent game providers.

So you can just remove the dependency to the recent game provider in `UserGameHistory` and perform the request directly with the filters.

veloce:
Thanks again for this!

ijm8710:
> 
> > would you be able to add a section for result: won/loss/draw
> 
> 
> 
> Not supported in the api yet.
> 

Hi @Mauritz8, I had inquired about win/loss/draw chips for the full game list filtering and at the time, you said api currently didn't support. 

This would also include chips for analyzed games, imported games and    rated/casual. Plus if you wanted to get a step further, perhaps the full suite of options that web has  as advanced filers to sort thru your full game list.

Right now, I was most curious about imported games! As far as I can tell, without it being implemented here, there's no tagging to be able to even get to imported games in the new app.

Do you happen to know if the list afforded by the api changed since to allow for filter by imported (or even the other options mentioned that had not been implemented yet)?

If there needs to be an issue ticket or if I ultimately need to ask Veloce, no worries but since you worked with the api here originally, figured best to start with you since your past work here may give a quick clue if the api has been expanded since the original work.

Mauritz8:
From what I can tell there seems to be support for filtering for analyzed games and rated games, but not for win/loss/draw nor imported games.

ijm8710:
> From what I can tell there seems to be support for filtering for analyzed games and rated games, but not for win/loss/draw nor imported games.

Hey mauritz any chance to put a Lila ticket to request that for api or how best do you think to proceed. Im happy to create a ticket on lila but I feel from An api perspective it may be better worded from you 
Or is there a better way?
Or would it make sense to add those filters that do exist in the interim or wait to see if there's a chance fuller api support may be imminent if requested

Or even, assuming this is technically possible. using the web view of advanced filters to build filter views in the app. Less native but does give more ways to filter which is nice

PR #884, opened 2024-07-24
feat: add setting to change language

### LINKED ISSUE #448: Add an option to choose the language
For now only the device default language is used.

tom-anders:
Closes #448

Not sure how to translate language codes to their localized name...? Looks like lichobile simply has a hardcoded map (see src/i18n.ts), do we just want to copy/paste that?

[language.webm](https://github.com/user-attachments/assets/07a8ded9-b1a4-4619-a134-3724d9aa656c)


lib/src/app.dart
```diff
@@ -204,6 +204,9 @@ class _AppState extends ConsumerState<Application> {
           localizationsDelegates: AppLocalizations.localizationsDelegates,
           supportedLocales: kSupportedLocales,
           onGenerateTitle: (BuildContext context) => 'lichess.org',
+          locale: ref.watch(
```
veloce, line 207: We have too many `generalPreferencesProvider.select` now. I think it's better to have one single
```dart
final generalPrefs = ref.watch(generalPreferencesProvider);
```

at the beginning of `build`. (It is better to put `ref.watch` at the beginning so we know what provider are listened to).

lib/src/model/settings/general_preferences.dart
```diff
@@ -14,6 +14,28 @@ part 'general_preferences.g.dart';
 
 const _prefKey = 'preferences.general';
 
+String? localeToJson(Locale? locale) {
+  return jsonEncode(
+    locale != null
+        ? {
+            'languageCode': locale.languageCode,
+            'countryCode': locale.countryCode,
```
veloce, line 22: We're not using it yet, but we should take into account the `Locale.scriptCode` in the json encoder.

lib/src/model/settings/general_preferences.dart
```diff
@@ -14,6 +14,28 @@ part 'general_preferences.g.dart';
...
+}
+
+Locale? localeFromJson(String? json) {
+  if (json == null) {
+    return null;
+  }
+  final map = jsonDecode(json) as Map<String, dynamic>;
+  return Locale(
+    map['languageCode'] as String,
+    map['countryCode'] as String,
```
veloce, line 35: that should be `as String?` since the field is nullable.

lib/src/model/settings/general_preferences.dart
```diff
@@ -82,13 +108,16 @@ class GeneralPrefsState with _$GeneralPrefsState {
...
+    @JsonKey(toJson: localeToJson, fromJson: localeFromJson)
+    required Locale? locale,
   }) = _GeneralPrefsState;
 
   static const defaults = GeneralPrefsState(
     themeMode: ThemeMode.system,
     isSoundEnabled: true,
     soundTheme: SoundTheme.standard,
     systemColors: true,
+    locale: null, // use system locale
```
veloce, line 120: Instead of this inline comment, prefer a doc comment above the field declaration.

lib/src/view/settings/settings_screen.dart
```diff
@@ -93,6 +95,10 @@ class _Body extends ConsumerWidget {
       generalPreferencesProvider.select((state) => state.systemColors),
     );
 
+    final locale = ref.watch(
+      generalPreferencesProvider.select((state) => state.locale),
```
veloce, line 99: Same remark as in app.dart file: we have too many `.select`. It is better now to use a single `ref.watch`.

lib/src/view/settings/settings_screen.dart
```diff
@@ -304,6 +310,33 @@ class _Body extends ConsumerWidget {
...
+          SettingsListTile(
+            icon: const Icon(Icons.language),
+            settingsLabel: Text(context.l10n.language),
+            settingsValue: localeL10n(context, locale),
+            onTap: () {
+              if (Theme.of(context).platform != TargetPlatform.android) {
+                showChoicePicker<Locale?>(
+                  context,
+                  choices: [
+                    null,
```
veloce, line 322: Instead of displaying a `null` choice for the system locale, I think we should be clever and show the system locale as selected if the settings value is `null` (and also remove the duplicate in the list).

veloce (commented):
Thanks for taking care of this.

There's something I wanted to do for iOS: since the OS allows changing the language in the app settings (available from iOS settings app), I think it is better to not allow the users to change the lang from the app.

So instead of pushing the language selection screen, on iOS it should open the settings app at the language section of lichess mobile app.
Looks like this package supports this: https://pub.dev/packages/app_settings

veloce:
Just to let you know this change: https://github.com/lichess-org/mobile/commit/0ba302919ad55ad1b40d6a78cbd751b1f56de268

According to the flutter doc this is how locales should be defined. For the languages names map, I think you can use the hardcoded map from lichobile, but I'm not sure it has all the langs. I think you can find one in the Lila repo too, but I don't know where.

veloce:
Here is the lang list from Lila: https://github.com/lichess-org/lila/blob/master/modules/i18n/src/main/LangList.scala @tom-anders

tom-anders:
> Thanks for taking care of this.
> 
> There's something I wanted to do for iOS: since the OS allows changing the language in the app settings (available from iOS settings app), I think it is better to not allow the users to change the lang from the app.
> 
> So instead of pushing the language selection screen, on iOS it should open the settings app at the language section of lichess mobile app. Looks like this package supports this: https://pub.dev/packages/app_settings

Done (but I have no way to test it)

tom-anders:
Apparently, some language code combinations are still missing from the lila list namely de-CH (added a translation for that myself), but also zh and tt-RU. Not sure we're to get the correct translations for those.

veloce:
Yes the `de-CH` is not available from lichess website, I guess we should remove it then. Same for `zh` (I left it as a fallback, but I guess we can remove it) and `tt-RU`.

They need to be removed from the `update-arb-from-crowdin` and the `constants.dart`.

HaonRekcef:
As far as I know, there is no mapping to the Swiss German language (DE-ch) because it is unusual for websites to have a Swiss German translation. Due to the many different accents, there are basically no standardized spelling or grammatical rules in the language.

That's why it was decided that people with DE-ch as their browser setting will see the German translation by default. However, there is an option to select the Swiss German translation in the settings. In my opinion, this is a good solution, and maybe we can do the same for the app.

tom-anders:
@HaonRekcef Thanks for the insight! https://github.com/lichess-org/lila/blob/27d6d15bb531a2ba5d85ec6e765e98415b3ce7c4/modules/i18n/src/main/LangList.scala#L45C1-L45C43 would be the relevant line, right?

@veloce I think then we simply add `gsw-CH` to the crowdin script to support this...?

HaonRekcef:
> @HaonRekcef Thanks for the insight! https://github.com/lichess-org/lila/blob/27d6d15bb531a2ba5d85ec6e765e98415b3ce7c4/modules/i18n/src/main/LangList.scala#L45C1-L45C43 would be the relevant line, right?

Yes, correct.

pubspec.lock
```diff
@@ -5,10 +5,10 @@ packages:
     dependency: transitive
     description:
       name: _fe_analyzer_shared
-      sha256: f256b0c0ba6c7577c15e2e4e114755640a875e885099367bf6e012b19314c834
+      sha256: "5aaf60d96c4cd00fe7f21594b5ad6a1b699c80a27420f8a837f4d68473ef09e3"
       url: "https://pub.dev"
     source: hosted
-    version: "72.0.0"
+    version: "68.0.0"
```
veloce, line 11: How come we have several versions downgraded? It looks like `app_settings` has no dependencies so it shouldn't affect it.

tom-anders: Huh, weird, I'll double-check this

tom-anders: ah, running `flutter upgrade` solved this

veloce:
One small conflict to solve @tom-anders , then I'll take care of testing and merging it :)

tom-anders:
> One small conflict to solve @tom-anders , then I'll take care of testing and merging it :)

Done, thanks!

tom-anders:
Whoops, missed two conflicts, now they should all be fixed actually

PR #887, opened 2024-07-26
fix: respect "hide rating" pref in puzzle solution screen

### LINKED ISSUE #874: With ratings not showing, puzzle ratings show after solve
### What is the problem?

After disabling the show ratings option, one can still see the rating of the puzzle once it has been solved.
image

1. Disable "show ratings" in the options
2. Solve a puzzle
3. Puzzle rating shows at the top

Rating shouldn't show

### What happened instead?

Rating showed

### Device Info

Oneplus 7 pro

### Operating System Version

Android 12

### Additional information

_No response_

tom-anders:
Closes #874

with ratings disabled:
Screenshot_1722017687

### DIFF
--- a/test/view/puzzle/puzzle_screen_test.dart
+++ b/test/view/puzzle/puzzle_screen_test.dart
@@ -5,6 +5,7 @@ import 'package:flutter/cupertino.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter_test/flutter_test.dart';
 import 'package:http/testing.dart';
+import 'package:lichess_mobile/src/model/account/account_preferences.dart';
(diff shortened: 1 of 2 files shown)


test/view/puzzle/puzzle_screen_test.dart
```diff
@@ -263,6 +265,7 @@ void main() {
         when(() => mockHistoryStorage.fetch(puzzleId: puzzle2.puzzle.id))
             .thenAnswer((_) async => puzzle2);
 
+        bool showRatings = true;
```
tom-anders, line 268: there's probably a more idiomatic way to do this...?

veloce (approved):

PR #889, opened 2024-07-28
feat: add setting to enable/disable piece magnification

### LINKED ISSUE #770: Opt-out Piece magnification
I understand V2 is an newer app and eventually features will be added it...just wanted to point out that magnification is on by default & there is no option to disable it.
Actually it's a nice feature for slower time controls but it's very bad for faster time controls as it creates unnecessary animations hence required unintended attention for it.

Option to disable it just like our old app would be nice.
liiiiiiiiiichhess_magnified_pieces

tom-anders:
Fixes #770

Is it worth adding a unit test for this? Or do we rely on chessground behaving correctly here?

### DIFF
--- a/lib/src/view/analysis/analysis_screen.dart
+++ b/lib/src/view/analysis/analysis_screen.dart
@@ -506,6 +506,9 @@ class _BoardState extends ConsumerState<_Board> {
         showLastMove: boardPrefs.boardHighlights,
         enableCoordinates: boardPrefs.coordinates,
         animationDuration: boardPrefs.pieceAnimationDuration,
+        dragFeedbackSize: boardPrefs.magnifyDraggedPiece ? 2.0 : 1.0,
+        dragFeedbackOffset:
+            Offset(0.0, boardPrefs.magnifyDraggedPiece ? -1.0 : 0.0),
         borderRadius: widget.isTablet
             ? const BorderRadius.all(Radius.circular(4.0))
             : BorderRadius.zero,
--- a/lib/src/view/analysis/analysis_screen.dart
+++ b/lib/src/view/analysis/analysis_screen.dart
@@ -499,27 +499,17 @@ class _BoardState extends ConsumerState<_Board> {
                     : IMap({sanMove.move.cg.to: annotation})
                 : null,
       ),
-      settings: cg.BoardSettings(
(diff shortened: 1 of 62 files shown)


translation/source/mobile.xml
```diff
@@ -43,4 +43,5 @@
   <string name="puzzleStormSubtitle" comment="Shown as a description for the 'Puzzle Storm' mode">Solve as many puzzles as possible in 3 minutes.</string>
   <string name="greeting" comment="A greeting that is shown on the app's home page, e.g. 'Hello, GM Magnus'. %s is substituted with the user's name">Hello, %s</string>
   <string name="greetingWithoutName" comment="Greeing that is shown on the app's home page when the user is not logged in.">Hello</string>
+  <string name="prefMagnifyDraggedPiece" comment="Title of a setting the settings screen. When enabled, the piece being dragged around the board will be magnified.">Magnify dragged piece</string>
```
tom-anders, line 46: Note: Since the setting already exists in the old app and is well known, I thought it would be okay to already add a translation.

tom-anders:
Demo:

[magnify.webm](https://github.com/user-attachments/assets/24f1e5d4-58f9-41fc-93cc-6c75e654961c)

lib/src/view/analysis/analysis_screen.dart
```diff
@@ -506,6 +506,9 @@ class _BoardState extends ConsumerState<_Board> {
         showLastMove: boardPrefs.boardHighlights,
         enableCoordinates: boardPrefs.coordinates,
         animationDuration: boardPrefs.pieceAnimationDuration,
+        dragFeedbackSize: boardPrefs.magnifyDraggedPiece ? 2.0 : 1.0,
```
tom-anders, line 509: Note: This logic is already duplicated (here and in the normal game screen), and once we have the board editor, we'll duplicate it again. Would it make sense to add a method to `BoardPreferences` that converts `BoardPreferences` to `chessground.ChessBoardSettings` and/or `chessground.ChessBoardEditorSettings` ?

tom-anders: Done, added `BoardPrefs.toBoardSettings()`

(also rebased to fix conflicts)

veloce (approved):
Thanks!

PR #891, opened 2024-07-29
fix: add start game sound

### LINKED ISSUE #890: No sound played when a game starts
### What is the problem?

Both the old app and lichess.org play the "bell" sound when a new game starts (e.g. after finding a game via the game matrix, or when your next pairing in a tournament is found). This is currently missing from the new app.

1. Open App
2. Press any time control in the game matrix (e.g. 5+0)
3. Wait for opponent
4. Once opponent is found, no bell sound is played

### App version

0.9.3

### Device Info

Pixel 4a, Android

### Additional information

Note that the bell sound is already played when a game is finished.

The correct place to implement this is probably somewhere in `game_controller.dart`? We could play the sound when receiving a new game with `game.status == started` and `game.steps.length == 0`.

Lonec-L:
fixes #890

Checks if the game just started in the first `'full'` game event handler and plays the start game sound. 

Since the first `'full'` event is always handled here, and we always receive a `'full'` event when connecting to a new game, this seemed like an appropriate place for playing the starting sound.

### DIFF
--- a/lib/src/model/game/game_controller.dart
+++ b/lib/src/model/game/game_controller.dart
@@ -103,6 +103,10 @@ class GameController extends _$GameController {
 
         _socketEventVersion = fullEvent.socketEventVersion;
 
+        if (game.status == GameStatus.started && game.steps.length == 1) {
+          ref.read(soundServiceProvider).play(Sound.dong);
+        }
+
         return GameState(
           gameFullId: gameFullId,
           game: game,


lib/src/model/game/game_controller.dart
```diff
@@ -103,6 +103,10 @@ class GameController extends _$GameController {
 
         _socketEventVersion = fullEvent.socketEventVersion;
 
+        if (game.status == GameStatus.started && game.steps.length == 1) {
```
veloce, line 106: The logic here should be more complicated than that.

One needs to ensure:
- you are playing this game and not spectating it
- you have not played a moved yet (so depending on the side, 0 or 1 move was played)

veloce (commented):
Thanks for contributing.

There is some logic to fix here. But [I'm still not sure](https://github.com/lichess-org/mobile/issues/890#issuecomment-2257616541) whether we want this or not.

tom-anders:
@Lonec-L Are you still planning to work on this? Otherwise I might take a look at this myself

lib/src/model/game/game_controller.dart
tom-anders: > you have not played a moved yet (so depending on the side, 0 or 1 move was played)

Not sure about that one - the web UI also seems to play the sound only if no move has been played at all yet.

Lonec-L:
I am not planning to work on this, so go ahead

Dne sob., 9. nov. 2024 ob 10:35 je Tom Praschan ***@***.***>
napisal/-a:

tom-anders:
@veloce you can close this PR as well (solved by #1159)

PR #896, opened 2024-08-02
feat: add setting to change shape colors

### LINKED ISSUE #841: Feature Request - Set/change default shape color
I'm really glad the new drawing mode is implemented and it works surprisingly well.

The default shape color is green and if you use the green board themes, the circles and arrows are not that visible. Also it's possible to prefer other colors, for example whenever I use arrows and circles on desktop, I always use the red color, no matter what board theme is set.

One way this could be implemented is to add a simple setting in the board/game settings. For example, this would work really well with the Puzzle screen in my opinion.

Implementing this the way I imagined sadly doesn't make it possible to have several different colored shapes on the board, but I'd say it's not that important on mobile. Having the option to set a custom default color for the shapes on the other hand could be useful.

Let me know what you think, if it's a good idea or not, if it's possible to add or not.

Thanks a lot as always and keep up the amazing work!

tom-anders:
Closes #841

[shapeColors.webm](https://github.com/user-attachments/assets/84ab5eef-5e28-442b-b97e-50657201a3a8)

### DIFF
--- a/lib/src/model/settings/board_preferences.dart
+++ b/lib/src/model/settings/board_preferences.dart
@@ -85,6 +85,10 @@ class BoardPreferences extends _$BoardPreferences {
     );
   }
 
+  Future<void> setShapeColor(ShapeColor shapeColor) {
+    return _save(state.copyWith(shapeColor: shapeColor));
+  }
+
   Future<void> _save(BoardPrefs newState) async {
     final prefs = ref.read(sharedPreferencesProvider);
     await prefs.setString(
@@ -118,6 +122,11 @@ class BoardPrefs with _$BoardPrefs {
     /// Whether to enable shape drawings on the board for games and puzzles.
     @JsonKey(defaultValue: true) required bool enableShapeDrawings,
     @JsonKey(defaultValue: true) required bool magnifyDraggedPiece,
+    @JsonKey(
+      defaultValue: ShapeColor.green,
+      unknownEnumValue: ShapeColor.green,
+    )
+    required ShapeColor shapeColor,
   }) = _BoardPrefs;
 
   static const defaults = BoardPrefs(
@@ -133,6 +142,7 @@ class BoardPrefs with _$BoardPrefs {
     pieceShiftMethod: PieceShiftMethod.either,
     enableShapeDrawings: true,
     magnifyDraggedPiece: true,
+    shapeColor: ShapeColor.green,
   );
 
   ChessboardSettings toBoardSettings() {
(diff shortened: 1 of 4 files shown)


lib/src/model/settings/board_preferences.dart
```diff
@@ -137,6 +148,23 @@ class BoardPrefs with _$BoardPrefs {
...
 
+enum ShapeColor {
+  green,
+  red,
+  blue,
+  cyan,
+  purple;
+
+  Color get color => switch (this) {
+        ShapeColor.green => LichessColors.green,
```
tom-anders, line 159: Are these the correct colors to use here? Or should we rather use the colors from here: https://github.com/lichess-org/chessground/blob/54a7e71bf88701c1109d3b9b8106b464012b94cf/src/state.ts#L178

veloce: I'd say let's use the colors defined in lichess website indeed.

Also I think showing the name of the color in the picker is not enough. Could you also show a coloured square so the user can look at the color? Thanks.

tom-anders: > I'd say let's use the colors defined in lichess website indeed.
👍 

> Also I think showing the name of the color in the picker is not enough. Could you also show a coloured square so the user can look at the color? Thanks.

Ah good idea! Maybe we could even show a small arrow instead of a square?

tom-anders:
Ok, made the following changes:

1) Using same colors as lichess.org now. IIRC blue, green, red and blue are the arrow colors supported on the website
2) Moved the setting to the new "Theme" screen, this way we can add the shapes to the preview board.
3) Added a colored box to the chouice picker entries, although arguably this is now not needed anymore, as the user already gets visual feedback on the preview board.

[shapeColors.webm](https://github.com/user-attachments/assets/d049f081-cd3d-4f59-8010-48cdaba562be)

veloce:
Great idea to put it in the theme screen! There are some weird conflicts in the game history screen @tom-anders can you see that please?

veloce:
> 3. Added a colored box to the choice picker entries, although arguably this is now not needed anymore, as the user already gets visual feedback on the preview board.

It is still useful for iOS users because the picker will be on a different screen. 

Another note: I don't think we should return the `/ blue` color in the settings value preview because:
1. it is not that clear this is about the color of the arrow
2. space is constrained on iOS devices and this should be kept short

tom-anders:
> Another note: I don't think we should return the `/ blue` color in the settings value preview because:

Done!

Also, fixed conflicts now.

PR #897, opened 2024-08-02
feat: coordinate trainer

tom-anders:
I think it makes sense to use the new `ChessboardEditor` widget here as well, since we can reuse the `onEditedSquare` callback to get notified when the user taps a square. With the normal `Chessboard` widget, this is currently not implemented, we only have callback when a (legal) move is made.

Haven't done any work on this yet, just opening the draft to let people know I'll start working on this next.


tom-anders:
sorry, clicked the wrong button, this should have been a draft PR - apparently github won't let me convert it back to a draft now...

Issue #459, opened 2024-01-08
Private messaging


veloce:
Add the private messaging feature, like in the old app.

TBestLittleHelper:
Something worth considering here, is treating messages from Lichess differently. This incudes ( but is not limited to!) warnings sent by moderators. For example, the website will show this when looking at the lobby :

![image](https://github.com/lichess-org/mobile/assets/19399877/93143a1f-ac29-42b4-af5f-bb0e583b7469)

PR #902, opened 2024-08-05
Add local notifications and listen for challenge requests

### LINKED ISSUE #432: Challenge a player to a real time game
The only way to create a game for now is with the lobby, against a random opponent.

Add a way to challenge directly a player, if allowed by their privacy settings. This issue is only for real-time game challenges. Correspondence will be done separately.

Lichess API: https://lichess.org/api#tag/Challenges/operation/challengeCreate

Will need a websocket connection (not the event stream) to get the message indicating the challenge is accepted or rejected.

Creating a real-time challenge should push a screen that cannot be exited. Only when canceling the challenge the screen is exited.

TODO: describe the challenge screen.

### LINKED ISSUE #438: Get notified when receiving a new real time game challenge
This is not related to correspondence game challenges, they work differently and are received through a push notification.

Here it need a persistent websocket connection with the following specifics:
- should be open when the app has focus and closed when the app is in background
- it should listen to the `challenges` event to be aware of new challenges

Opening a persistent websocket connection is also the way to notify other users that the player is online. Closing #404 thus.

We'll also need a `challengesProvider`, which is a [NotifierProvider](https://pub.dev/documentation/riverpod/latest/riverpod/NotifierProvider.html). 

Upon receiving a new challenge one should:
- display a notification, using the [flutter_local_notifications](https://github.com/MaikuB/flutter_local_notifications/tree/master) plugin
- have a "decline" action button in the notification (clicking on the notification would mean accepting the challenge)
- update the `challengesProvider` state 

The `challenges` message always contains the current list of all challenges, so we're notified when a challenge has been cancelled. When a challenge is canceled:
- cancel the corresponding challenge notification
- update the `challengesProvider` state

### LINKED ISSUE #439: Real time challenges widget and screen
A new screen is needed to display the current list of challenges (in and out).

It should be similar to the already existing correspondence challenges list (see: https://github.com/lichess-org/mobile/blob/b03cffa2aa32bbf2a87b4a6dad888127e65d5e9e/lib/src/view/play/create_correspondence_game_screen.dart#L177)

- User created (outgoing) challenges can be canceled.
- Incoming challenges can be declined or accepted

This screen should be accessed from a widget (which would look like a ListTile) on the "Play" tab main screen, below the "X games in play". This widget only show the first challenge, and has a "More" button to display the full screen if needed.
Capture d’écran 2024-01-04 à 18 23 23

Screen and widget data is handled by a `challengesProvider` which is roughly described in #438 .

TODO: describe the widget.

### LINKED ISSUE #640: Move notification action does not go to current state of game
### What is the problem?

When your opponent makes a move in a correspondence game if you were in the game still, clicking the app / opening the app does not go to the current state of the game but to the state before the move (i.e. the state you left the game in the background). 

1. Join correspondence game
2.  Make a couple of moves
3.  After your last move, minimise the app (i.e. put to background)
4. Have opponent make a move
5. Observe badge indicating it is your turn
6. Click / open the app 
7. State dos not update

State of game should be updated

### What happened instead?

The game went back to the last "seen" state, so I had to go back > back > into the game

### Device Info

iPhone 12 mini

### Operating System Version

iOS 17.4,1

### Additional information

_No response_

incogg:
Closes #432 
Closes #438 
Closes #439 

Adds an api for interacting with the flutter_local_notifications package
Adds persistant websocket connection to listen for challenges in the form of a notifierProvider
Adds a screen to show, accept, and cancel incoming and outgoing challenges, both realtime and correspondance

Committing the locales makes the PR huge rip

EDIT:
- Closes #640


veloce:
Whoa! This is a lot of work, I did not expect to receive such a PR :)

I'll take the time to review it properly.

I don't think adding a new translation was necessary here. As explained in [the doc](https://github.com/lichess-org/mobile/blob/main/docs/internationalisation.md#adding-mobile-specific-translations), we should not translate new strings for new feature right away.

Could you also solve the conflicts please? If you remove the added translation, only the `ios/Runner/AppDelegate.swift` conflict should be left.

veloce:
Also, can you post a screen recording showcasing this new feature? Thanks.

incogg:
Resolving that was a pain

incogg:
There is a bug that I am aware of but I don't really have a good solution for at the moment, I would appreciate some advice.

When sending a notification with flutter_local_notifications, if the notification was interacted with in a way that doesnt open the application, for example when the user presses decline on the challenge notification, the resulting callback is spawned in a background isolate. From the background isolate I can send a request to the server to decline the challenge, but I have no idea how to go about updating the state of the application for when the user goes back to it. This results in declining a challenge with the notification and then it still showing in the challenges list as the state of the challengesProvider hasnt been updated.

Do you have any idea how I could send a message from the background isolate to the main isolate, I have looked around and I cannot find any method of doing so. Perhaps I am just approaching the problem the wrong way.

Would love to hear your thoughts Veloce.

incogg:
I'll record the videos and see if i can fix the checks failing tomorrow 😃.

incogg:
> There is a bug that I am aware of but I don't really have a good solution for at the moment, I would appreciate some advice.
> 
> When sending a notification with flutter_local_notifications, if the notification was interacted with in a way that doesnt open the application, for example when the user presses decline on the challenge notification, the resulting callback is spawned in a background isolate. From the background isolate I can send a request to the server to decline the challenge, but I have no idea how to go about updating the state of the application for when the user goes back to it. This results in declining a challenge with the notification and then it still showing in the challenges list as the state of the challengesProvider hasnt been updated.
> 
> Do you have any idea how I could send a message from the background isolate to the main isolate, I have looked around and I cannot find any method of doing so. Perhaps I am just approaching the problem the wrong way.
> 
> Would love to hear your thoughts Veloce.

Nevermind the IsolateNameServer exists and i am aware of it now.
Working on a fix

incogg:
> Hey @incogg do you think the issue you described would be similar to the one mentioned in my first bullet [here (the first reply by me in that issue ticket)?](https://github.com/lichess-org/mobile/issues/640#issuecomment-2066750671)
> 
> I've outlined a few correspondence game issues for some time and I feel it matches what you outlined

Doesnt seem related no

veloce:
> Do you have any idea how I could send a message from the background isolate to the main isolate, I have looked around and I cannot find any method of doing so. Perhaps I am just approaching the problem the wrong way.

I cannot answer you atm sorry, and I don't know how to send a message to the main isolate. I will need to check this more thoroughly but I think the solution to this problem is to react to the app lifecycle events. We need to update the challenge list each time the app goes back to foreground (or is started)).

Oh I'm just seeing your next answer about `IsolateNameServer` as I'm writing this. I still think the easiest way is app lifecycle events. Sending messages between isolates looks too complex to me.

lib/l10n/app_en.arb
veloce, line 1: It would be easier to me not to have the localisation files changed. Since the PR is not dealing with translations or is not adding new translations all the l10n_* files should not be changed. Can you just cancel them? Thanks.

translation/source/mobile.xml
veloce, line 1: Can you make sure not to change indentation of existing files? Thanks for that.

veloce (commented):
Just a stylistic review for now. I'll review the rest asap.

incogg:
> > Do you have any idea how I could send a message from the background isolate to the main isolate, I have looked around and I cannot find any method of doing so. Perhaps I am just approaching the problem the wrong way.
> 
> 
> 
> I cannot answer you atm sorry, and I don't know how to send a message to the main isolate. I will need to check this more thoroughly but I think the solution to this problem is to react to the app lifecycle events. We need to update the challenge list each time the app goes back to foreground (or is started)).
> 
> 
> 
> Oh I'm just seeing your next answer about `IsolateNameServer` as I'm writing this. I still think the easiest way is app lifecycle events. Sending messages between isolates looks too complex to me.

Sounds like that would work in this case. 
I feel like messaging the main isolate is a little more flexible as you can just register any old callback to modify the applications state.
Seems like this way writing different types of notifications in future will be simpler.

incogg:
https://github.com/user-attachments/assets/75556c3b-b9d1-413d-aada-2cc31a9853f5

Here is a video of the feature

lib/src/view/play/challenge_screen.dart
```diff
@@ -406,15 +407,47 @@ class _ChallengeBodyState extends ConsumerState<_ChallengeBody> {
...
-                              : () async {
+                          : timeControl ==
+                                  ChallengeTimeControlType.correspondence
+                              ? () async {
+                                  final createGameService =
+                                      ref.read(createGameServiceProvider);
                                   showPlatformSnackbar(
                                     context,
-                                    'Correspondence time control is not supported yet',
+                                    'Sent challenge to ${widget.user.name}',
```
veloce, line 417: Have you implemented sending correspondence challenges?

I would prefer to keep one PR per feature, to simplify the review process. This one is about receiving challenges, not sending them.

Could you open another one later with the challenge creation, and also send a screen recording to show how it works? thanks.

lib/src/widgets/challenge_display.dart
```diff
@@ -0,0 +1,113 @@
...
+import 'package:lichess_mobile/src/model/common/speed.dart';
+import 'package:lichess_mobile/src/model/lobby/correspondence_challenge.dart';
+import 'package:lichess_mobile/src/model/user/user.dart';
+import 'package:lichess_mobile/src/styles/lichess_icons.dart';
+import 'package:lichess_mobile/src/styles/styles.dart';
+import 'package:lichess_mobile/src/utils/l10n_context.dart';
+import 'package:lichess_mobile/src/widgets/list.dart';
+import 'package:lichess_mobile/src/widgets/user_full_name.dart';
+
+class ChallengeDisplay extends StatelessWidget {
```
veloce, line 15: Looks like it is a challenge item in a list? It would be more clear to name it `ChallengeListItem` then.

lib/src/model/challenge/challenge_repository.dart
```diff
@@ -82,3 +102,109 @@ class ChallengeRepository {
...
+
+    ref.read(localNotificationServiceProvider).show(
+          ChallengeNotification(
+            challenge,
+            l10n,
+            onPressed: (action, id) {
+              switch (action) {
+                case ChallengeNotificationAction
+                      .accept: // accept the game and open board
+                  final repo = ref.read(challengesProvider.notifier);
```
veloce, line 188: As I explained in the other comment I don't think the challenges notifier should be used here.

It is better to use the challenge repository directly to make a server request, or use a dedicated service.

The provider (and associated notifier) is instantiated from the view (using ref.watch) and should only be used as a view controller (the term is perhaps not accurate but you get the idea).

I think this will simplify things a lot, since you don't need anymore to pass a message from isolates. The view needs only to be updated when the app is in foreground, and that should be done with the websocket only because the server is the source of truth. We don't want to modify the state locally from a user action, but just listen to server to update view state.

veloce (commented):
Thanks again for your efforts!

Most the review is about renaming but the most important aspect is about riverpod usage which should be refactored (cf review comments).

lib/src/model/challenge/challenge_repository.dart
```diff
@@ -82,3 +102,109 @@ class ChallengeRepository {
...
+    _subscription = _socketClient.stream.listen(_handleSocketEvent);
+
+    ref.onDispose(() {
+      _subscription?.cancel();
+    });
+
+    return ref.read(challengeRepositoryProvider).list();
+  }
+
+  Future<GameFullId?> accept(ChallengeId id) async {
```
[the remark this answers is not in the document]
incogg: I did this becuase the GameScreen requires a GameFullId to join the game.
You could make it return void but assuming you want to join the game you just accepted you would then need to query the repo with repo.show(id) which returns a Challenge which will contain the gameFullId.

lib/src/view/play/challenge_screen.dart
incogg: Yeah i think when i was rebasing to remove the localisation changes i accidentally removed the commit that modified the user screen to actually implement this. the implementation is already there in the comment from before left by someone else it just needs to be uncommented out. 
If you feel it needs its own PR im happy to open another one.

lib/src/model/challenge/challenge.dart
```diff
@@ -45,7 +45,7 @@ class Challenge with _$Challenge, BaseChallenge implements BaseChallenge {
     'Either clock or days must be set but not both.',
   )
   const factory Challenge({
-    required int socketVersion,
+    int? socketVersion,
```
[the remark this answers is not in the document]
incogg: I believe it is included in the api request but over the websocket it is not included

lib/src/model/challenge/challenge_repository.dart
```diff
@@ -82,3 +102,109 @@ class ChallengeRepository {
     }
   }
 }
+
+@Riverpod(keepAlive: true)
+class Challenges extends _$Challenges {
```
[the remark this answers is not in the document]
incogg: now that im thinking about it cant we just replace the challenge repository with the notifier provider.
essentially its the same thing except it caches some local state
it wouldn't break anything in the only place its used which is in CreateGameService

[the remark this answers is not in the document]
incogg: actually we cant becuase if you are saying that we dont define methods that return values on notifiers then show and create wouldnt work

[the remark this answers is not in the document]
veloce: yep; also I'd like to keep the consistency within the app. I'm not too happy with how the repository classes are done right now, but I still want to keep the separation.

lib/src/model/challenge/challenge.dart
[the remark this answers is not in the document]
veloce: hmm, ok I guess that makes sense.

lib/src/model/challenge/challenge_repository.dart
```diff
@@ -82,3 +102,109 @@ class ChallengeRepository {
...
+    ref.onDispose(() {
+      _subscription?.cancel();
+    });
+
+    return ref.read(challengeRepositoryProvider).list();
+  }
+
+  Future<GameFullId?> accept(ChallengeId id) async {
+    final repo = ref.read(challengeRepositoryProvider);
+    _stateRemove(id);
```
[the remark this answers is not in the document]
incogg: yeah i realised now that its actually completely redundant as the an updated list is sent on the socket any time something changes anyway
programming at 3am moment

lib/src/view/play/challenge_screen.dart
incogg: i just uncommented it out as its not a big change

incogg:
Thanks for all the feedback Veloce, this is the first time I have ever used flutter or riverpod, it seems like I didn't fully comprehend how to use riverpod correctly. I feel like I have a much better grasp now.

incogg:
https://github.com/user-attachments/assets/42eb84cc-7de4-465f-8063-be1134e04ba2

And here is a video showing the now enabled user challenges

lib/src/app.dart
```diff
@@ -106,6 +112,56 @@ class _AppState extends ConsumerState<Application> {
       setOptimalDisplayMode();
     }
 
+    ref.read(localNotificationServiceProvider).init();
+
+    ref.listenManual(challengesProvider, (prev, current) {
+      final prevIds = prev!.value!.inward.map((challenge) => challenge.id);
```
veloce, line 118: I don't think we can assume the `prev` and the `value` are never null. I'd write that if you want prevIds to be defined no matter what:
```dart
final prevIds = prev?.value?.inward.map((challenge) => challenge.id) ?? [];
```

lib/src/app.dart
```diff
@@ -106,6 +112,56 @@ class _AppState extends ConsumerState<Application> {
...
+      final inward = current.value!.inward;
```
veloce, line 119: Same here, let's not assume `current.value` is never null. I guess here we can simply do:
```dart
if (!current.hasValue) return;
```

lib/src/app.dart
```diff
@@ -106,6 +112,56 @@ class _AppState extends ConsumerState<Application> {
...
+      final repo = ref.read(challengeRepositoryProvider);
+      final l10n = ref.read(l10nProvider).strings;
+
+      inward
+          .map((element) => element.id)
+          .where((id) => !prevIds.contains(id))
+          .map((id) => inward.firstWhere((element) => element.id == id))
```
veloce, line 126: Both `.map()` are unnecessary, you could simply write `.where((element) => !prevIds.contains(element.id))` and remove the maps.

lib/src/app.dart
```diff
@@ -106,6 +112,56 @@ class _AppState extends ConsumerState<Application> {
...
+                onPressed: (action, id) async {
+                  switch (action) {
+                    case ChallengeNotificationAction
+                          .accept: // accept the game and open board
+                      await repo.accept(challenge.id);
+                      final fullId = await repo.show(challenge.id).then(
+                            (challenge) => challenge.gameFullId,
+                          );
+                      pushPlatformRoute(
+                        ref.read(currentNavigatorKeyProvider).currentContext!,
```
veloce, line 141: I wonder if that is necessary to give the current navigator context. I suppose simply passing `context` is enough?

lib/src/app.dart
```diff
@@ -106,6 +112,56 @@ class _AppState extends ConsumerState<Application> {
...
+
+    ref.listenManual(challengesProvider, (prev, current) {
+      final prevIds = prev!.value!.inward.map((challenge) => challenge.id);
+      final inward = current.value!.inward;
+      final repo = ref.read(challengeRepositoryProvider);
+      final l10n = ref.read(l10nProvider).strings;
+
+      inward
+          .map((element) => element.id)
+          .where((id) => !prevIds.contains(id))
```
veloce, line 125: I wonder if this check is enough to prevent sending twice the notification for a same challenge. I assume there are other checks, like ensuring the notification ID is linked to a specific challenge to not send it twice?

lib/src/model/notifications/local_notification_service.dart
```diff
@@ -0,0 +1,156 @@
...
+      notification.body,
+      notification.notificationDetails,
+      payload: notification.payload,
+    );
+    _log.info(
+      '[Local Notifications] Sent notification: ($id | ${notification.title}) ${notification.body} (Payload: ${notification.payload})',
+    );
+
+    // if the notification implements LocalNotificaitonCallback
+    if (notification is LocalNotificationCallback) {
```
veloce, line 87: Instead of checking an interface, you could simply add a nullable `callback` field to the `LocalNotification` class and check for its presence. That would avoid unnecessary casts.

lib/src/model/notifications/local_notification_service.dart
```diff
@@ -0,0 +1,156 @@
...
+  void _updateLocalisations() {
+    final l10n = _ref.read(l10nProvider);
+    InfoNotificationDetails(l10n.strings);
+    ChallengeNotificationDetails(l10n.strings);
+  }
+
+  Future<int> show(LocalNotification notification) async {
+    final id = currentId++;
+    await _notificationPlugin.show(
+      id,
```
veloce, line 76: Just thinking out loud here, dunno yet if it's a good idea. 
But if you use several time `_plugin.show` with the same ID, does the plugin ignore further calls?
If we had a way to map a challenge Id to a unique numeric id we'd ensure a challenge notification is not shown twice.

lib/src/model/notifications/local_notification_service.dart
```diff
@@ -0,0 +1,156 @@
...
+
+    return id;
+  }
+
+  Future<void> cancel(int id) async {
+    _log.info('[Local Notifications] canceled notification id: [$id]');
+    return _notificationPlugin.cancel(id);
+  }
+
+  void _onReceivePortMessage(dynamic message) {
```
veloce, line 100: I don't understand why it is still needed to send message to the main isolate. Can you explain why it is useful? Thanks.

I still feel it should not be necessary, but maybe I'm missing sth.

lib/src/model/notifications/local_notification_service.dart
```diff
@@ -0,0 +1,156 @@
...
+    }
+  }
+
+  void _notificationResponse(NotificationResponse response) {
+    final callback = _callbacks[response.id];
+    if (callback != null) callback(response.actionId, response.payload);
+  }
+
+  @pragma('vm:entry-point')
+  static void _notificationBackgroundResponse(NotificationResponse response) {
```
veloce, line 125: Here instead of sending a message, could we not just use the server challenge API to react to user input?

I know we can't have access to riverpod scope here so we're limited. But if the only action is a network request you can instantiate a one time client. Sth like:
```dart
final client = LichessClient(httpClientFactory());
try {
    // do sth
} finally
  client.close()
}
```

lib/src/view/play/challenge_screen.dart
```diff
@@ -406,15 +407,47 @@ class _ChallengeBodyState extends ConsumerState<_ChallengeBody> {
...
                                   );
-                                },
+                                  final response =
```
veloce, line 419: You should use the `_pendingCreateGame` of the `FutureBuilder` to ensure the button cannot be clicked twice (which should be renamed to `_pendingCreateChallenge`). So:

```dart
_pendingCreateChallenge = createGameService.newChallgenge(...);

await _pendingCreateChallenge;

if (context.mounted) {

 ... do stuff
}
```


This file starts to be hard to read, and the indentation is horrible. Some widgets should be extracted, but don't worry about it I'll refactor it later after this PR is merged.

lib/src/widgets/challenge_display.dart
veloce, line 1: the file itself should be named `challenge_list_item.dart`

veloce (commented):
This is getting in the right direction, thanks again for your work!

lib/src/app.dart
incogg: i think we just want to exit if prev is null though.
if prev is null it means that we have the initial state of the provider, or the app has just opened, and if there is an incoming challenge request when we open the app for the first time, then immediately after opening the app the user will get a notification for each challenge request they currently have.
i think that would be pretty annoying, and if they are opening the app for the first time they will just see that they have challenge requests based on the notification pip on the challenge button.

incogg: maybe i missunderstand how the ! works here, i assumed it just exited on a null value, does it also throw an error?

lib/src/app.dart
incogg: That id is from the server, i would assume that lichess ensures uniqueness .

lib/src/app.dart
incogg: i misunderstood how the ! operator was used, i thought it just exited i didnt realise it also threw an error.

lib/src/model/notifications/local_notification_service.dart
incogg: it actually overwrites the old notification with new content.

lib/src/model/notifications/local_notification_service.dart
incogg: the thing is we dont have much information about the notification, what you suggest below with directly sending a message to the api in _notificationBackgroundResponse() is definitely possible, but this is a very basic example we have here with only one type of notification.

with multiple notification types that have to do some action in the background isolate the only real meaningful data that we have access to is the actionId and the payload.

this means that we will have to somehow encode the type of notification that is being interacted with in one of these two things (probably the payload) which could look something like
"{notificationType}:{payload}"
as soon as we start dealing with 4 or 5 different types of notifications itll get pretty messy

sending messages between isolates also makes defining new notifications easier as all the behaviour associated with that notification is declared in that one function, instead of two different places which i think improves the code readability.

i dont think its impossible to have it separate and to keep it isolated, but i also dont understand the tradeoff. it seems like forwarding the notificationResponse to the main isolate to handle just makes things simpler and more readable without any downside.

lib/src/model/notifications/local_notification_service.dart
incogg: now the notification gets to define its own id. just using hashid will probably avoid collisions

lib/src/model/notifications/local_notification_service.dart
veloce: When threading is involved I'm not sure things are more readable (or understandable) but I get your point.

When your speaking about multiple notification type your thinking about future usage? But we don't know yet what that will be yet.

We can define an interface for the payload, so that we always know the type of notification:
```dart
class NotificationPayload<T> {
   final String notifType;
   final T data;
}
```

Used in all notifications, (de)serialised to JSON. That's also how it is done with push notifications.
   
Also there's another thing that I'm not sure about. When you receive a notification when the app is in background there are 2 possibilities:
- user declines: then the app should stay in background and the cancel request should be send to server
- user accepts: app should be bring to foreground

I was seeing things like that, but maybe the flutter local notification plugin doesn't work that way, you'll tell me:
- in the decline case we have to deal with the limited background isolate where we don't have access to the whole app but that's ok if we only need to send a request.
- in the accept case, I'd expected a similar mechanism as the [handleMessage callback](https://github.com/lichess-org/mobile/blob/af96a89efb6cbb1e0ea67b3c20845b262a32bd6f/lib/src/app.dart#L374) of the firebase push plugin (that fires when a message clicked caused the app to open). Do we have a similar thing in the local notification plugin?

lib/src/model/notifications/local_notification_service.dart
incogg: LichessClient requires a ref to create.
so now a static instance of the LocalNotificationService is required
```dart
  @pragma('vm:entry-point')
  static void _notificationBackgroundResponse(NotificationResponse response) {
    if (response.payload == null) return;
    final instance = LocalNotificationService.instance;
    if (instance == null) return;

    final payload = NotificationPayload.fromJson(
      jsonDecode(response.payload!) as Map<String, Object?>,
    );
    switch (payload.type) {
      case NotificationType.info:
      case NotificationType.challenge:
        final id = payload.data['id'] as String;
        instance._ref
            .read(challengeRepositoryProvider)
            .decline(ChallengeId(id));
    }
  }
```

lib/src/model/notifications/local_notification_service.dart
incogg: >When threading is involved I'm not sure things are more readable (or understandable) but I get your point.

i really think it does.
now the behaviour relating to the challenge notification is split up into 3 different files, and there is a bunch of casting and null checks to extract data from the notification payload.
i think that with your solution of keeping things separate it becomes way more confusing as now you actually **do** have to think about different threads as with some reponses youre in a background isolate and with some you arent.

not only that but as seen in this https://github.com/lichess-org/mobile/pull/902#discussion_r1711238492 we dont have access to any refs, so now a static instance is required somewhere to get access to one as doing anything with the lichess servers requires a LichessClient, which in and of itself requires a ref.

with messaging the main isolate everything is just abstracted away and in future it doesnt have to be thought about again. maybe the current implementation could be a little cleaner, maybe by turning the NotificationReponse into json and then sending over the RecievePort and deserializing it there.
this means that all the behaviour relating to a notification is just included in the same file it's declared in and when implementing it you dont need to think about whether or not a particular response will open the application or not because it has already been handled for you.

> I was seeing things like that, but maybe the flutter local notification plugin doesn't work that way, you'll tell me:

essentially how the flutter_local_notfication package works is that there are two different types of interactions a user can have with a notification. One that opens the application and one that doesnt.

in the case that the application is opened _notificationReponse is called,
and in the case that the applciation remains closed _notificationBackgroundResponse is executed.

we can handle it similarly to how handleMessage does with firebase notifications as seen below, but this function is going to get very big very quickly.

```dart

  /// Handle a notification response that caused the application to open
  Future<void> _notificationResponse(NotificationResponse response) async {
    if (response.payload == null) return;
    final payload = NotificationPayload.fromJson(
      jsonDecode(response.payload!) as Map<String, Object?>,
    );

    switch (payload.type) {
      case NotificationType.info:
      case NotificationType.challenge:
        switch (response.actionId) {
          case 'accept':
            final repo = ref.read(challengeRepositoryProvider);
            final id = payload.data['id'] as String;
            final challengeId = ChallengeId(id);

            await repo.accept(challengeId);
            final fullId = await repo
                .show(challengeId)
                .then((challenge) => challenge.gameFullId);

            _popNavState();
            pushPlatformRoute(
              ref.read(currentNavigatorKeyProvider).currentContext!,
              rootNavigator: true,
              builder: (BuildContext context) =>
                  GameScreen(initialGameId: fullId),
            );
          case null:
            _popNavState();
            pushPlatformRoute(
              context,
              builder: (BuildContext context) =>
                  const ChallengeRequestsScreen(),
            );
        }
    }
  }
```

>When your speaking about multiple notification type your thinking about future usage? But we don't know yet what that will be yet.

i have already started working on #459 and ill probably add notifications to that as well as the ability to respond to a message within the notification itself.
basically i dont think it matters what type of notifications are added in the future. As long as there are more notifications in the future and we handle them like this the _notificationReponse functions will become more messy regardless.

veloce: Ok, let's keep for now the message to main isolate, I prefer to focus on the more important matters. A couple of questions about it though:
- what happens if the use kills the app, then for instance, cancels a challenge through a notification? does the notification disappear with the app being killed? or is the message just being ignored?
- a scheduled notification cannot send messages to main isolate because we cannot assume the app will be launched when the notification action is tapped.

Now, let's deal with the notifications handlers. Here I'm sure that I don't want the callback based system. My main concern is code readability and maintenance over years. With callbacks it is definitely harder to understand, especially if you come back to the code months after, or for anybody not familiar with the codebase.

So we definitely need a payload type as I described in my other comment:
```dart
class NotificationPayload<T> {
   final PayloadType payloadType;
   final T data;
}
```

It has the added benefit of a common interface for all notification messages the app will have to deal with in the future. And it adds another type safety layer, over just using `String` and let the message handler define their own format.

> we can handle it similarly to how handleMessage does with firebase notifications as seen below, but this function is going to get very big very quickly.

I won't be big if we use the following approach:
```dart
  /// Handle a notification response that caused the application to open
  Future<void> _notificationResponse(NotificationResponse response) async {
    if (response.payload == null) return;
    final payload = NotificationPayload.fromJson(
      jsonDecode(response.payload!) as Map<String, Object?>,
    );

    switch (payload.type) {
      case PayloadType.challenge:
         ref.read(challengeService).onNotificationResponse(response, payload);
      case PayloadType.message:
         ref.read(messagingService).onNotificationResponse(response, payload);
    }
  }
```

That way we still have a clean separation of concerns: the notification service only job is to receive and dispatch the notifications to the proper handlers. It is not messy and it is easy to follow. All message types that the app deal with are defined in only one place.

lib/src/model/notifications/local_notification_service.dart
veloce: We certainly don't want to do that, I agree on that point.

lib/src/model/notifications/local_notification_service.dart
```diff
@@ -0,0 +1,152 @@
...
+    this.payload,
+  });
+
+  int get id;
+
+  final String title;
+  final String? body;
+  final String? payload;
+  final NotificationDetails notificationDetails;
+  final void Function(String? actionId, String? payload)? callback = null;
```
veloce, line 151: As explained in the other comment, let's remove the `callback` here.

incogg:
okay i think i have solved all the problems.

i got rid of the messaging between isolates and i fixed the problems with not having access to refs by creating a separate riverpod provider scope that exists on the background isolate

no more dealing with threads and the notification declarations are all in one file

lib/src/app.dart
```diff
@@ -314,6 +318,31 @@ class _EntryPointState extends ConsumerState<_EntryPointWidget> {
         }
       }
     });
+
+    ref.read(localNotificationServiceProvider).init();
+
+    ref.listenManual(challengesProvider, (prev, current) {
```
veloce, line 324: You should not listen to a provider that should be only linked to a specific view to achieve that. (Cf. other comment of challengesProvider).

`challenges` socket event is a global event, which means it can appear on any socket route of lichess. The SocketPool already makes sure the app is connected to a socket, so you don't need to care about which socket it is connected to. What you want here is to listen to the [global socket event stream](https://github.com/lichess-org/mobile/blob/ba871fe83aeffd16da6878f0fc994420f41c84f9/lib/src/model/common/socket.dart#L52). (You'll need to add 'messages' to the white list).

You could listen to the stream here, but it is probably better to do that in the `challengeService` (where you'd start listening in the `init` method).

lib/src/model/notifications/challenge_notification.dart
```diff
@@ -0,0 +1,205 @@
...
+        await repo.accept(payload.id);
+
+        final fullId = await repo
+            .show(payload.id)
+            .then((challenge) => challenge.gameFullId);
+
+        final context = ref.read(currentNavigatorKeyProvider).currentContext!;
+        if (!context.mounted) break;
+
+        final navState = Navigator.of(context);
```
veloce, line 45: Not sure to understand the reason to pop all routes. Is it to prevent pushing twice the screen?

lib/src/model/notifications/challenge_notification.dart
```diff
@@ -0,0 +1,205 @@
...
+        if (navState.canPop()) {
+          navState.popUntil((route) => route.isFirst);
+        }
+
+        // sometimes becuase GameScreen connects to a socket we dont recieve a new event on the event listener in time
+        // so we just make sure to refresh this value when we go back to the home tab
+        ref.invalidate(challengesProvider);
```
veloce, line 52: This is not needed anymore if we listen to the global event stream.

lib/src/model/notifications/challenge_notification.dart
```diff
@@ -0,0 +1,205 @@
...
+        : '${_l10n.casual} • $time';
+  }
+}
+
+class ChallengePayload {
+  const ChallengePayload(this.id);
+
+  final ChallengeId id;
+
+  NotificationPayload toNotifiationPayload() {
```
veloce, line 125: Use a getter here instead of a method.

lib/src/model/notifications/local_notification_service.dart
```diff
@@ -0,0 +1,178 @@
...
+  }
+
+  Future<void> cancel(int id) async {
+    _log.info('canceled notification id: [$id]');
+    return _notificationPlugin.cancel(id);
+  }
+
+  void _handleResponse(int id, String? actionId, NotificationPayload payload) {
+    switch (payload.type) {
+      case PayloadType.info:
```
veloce, line 92: missing `break` here

lib/src/utils/system.dart
```diff
@@ -14,12 +14,11 @@ class System {
   static const instance = System._();
 
   /// Returns the system total RAM in megabytes.
-  Future<int?> getTotalRam() async {
+  Future<int> getTotalRam() async {
     try {
-      return await _channel.invokeMethod<int>('getTotalRam');
-    } on PlatformException catch (e) {
-      debugPrint('Failed to get total RAM: ${e.message}');
```
veloce, line 21: Why this change? Now if the RAM cannot be retrieved it fails silently.

lib/src/widgets/challenge_list_item.dart
veloce, line 1: I tend to put in that folder only generic widgets. Here I feel this one belongs to `view/play` (or `view/challenge`).

lib/src/widgets/challenge_list_item.dart
```diff
@@ -0,0 +1,141 @@
...
+  });
+
+  final Challenge challenge;
+  final LightUser user;
+  final VoidCallback? onPressed;
+  final VoidCallback? onCancel;
+
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {
+    final me = ref.read(authSessionProvider)?.user;
```
veloce, line 34: Always use `ref.watch` in widgets `build` methods, as recommended in Riverpod's documentation.

veloce (commented):
Nicely done with the `ProviderScope`. Thanks! Now there are a couple of things to fix related to the socket listening.

veloce:
I must thank you again for your work and your patience @incogg , this has probably taken longer than expected on your side. This feature is essential to the app and is quite tricky, that's why the review process takes more time here. But I think we're heading in the right direction now.

lib/src/model/notifications/challenge_notification.dart
incogg: yeah it just makes sure the same screen isnt pushed twice
same thing is done in app.dart with the firebase notifications

lib/src/utils/system.dart
incogg: i should have changed it back its just an artifact from testing i was doing.
basically method channels dont work on the background isolate and it was causing the `appInitializationProvider` to fail 
i changed it back and added a new check for the error that occurs when you call a method channel on the background isolate

incogg:
no worries 
maybe a little longer than expected but its okay. I picked an issue in the release milestone because i want to help get the app closer to release. I expected the work.

lib/src/model/notifications/challenge_notification.dart
```diff
@@ -0,0 +1,226 @@
...
+  return ChallengeService(ref);
+}
+
+class ChallengeService {
+  ChallengeService(this.ref);
+
+  final ChallengeServiceRef ref;
+
+  void init() {
+    ref.listen(challengesProvider, (prev, current) {
```
veloce, line 30: Here you should listen `socketGlobalStream`, the server holds all the state and is the source of truth. Also we don't want the service to depend on `challengesProvider`.

If you want to avoid parsing twice the socket json object, I'd do it the other way around: export a `Stream<ChallengesList>` in the `ChallengeService` (this one is kept alive because it is not bound to a view, and so we can access it anywhere), that you'd consume in the `challengesProvider`.

veloce (commented):
Only a last refactoring related to not listening to the riverpod provider state, and we're good in my opinion.

lib/src/model/challenge/challenge.dart
```diff
@@ -40,10 +40,6 @@ abstract mixin class BaseChallenge {
 class Challenge with _$Challenge, BaseChallenge implements BaseChallenge {
   const Challenge._();
 
-  @Assert(
```
incogg, line 43: as far as i can tell this doesnt break anything, but im not sure if some checks need to be done instead of this
thoughts @veloce

incogg: needed the change becuase otherwise the challenge service breaks when trying to parse the list if there is an unlimited challenge in it.

veloce: So an unlimited challenge will have no clock and no days, that's fine. I suppose then you could replace the check to ensure that both days and clock cannot be set at the same time.

lib/src/view/user/challenge_requests_screen.dart
```diff
@@ -0,0 +1,128 @@
...
+      navigationBar: const CupertinoNavigationBar(),
+      child: _Body(),
+    );
+  }
+}
+
+class _Body extends ConsumerWidget {
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {
+    final challengeRepo = ref.read(challengeRepositoryProvider);
```
veloce, line 50: Don't use ref.read inside build. This is a one time read, it should be done in the callbacks directly.

lib/src/view/play/challenge_list_item.dart
```diff
@@ -0,0 +1,141 @@
...
+      ChallengeTimeControlType.unlimited => '∞',
+    };
+
+    final subtitle = challenge.rated
+        ? '${context.l10n.rated} • $time'
+        : '${context.l10n.casual} • $time';
+
+    final color = isMyChallenge ? null : LichessColors.green.withOpacity(0.2);
+
+    return Container(
```
veloce, line 60: The Container could be replaced with a `ColoredBox`.

veloce:
I'll test your PR tomorrow @incogg . Really excited to have the challenges in the app!

veloce:
To see the correct diff without extra main commits: https://github.com/lichess-org/mobile/compare/main...incogg:mobile:feat/challenges

PR #905, opened 2024-08-06
fix: flip players on flip board action in archived game screen

Mauritz8:


veloce:
I originally wanted to only turn the board and not the players, but since it's done in the normal game screen, we should do it here for consistency.

PR #912, opened 2024-08-10
fix: use correct piece assets when visualizing promition in analysis

tom-anders:
Needs https://github.com/lichess-org/flutter-chessground/pull/48

Screenshot_1723328086


veloce:
Closing it since I've dealt with it in #961

PR #922, opened 2024-08-12
fix: ignore pointer events when editing widgets

### LINKED ISSUE #915: Editing the Home Screen should disable Quick pairing
### Request

When editing the Home screen's appearance, the buttons on the grid should be disabled/no longer be interactive. They should only be re-enabled outside the editing view.

### Rationale

1. I noticed the new feature to edit the Home screen.
2. I used it and thought I could customise the tiles that appear in the Quick Pairing grid.
3. I pressed on 1+0 to take it off. Instead, it launched me into a game I was not prepared for and lost quickly - not really a bullet player.

I believe other users may also expect to not be able to interact with the controls on that screen while in Editing view, until they have saved their new configuration.

tom-anders:
Closes #915

[editing.webm](https://github.com/user-attachments/assets/d781a7d8-1fa1-4fc7-9f08-bb7e43924f6a)

### DIFF
--- a/lib/src/view/home/home_tab_screen.dart
+++ b/lib/src/view/home/home_tab_screen.dart
@@ -372,7 +372,12 @@ class _EditableWidget extends ConsumerWidget {
                   },
                 ),
               ),
-              Expanded(child: child),
+              Expanded(
+                child: IgnorePointer(
+                  ignoring: isEditing,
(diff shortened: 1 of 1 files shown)


veloce (approved):
Thanks!

PR #924, opened 2024-08-12
feat: add offline games ("Over the board")

### LINKED ISSUE #287: [Feature] Offline play
Please add the possibility to play in offline on the same device for two people, like on the old mobile app (the "Over the board" thing).

It should have the same settings, those being : 
- the ability to flip the piece when a player has played
- the ability to use symetric pieces.

tom-anders:
Closes #287

Design is pretty close to the old lichess app and the basic functionality is implemented. Some minor features from the old app are missing and could be added in follow-up PRs:

- Different clock modes (Bronstein delay, handicap etc.)
- Share game as PGN 
- Import game on lichess
- Claim threefold repitition
- Symmetric pieces
- Flip pieces after move


tom-anders:
Demonstration video:

[over_the_board.webm](https://github.com/user-attachments/assets/b31f1889-fdfa-4d24-9161-16419186a017)

EmmetSchuler:
Everything looks really good so far!
Just some quick feedback that I thought might be useful:

I think there should be an option to pick the piece color for the main player, so that you don't have to rotate the phone and give all the menus / options to the secondary player. 

There's an option in the original otb mode that I find particularly useful, where it flips all pieces and info for the person whose turn it is to move. Would that be something to consider?

Having the option to add +15 seconds of time to your opponent's clock would be nice to have too.

tom-anders:
Thanks!

> I think there should be an option to pick the piece color for the main player, so that you don't have to rotate the phone and give all the menus / options to the secondary player.

Sounds useful, I'll add it to the settings as well

> There's an option in the original otb mode that I find particularly useful, where it flips all pieces and info for the person whose turn it is to move. Would that be something to consider?

Yeah, looks like it needs to be implemented in `flutter-chessground`  first, I opened a PR :+1: 

> Having the option to add +15 seconds of time to your opponent's clock would be nice to have too.

Also a useful feature, but I'm not sure where we would put this button - Probably in the bottom bar? any other ideas?

tom-anders:
Rebased against main and adapted to chessground API changes.

Some updates on the comments above:

> Thanks!
> 
> > I think there should be an option to pick the piece color for the main player, so that you don't have to rotate the phone and give all the menus / options to the secondary player.
> 
> Sounds useful, I'll add it to the settings as well

Added the usual "flip board" button in the bottom bar for this.

> > There's an option in the original otb mode that I find particularly useful, where it flips all pieces and info for the person whose turn it is to move. Would that be something to consider?
> 

Done :+1: 

> > Having the option to add +15 seconds of time to your opponent's clock would be nice to have too.

Since this is now planned for normal/online games as well (#936), let's wait for that to be done and then add it here as well

veloce (approved):
Outstanding work! Thank you.

PR #938, opened 2024-08-22
refactor: add common widget for BottomBar

tom-anders:
This gets rid of a lot of code duplication and makes it more straight-forward to build new screens for new features.

Note that the Row in BottomBar now has MainAxisAlignment.spaceAround by default, so it's not necessary anymore to wrap buttons in `Expanded`.

### DIFF
--- a/lib/src/view/analysis/analysis_screen.dart
+++ b/lib/src/view/analysis/analysis_screen.dart
@@ -33,6 +33,7 @@ import 'package:lichess_mobile/src/view/engine/engine_gauge.dart';
 import 'package:lichess_mobile/src/view/opening_explorer/opening_explorer_screen.dart';
 import 'package:lichess_mobile/src/widgets/adaptive_action_sheet.dart';
 import 'package:lichess_mobile/src/widgets/adaptive_bottom_sheet.dart';
+import 'package:lichess_mobile/src/widgets/bottom_bar.dart';
 import 'package:lichess_mobile/src/widgets/bottom_bar_button.dart';
 import 'package:lichess_mobile/src/widgets/buttons.dart';
 import 'package:lichess_mobile/src/widgets/feedback.dart';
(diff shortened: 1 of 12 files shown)


veloce:
While I think it is a good change, we must be careful it doesn't break some screens. We should make a list of all impacted screens and check they still look fine.

Space around is the same as `Expanded` if all the children are wrapped in an `Expanded`. This may be not the case everywhere (and is why I used `Expanded` instead of `spaceAround` I believe), as you might want to expand some buttons only.

tom-anders:
> While I think it is a good change, we must be careful it doesn't break some screens. We should make a list of all impacted screens and check they still look fine.

I'll add Screenshots for all impacted screens. 👍 

> Space around is the same as `Expanded` if all the children are wrapped in an `Expanded`. This may be not the case everywhere (and is why I used `Expanded` instead of `spaceAround` I believe), as you might want to expand some buttons only.

Yeah that's why I exposed the `mainAxisAlignment` property, so expanding only one of the buttons should still be possible. However, all the current bottoms bars seems to either use `spaceAround` alignment or `Expanded` for all buttons. I can double-check to make sure.

tom-anders:
Analysis:

Screenshot_1725200820

Opening Explorer:

Screenshot_1725200994

Board Editor:

Screenshot_1725200889

Offline Correspondence:

Screenshot_1725200974

Archived Game:

Screenshot_1725201013

Game Loading:

Screenshot_1725201066

Online Game:

Screenshot_1725201122

Puzzle Screen:

Screenshot_1725201175

Puzzle Storm:

Screenshot_1725201191

Puzzle Streak:

Screenshot_1725201210

TV:

Screenshot_1725201308

veloce:
Thanks for the screenshots @tom-anders !

veloce (approved):

PR #944, opened 2024-08-26
More progress on broadcast feature

julien4215:
https://github.com/user-attachments/assets/040a1dc0-783d-4e6d-b4a8-174488d419ee

### DIFF
--- a/lib/src/view/broadcast/broadcast_game_screen.dart
+++ b/lib/src/view/broadcast/broadcast_game_screen.dart
@@ -160,7 +160,6 @@ class _BroadcastBoardWithHeaders extends ConsumerWidget {
 
   @override
   Widget build(BuildContext context, WidgetRef ref) {
-    // TODO use borderRadius with players widget
     return Column(
       children: [
         _PlayerWidget(
@@ -188,13 +187,13 @@ class _BroadcastBoard extends ConsumerStatefulWidget {
     this.roundId,
     this.gameId,
     this.boardSize,
-    this.hasRadius,
+    this.hasShadow,
   );
 
   final BroadcastRoundId roundId;
   final BroadcastGameId gameId;
   final double boardSize;
-  final bool hasRadius;
+  final bool hasShadow;
 
   @override
   ConsumerState<_BroadcastBoard> createState() => _BroadcastBoardState();
@@ -263,7 +262,7 @@ class _BroadcastBoardState extends ConsumerState<_BroadcastBoard> {
                   : IMap({sanMove.move.to: annotation})
               : null,
       settings: boardPrefs.toBoardSettings().copyWith(
-            boxShadow: widget.hasRadius ? boardShadows : const <BoxShadow>[],
+            boxShadow: widget.hasShadow ? boardShadows : const <BoxShadow>[],
             drawShape: DrawShapeOptions(
               enable: boardPrefs.enableShapeDrawings,
               onCompleteShape: _onCompleteShape,


ijm8710:
It's minor but not sure tournament is the right word to switch between open and gm specific.
Maybe draw. Veloce may know the best word but I think tournament may falsely induce the user to think that's a dropdown to switch to a whole new tournament altogether when it's not.

ornicar:
Very impressive work Julien! :+1:

tom-anders:
Looks really cool!

In the analysis screen (final part of the video), would it make sense to display the player's names above/below the board? The mobile site also does this.

julien4215:
> Looks really cool!
> 
> In the analysis screen (final part of the video), would it make sense to display the players' names above/below the board? The mobile site also does this.

Sure, I also think it would be better with the players' names and clock/result.

veloce:
Yeah, awesome work @julien4215 . I think we can release like that indeed, even without the eval bar. I'll deal with it asap. Right now I'm focusing on releasing the new editor and opening explorer so I think it'll be part of the next release.

veloce:
I'm focusing right now on the next release but I'll review asap. Can you please fix the test and solve the conflict in the meantime? Thanks!

julien4215:
I'll fix the tests and the conflict once the PR is fully ready. I have to fix some problems with the clock and the analysis screen.

veloce:
> I'll fix the tests and the conflict once the PR is fully ready. I have to fix some problems with the clock and the analysis screen.

In that case I prefer that the PR is a draft. Easier for me to choose PR to review. Thanks!

lib/src/model/analysis/analysis_controller.dart
```diff
@@ -46,6 +49,7 @@ class AnalysisOptions with _$AnalysisOptions {
     int? initialMoveCursor,
     LightOpening? opening,
     Division? division,
+    @Default(false) bool isBroadcast,
```
veloce, line 52: I think we should not re-use the analysis controller for broadcast games.

I know there is a lot of logic in common, but still, they have complete different purposes and enough differences so that we should split them.

For now I'd say let's copy/paste all the logic we need from analysis controller into `BroadcastGameController`. We'll figure out later how to factor out common logic properly. 

I said the same thing to @tom-anders about study controller. I guess both will share a good amount of logic too.

(One thing we can extract now I think, is the engine preferences. We should separate what is purely engine settings (nb of threads, memory, etc) from the analysis settings.)

lib/src/model/broadcast/broadcast_preferences.dart
```diff
@@ -0,0 +1,50 @@
...
+class BroadcastPrefState with _$BroadcastPrefState {
+  const factory BroadcastPrefState({
+    required bool showEvaluationBar,
+  }) = _BroadcastPrefState;
+
+  factory BroadcastPrefState.defaults() => const BroadcastPrefState(
+        showEvaluationBar: true,
+      );
+
+  factory BroadcastPrefState.fromJson(Map<String, dynamic> json) =>
```
veloce, line 48: Perhaps we can put a `try/catch` here so it is "safe" in the future?

lib/src/model/broadcast/broadcast_repository.dart
```diff
@@ -51,34 +71,75 @@ BroadcastsList _makeBroadcastResponseFromJson(
...
+      name: pick('name').asStringOrThrow(),
+      imageUrl: pick('image').asStringOrNull(),
+      description: pick('description').asStringOrNull(),
+      information: (
+        format: pick('info', 'format').asStringOrNull(),
+        timeControl: pick('info', 'tc').asStringOrNull(),
+        players: pick('info', 'players').asStringOrNull(),
+        dates: pick('dates').letOrNull(
+          (pick) => (
+            startsAt: pick(0).asDateTimeFromMillisecondsOrThrow().toLocal(),
```
veloce, line 99: I realise now the
```dart
      return DateTime.fromMillisecondsSinceEpoch(value, isUtc: true);
```

from `asDateTimeFromMillisecondsOrThrow` is a mistake; We should always return local time, not UTC. So I'd rather remove the `isUtc: true` to have all the date handled the same (and remove the `toLocal()` here).

lib/src/model/broadcast/broadcast_round_controller.dart
```diff
@@ -21,22 +21,42 @@ class BroadcastRoundController extends _$BroadcastRoundController {
...
       _subscription?.cancel();
+      _timer?.cancel();
     });
 
-    return await ref.withClient(
-      (client) => BroadcastRepository(client).getRound(broadcastRoundId),
+    final games =
+        await ref.watch(broadcastRoundProvider(broadcastRoundId).future);
+
+    _timer = Timer.periodic(
```
veloce, line 44: I don't understand why you chose to implement the clocks like that, instead of relying on the `CountdownClock` widget.

Here, I guess it works, but:
- it is not precise (second precision)
- it is not efficient: you rebuild the whole screen and all the games each second; if you have a local state instead, only the clock widget is rebuilt

lib/src/utils/current_locale.dart
```diff
@@ -0,0 +1,20 @@
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+import 'package:intl/intl.dart';
+import 'package:lichess_mobile/src/model/settings/general_preferences.dart';
+import 'package:riverpod_annotation/riverpod_annotation.dart';
+
+part 'current_locale.g.dart';
+
+@riverpod
+String currentLocale(CurrentLocaleRef ref) {
```
veloce, line 9: I think this is not needed and we should not handle it this way. I presume you made it to pass the locale to the date formatters. But these are designed to use the default locale, and we just need to correctly set the default locale elsewhere.

I know currently this is broken in `main` branch, but I'm taking care of it in the challenges PR.

Anyway I'd prefer you remove that logic from this PR so we can fix it globally in another PR.

lib/src/view/analysis/tree_view.dart
```diff
@@ -151,23 +158,19 @@ class _InlineTreeViewState extends ConsumerState<AnalysisTreeView> {
       );
     }
 
-    return CustomScrollView(
-      slivers: [
+    return ListView(
```
veloce, line 161: Why this change?

veloce (commented):
Partial review for now; I stopped at `broadcast_analysis_view.dart for now.

Great job so far!

lib/src/model/broadcast/broadcast_repository.dart
julien4215: The server returns UTC so we need to keep `isUtc: true`, right ?

However I can move the `toLocal()` in the pick extension

lib/src/model/broadcast/broadcast_round_controller.dart
julien4215: We don't need more than second precision for broadcast and I need to share the clock state between two screens (broadcast_screen.dart and broadcast_analysis_screen.dart) so I don't know how you would do it with local state.

lib/src/view/analysis/tree_view.dart
julien4215: To remove the blank space when there is no opening header

lib/src/utils/current_locale.dart
julien4215: Since the users can now change the language in settings shouldn't we format dates accordingly ?

lib/src/model/broadcast/broadcast_round_controller.dart
veloce: It is not obligatory to share the state. Even more, this state shouldn't be part of the model (of riverpod), but it is best to keep it local. It know it is debatable, but to me it falls in the [ephemeral state](https://riverpod.dev/docs/essentials/do_dont#avoid-using-providers-for-ephemeral-state) category, especially since the clocks are read only.

You need to distinguish the timer state (when the clock is running down) from the clock state (the real clock time which is held by the server).

julien4215: Let's say the user is on the broadcast screen and tap on a board, that brings it to the broadcast analysis screen and then he goes back to the broadcast screen. All I have is the time a player had when the server sent the clock because when the user moved to the broadcast analysis screen I lost the timer state. So I would need to make another request to the server just to get the time of the clock.

lib/src/utils/current_locale.dart
veloce: Yes but as explained it should work without passing an argument to the formatter. And we should not deal with that in this PR.

lib/src/model/broadcast/broadcast_round_controller.dart
veloce: I still think it is preferable to make this extra request versus having a shared state between screen and a potentially costly build of a screen that can contain a list with many board widgets (think of cheap phones).

lib/src/utils/current_locale.dart
julien4215: Yes I'll remove it from this PR but just to let you know, if I remember correctly I chose to do it like this because if you just change the default locale it doesn't apply to already loaded pages. I remember that I tried both solutions and I ended it with this one but I understand that it doesn't add much for the problems it might cause.

lib/src/model/broadcast/broadcast_round_controller.dart
julien4215: I thought rebuilding a widget is cheap if it is not repainted. In that case each second the widget is rebuilded but only the clocks part get repainted. Making an extra request each time add a load on the server that could be avoided so I guess none of these solutions works great. 

Maybe I could create a different provider that stores the elapsed time since the request has been sent which is only watched in the clocks widgets ?

lib/src/utils/current_locale.dart
veloce: The solution, I think, is to set again the `Intl.defaultLocale`, I'm going to do it in the challenge PR.

julien4215: That's what I meant when I said > change the default locale

lib/src/model/broadcast/broadcast_preferences.dart
julien4215: done [[f6666b9](https://github.com/lichess-org/mobile/pull/944/commits/f6666b9d194090173c99dd256910b9db851a4dec)

lib/src/model/analysis/analysis_controller.dart
julien4215: I believe I need all the logic of the analysis controller except maybe the method to request a server analysis (I don't know if there are people who can ask a server analysis for a broadcast game or if it is automatic). Am I missing some other parts that I do not need for the broadcast analysis controller ?

veloce: Right, I think you need almost everything. I just don't want to add methods and logic belonging to broadcast (or any other feature) to the analysis controller. If we continue to do that for all features (study, explorer, broadcast, etc.), it can become a mess I fear.
Also we shouldn't use `standaloneBroadcastId`. A broadcast is never standalone, but I understand why you chose this approach, it was the simplest way to integrate with the analysis controller.

So you can leave it like that for now, let's focus on getting the broadcast feature out. I'll figure later how to refactor this.

lib/src/model/broadcast/broadcast_repository.dart
veloce: The server returns a timestamp representing the number of milliseconds since epoch, which is always using UTC as a reference, that is true.

But the [dart constructor](https://api.dart.dev/stable/3.5.3/dart-core/DateTime/DateTime.fromMillisecondsSinceEpoch.html) `isUtc: true` parameter is a configuration to decide whether we want a `DateTime` in local time or in UTC. We always want to display the time in local time that's why the `isUtc` has to be false, which is the default.

lib/src/model/broadcast/broadcast_round_controller.dart
veloce: Testing right now the web version, each time we are changing the board focus, this xhr request is made: 

https://lichess.org/study/FOEszqAH/KBwrgngN

The response contains all the data you need to update the clock as far as I can tell. So if the browser does it, we can do also this extra request. This solution is by far the one which adds the less complexity and I really think we should keep things simple.

veloce:
I just saw your replies to my comments now @julien4215 . I've responded to all of them, sorry for the late reply.

lib/src/model/broadcast/broadcast_repository.dart
julien4215: Oh right, I'll remove the `isUtc: true`

lib/src/model/broadcast/broadcast_round_controller.dart
julien4215: The extra request I would need to make is https://lichess.org/api/broadcast/{broadcastTournamentSlug}/{broadcastRoundSlug}/{broadcastRoundId} which can be quite heavy if there are a lot of boards.

veloce:
```
{"study":{"id":"hHsVn1AO","name":"Round 5","members":{},"position":{"chapterId":"QIJJTYSZ","path":""},"ownerId":"linuxbrickie","settings":{"explorer":"everyone","description":false,"computer":"everyone","chat":"everyone","sticky":false,"shareable":"everyone","cloneable":"everyone"},"visibility":"public","createdAt":1726870475851,"secondsSinceUpdate":13,"from":{"relay":null},"likes":1,"liked":false,"features":{"cloneable":true,"shareable":true,"chat":true},"topics":["Broadcast"],"chapters":[{"id":"QIJJTYSZ","name":"Zhuang, Kyle - Khanna, Nalin","fen":"2r5/pp2bk2/4pp2/3p3p/1n1P1B1P/1P6/P3RPP1/4N1K1 w - - 4 27","players":[{"name":"Zhuang, Kyle","rating":1987,"fideId":30957486,"fed":"USA","clock":235800},{"name":"Khanna, Nalin","rating":2113,"fideId":30919550,"fed":"USA","clock":179900}],"lastMove":"c6c8","thinkTime":330,"status":"*"},{"id":"tkYyyGnL","name":"Badiee, Barzin - Diao, Matthew Guo","fen":"r1r3k1/1p3ppp/3q1n2/nP1p2N1/N2Pp1bP/PQ2P3/5PP1/R3KB1R w KQ - 1 17","players":[{"name":"Badiee, Barzin","rating":1860,"fideId":30962374,"fed":"USA","clock":213300},{"name":"Diao, Matthew Guo","rating":2097,"fideId":30970547,"fed":"USA","clock":118400}],"lastMove":"c6a5","thinkTime":68,"status":"*"},{"id":"OdX0sDuk","name":"ADU, Oladapo - Yang, Daniel","fen":"2r3k1/3q1pb1/p5pp/1p1np3/1P1n4/P2PN1PP/1B3PB1/2Q1R1K1 w - - 1 28","players":[{"name":"ADU, Oladapo","title":"IM","rating":2024,"fideId":8500258,"fed":"NGR","clock":298000},{"name":"Yang, Daniel","rating":2091,"fideId":30976340,"fed":"USA","clock":95800}],"lastMove":"e8c8","thinkTime":55,"status":"*"},{"id":"i0umHUIG","name":"Samuelson, Andrew - Xiong, Lang","fen":"1Q6/K2k2N1/8/8/8/8/8/8 b - - 0 78","players":[{"name":"Samuelson, Andrew","title":"FM","rating":2051,"fideId":2020580,"fed":"USA","clock":528800},{"name":"Xiong, Lang","rating":2000,"fideId":30957435,"fed":"USA","clock":341600}],"lastMove":"b7b8q","status":"1-0"},{"id":"QG0I72q3","name":"Ismayilova, Khanim - Baron, Tal","fen":"1n1r2k1/p2b1pbp/3Pp2p/1B2P3/3p4/5N2/q2N1PPP/1R2Q1K1 w - - 0 21","players":[{"name":"Ismayilova, Khanim","title":"WCM","rating":1960,"fideId":13429019,"fed":"AZE","clock":149000},{"name":"Baron, Tal","title":"GM","rating":2459,"fideId":2809958,"fed":"ISR","clock":245500}],"lastMove":"c5d4","thinkTime":468,"status":"*"},{"id":"FGsWkzuW","name":"Swaminathan, Pranav - Hebbar, Eshaan","fen":"2r1r3/pb3pk1/3Q1bp1/1P6/1Pp5/5N1P/P4PP1/2R3K1 w - - 0 24","players":[{"name":"Swaminathan, Pranav","rating":2056,"fideId":30970482,"fed":"USA","clock":221500},{"name":"Hebbar, Eshaan","title":"CM","rating":2208,"fideId":30974690,"fed":"USA","clock":235800}],"lastMove":"c5c4","thinkTime":914,"status":"*"},{"id":"qTNxh3Gg","name":"Li, Sophie - Feng, Andrew","fen":"1r4k1/6p1/4p2p/pq2Pp2/2pBnP2/1pP3P1/PPQ4P/3R2K1 w - - 0 35","players":[{"name":"Li, Sophie","title":"WCM","rating":1865,"fideId":30981107,"fed":"USA","clock":359300},{"name":"Feng, Andrew","rating":2091,"fideId":30943256,"fed":"USA","clock":76100}],"lastMove":"b4b3","thinkTime":55,"status":"*"},{"id":"GpSrbymH","name":"Brinkmann, Nicholas - Moorhouse, Will","fen":"1r6/4k2p/1pppb1p1/5p2/2P5/1B2P2P/P4KP1/3R4 b - - 2 29","players":[{"name":"Brinkmann, Nicholas","rating":1695,"fideId":39905314,"fed":"USA","clock":159900},{"name":"Moorhouse, Will","rating":2091,"fideId":30989531,"fed":"USA","clock":238800}],"lastMove":"f1f2","thinkTime":13,"status":"*"}],"chapter":{"id":"tkYyyGnL","ownerId":"linuxbrickie","setup":{"variant":{"key":"standard","name":"Standard"},"orientation":"white"},"tags":[["White","Badiee, Barzin"],["WhiteElo","1860"],["WhiteFideId","30962374"],["Black","Diao, Matthew Guo"],["BlackElo","2097"],["BlackFideId","30970547"],["Result","*"],["Round","5.2"]],"features":{"computer":true,"explorer":true},"relayPath":"-=aP$5WG)8G?8IUM.>VF=FMF2B\\M%@`N@N^N/7]A&;_b+3SK,<`];4KC<DC;5;MC"},"federations":{"USA":"United States of America","NGR":"Nigeria","ISR":"Israel","AZE":"Azerbaijan"},"chat":{"lines":[{"u":"dragonsfire66","t":"good luck to all","p":true,"f":"nature.crocodile"},{"u":"Archovie16","t":"board 4 glitch?"},{"u":"learnedMachine","t":"yep should update in a bit"}],"writeable":true}},"analysis":{"game":{"id":"synthetic","variant":{"key":"standard","name":"Standard","short":"Std"},"opening":null,"fen":"rnbqkbnr/pppppppp/8/8/8/8/PPPPPPPP/RNBQKBNR w KQkq - 0 1","turns":0,"player":"white","status":{"id":10,"name":"created"},"initialFen":"rnbqkbnr/pppppppp/8/8/8/8/PPPPPPPP/RNBQKBNR w KQkq - 0 1"},"player":{"id":null,"color":"white"},"opponent":{"color":"black","ai":null},"orientation":"white","pref":{"animationDuration":144,"coords":2,"moveEvent":2,"showCaptured":true,"keyboardMove":false,"rookCastle":true,"highlight":true,"destination":true},"userAnalysis":true,"treeParts":[{"ply":0,"fen":"rnbqkbnr/pppppppp/8/8/8/8/PPPPPPPP/RNBQKBNR w KQkq - 0 1","dests":"ksA owE bqs gvx jrz nvD muC ltB iqy pxF"},{"ply":1,"fen":"rnbqkbnr/pppppppp/8/8/2P5/8/PP1PPPPP/RNBQKBNR b KQkq - 0 1","id":"-=","uci":"c2c4","san":"c4","clock":545700},{"ply":2,"fen":"rnbqkb1r/pppppppp/5n2/8/2P5/8/PP1PPPPP/RNBQKBNR w KQkq - 1 2","id":"aP","uci":"g8f6","san":"Nf6","clock":500500},{"ply":3,"fen":"rnbqkb1r/pppppppp/5n2/8/2P5/2N5/PP1PPPPP/R1BQKBNR b KQkq - 2 2","id":"$5","uci":"b1c3","san":"Nc3","clock":547900},{"ply":4,"fen":"rnbqkb1r/pppp1ppp/5n2/4p3/2P5/2N5/PP1PPPPP/R1BQKBNR w KQkq - 0 3","id":"WG","uci":"e7e5","san":"e5","clock":502200},{"ply":5,"fen":"rnbqkb1r/pppp1ppp/5n2/4p3/2P5/2N2N2/PP1PPPPP/R1BQKB1R b KQkq - 1 3","id":")8","uci":"g1f3","san":"Nf3","clock":550000},{"ply":6,"fen":"rnbqkb1r/pppp1ppp/5n2/8/2P1p3/2N2N2/PP1PPPPP/R1BQKB1R w KQkq - 0 4","id":"G?","uci":"e5e4","san":"e4","clock":504800},{"ply":7,"fen":"rnbqkb1r/pppp1ppp/5n2/6N1/2P1p3/2N5/PP1PPPPP/R1BQKB1R b KQkq - 1 4","id":"8I","uci":"f3g5","san":"Ng5","clock":552100},{"ply":8,"fen":"rnbqkb1r/pp1p1ppp/2p2n2/6N1/2P1p3/2N5/PP1PPPPP/R1BQKB1R w KQkq - 0 5","id":"UM","uci":"c7c6","san":"c6","clock":507000},{"ply":9,"fen":"rnbqkb1r/pp1p1ppp/2p2n2/6N1/2PPp3/2N5/PP2PPPP/R1BQKB1R b KQkq d3 0 5","id":".>","uci":"d2d4","san":"d4","clock":524400},{"ply":10,"fen":"rnbqkb1r/pp3ppp/2p2n2/3p2N1/2PPp3/2N5/PP2PPPP/R1BQKB1R w KQkq - 0 6","id":"VF","uci":"d7d5","san":"d5","clock":428000},{"ply":11,"fen":"rnbqkb1r/pp3ppp/2p2n2/3P2N1/3Pp3/2N5/PP2PPPP/R1BQKB1R b KQkq - 0 6","id":"=F","uci":"c4d5","san":"cxd5","clock":496500},{"ply":12,"fen":"rnbqkb1r/pp3ppp/5n2/3p2N1/3Pp3/2N5/PP2PPPP/R1BQKB1R w KQkq - 0 7","id":"MF","uci":"c6d5","san":"cxd5","clock":429600},{"ply":13,"fen":"rnbqkb1r/pp3ppp/5n2/3p2N1/3Pp2P/2N5/PP2PPP1/R1BQKB1R b KQkq - 0 7","id":"2B","uci":"h2h4","san":"h4","clock":497900},{"ply":14,"fen":"r1bqkb1r/pp3ppp/2n2n2/3p2N1/3Pp2P/2N5/PP2PPP1/R1BQKB1R w KQkq - 1 8","id":"\\M","uci":"b8c6","san":"Nc6","clock":392500},{"ply":15,"fen":"r1bqkb1r/pp3ppp/2n2n2/3p2N1/3PpB1P/2N5/PP2PPP1/R2QKB1R b KQkq - 2 8","id":"%@","uci":"c1f4","san":"Bf4","clock":459000},{"ply":16,"fen":"r1bqk2r/pp3ppp/2nb1n2/3p2N1/3PpB1P/2N5/PP2PPP1/R2QKB1R w KQkq - 3 9","id":"`N","uci":"f8d6","san":"Bd6","clock":280600},{"ply":17,"fen":"r1bqk2r/pp3ppp/2nB1n2/3p2N1/3Pp2P/2N5/PP2PPP1/R2QKB1R b KQkq - 0 9","id":"@N","uci":"f4d6","san":"Bxd6","clock":445700},{"ply":18,"fen":"r1b1k2r/pp3ppp/2nq1n2/3p2N1/3Pp2P/2N5/PP2PPP1/R2QKB1R w KQkq - 0 10","id":"^N","uci":"d8d6","san":"Qxd6","clock":273500},{"ply":19,"fen":"r1b1k2r/pp3ppp/2nq1n2/3p2N1/3Pp2P/2N1P3/PP3PP1/R2QKB1R b KQkq - 0 10","id":"/7","uci":"e2e3","san":"e3","clock":444200},{"ply":20,"fen":"r3k2r/pp3ppp/2nq1n2/3p2N1/3Pp1bP/2N1P3/PP3PP1/R2QKB1R w KQkq - 1 11","id":"]A","uci":"c8g4","san":"Bg4","clock":245200},{"ply":21,"fen":"r3k2r/pp3ppp/2nq1n2/3p2N1/Q2Pp1bP/2N1P3/PP3PP1/R3KB1R b KQkq - 2 11","id":"&;","uci":"d1a4","san":"Qa4","clock":406700},{"ply":22,"fen":"r4rk1/pp3ppp/2nq1n2/3p2N1/Q2Pp1bP/2N1P3/PP3PP1/R3KB1R w KQ - 3 12","id":"_b","uci":"e8h8","san":"O-O","clock":237600},{"ply":23,"fen":"r4rk1/pp3ppp/2nq1n2/3p2N1/Q2Pp1bP/P1N1P3/1P3PP1/R3KB1R b KQ - 0 12","id":"+3","uci":"a2a3","san":"a3","clock":400400},{"ply":24,"fen":"r4rk1/1p3ppp/p1nq1n2/3p2N1/Q2Pp1bP/P1N1P3/1P3PP1/R3KB1R w KQ - 0 13","id":"SK","uci":"a7a6","san":"a6","clock":204400},{"ply":25,"fen":"r4rk1/1p3ppp/p1nq1n2/3p2N1/QP1Pp1bP/P1N1P3/5PP1/R3KB1R b KQ - 0 13","id":",<","uci":"b2b4","san":"b4","clock":346500},{"ply":26,"fen":"r1r3k1/1p3ppp/p1nq1n2/3p2N1/QP1Pp1bP/P1N1P3/5PP1/R3KB1R w KQ - 1 14","id":"`]","uci":"f8c8","san":"Rfc8","clock":142500},{"ply":27,"fen":"r1r3k1/1p3ppp/p1nq1n2/3p2N1/1P1Pp1bP/PQN1P3/5PP1/R3KB1R b KQ - 2 14","id":";4","uci":"a4b3","san":"Qb3","clock":321600},{"ply":28,"fen":"r1r3k1/1p3ppp/2nq1n2/p2p2N1/1P1Pp1bP/PQN1P3/5PP1/R3KB1R w KQ - 0 15","id":"KC","uci":"a6a5","san":"a5","clock":126000},{"ply":29,"fen":"r1r3k1/1p3ppp/2nq1n2/pP1p2N1/3Pp1bP/PQN1P3/5PP1/R3KB1R b KQ - 0 15","id":"<D","uci":"b4b5","san":"b5","clock":227200},{"ply":30,"fen":"r1r3k1/1p3ppp/2nq1n2/1P1p2N1/p2Pp1bP/PQN1P3/5PP1/R3KB1R w KQ - 0 16","id":"C;","uci":"a5a4","san":"a4","clock":117100},{"ply":31,"fen":"r1r3k1/1p3ppp/2nq1n2/1P1p2N1/N2Pp1bP/PQ2P3/5PP1/R3KB1R b KQ - 0 16","id":"5;","uci":"c3a4","san":"Nxa4","clock":213300},{"ply":32,"fen":"r1r3k1/1p3ppp/3q1n2/nP1p2N1/N2Pp1bP/PQ2P3/5PP1/R3KB1R w KQ - 1 17","id":"MC","uci":"c6a5","san":"Na5","clock":118400}]}}
```

Here is the JSON returned by the study chapter XHR from the browser (same request as mentioned above). It looks like the data returned is pretty similar to the one you mentioned: https://lichess.org/api#tag/Broadcasts/operation/broadcastRoundGet 

Instead of `games` you have `chapters`, but actually inside it is the same data and you have one chapter per board.

You're right to be caution about extra requests and server work. But again, in our case, if the browser does it, we can do it. Also this is a first version, there's room for improvement and optimisation later. Early optimisation... you know the saying.

lib/src/model/broadcast/broadcast_preferences.dart
```diff
@@ -0,0 +1,59 @@
...
+import 'package:lichess_mobile/src/model/common/id.dart';
+import 'package:riverpod_annotation/riverpod_annotation.dart';
+
+part 'broadcast_preferences.freezed.dart';
+part 'broadcast_preferences.g.dart';
+
+const _prefKey = 'broadcast.preferences';
+
+@riverpod
+class BroadcastPreferences extends _$BroadcastPreferences {
```
veloce, line 16: This should be refactored according to the new preferences model. See https://github.com/lichess-org/mobile/blob/8f0e3adce38374fe52180736243b2d135d74e839/lib/src/model/settings/preferences.dart#L15

veloce (commented):
Coming back on it @julien4215 

Have you found a way to handle the clocks?

I think the most obvious way to share state between 2 screen is to use a separate provider.

So you need a provider family that can instantiate the clock of each game. Something like:

```dart
@riverpod
Future<BroadcastGameClockState> gameClock(BroadcastRoundId roundId, BroadcastGameId gameId) {
   // get clock data from game controller
 
   // listen to socket to get clock update events

  // instantiate the local timer

  return clockData;
}

So you can use this provider in both screens.

**Comment by julien4215** [comment]
I didn't have time to think about it but by using [select](https://riverpod.dev/docs/concepts/reading#using-select-to-filter-rebuilds) I could avoid rebuilding when it is not needed or splitting the provider as you said would also work. I'll try to work on it next week.

**Comment by veloce** [comment]
> I didn't have time to think about it but by using [select](https://riverpod.dev/docs/concepts/reading#using-select-to-filter-rebuilds) I could avoid rebuilding when it is not needed or by splitting the provider as you said would also work. I'll try to work on it next week.

I don't think `select` is a good idea in that case, because notifying a change every second in the broadcast list controller is not a good thing first.

**Comment by julien4215** [comment]
> Coming back on it @julien4215
> 
> Have you found a way to handle the clocks?
> 
> I think the most obvious way to share state between 2 screen is to use a separate provider.
> 
> So you need a provider family that can instantiate the clock of each game. Something like:
> 
> ```dart
> @riverpod
> Future<BroadcastGameClockState> gameClock(BroadcastRoundId roundId, BroadcastGameId gameId) {
>    // get clock data from game controller
>  
>    // listen to socket to get clock update events
> 
>   // instantiate the local timer
> 
>   return clockData;
> }
> ```
> 
> Then you can use this provider in both screens.

Is it a good idea to create as many timers as there are games ? I only need one timer to get clocks of each player.

**Comment by veloce** [comment]
I know what you mean: the elapsed seconds are the same for everyone. 

But it is not a problem to create one timer per game if it allows us to solve this problem we're having. In the real tournament, there is also one clock per game. We're just doing the same here.

I haven't looked deeply at the broadcast server side and API yet, but I suppose you'd receive a socket event for a clock update for one game at a time? It makes sense to have separate timers for each clock for that reason too.

Currently in TV and game screens, each clock you see has its own timer as well.

**Comment by julien4215** [comment]
> I know what you mean: the elapsed seconds are the same for everyone.
> 
> But it is not a problem to create one timer per game if it allows us to solve this issue we're having. In the real tournament, there is also one clock per game. We're just doing the same here.
> 
> I haven't looked deeply at the broadcast server side and API yet, but I suppose you'd receive a socket event for a clock update for one game at a time? It makes sense to have separate timers for each clock for that reason too.
> 
> Currently in TV and game screens, each clock you see has its own timer as well.

Yes each time a move is played in a game I receive from the websocket the new clock of the player that made the move.

**Comment by tom-anders** [comment]
@julien4215 just as a heads up, once https://github.com/lichess-org/mobile/pull/1076 lands you'll have to merge again (the PR simply moves the tree to a separate file, so you'll need to reapply all your changes in that file)

**Comment by julien4215** [comment]
> @julien4215 just as a heads up, once #1076 lands you'll have to merge again (the PR simply moves the tree to a separate file, so you'll need to reapply all your changes in that file)

Okay thanks for letting me know

File: lib/src/view/broadcast/broadcast_analysis_screen.dart [hunk 8]
```diff
@@ -0,0 +1,408 @@
...
+            boardSide: _PlayerWidgetSide.bottom,
+            playingSide: playingSide,
+            playClock: currentPath == broadcastLivePath,
+          ),
+      ],
+    );
+  }
+
+  Widget _androidBuilder(BuildContext context, WidgetRef ref) {
+    final pgn = ref.watch(
```
veloce, line 95: Here again this is rather confusing. We're seeing a ref.watch, but after investigation I realise the pgn is never changing. It's only here to feed the analysis controller.

This is indicating something is wrong in the architecture. I told you in our last conversation that it was ok to keep the analysis controller and we'd refactor later, but I changed my mind. If there wasn't the existing analysis controller, I'm pretty sure you'd have done it differently, because without the prior implementation it would have made much more sense to do like what I suggest below:

A `BroadcastGameController` with this signature:
```dart
@riverpod
class BroadcastGameController extends $BroadcastGameController {
   Future<BroadcastGameState> build(BroadcastGameId id) {
      ....
      // here you get and watch the game through http / websocket
   }
}
```

In the broadcast screen you only watch this controller to get changes (as opposed to 3 different ones right now).
The live clock in the `broadcast_analysis_screen` must be implemented with a `CountdownClock` widget (appearance may need adjustment). It should work like it works right now in the TV screen.

It is ok to copy/paste all the logic from analysis controller for now (I'll work on mixins later for shared logic). It should look like the new `StudyController` which has a similar signature.

lib/src/model/broadcast/broadcast_game_controller.dart
```diff
@@ -0,0 +1,112 @@
...
+class BroadcastGameController extends _$BroadcastGameController {
+  static Uri broadcastSocketUri(BroadcastRoundId broadcastRoundId) =>
+      Uri(path: 'study/$broadcastRoundId/socket/v6');
+
+  StreamSubscription<SocketEvent>? _subscription;
+
+  late SocketClient _socketClient;
+
+  @override
+  Future<String> build(BroadcastRoundId roundId, BroadcastGameId gameId) async {
```
veloce, line 25: The signature of this controller indicates something is wrong. 

It is returning a PGN that never changes. A riverpod notifier that does not update its state should not exist really. See the comment below where I explain in more details what should be done instead.

 Under the hood it is imperatively changing the analysis controller state. That is another pattern that we should avoid with riverpod. This is a reactive framework so instead of:

```
Provider A
----------
subs to an event source
on each event --> modify Provider B
```

we should do:
```
Provider A
----------
subs to an event source
on each event: --> update and notifies its state change

Provider B
----------
subs to provider A changes
on each change: do whatever needed
```

A provider modifying another one imperatively is almost never what we want. I've done it myself a couple of times too and realised later it was a mistake. It is possible there is still a couple of bad patterns like that in the app code but it should be fixed really.

veloce (commented):
There's a lot in this review. If you have any question don't hesitate to ping me.

lib/src/model/broadcast/broadcast_round_controller.dart
```diff
@@ -22,22 +22,42 @@ class BroadcastRoundController extends _$BroadcastRoundController {
...
+
+    return games;
+  }
+
+  void _updateClocks() {
+    state = AsyncData(
+      state.requireValue.map((gameId, game) {
+        if (!game.isPlaying) return MapEntry(gameId, game);
+        final thinkTime = game.thinkTime;
+        final newThinkTime = thinkTime + const Duration(seconds: 1);
```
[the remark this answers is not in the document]
julien4215: Why is `Timer.periodic not precise ?

The doc says
> No more than n callbacks will be made in `duration * n` time, but the time between two consecutive callbacks can be shorter and longer than duration.

I understand it as it is accurate at the second which is what is needed for broadcast.

lib/src/model/broadcast/broadcast.dart
```diff
@@ -45,25 +78,37 @@ class BroadcastRound with _$BroadcastRound {
...
+  const BroadcastGame._();
 
-  const factory BroadcastGameSnapshot({
+  const factory BroadcastGame({
+    required BroadcastGameId id,
     required IMap<Side, BroadcastPlayer> players,
     required String fen,
     required Move? lastMove,
-    required String status,
+    required String? status,
```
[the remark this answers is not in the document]
julien4215: This is the backend type `Option[Option[Outcome]]` with the following comment 
```
/* None = No Result PGN tag, the chapter may not be a game
* Some(None) = Result PGN tag is "*", the game is ongoing
* Some(Some(Outcome)) = Game is over with a result
*/
```
Not sure what that should be in dart. Maybe I can extend the Dart class Outcome ?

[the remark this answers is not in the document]
julien4215: I went with creating an enum specific to the broadcast result

lib/src/view/broadcast/broadcast_analysis_screen.dart
julien4215: The pgn is changing. On each move received from the websocket I edit the pgn to add the new move.

lib/src/view/broadcast/broadcast_analysis_screen.dart
```diff
@@ -0,0 +1,408 @@
...
+            icon: const Icon(Icons.settings),
+          ),
+        ],
+      ),
+      body: pgn.when(
+        data: (pgn) => AnalysisScreenBody(
+          pgn: pgn,
+          options: AnalysisState.broadcastOptions,
+          enableDrawingShapes: true,
+          broadcastWrapBoardBuilder: broadcastWrapBoardBuilder,
```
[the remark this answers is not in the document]
julien4215: What I intend to do with this parameter is to share the same code for the layout of the body (which includes the logic to adapt to different screen sizes). I would need to duplicate that code if I make two separate bodies.

lib/src/model/broadcast/broadcast_round_controller.dart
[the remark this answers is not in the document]
veloce: The time between 2 consecutive callbacks is not precisely one second as explained:
>  the time between two consecutive callbacks can be shorter and longer than duration

So if you remove a second each time when not necessarily a second has elapsed, it is not really precise. That's why we never implement clocks like that.

lib/src/view/broadcast/broadcast_analysis_screen.dart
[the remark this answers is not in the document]
veloce: Unless you write an `AnalysisScreenLayout` widget, which is another way to abstract it (but which I prefer, as it's more idiomatic in flutter).

lib/src/model/broadcast/broadcast_round_controller.dart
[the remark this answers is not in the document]
julien4215: I'd like the doc to state that more than n-1 callbacks will be made in duration * n time but that's obviously not the case. With only what they say in the doc I agree that there is zero guarantee on the accuracy.

I'll try to use the `Stopwatch` but it might be a bit more complicated because it has to be shared with all boards with the current structure. I need to think more about it.

lib/src/model/broadcast/broadcast_game_controller.dart
julien4215: I didn't even realize that I was not updating the state of this controller. I think this is due the way I was reasoning. My goal was to update the analysis controller from moves received by a broadcast like it is for moves received by user inputs.

I am not sure about changing the signature to `Future<BroadcastAnalysisState>`. If it is a future it will probably make it harder to implement a mixin for both controller (`BroadcastGameController` and `AnalysisController`)

tom-anders: The `StudyController` from https://github.com/lichess-org/mobile/pull/1111 already returns Future<StudyState> as well, so you won't make refactoring harder

lib/src/model/broadcast/broadcast_round_controller.dart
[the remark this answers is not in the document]
veloce: That is less important than the other comments I made (on the code architecture). So you can ignore it for now.

lib/src/model/broadcast/broadcast_game_controller.dart
veloce: Indeed `StudyController` and `BroadcastController` will be very similar. We'll be able to refactor all this step by step. 

Starting with a mixin for engine evaluation I think, as it looks straightforward.

julien4215: I created a controller that is separated of the analysis controller but I still believe that is not the right way to do it. Now the code is full of `if (!state.hasValue) return;` and `state.requireValue` whereas the state will never be a future expect during the loading of the pgn which is only when a user taps on a game. 

Also the factoring of `BroadcastController` methods and `AnalysisController` methods will be harder since the methods are not the same anymore because of the check for a value in the state (`if (!state.hasValue) return;` and `state.requireValue` )

veloce: Yes I know it's annoying, having to deal with `AsyncValue`. But actually there is no other way, is it?

How do you deal with the absence of value (while it is loading) otherwise?

We'll have to make the `AnalysisController` async too.

In order to make this refactoring, we'd have a `GameAnalysisController` which would returns a Future<GameAnalysisState> and takes a game id + options. And a `StandaloneAnalysisController` which takes a PGN string (but has to be async too to be compatible with the common interface).

I don't have a better idea for now. What do you think @tom-anders ? Since you've worked on studies you might have an idea on how to deal with this?

tom-anders: So for studies, `AsyncValue` actually feels really natural - For broadcast, the `state` is `loading` only at the very start, but for studies it's actually also `loading` when switching between chapters.

I already did some experimentation locally to factor out common code between `StudyController` and `AnalysisController`, and while it does somehow work, the fact that one had `State` while the other had `AsyncValue<State>` was really annoying, meant I couldn't reuse as much code as I wanted to, and also made readability worse in some places. 

I get that making `AnalysisController` async (even though it technically does not need to be) feels weird, but I think the benefits of improving code reusability would outweigh the benefits here. I already caused a bug in `StudyController` because I incorrectly ported a change I made in the `AnalysisController` on another branch. And yeah, there will be a lot of `state.requireValue` (the `StudyController` has them as well), but at least they will be in a single place.

veloce: So the plan is to merge study and broadcast branches as-is (with separated code). Then we'll refactor study, broadcast and analysis together, making everything async, and write a lot of tests while doing it. cc @julien4215

tom-anders: Sounds good to me, let me know if you need help :)

julien4215: What I could do to make the `broadcastAnalysisController` not async is giving it directly the pgn instead of the broadcastGameId like it is done for the `AnalysisController`. I would get the pgn in a loading screen and then give it to broadcast game controller. Would it be hard to do the same for study ?

tom-anders: Not sure if that would work:

1) For studies there's the additional complexity of switching between chapters
2) I think the major problem is that for studies, the PGN does not actually contain all the information that the `StudyController` needs (for example the hints in interactive studies, or whether analysis is allowed or not, etc..), so it would also need to be passed the `Study` object - but IIRC passing such heavy weight objects to a provider family is considered an antipattern in Riverpod

tom-anders: On the other hand, if we're making the `AnalysisController` async anyway, then we **could** actually give it the `id` instead of `pgn` and let it internally fetch the game itself from the server — Then it also wouldn't feel as pointless to slap a `Future` around its state

julien4215: Is it possible to do that ? Maybe there are use cases where we can't use an `id` but only a `pgn`.

tom-anders: Hmm, the `AnalysisScreen` already uses a `String pgnOrId` field so it should be possible - But separating `GameAnalysisController` and `StandaloneAnalysisController` as veloce suggested makes sense as well  I  guess

veloce: Yes we need to make everything async. And so a GameAnalysisController that takes no PGN but the GameId makes more sense.

julien4215: So the plan is to make two separate base class : a `GameAnalysisController` and a `StandaloneAnalysisController` ?

The `BroadcastGameController` requires two ids : the round id is to listen to the websocket and the game id to get the pgn.

julien4215:
I am having an issue with the broadcast game screen. Because I am using a different controller I can't refactor the analysis screes since most widgets of the analysis screen takes the controller as an argument.

I am seeing three options:
- duplicate the analysis screen code
- make all widgets independent of the analysis controller
- make the common interface for analysis controller now

Which one should I go for ?

veloce:
> I am having an issue with the broadcast game screen. Because I am using a different controller I can't refactor the analysis screes since most widgets of the analysis screen takes the controller as an argument.
> 
> I am seeing three options:
> 
> * duplicate the analysis screen code
> * make all widgets independent of the analysis controller
> * make the common interface for analysis controller now
> 
> Which one should I go for ?

@tom-anders has really done it for study screen. So you should do the same. In this PR I should see no modification of the analysis screen nor the analysis controller.

Then you won't copy all the analysis screen, since we don't need everything for this first broadcast version. So for now I would expect the `broadcast_screen.dart` to contain more or less the same number of lines as the study screen file. That is I should not see code related to "server analysis summary", "acpl chart". We only want to show the moves for now.

That way, it will be easier for me to review, and then to refactor.

julien4215:
So we don't want to include local analysis right now but in a future PR ?

veloce:
You can include local analysis, but not server side analysis (game summary, acpl chart, etc).

Study screen PR is now merged, and it includes local analysis too.

veloce:
Let's remove the opening header and explorer for now.

lib/src/model/broadcast/broadcast_round_controller.dart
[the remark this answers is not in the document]
julien4215: I think that I fixed the code architecture now. What should I do with the clocks ?

I merged main and tried to use the new `CountdownClockBuilder` widget but I still have the issue of sharing clocks between two screens.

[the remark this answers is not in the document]
veloce: What issue? I think you can maintain a `Map<BroadcastGameId, DateTime>` in the controller, and then pass the `clockUpdateAt` date time to `CountdownClockBuilder`.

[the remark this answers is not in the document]
julien4215: Oh right I didn't think of using the `updetedAt` field like this. Thanks.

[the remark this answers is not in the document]
julien4215: I had to change the `didUpdateWidget` method of the `CountdownClockBuilder` because it wasn't changing the `timeLeft` when switching between inactive clocks which is the case when going through moves on the broacast game screen.

julien4215:
@veloce I have addressed all of your comments.

There are still some annoying bugs that I need to fix but the code should not change much.

julien4215:
I fixed all the bugs that I had noticed

veloce:
So it is ready to review? Awesome. There's just a couple of conflicts to solve and I'll review this.

julien4215:
Yes it is ready to review.

I fixed the conflicts and I modified the code to use the new `AnalysisLayout` widget in the broadcast game screen.

lib/src/view/broadcast/broadcast_boards_tab.dart
```diff
@@ -0,0 +1,296 @@
...
+            footer: _PlayerWidgetLoading(width: boardWithMaybeEvalBarWidth),
+            showEvaluationBar: showEvaluationBar,
+          );
+        }
+
+        final game = games![index];
+        final playingSide = Setup.parseFen(game.fen).turn;
+
+        return BoardThumbnail(
+          onTap: () {
```
veloce, line 139: Here I think you can pass an `animationDuration` to the `BoardThumbnail` to animate position change.

lib/src/view/broadcast/broadcast_game_screen.dart
```diff
@@ -0,0 +1,509 @@
...
+        broadcastGameControllerProvider(widget.roundId, widget.gameId);
+    final broadcastAnalysisState = ref.watch(ctrlProvider).requireValue;
+    final boardPrefs = ref.watch(boardPreferencesProvider);
+    final showBestMoveArrow = ref.watch(
+      analysisPreferencesProvider.select(
+        (value) => value.showBestMoveArrow,
+      ),
+    );
+    final showAnnotationsOnBoard = ref.watch(
+      analysisPreferencesProvider.select((value) => value.showAnnotations),
```
veloce, line 219: Instead of using 2 `.select` prefer, watching directly the whole analysis preferences state.

lib/src/view/broadcast/broadcast_game_tree_view.dart
```diff
@@ -0,0 +1,50 @@
...
+  );
+
+  final BroadcastRoundId roundId;
+  final BroadcastGameId gameId;
+
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {
+    final ctrlProvider = broadcastGameControllerProvider(roundId, gameId);
+
+    final root =
```
veloce, line 23: Here you can also watch the whole state at once, instead of using several `.select`

lib/src/view/broadcast/broadcast_screen.dart
```diff
@@ -0,0 +1,486 @@
...
+
+  @override
+  Widget build(BuildContext context) {
+    return Scaffold(
+      appBar: AppBar(
+        title: Text(widget.broadcast.title),
+        bottom: TabBar(
+          controller: _tabController,
+          tabs: const <Widget>[
+            Tab(text: 'Overview'),
```
veloce, line 88: Translations are available here.

lib/src/widgets/evaluation_bar.dart
```diff
@@ -0,0 +1,49 @@
+import 'package:flutter/material.dart';
+
+const evaluationBarAspectRatio = 1 / 20;
+
+class EvaluationBar extends StatelessWidget {
```
veloce, line 5: A doc comment would be welcome here. What it is? How is it different from the existing `EvaluationGauge`?

lib/src/widgets/pgn.dart
```diff
@@ -177,14 +185,20 @@ class _DebouncedPgnTreeViewState extends ConsumerState<DebouncedPgnTreeView> {
...
-          pathToCurrentMove = widget.currentPath;
+          if (oldWidget.currentPath != widget.currentPath) {
+            pathToCurrentMove = widget.currentPath;
+          }
+          if (oldWidget.broadcastLivePath != widget.broadcastLivePath) {
+            pathToBroadcastLiveMove = widget.broadcastLivePath;
+          }
         });
-        WidgetsBinding.instance.addPostFrameCallback((_) {
+        if (oldWidget.currentPath != widget.currentPath) {
```
veloce, line 201: Should should not remove the postFrameCallback here. This change looks like a mistake.

lib/src/widgets/clock.dart
```diff
@@ -287,7 +287,8 @@ class _CountdownClockState extends State<CountdownClockBuilder> {
   void didUpdateWidget(CountdownClockBuilder oldClock) {
     super.didUpdateWidget(oldClock);
 
-    if (widget.clockUpdatedAt != oldClock.clockUpdatedAt) {
+    if (widget.timeLeft != oldClock.timeLeft ||
```
veloce, line 290: I don't think that change is a good idea. The way to clock is designed is to rely only on `clockUpdatedAt` to know when the clock has changed. We want to force update `clockUpdateAt`, and here you'd broke that contract.

test/widgets/clock_test.dart
```diff
@@ -115,7 +115,7 @@ void main() {
       expect(find.text('0:00.0'), findsOneWidget);
     });
 
-    testWidgets('do not update if clockUpdatedAt is same',
+    testWidgets('do not update if timeLeft and clockUpdatedAt are same',
```
veloce, line 118: Cf. other comment, this change need to be removed.

veloce (commented):
Excellent work! I think we're getting close to release this. Only have a couple of comments here.

I'll test this next week, and figure out something about the dropdown menu bug, since I don't really want to patch flutter.

lib/src/widgets/pgn.dart
julien4215: Completely a mistake. Good catch

lib/src/view/broadcast/broadcast_game_screen.dart
julien4215: Why do you think it is better to watch the whole analysis preferences state ? I'd like to understand so I don't do the same mistakes again.

From what I understand, by listening to the whole state it will rebuild the board even if an analysis preference shouldn't change the board.

lib/src/widgets/evaluation_bar.dart
julien4215: It is for board thumbnail like there is in the website at the right of the board.

20241123_16h06m38s_grim

I don't know if I could reuse the `EvaluationGauge` for that instead.

Anyway I'll remove it from this PR as I should have it done it before.

lib/src/view/analysis/analysis_board.dart
```diff
@@ -17,14 +17,14 @@ class AnalysisBoard extends ConsumerStatefulWidget {
...
     this.boardSize, {
-    this.borderRadius,
+    this.radius,
     this.enableDrawingShapes = true,
     this.shouldReplaceChildOnUserMove = false,
   });
 
   final AnalysisOptions options;
   final double boardSize;
-  final BorderRadiusGeometry? borderRadius;
```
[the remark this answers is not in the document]
julien4215: Are you talking only about the name of the variable or the type ?

lib/src/view/broadcast/broadcast_game_screen.dart
veloce: Because the code is less readable (several watch instead of one) and the "optimisation" is not worth it. Analysis prefs will rarely change and it is ok to rebuild even if one of the change is not related to this screen.

lib/src/view/analysis/analysis_board.dart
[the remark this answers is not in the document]
veloce: both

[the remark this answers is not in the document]
julien4215: To add rounded corners to the players widgets instead of the board for broadcast like on the website

image

The change of the name is just to match the type

lib/src/widgets/clock.dart
julien4215: Not sure to understand your comment. Do you mean that the only to change the clock should be the `clockUpdateAt` field ?

julien4215:
> I'll test this next week, and figure out something about the dropdown menu bug, since I don't really want to patch flutter.

Flutter fixed the issue on the master branch but I don't know when it will land in the beta.

If you don't want to patch Flutter maybe I can just copy the fixed dropdown menu widget in the repo for now and we remove it once it lands on the beta channel ?

lib/src/widgets/clock.dart
veloce: The only way to indicate the clock has changed is to update `clockUpdatedAt`. Meaning if you want to update `timeLeft` you *have* to update `clockUpdatedAt`. We want to support updating the clock with the same time left, this is the only way.
If you don't provide at all the `clockUpdatedAt` it means `timeLeft` will never change.

veloce: You can read the tests in `clock_test.dart` to understand also, it contains the specification.

veloce:
> > I'll test this next week, and figure out something about the dropdown menu bug, since I don't really want to patch flutter.
> 
> Flutter fixed the issue on the master branch but I don't know when it will land in the beta.
> 
> If you don't want to patch Flutter maybe I can just copy the fixed dropdown menu widget in the broadcast folder for now and we remove it once it lands on the beta channel ?

Not sure about that. The problem with the dropdown menu is that we can't show more detailed info it seems. I think it would be very nice to show the round date as in the website, it is an important information. And it looks like we can't do it with `DropdownMenu`.

I wonder what is the alternative? Maybe `showMenu` and `PopupMenuItem`? https://api.flutter.dev/flutter/material/PopupMenuItem-class.html

julien4215:
The dropdown menu entries can be larger than the button so that should be possible to show more detailed information.

Maybe that can be done in a different pull request after merging this one ? This one is already bigger than what I planned to do.

tom-anders:
> The dropdown menu entries can be larger than the button so that should be possible to show more detailed information.
> 
> Maybe it can be done in a different pull request after merging this one ? This one is already bigger than what I planned to do.

Plus, people would probably very excited if this feature is released before the WCC ends, even if some features are missing

veloce:
You're right we'll see to improve the menu after. So you can put a copy of the widget in the meantime.

lib/src/widgets/clock.dart
julien4215: I don't understand why you are saying that `We want to support updating the clock with the same time left, so this is the only way to ensure it.`

With what I have done you can still update the clock with the same time left.

julien4215: The reason I needed to also update the clock when `timeLeft` change (which makes sense I think) is that on the broadcast game screen when going through the moves I need to display the clocks and the clocks are always inactive with no `clockUpdatedAt` expect for the current broadcast path if the game is still ongoing.

julien4215: What I could do instead is simply use a text widget but then why add an `inactive` field.

julien4215:
> So you can put a copy of the widget in the meantime.

Done. I also merged main to get the new search time setting and the new ui for settings.

veloce:
🎉 finally!

I must thank you again @julien4215 , this was a long journey but we finally made it! Thanks to your dedication and hard work. Congrats!

julien4215:
Happy to see it merged into main! And thanks for you thorough review.

I'll try to make smaller PR in the future. This one got a bit out of hand.

PR #949, opened 2024-08-30
Use badge widget for icons

julien4215:
https://api.flutter.dev/flutter/material/Badge/Badge.html

### DIFF
--- a/lib/src/view/game/game_body.dart
+++ b/lib/src/view/game/game_body.dart
@@ -1,5 +1,4 @@
 import 'dart:async';
-import 'dart:math' as math;
 
 import 'package:chessground/chessground.dart';
 import 'package:collection/collection.dart';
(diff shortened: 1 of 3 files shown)


veloce:
Hi Julien, thanks for this! Can you share a screenshot to show before/after? Thanks!

julien4215:
Before :

After :

veloce:
Nice catch! I didn't know this widget existed.

We need the same color as before I'd say, at least for the chat. The red is too agressive here.

julien4215:
I put the same color as before for the bottom bar button. The font was also heavier before, I don't know if you want to keep it like that.

julien4215:
I changed the style again. Now this is the same as before for both badges. Same font and same color.

veloce (approved):
👍

PR #950, opened 2024-08-30
Add en passant option in Editor

HaonRekcef:
Adds the option to enpassant in the editor, similar as its done on the website, as well as corresponding tests:

https://github.com/user-attachments/assets/5646a5f0-6520-420d-acf6-6959a1392565

remarks:
- if no en passant could be possible the option is not shown
- it is checked if there is an adjacent opposing pawn and if the two squares where he would have moved through are empty


lib/src/view/board_editor/board_editor_menu.dart
```diff
@@ -15,6 +15,9 @@ class BoardEditorMenu extends ConsumerWidget {
     final editorController = boardEditorControllerProvider(initialFen);
     final editorState = ref.watch(editorController);
 
+    final enPassantSquares =
+        ref.read(editorController.notifier).calculateEnPassantOptions();
```
veloce, line 19: We usually don't do that in riverpod. It is better to put the en passant options directly in the state instead.

lib/src/model/board_editor/board_editor_controller.dart
```diff
@@ -65,6 +66,56 @@ class BoardEditorController extends _$BoardEditorController {
     _updatePosition(readFen(fen).lock);
   }
 
+  /// Calculates the squares where an en passant capture could be possible.
+  SquareSet calculateEnPassantOptions() {
```
veloce, line 70: Instead of a public method, make a private one and add an `enPassantSquares` field to the state that exposes this computation.

veloce (commented):
Nice work!

lib/src/model/board_editor/board_editor_controller.dart
```diff
@@ -59,14 +61,66 @@ class BoardEditorController extends _$BoardEditorController {
     state = state.copyWith(
       sideToPlay: side,
     );
+    _calculateEnPassantOptions();
```
veloce, line 64: It is better to have atomic state changes, so:
```dart
state = state.copyWith(
  sideToPlay: side,
  enPassantOptions: _calculateEnPassantOptions(),
);
```

veloce: Done here: https://github.com/lichess-org/mobile/commit/9dc7df076e8e1eb9a5b1e03204e5dfcd8b88de74

olivertzeng:
Holy hell!
New update just dropped

PR #951, opened 2024-08-31
Improve share pgn

### LINKED ISSUE #948: Share PGN button doesn't save inputted info
### What is the problem?

As per a report from the forum, the share pgn function doesn't save the majority of the info that it prompts you to fill in, such as White and Black, Round, site, etc.
Thus when you click on the share pgn button, you copy a pgn that is mostly blank.
Additionally I've noticed that when inputting information down the list previous info won't save. For example I put down the names of White and Black, and when I input the game result, the names I put in for White and Black aren't saved.

All of the above is regarding creating a pgn from the analysis board, but it also applies after playing a game on Lichess, where you cannot edit any information because it will revert to the info that was already provided.

Go to the Analysis Board
Make a couple moves
Click on the share pgn button 
Fill in the information

Expected: Information to save when I input it, and when I click to share the pgn that info is copied.

What Happens: The information is not saved, and sharing the pgn leaves most of the items blank.

### App version

0.9.7

### Device Info

Pixel 6 and Android 14

### Additional information

Even just closing the keyboard wipes all the information, example shown:

https://github.com/user-attachments/assets/5927db6c-7659-49c7-ae53-379e30dcb846

HaonRekcef:
fix #948 
This pull request improves the sharing of PGNs:

- The values of entered text are now saved automatically, so the user doesn't need to submit to save the text.
- Submitting a field automatically jumps to the next one and selects all text. If the next field is the "Result" field, a choice picker will appear.
- Tapping on a field automatically selects all the text.

https://github.com/user-attachments/assets/4b33655a-6684-4d24-a47f-569b50a15b6b

Remarks:
- not tested on iOS

### DIFF
--- a/lib/src/view/analysis/analysis_share_screen.dart
+++ b/lib/src/view/analysis/analysis_share_screen.dart
@@ -66,6 +66,26 @@ class _EditPgnTagsFormState extends ConsumerState<_EditPgnTagsForm> {
   final Map<String, FocusNode> _focusNodes = {};
 
   @override
+  void initState() {
+    super.initState();
+    final ctrlProvider = analysisControllerProvider(widget.pgn, widget.options);
+    final pgnHeaders = ref.read(ctrlProvider).pgnHeaders;
+
+    for (final entry in pgnHeaders.entries) {
+      _controllers[entry.key] = TextEditingController(text: entry.value);
+      _focusNodes[entry.key] = FocusNode();
+      _focusNodes[entry.key]!.addListener(() {
+        if (!_focusNodes[entry.key]!.hasFocus) {
+          ref.read(ctrlProvider.notifier).updatePgnHeader(
+                entry.key,
+                _controllers[entry.key]!.text,
+              );
+        }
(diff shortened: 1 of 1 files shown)


EmmetSchuler:
Thanks for the quick fix!

lib/src/view/analysis/analysis_share_screen.dart
```diff
@@ -75,6 +123,20 @@ class _EditPgnTagsForm extends ConsumerWidget {
                     (e) => showRatings || !_ratingHeaders.contains(e.key),
                   )
                       .mapIndexed((index, e) {
+                    if (!_controllers.containsKey(e.key)) {
+                      _controllers[e.key] =
```
veloce, line 127: Not a big fan of doing side effects like that in the build method, but this is by far the easiest way here I reckon.

HaonRekcef: I see. I guess the initialization belongs to initState. 😁

PR #953, opened 2024-09-01
improve display and sharing of master games

HaonRekcef:
Improves the display of master games opened from the masters database in the opening explorer:

- The rating is now parsed (f.ex Carlsen, M. (2882) gets split up) and saved in the rating field.
- The offline icon is no longer shown as it doesn’t apply to master games.
- When opening a game from the database, then viewing it in the analysis board and sharing it as a PGN, the player names and ratings are now included in the PGN.

| **Before** | **After** |
|------------|-----------|
| Screenshot_20240901-144106 | Screenshot_20240901-144012 |
| Screenshot_20240901-144117 | Screenshot_20240901-144023 |


veloce (approved):
Thanks!

PR #954, opened 2024-09-01
refactor: add PlatformScaffold and PlatformAppBar widgets

tom-anders:
We had many screens where there was a lot of code duplication between androidBuilder and iosBuilder when using PlatformWidget. With these two new classes we clean a lot of this up.

Custom screens that need to do something different for Android vs iOS still use PlatformWidget.

This may be a bit more controversial than #938, and I'm totally fine if you don't want to merge this, I had a lot of fun refactoring and learned a lot anyway. I think this change would improve readability of the code base, especially for new contributors.

If you agree, I can again make some screenshots, confirming that everything still looks as before (I can only test Android though, not iOS)

### DIFF
--- a/lib/src/widgets/platform_scaffold.dart
+++ b/lib/src/widgets/platform_scaffold.dart
@@ -1,0 +1,177 @@
+import 'package:flutter/cupertino.dart';
+import 'package:flutter/material.dart';
+import 'package:lichess_mobile/src/styles/styles.dart';
+import 'package:lichess_mobile/src/widgets/platform.dart';
+
+/// Displays an [AppBar] for Android and a [CupertinoNavigationBar] for iOS.
+///
+/// Intended to be passed to [PlatformScaffold].
+class PlatformAppBar extends StatelessWidget {
+  const PlatformAppBar({
+    super.key,
+    required this.title,
+    this.centerTitle = false,
+    this.leading,
+    this.actions = const [],
+    this.cupertinoPadding,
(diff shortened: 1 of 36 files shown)


lib/src/widgets/platform_scaffold.dart
```diff
@@ -0,0 +1,177 @@
...
+  /// The main content of the screen, displayed below the navigation bar.
+  final Widget body;
+
+  /// See [Scaffold.resizeToAvoidBottomInset] and [CupertinoPageScaffold.resizeToAvoidBottomInset]
+  final bool resizeToAvoidBottomInset;
+
+  Widget _androidBuilder(BuildContext context) {
+    return Scaffold(
+      resizeToAvoidBottomInset: resizeToAvoidBottomInset,
+      appBar: PreferredSize(
```
veloce, line 154: Instead of wrapping with `PreferredSize`, would it be possible to have `PlatformAppBar` implements `ObstructingPreferredSizeWidget` and `PreferredSizeWidget`?

veloce (commented):
I thinks it's cool, we'll keep that refactoring. Thanks for the initiative!

lib/src/widgets/platform_scaffold.dart
tom-anders: I thought this was not possible because Android and iOS need different numeric values here, and we can't use `Theme.of(context).platform` because the `context` is not available.

veloce (approved):

Issue #945, opened 2024-08-28
[Crowdin] Improve the process of updating translations


veloce:
We cannot anymore update translation with `./script/update-arb-from-crowdin.mjs`.

To be future proof we should:
- leverage [crowdin API](https://developer.crowdin.com/api/v2/#tag/Translations)
- automate more the process

The script now only download the translations, it does not build them. A better solution would be a GitHub action than can be triggered manually and that would:
- build the project using crowdin API
- download the project (via API) and update ARB file (what is doing the script right now)
- run `flutter gen-l10n` to update the dart files
- create a PR with the updated translations so we can review the changes

comradekingu:
The more obvious course of action seems to be to use Weblate.
Lichess prides itself on being open and no spying.

tom-anders:
> The more obvious course of action seems to be to use Weblate. Lichess prides itself on being open and no spying.
> 
> The translations I see now are awful.

That's probably a discussion that should be raised over at https://github.com/lichess-org/lila, since the app reuses a lot of the existing Lila's crowdin translations, so it's not really feasible to switch the backend just for the app.

veloce:
Done:
- bb854b7a3ccbf4905912911689a8848b0d1bcc3e
- 216d742a0594a2fb49560d6544bf653222a3d7a6

PR #963, opened 2024-09-03
Update lila_zh_TW.arb from crowdin

olivertzeng:
I've tried to sync some essential ones from crowdin but please update these string from there asap thanks🙏


veloce:
Sorry, translation files cannot be updated individually. Currently the script doesn't work and [some work is needed](https://github.com/lichess-org/mobile/issues/945) to be able to again update translations.

PR #966, opened 2024-09-03
refactor: add PlatformAlertDialog

tom-anders:
In a similar spirit to `PlatformScaffold`, but implementation was much more straight-forward. So I think one screenshot is enough:

Screenshot_1725396066


veloce (approved):
Very nice! I love these refactoring :)

tom-anders:
> Very nice! I love these refactorings :)

Yay, I'll keep going then :D

Next place would probably be the settings, where we usually display a choice picker for Android, but a new screen for ios. Looks to me like this has some potential for a similar refactoring as well

PR #971, opened 2024-09-04
docs: various clean ups and improvements

tom-anders:
- Move dev-setup related docs from README.md to setting_dev_env.md, some of this was duplicated between the two.
- Add "Before submitting a Pull Request" to CONTRIBUTING.md
- Add section about `PlatformWidget` to coding_style.md


veloce (approved):
Thanks for that, it is better now.

PR #990, opened 2024-09-06
feat: add study list screen

tom-anders:
@veloce Since studies are such a huge feature, I'd suggest we split it up and start with this PR that only will add the study list screen. This also makes it easier for others to take over development from me if necessary. After review, we can hide the button until the study feature is fully implemented.

### DIFF
--- a/lib/src/view/study/study_list_screen.dart
+++ b/lib/src/view/study/study_list_screen.dart
@@ -1,0 +1,467 @@
+import 'package:cached_network_image/cached_network_image.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+import 'package:lichess_mobile/src/model/auth/auth_session.dart';
+import 'package:lichess_mobile/src/model/study/study.dart';
+import 'package:lichess_mobile/src/model/study/study_filter.dart';
+import 'package:lichess_mobile/src/model/study/study_list_paginator.dart';
+import 'package:lichess_mobile/src/styles/lichess_icons.dart';
+import 'package:lichess_mobile/src/styles/styles.dart';
+import 'package:lichess_mobile/src/utils/l10n_context.dart';
+import 'package:lichess_mobile/src/utils/lichess_assets.dart';
+import 'package:lichess_mobile/src/widgets/adaptive_bottom_sheet.dart';
+import 'package:lichess_mobile/src/widgets/buttons.dart';
+import 'package:lichess_mobile/src/widgets/filter.dart';
+import 'package:lichess_mobile/src/widgets/list.dart';
+import 'package:lichess_mobile/src/widgets/platform_scaffold.dart';
+import 'package:lichess_mobile/src/widgets/user_full_name.dart';
+import 'package:logging/logging.dart';
+import 'package:timeago/timeago.dart' as timeago;
+
+final _logger = Logger('StudyListScreen');
+
+// TODO l10n
+String studyCategoryL10n(StudyCategory category, BuildContext context) =>
+    switch (category) {
+      StudyCategory.all => 'All',
+      StudyCategory.mine => 'Mine',
+      StudyCategory.member => 'Member',
+      StudyCategory.public => 'Public',
+      StudyCategory.private => 'Private',
+      StudyCategory.likes => 'Liked',
+    };
+
+// TODO l10n
+String studyListOrderL10n(StudyListOrder order, BuildContext context) =>
+    switch (order) {
+      StudyListOrder.hot => 'Hot',
+      StudyListOrder.newest => 'Newest',
+      StudyListOrder.oldest => 'Oldest',
+      StudyListOrder.updated => 'Updated',
+      StudyListOrder.popular => 'Popular',
+    };
+
+/// A screen that displays a paginated list of studies
+class StudyListScreen extends ConsumerWidget {
+  const StudyListScreen({super.key});
+
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {
+    final isLoggedIn = ref.read(authSessionProvider)?.user.id != null;
+
+    final filter = ref.watch(studyFilterProvider);
+    final categorySection =
+        isLoggedIn ? ' • ${studyCategoryL10n(filter.category, context)}' : '';
+    final title = Text(
+      '${context.l10n.studyMenu}$categorySection • ${studyListOrderL10n(filter.order, context)}',
+    );
+
+    return PlatformScaffold(
+      appBar: PlatformAppBar(
+        title: title,
+        actions: [
+          AppBarIconButton(
+            icon: const Icon(Icons.tune),
+            semanticsLabel: 'Filter studies',
+            onPressed: () => showAdaptiveBottomSheet<void>(
+              context: context,
+              builder: (_) => _StudyFilterSheet(
+                isLoggedIn: isLoggedIn,
+              ),
+            ),
+          ),
+        ],
+      ),
+      body: _Body(
+        filter: ref.watch(studyFilterProvider),
+      ),
+    );
+  }
+}
+
+class _StudyFilterSheet extends ConsumerWidget {
+  const _StudyFilterSheet({required this.isLoggedIn});
+
+  final bool isLoggedIn;
+
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {
+    final filter = ref.watch(studyFilterProvider);
+
+    return SafeArea(
+      child: Padding(
+        padding: const EdgeInsets.all(16.0),
+        child: Column(
+          mainAxisSize: MainAxisSize.min,
+          children: [
+            const SizedBox(height: 12.0),
+            // If we're not logged in, the only category available is "All"
+            if (isLoggedIn) ...[
+              Filter<StudyCategory>(
+                // TODO l10n
+                filterName: 'Category',
+                filterType: FilterType.singleChoice,
+                choices: StudyCategory.values,
+                choiceSelected: (choice) => filter.category == choice,
+                choiceLabel: (category) =>
+                    Text(studyCategoryL10n(category, context)),
+                onSelected: (value, selected) =>
+                    ref.read(studyFilterProvider.notifier).setCategory(value),
+              ),
+              const PlatformDivider(thickness: 1, indent: 0),
+              const SizedBox(height: 10.0),
+            ],
+            Filter<StudyListOrder>(
+              // TODO l10n
+              filterName: 'Sort by',
+              filterType: FilterType.singleChoice,
+              choices: StudyListOrder.values,
+              choiceSelected: (choice) => filter.order == choice,
+              choiceLabel: (order) => Text(studyListOrderL10n(order, context)),
+              onSelected: (value, selected) =>
+                  ref.read(studyFilterProvider.notifier).setOrder(value),
+            ),
+          ],
+        ),
+      ),
+    );
+  }
+}
+
+class _Body extends StatefulWidget {
+  const _Body({
+    required this.filter,
+  });
+
+  final StudyFilterState filter;
+
+  @override
+  State<_Body> createState() => _BodyState();
+}
+
+class _BodyState extends State<_Body> {
+  String? search;
+
+  final _searchController = SearchController();
+
+  @override
+  void dispose() {
+    super.dispose();
+    _searchController.dispose();
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    return SafeArea(
+      child: Column(
+        children: [
+          Padding(
+            padding: Styles.bodySectionPadding,
+            child: SearchBar(
+              controller: _searchController,
+              leading: const Icon(Icons.search),
+              trailing: [
+                if (search != null)
+                  IconButton(
+                    onPressed: () => setState(() {
+                      search = null;
+                      _searchController.clear();
+                    }),
+                    tooltip: 'Clear',
+                    icon: const Icon(
+                      Icons.close,
+                    ),
+                  ),
+              ],
+              hintText: search ?? context.l10n.searchSearch,
+              onSubmitted: (term) {
+                setState(() {
+                  search = term;
+                });
+              },
+            ),
+          ),
+          _StudyList(
+            paginatorProvider: StudyListPaginatorProvider(
+              filter: widget.filter,
+              search: search,
+            ),
+          ),
+        ],
+      ),
+    );
+  }
+}
+
+class _StudyList extends ConsumerStatefulWidget {
+  const _StudyList({
+    required this.paginatorProvider,
+  });
+
+  final StudyListPaginatorProvider paginatorProvider;
+
+  @override
+  ConsumerState<ConsumerStatefulWidget> createState() => _StudyListState();
+}
+
+class _StudyListState extends ConsumerState<_StudyList> {
+  final _scrollController = ScrollController();
+
+  @override
+  void initState() {
+    super.initState();
+    _scrollController.addListener(_scrollListener);
+  }
+
+  @override
+  void dispose() {
+    _scrollController.removeListener(_scrollListener);
+    _scrollController.dispose();
+    super.dispose();
+  }
+
+  void _scrollListener() {
+    if (_scrollController.position.pixels ==
+        _scrollController.position.maxScrollExtent) {
+      final studiesList = ref.read(widget.paginatorProvider);
+
+      if (!studiesList.isLoading) {
+        ref.read(widget.paginatorProvider.notifier).next();
+      }
+    }
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    final studiesAsync = ref.watch(widget.paginatorProvider);
+
+    return studiesAsync.when(
+      data: (studies) {
+        return Expanded(
+          child: ListView.separated(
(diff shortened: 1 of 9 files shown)


ijm8710:
I like the big icons they show on web.
Makes it easier to quickly differentiate between studies

tom-anders:
[study_list.webm](https://github.com/user-attachments/assets/87b4d662-3549-4eb6-bb63-beeaf62337df)

veloce:
> @veloce Since studies are such a huge feature, I'd suggest we split it up and start with this PR that only will add the study list screen. This also makes it easier for others to take over development from me if necessary. After review, we can hide the button until the study feature is fully implemented.

@tom-anders yeah great initiative! we should split this one as much as possible; I prefer to review small PRs :)

tom-anders:
> > @veloce Since studies are such a huge feature, I'd suggest we split it up and start with this PR that only will add the study list screen. This also makes it easier for others to take over development from me if necessary. After review, we can hide the button until the study feature is fully implemented.
> 
> @tom-anders yeah great initiative! we should split this one as much as possible; I prefer to review small PRs :)

Perfect 👍 I'll further split up this PR as well.

ijm8710:
Thanks for adding the icons :) 

2 more

A few times in video the names become ... prob because too long. But on mobile web I don't see it. So I'm assuming (because names are prob capped anyway) should be able to fit them always without truncation

I do slightly prefer how mobile web the chosen filter sort is actively shown (ie "hot" studies) rather than having to open the filter to see the specified choice

tom-anders:
> Thanks for adding the icons :)
> 
> 2 more
> 
> A few times in video the names become ... prob because too long. But on mobile web I don't see it. So I'm assuming (because names are prob capped anyway) should be able to fit them always without truncation
> 

I guess that depends on the on the screen size, you probably always have cases where names are to long. But we can tweak the layout and font size too make it a bit less likely. 

> I do slightly prefer how mobile web the chosen filter sort is actively shown (ie "hot" studies) rather than having to open the filter to see the specified choice

Good point - I don't want the buttons to be visible always, but adjusting the title of the top bar would probably be the best solution. Something like "Study - All - Popular"

tom-anders:
(converted back to a draft, as I'll split some code into smaller independent PRs

veloce:
I think the `StudyController` should be completely separate from the analysis controller and should have its own logic.

Over reusing a controller is not a good idea. I plan to write a dedicated controller for the opening explorer for instance, because almost all the logic of the analysis controller is not needed in the explorer.

Once we have several controller that share some logic, we can think about refactoring, using mixins probably. This kind of refactoring is not that easy to do, so we'd better way for things to stabilise.

veloce:
That being said, if you fancy that kind of refactoring and want to create a `Notififier` mixin to handle basic operations like navigating in the tree, fell free to try. But I'm not sure it is easy, because you'd have differences in each app screen.

> Refactor analysisControllerProvider, so that you can either pass a PGN, or an existing Root object (that we convert the study json into). I'm not really sure what's the most idiomatic way to do this though, do we simply use two nullable parameters and require that one of them has to be non-null?

Is there a direct way to convert the study JSON to a `Node`? We need a tree structure in study for sure, but study doesn't have the same need as analysis controller. A couple of notes:
- analysis controller takes a PGN because it aims to support lichess analysis and standalone analysis (PGN input)
- performance wise, you don't want to supply large objects to a Provider `Notifier` family instance (such a an `IList`, or a large object)

For this reason, the study notifier should take only the study ID as parameter, and be asynchronous.

tom-anders:
Thanks for the quick and helpful reply! 

> I think the `StudyController` should be completely separate from the analysis controller and should have its own logic.
> 
> Over reusing a controller is not a good idea. I plan to write a dedicated controller for the opening explorer for instance, because almost all the logic of the analysis controller is not needed in the explorer.



> Once we have several controller that share some logic, we can think about refactoring, using mixins probably. This kind of refactoring is not that easy to do, so we'd better way for things to stabilise. 

Ah, the opening explorer was actually the code that gave me this idea of reusing the controller in the first place. Ok, then I agree that we should write a separate study controller, and (maybe) do the refactoring in the future. 


> Is there a direct way to convert the study JSON to a `Node`?

Haven't tried yet, but the layout looked similar enough. But maybe it's better to use a dedicated StudyNode class instead, we'll see. 

> * performance wise, you don't want to supply large objects to a riverpod `Notifier` family instance (such a an `IList`, or a large object), because `provider` will compute the `hash` each time the widget is built, and the `hash` of a large object is expensive to compute. Riverpod's doc warns agains this also (should only provide primitives to a provider family)
> 
> For this reason, the study notifier should take only the study ID as parameter, and be asynchronous.

Good hint, will keep this in mind 👍

julien4215:
For the broadcast feature, I made a different screen for the analysis board but I reused the analysis controller because I thought the logic was almost the same. I also went for this because the api returns a pgn https://lichess.org/api#tag/Studies/operation/studyChapterPgn so I don't understand why you said tom-anders that you need to pass a `Root` object to the analysis controller. By the way I think the study and broadcast feature might have some code in common because the broadcast is implemented as a study on the backend side.

tom-anders:
> For the broadcast feature, I made a different screen for the analysis board but I reused the analysis controller because I thought the logic was almost the same. I also went for this because the api returns a pgn https://lichess.org/api#tag/Studies/operation/studyChapterPgn so I don't understand why you said tom-anders that you need to pass a `Root` object to the analysis controller. 

Oh wow, I must have completely missed that API endpoint when looking through the docs. Currently I'm using `lichess.org/study/<id>/<chapter>` (run `curl -X GET 'https://lichess.org/study/U3guoz56/byuyr1dH' -H "Accept: application/json" | jq --color-output | less` for an example), which returns a tree-like structure instead of a PGN. That's why I was talking about nodes etc.

In that case, we probably _do_ want to reuse the `AnalysisController` then.

tom-anders:
@veloce what's you opinion on making the analysis tree view more "tree-like", especially for nested sidelines, like in the old app and Web interface? I think especially for studies this would be much nicer. Attached a comparison of old and new app for reference. The current interface can get pretty confusing once you have multiple levels of nested sidelines (plus comments) . I have a WIP branch for studies that is more close to the old app, should we do this for analysis as well? 

Screenshot_20240915-175823
Screenshot_20240915-175633

tom-anders:
Here's what it currently looks like with my refactoring:

Screenshot_1726420817

tom-anders:
Marking this as "Ready to Review" again, as I don't think I can split this up further

tom-anders:
I rebased against `main` to fix conflicts with the build_runner PR, also adjusted the formatting of study chapters/members a bit, to avoid the names getting cut off.

Here's an updated video:

[study_list.webm](https://github.com/user-attachments/assets/5c9bb447-15d0-47d4-b60a-39c2b0f54050)

ijm8710:
Excited for this Tom. The one thing I find that's goofy is the open circle bullets. I know that's how it is on mobile web but as Veloce says they don't always have to mirror.

The one thing I'd ask is do they serve any point. If those circles get filled in or checked off when a study is completed then they do make some sense. If they don't currently, is that even possible to do even if web doesn't since it adds decent utility? If not, do you think there's a better formatting marker. I know that's two decently long paragraphs for some a minor formatting marker and no clue when they look so goofy to me lol

tom-anders:
> Thanks Tom,
> 
> Any chance to see both the - and the • as a comparison 😀

I'll wait for @veloce's input here first

> Is having circles get filled as you complete a chapter even possible? I think it'd be really cool!! If not no worries

Does not look like the API supports it

> Awesome about the favorite studies comment. So then what's a small preview of what's left for studies. You hinted it's a huge pr and this was just the first piece of it, or did you actually do most of it already since you're awesome

I have a branch that implements the study screen, but it's not ready for review yet. Next would be a follow-up PR adding support for interactive studies.

lib/src/view/study/study_list_screen.dart
```diff
@@ -0,0 +1,472 @@
...
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    return SafeArea(
+      child: Column(
+        children: [
+          Padding(
+            padding: Styles.bodySectionPadding,
+            child: SearchBar(
```
veloce, line 162: We should display a different bar style on iOS.

See how it's done in search scree: https://github.com/lichess-org/mobile/blob/main/lib/src/view/user/search_screen.dart#L108-L115 

Maybe we can write a `PlatformSearchBar` widget?

lib/src/view/study/study_list_screen.dart
```diff
@@ -0,0 +1,472 @@
...
+
+  @override
+  void dispose() {
+    _scrollController.removeListener(_scrollListener);
+    _scrollController.dispose();
+    super.dispose();
+  }
+
+  void _scrollListener() {
+    if (_scrollController.position.pixels ==
```
veloce, line 226: I think we should trigger the next page before the scroll reaches the end, so the scrolling will feel smoother. I know it's not done like that in other paginated view but that's something I wanted to change (will write an issue).

I don't know what is a good threshold, perhaps 75% of the max scroll extent?

lib/src/view/study/study_list_screen.dart
```diff
@@ -0,0 +1,472 @@
...
+      StudyListOrder.popular => 'Popular',
+    };
+
+/// A screen that displays a paginated list of studies
+class StudyListScreen extends ConsumerWidget {
+  const StudyListScreen({super.key});
+
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {
+    final isLoggedIn = ref.read(authSessionProvider)?.user.id != null;
```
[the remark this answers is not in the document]
tom-anders: huh, it's been some time since I wrote this, but I probably assumed that the user can never change while we show this screen..? Not a good enough reason though, I changed it to `watch` now

lib/src/view/study/study_list_screen.dart
tom-anders: Good idea, done :+1:

lib/src/view/study/study_list_screen.dart
tom-anders: 75% sounds good :)
 
I had to introduce an additional `bool` to the state to avoid triggering the request multiple times before the next page is loaded.

veloce (commented):
I've noticed a strange UI freeze when pushing to study screen:

https://github.com/user-attachments/assets/d792d7bd-f098-461e-a615-a9b702dc3298

I don't know what is causing this. It is also on android.

Also if you want to use the real study icon from lichess, I've pushed it, you can rebase to main to get it.

tom-anders:
> I've noticed a strange UI freeze when pushing to study screen:
> Enregistrement.de.l.ecran.2024-10-29.a.09.34.22.mov
> 
> I don't know what is causing this. It is also on android.
> 
> Also if you want to use the real study icon from lichess, I've pushed it, you can rebase to main to get it.

Oh, that's not actually a freeze, sorry for not mentioning that, it's just a dummy study screen I added here: https://github.com/lichess-org/mobile/pull/990/files#diff-d6c82c9cf2716bd768ab5fc1abfc30ce85567b52d36230ebdf68f2c616388d56

The intention was to make https://github.com/lichess-org/mobile/pull/1111 (which adds the "real" study screen) independent from the study list screen PR.

veloce:
> Oh, that's not actually a freeze, sorry for not mentioning that, it's just a dummy study screen I added here: https://github.com/lichess-org/mobile/pull/990/files#diff-d6c82c9cf2716bd768ab5fc1abfc30ce85567b52d36230ebdf68f2c616388d56
> 
> The intention was to make #1111 (which adds the "real" study screen) independent from the study list screen PR.

I know it's an empty study screen. What I meant by freeze is that the animation starts then stops abruptly at about 50% after it started, it stays blocked during a short time then resume. That's not normal, it should be smooth and continuous all the time.

tom-anders:
> > Oh, that's not actually a freeze, sorry for not mentioning that, it's just a dummy study screen I added here: https://github.com/lichess-org/mobile/pull/990/files#diff-d6c82c9cf2716bd768ab5fc1abfc30ce85567b52d36230ebdf68f2c616388d56
> > The intention was to make #1111 (which adds the "real" study screen) independent from the study list screen PR.
> 
> I know it's an empty study screen. What I meant by freeze is that the animation starts then stops abruptly at about 25% after it started, it stays blocked during a short time then resume. That's not normal, it should be smooth and continuous all the time.

Ah, got it, I can reproduce as well. But I still think it's related to the empty screen - If I take #1111 and rebase it onto this branch (Pass `-Xtheirs` to `git rebase`), then the freeze is gone

test/view/study/study_list_screen_test.dart
```diff
@@ -0,0 +1,694 @@
...
+        tester,
+        home: const StudyListScreen(),
+        overrides: [
+          lichessClientProvider.overrideWith((ref) {
+            return LichessClient(mockClient, ref);
+          }),
+        ],
+      );
+      await tester.pumpWidget(app);
+      await tester.pumpAndSettle();
```
veloce, line 77: Why is `pumpAndSettle` needed here? it is usually needed when animations need to finish, but I prefer put a comment each time we use `pump()` or `pumpOrSettle` to explain what are we waiting for, to help future readers.

test/view/study/study_list_screen_test.dart
```diff
@@ -0,0 +1,694 @@
...
+      await tester.scrollUntilVisible(
+        find.text('First Study Page 2'),
+        200,
+        scrollable: find.descendant(
+          of: find.byType(ListView),
+          matching: find.byType(Scrollable),
+        ),
+      );
+
+      await tester.pumpAndSettle();
```
veloce, line 92: Also explain why there is a `pumpAndSettle` at the end.

Would be good to track the requests made and add a check to confirm that the requests match the expected.

veloce (commented):
The search bar is sticky so it should be part of the page header, it will look much better. On Android it is very easy, on iOS we have to workaround a little. You can see how it's done in the opening explorer screen.

veloce:
> The search bar is sticky so it should be part of the page header, it will look much better. On Android it is very easy, on iOS we have to workaround a little. You can see how it's done in the opening explorer screen.

I must add that another simpler solution is to make the search bar non sticky. This is perhaps even better.

test/view/study/study_list_screen_test.dart
tom-anders: Ah, this `pumpAndSettle()` was needed because

```
WidgetsBinding.instance.addPostFrameCallback((_) {
   setState(() {
      requestedNextPage = false;
   });
});
```
causes a crash otherwise. But, thanks to the helpful flutter error message, we can add an `if (mounted)` check to fix this instead.

test/view/study/study_list_screen_test.dart
tom-anders: ah this one is needed, I added a comment

tom-anders: (although `pump` seems to also be sufficient)

test/view/study/study_list_screen_test.dart
tom-anders: > Would be good to track the requests made and add a check to confirm that the requests match the expected.

Added helper variables to track the requests, not sure if this is the most elegant solution though

tom-anders:
> > The search bar is sticky so it should be part of the page header, it will look much better. On Android it is very easy, on iOS we have to workaround a little. You can see how it's done in the opening explorer screen.
> 
> I must add that another simpler solution is to make the search bar non sticky. This is perhaps even better.

Done, I wrapped the column with a `SingleChildScrollView` such that the search bar is scrolled as well.  This broke the `ScrollListener` in the ListView though, so I had to move it up to `_Body`

veloce:
I tested it on a low end device to check if embedding a `ListView` inside a `SingleChildScrollView` would not cause performance issues.

Turn out that the answer is yes... I suppose the `ListView` has to be the primary scrollable to be efficient.

A simple way to deal with it I think:
```dart
itemBuilder: (context, index) {
  if (index == 0) {
    return SearchBar();
  } else {
      final study = studies.studies[index - 1];
      ....
  }
```

tom-anders:
> I tested it on a low end device to check if embedding a `ListView` inside a `SingleChildScrollView` would not cause performance issues.
> 
> Turn out that the answer is yes... I suppose the `ListView` has to be the primary scrollable to be efficient.
> 
> A simple way to deal with it I think:
> 
> ```dart
> itemBuilder: (context, index) {
>   if (index == 0) {
>     return SearchBar();
>   } else {
>       final study = studies.studies[index - 1];
>       ....
>   }
> ```

Ah yes, that seems more elegant anyway. Done.

lib/src/view/study/study_list_screen.dart
```diff
@@ -0,0 +1,472 @@
...
+      child: Padding(
+        padding: const EdgeInsets.all(16.0),
+        child: Column(
+          mainAxisSize: MainAxisSize.min,
+          children: [
+            const SizedBox(height: 12.0),
+            // If we're not logged in, the only category available is "All"
+            if (isLoggedIn) ...[
+              Filter<StudyCategory>(
+                // TODO mobile l10n
```
tom-anders, line 101: This one (and also "Sort by") do not exist in the Web UI and need will mobile-specific translations at some point in the future

PR #1002, opened 2024-09-09
[Study] add models for listing available studies

tom-anders:
Do we need separate test for json parsing here? Or is it enough to have a end-to-end test for the final study list screen?


veloce:
Some failing tests here @tom-anders

tom-anders:
ah, classic unused imports, fixed now :+1:

tom-anders:
oh looks like it's time to fix lints after the latest flutter update again though

veloce (approved):

PR #1003, opened 2024-09-09
[Study] refactor: add shared model for chat messages

tom-anders:
We'll reuse it for studies.

Short recording to verify chat still works:

[chat.webm](https://github.com/user-attachments/assets/52ec4c2d-4392-4cc6-bdbf-c7ea2d758f47)


veloce:
For this one I need to step back a little and think. We've having issues related to chat controller and the websocket in #825

PR #1025, opened 2024-09-14
🧹  Update FVM config files for FVM 3.0

jas14:
Hello again 🙂

FVM 3.x is available as of 2024-02-19, so it seems it's probably been enough time for folks to upgrade. As described in the [changelog](https://github.com/leoafarias/fvm/blob/main/CHANGELOG.md#breaking-changes), `.fvmrc` is the new config file; `.fvm/fvm_config.json` is now a legacy file.

So this PR is just a bit of tidying to upgrade to FVM 3.0+. Looking forward to your feedback!

### DIFF
--- a/.fvm/fvm_config.json
+++ b/.fvm/fvm_config.json
@@ -1,4 +1,0 @@
-{
-  "flutterSdkVersion": "beta",
-  "flavors": {}
-}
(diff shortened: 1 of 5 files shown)


veloce:
I'm not using FVM, so I'm trusting you and merging it without testing :) Thanks!

Issue #1010, opened 2024-09-11
The board editor is missing the "Start position" and "clear board" buttons


aslyamov:
The board editor is missing the "Start position" and "clear board" buttons

Can we add these buttons, for example, to the board editor settings?

![Screenshot_20240911-144721](https://github.com/user-attachments/assets/88f41d30-d451-4597-a899-bf1d79ca89e1)

![Снимок экрана 2024-09-11 144455](https://github.com/user-attachments/assets/9f1eb3f0-c7b8-4409-ba83-7f7a322eacd2)

bhavikpanchal1111:
I am willing to solve this issue, so I would need developer access

PR #1035, opened 2024-09-17
feat: [BoardEditor] add reset board and starting position option

sid0-0:
Implements #1010 

Screenshot:
image

### DIFF
--- a/lib/src/model/board_editor/board_editor_controller.dart
+++ b/lib/src/model/board_editor/board_editor_controller.dart
@@ -76,7 +76,7 @@ class BoardEditorController extends _$BoardEditorController {
       setSideToPlay(splits[1].toLowerCase() == 'w' ? Side.white : Side.black);
     }
 
-    if(loadCastling && splits.length >= 3) {
+    if (loadCastling && splits.length >= 3) {
       final castling = splits[2];
       setCastling(Side.white, CastlingSide.king, castling.contains('K'));
       setCastling(Side.white, CastlingSide.queen, castling.contains('Q'));
--- a/lib/src/model/board_editor/board_editor_controller.dart
+++ b/lib/src/model/board_editor/board_editor_controller.dart
@@ -64,8 +64,25 @@ class BoardEditorController extends _$BoardEditorController {
     );
   }
 
-  void loadFen(String fen) {
-    _updatePosition(readFen(fen).lock);
+  void loadFen(
+    String fen, {
+    bool loadSideToPlay = false,
+    bool loadCastling = false,
+  }) {
+    final splits = fen.split(' ');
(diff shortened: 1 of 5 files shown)


sid0-0:
Should I also add an option to `Load from FEN`?

veloce:
These new buttons should not be placed here, and should not look like a chip. They are action buttons (clear board) where the other ones are toggles on the current position.

We should rearrange as follow imo:
- add a 'menu' button (menu icon 3 bars as in game screen)
  - flip board, clear board and reset board should be part of this new menu
- the `BoardEditorMenu` should be renamed `BoardEditorPositionSettings`
- the button icon for the position settings should be changed to a 'settings' (cog) icon

veloce:
> Should I also add an option to `Load from FEN`

We could, but we already have the "Load position" screen which achieves the same.

More interesting is to add the choice picker to choose a preconfigured opening or endgame position like in the website.

sid0-0:
Made the suggested changes.
Screenshots:
Screenshot_1727122329
Screenshot_1727122337

While we're modifying menu(now settings), I also think we should move the castling options around for better clarity. I would suggest something like this:
Screenshot_1727123068

IEncinas10:
Hello! I was missing this from the app and decided to try and see if I could implement it without noticing there is a PR open. 

It doesn't look like there is much activity around this, so I'll share my proposal:

image

which I think looks reasonably similar to the webpage:

image

The implementation is as simple as:

```diff
diff --git a/lib/src/view/board_editor/board_editor_screen.dart b/lib/src/view/board_editor/board_editor_screen.dart
index e5a66bd1879d..3a76be5579f5 100644
--- a/lib/src/view/board_editor/board_editor_screen.dart
+++ b/lib/src/view/board_editor/board_editor_screen.dart
@@ -315,6 +315,20 @@ class _BottomBar extends ConsumerWidget {
           ),
           icon: Icons.tune,
         ),
+        BottomBarButton(
+          label: context.l10n.startPosition,
+          onTap: () => ref
+              .read(boardEditorControllerProvider(initialFen).notifier)
+              .loadFen(kInitialFEN),
+          icon: CupertinoIcons.arrow_counterclockwise,
+        ),
+        BottomBarButton(
+          label: context.l10n.clearBoard,
+          onTap: () => ref
+              .read(boardEditorControllerProvider(initialFen).notifier)
+              .loadFen(kEmptyFEN),
+          icon: CupertinoIcons.delete_simple,
+        ),
         BottomBarButton(
           key: const Key('flip-button'),
           label: context.l10n.flipBoard,
```

Let me know if I should open a PR. Thanks for your hard work!

veloce:
Closed by https://github.com/lichess-org/mobile/pull/1394

PR #1040, opened 2024-09-19
fix: [TV] add last move highlights

### LINKED ISSUE #1026: Lichess TV games don't display board highlights
### What is the problem?

From the Forum:
On Lichess TV games, board highlights, king check highlights, etc don't display.

Enter Lichess TV
Expected: Board highlights are shown
What Happened: board highlights are not shown

### App version

0.11.0

### Device Info

Pixel 6 and Android 14

### Additional information

_No response_

sid0-0:
~~Fixes #1026~~
feat: add last move highlights in TV

Created and passed `GameData` to display last move highlights and checks.

### DIFF
--- a/lib/src/view/watch/tv_screen.dart
+++ b/lib/src/view/watch/tv_screen.dart
@@ -1,6 +1,4 @@
-import 'package:chessground/chessground.dart';
 import 'package:dartchess/dartchess.dart';
-import 'package:fast_immutable_collections/fast_immutable_collections.dart';
 import 'package:flutter/cupertino.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:lichess_mobile/src/constants.dart';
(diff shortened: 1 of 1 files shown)


lib/src/view/watch/tv_screen.dart
```diff
@@ -111,6 +113,21 @@ class _Body extends ConsumerWidget {
                       : null,
                   materialDiff: game.lastMaterialDiffAt(Side.white),
                 );
+
+                final gameAtCursor = game.positionAt(gameState.stepCursor);
+
+                final gameDataAtCursor = GameData(
```
veloce, line 119: The `GameData` was meant only for boards that are playable. 

For now we can only display the `lastMove` which is already a great improvement. We'll only be missing the king check highlight which is not a big deal. I'll figure out what we can do with the chessground API later, to improve that.

sid0-0: updated

veloce (approved):

PR #1000, opened 2024-09-09
[Study] refactor: make _Filter from game filter public

tom-anders:
I'll reuse this in the study list screen. I think the coordinate trainer config screen can even reuse this as well.

PR #1041, opened 2024-09-19
Coordinate trainer adjustements

julien4215:
- Style changes for the time progress bar and the settings
- Show score at the end
- New training in random mode recomputes a random side (consistent with the website)


lib/src/view/coordinate_training/coordinate_training_screen.dart
```diff
@@ -351,61 +404,54 @@ class _SettingsState extends ConsumerState<_Settings> {
...
-                        .read(coordinateTrainingPreferencesProvider.notifier)
-                        .setSideChoice(choice);
-                  },
-                );
-              }).toList(),
-            ),
-          ),
+        Wrap(
+          spacing: 8.0,
+          children: SideChoice.values.map((choice) {
```
tom-anders, line 412: While you're at it, I think we could use the `Filter` class here that I made public in https://github.com/lichess-org/mobile/pull/1000

julien4215: The problem is that I removed the text because I think it looks better without it and the `Filter` widget requires it.

I'll post a before/after video so you can see what I mean.

julien4215:
https://github.com/user-attachments/assets/19a86102-6c2f-4793-a046-5b6fd6fb62ec

https://github.com/user-attachments/assets/291d8907-68c1-4e7b-955c-30683b13e179

julien4215:
https://github.com/user-attachments/assets/8d27fc60-bf69-4a8a-9e0d-1d88a60f0bb

https://github.com/user-attachments/assets/5370f6f3-0ccf-4d1a-84fc-5166faa12cdb

lib/src/view/coordinate_training/coordinate_training_screen.dart
tom-anders: Ah got it 👍 (impressive speed btw)

lib/src/view/coordinate_training/coordinate_training_screen.dart
```diff
@@ -47,17 +47,31 @@ class _Body extends ConsumerStatefulWidget {
 class _BodyState extends ConsumerState<_Body> {
   late Side orientation;
 
+  late bool computeRandomOrientation;
```
veloce, line 50: This new `boolean` and the `_maybeSetOrientation` makes the code harder to read and understand.

This is a sign that now this logic shouldn't be in the widget itself, as it is not an ephemeral state.

Adding the `orientation` as a property of the `CoordinateTrainingState`, and recompute the orientation at each new session if `random` pref is chosen, would be much clearer.

julien4215: Even if I move this to the state, I'll still need the boolean but what I can do instead is to choose the random side when the user either aborts or ends a training.

This way the boolean would not be needed and I think it is a better experience for the user to have the orientation before starting the training. This will differ from the website but I believe it is better for the user and easier to code.

I think it still remains a good idea to move this to the coordinate training controller.

julien4215:
I am trying to find a better layout to avoid overflows.
[layout.webm](https://github.com/user-attachments/assets/9b5b6dd7-3953-4645-8d2a-b06bcd6d3847)

veloce:
I don't really understand the conflicts here @julien4215 . Can you have a look? Thanks :)

veloce:
> I still didn't push my changes. I'd like to know first if you prefer the new layout I posted in my last message. I am not sure about it since it makes the side and time choices less accessible.

Oops I missed that, I'm sorry. Yes I think it is fine as it is in your last recording.

veloce (approved):

PR #1053, opened 2024-09-28
feat: make analysis tree view more "tree like"

tom-anders:
Especially for heavily commented PGNs this is now much more readable. We're also more consistent with the website now. I added screenhots for comparison below.

I know this is a rather big refactoring, and the recursive nature of this stuff is quite complex, so let me know if anything is unclear. 

I also made a few style adaptions to be a bit more consistent with the the Web UI, but I don't have any strong opinions here, feel free to tweak the styling again after merging.

(Left: Before, Right: This PR)

<p float="left">
  
   
</p>

lichess.org for comparison:

Screenshot_2024-09-28_11-23-42

### DIFF
--- a/lib/src/view/analysis/tree_view.dart
+++ b/lib/src/view/analysis/tree_view.dart
@@ -239,6 +239,7 @@ class _PgnTreeView extends StatelessWidget {
                     parent: nodes.last,
                     params: params,
                     initialPath: sidelineInitialPath,
+                    nesting: 1,
                   ),
               ];
             },
@@ -490,6 +491,7 @@ class _SideLines extends StatelessWidget {
     required this.firstMoveKey,
     required this.initialPath,
     required this.params,
+    required this.nesting,
   });
 
   final ViewBranch firstNode;
@@ -497,6 +499,7 @@ class _SideLines extends StatelessWidget {
   final GlobalKey firstMoveKey;
   final UciPath initialPath;
   final _PgnTreeViewParams params;
+  final int nesting;
 
   @override
   Widget build(BuildContext context) {
@@ -528,6 +531,7 @@ class _SideLines extends StatelessWidget {
               UciPath.fromIds(sidelineNodes.map((node) => node.id)),
             ),
             params: params,
+            nesting: nesting + 1,
           ),
       ],
     );
@@ -581,6 +585,7 @@ class _IndentedSideLines extends StatefulWidget {
     required this.parent,
     required this.initialPath,
     required this.params,
+    required this.nesting,
   });
 
   final Iterable<ViewBranch> sideLines;
@@ -590,6 +595,8 @@ class _IndentedSideLines extends StatefulWidget {
   final UciPath initialPath;
 
   final _PgnTreeViewParams params;
+
+  final int nesting;
 
   @override
   State<_IndentedSideLines> createState() => _IndentedSideLinesState();
@@ -649,14 +656,15 @@ class _IndentedSideLinesState extends State<_IndentedSideLines> {
             firstMoveKey: _keys[i],
             initialPath: widget.initialPath,
             params: widget.params,
+            nesting: widget.nesting,
           ),
         )
         .toList();
 
-    const padding = 12.0;
+    final padding = widget.nesting < 6 ? 12.0 : 0.0;
     return RepaintBoundary(
       child: Padding(
-        padding: const EdgeInsets.only(left: padding),
+        padding: EdgeInsets.only(left: padding),
         child: CustomPaint(
           painter: _IndentPainter(
             sideLineStartPositions: _sideLineStartPositions,
--- a/lib/src/view/analysis/tree_view.dart
+++ b/lib/src/view/analysis/tree_view.dart
@@ -223,9 +223,7 @@ class _PgnTreeView extends StatelessWidget {
               }
 
               // Skip the first node which is the continuation of the mainline
-              final sidelineNodes = nodes.last.children.skip(1).whereNot(
-                    (node) => node.isHidden,
-                  );
+              final sidelineNodes = nodes.last.children.skip(1);
 
               return [
                 _MainLinePart(
@@ -509,8 +507,7 @@ class _SideLines extends StatelessWidget {
       sidelineNodes.add(sidelineNodes.last.children.first);
     }
 
-    final children =
-        sidelineNodes.last.children.whereNot((node) => node.isHidden);
+    final children = sidelineNodes.last.children;
 
     return Column(
       crossAxisAlignment: CrossAxisAlignment.start,
@@ -646,9 +643,16 @@ class _IndentedSideLinesState extends State<_IndentedSideLines> {
 
   @override
   Widget build(BuildContext context) {
-    _keys = List.generate(widget.sideLines.length, (_) => GlobalKey());
+    final hasHiddenLines = widget.sideLines.any((node) => node.isHidden);
 
-    final sideLineWidgets = widget.sideLines
+    final visibleSideLines = widget.sideLines.whereNot((node) => node.isHidden);
+
+    _keys = List.generate(
+      visibleSideLines.length + (hasHiddenLines ? 1 : 0),
+      (_) => GlobalKey(),
+    );
+
+    final sideLineWidgets = visibleSideLines
         .mapIndexed(
           (i, firstSidelineNode) => _SideLines(
             firstNode: firstSidelineNode,
@@ -674,7 +678,23 @@ class _IndentedSideLinesState extends State<_IndentedSideLines> {
           child: Column(
             key: _columnKey,
             crossAxisAlignment: CrossAxisAlignment.start,
-            children: sideLineWidgets,
+            children: [
+              ...sideLineWidgets,
+              if (hasHiddenLines)
+                GestureDetector(
+                  child: Icon(
+                    Icons.add_box,
+                    color: _textColor(context, 0.6),
+                    key: _keys.last,
+                    size: _baseTextStyle.fontSize,
+                  ),
+                  onTap: () {
+                    widget.params
+                        .notifier()
+                        .expandVariations(widget.initialPath);
+                  },
+                ),
+            ],
           ),
         ),
       ),
@@ -917,17 +937,11 @@ class _MoveContextMenu extends ConsumerWidget {
             ),
           ),
         const PlatformDivider(indent: 0),
-        if (parent.children.any((c) => c.isHidden))
-          BottomSheetContextMenuAction(
-            icon: Icons.subtitles,
-            child: Text(context.l10n.mobileShowVariations),
-            onPressed: () => notifier().showAllVariations(path),
-          ),
         if (isSideline) ...[
           BottomSheetContextMenuAction(
             icon: Icons.subtitles_off,
-            child: Text(context.l10n.mobileHideVariation),
-            onPressed: () => notifier().hideVariation(path),
+            child: Text(context.l10n.collapseVariations),
+            onPressed: () => notifier().collapseVariations(path),
           ),
           BottomSheetContextMenuAction(
             icon: Icons.expand_less,
--- a/lib/src/view/analysis/tree_view.dart
+++ b/lib/src/view/analysis/tree_view.dart
@@ -615,6 +615,10 @@ class _IndentedSideLinesState extends State<_IndentedSideLines> {
   final GlobalKey _columnKey = GlobalKey();
 
   void _redrawIndents() {
+    _keys = List.generate(
+      _visibleSideLines().length + (_hasHiddenLines() ? 1 : 0),
+      (_) => GlobalKey(),
+    );
     WidgetsBinding.instance.addPostFrameCallback((_) {
       final RenderBox? columnBox =
           _columnKey.currentContext?.findRenderObject() as RenderBox?;
@@ -635,6 +639,11 @@ class _IndentedSideLinesState extends State<_IndentedSideLines> {
     });
   }
 
+  bool _hasHiddenLines() => widget.sideLines.any((node) => node.isHidden);
+
+  Iterable<ViewBranch> _visibleSideLines() =>
+      widget.sideLines.whereNot((node) => node.isHidden);
+
   @override
   void initState() {
     super.initState();
@@ -651,16 +660,7 @@ class _IndentedSideLinesState extends State<_IndentedSideLines> {
 
   @override
   Widget build(BuildContext context) {
-    final hasHiddenLines = widget.sideLines.any((node) => node.isHidden);
-
-    final visibleSideLines = widget.sideLines.whereNot((node) => node.isHidden);
-
-    _keys = List.generate(
-      visibleSideLines.length + (hasHiddenLines ? 1 : 0),
-      (_) => GlobalKey(),
-    );
-
-    final sideLineWidgets = visibleSideLines
+    final sideLineWidgets = _visibleSideLines()
         .mapIndexed(
           (i, firstSidelineNode) => _SideLines(
             firstNode: firstSidelineNode,
@@ -688,7 +688,7 @@ class _IndentedSideLinesState extends State<_IndentedSideLines> {
             crossAxisAlignment: CrossAxisAlignment.start,
             children: [
               ...sideLineWidgets,
-              if (hasHiddenLines)
+              if (_hasHiddenLines())
                 GestureDetector(
                   child: Icon(
                     Icons.add_box,
--- a/lib/src/view/analysis/tree_view.dart
+++ b/lib/src/view/analysis/tree_view.dart
@@ -686,7 +686,7 @@ class _IndentedSideLinesState extends State<_IndentedSideLines> {
                     Icons.add_box,
                     color: _textColor(context, 0.6),
                     key: _keys.last,
-                    size: _baseTextStyle.fontSize,
+                    size: _baseTextStyle.fontSize! + 5,
                   ),
                   onTap: () {
                     widget.params
--- a/lib/src/view/analysis/tree_view.dart
+++ b/lib/src/view/analysis/tree_view.dart
@@ -294,7 +294,7 @@ class _PgnTreeViewState extends State<_PgnTreeView> {
             containsCurrentMove: containsCurrentMove,
           );
         } else {
-          // Avoid expensive rebuilds by caching parts of the tree that did not change across a path change
+          // Avoid expensive rebuilds ([State.build]) of the entire PGN tree by caching parts of the tree that did not change across a path change
           return subtrees[i];
         }
       },
--- a/lib/src/view/analysis/tree_view.dart
+++ b/lib/src/view/analysis/tree_view.dart
@@ -203,7 +203,7 @@ class _PgnTreeView extends StatelessWidget {
               rootComments?.any((c) => c.text?.isNotEmpty == true) == true)
             Text.rich(
               TextSpan(
-                children: comments(rootComments!, textStyle: _baseTextStyle),
+                children: _comments(rootComments!, textStyle: _baseTextStyle),
               ),
             ),
           ..._mainlineParts(root).map(
@@ -278,7 +278,7 @@ List<InlineSpan> _buildInlineSideLine({
               style: textStyle,
             ),
           ],
-          ...moveWithComment(
+          ..._moveWithComment(
             node,
             parent: i == 0 ? parent : sidelineNodes[i - 1],
             lineInfo: (
@@ -322,7 +322,7 @@ enum _LineType {
 /// Metadata about a move's role in the tree view.
 typedef _LineInfo = ({_LineType type, bool startLine});
 
-List<InlineSpan> moveWithComment(
+List<InlineSpan> _moveWithComment(
   ViewBranch branch, {
   required ViewNode parent,
   required TextStyle textStyle,
@@ -345,7 +345,7 @@ List<InlineSpan> moveWithComment(
       ),
     ),
     if (params.shouldShowComments && branch.hasTextComment)
-      ...comments(branch.comments!, textStyle: textStyle),
+      ..._comments(branch.comments!, textStyle: textStyle),
   ];
 }
 
@@ -377,7 +377,7 @@ class _SideLinePart extends ConsumerWidget {
 
     var path = initialPath + nodes.first.id;
     final moves = [
-      ...moveWithComment(
+      ..._moveWithComment(
         nodes.first,
         parent: parent,
         lineInfo: (
@@ -392,7 +392,7 @@ class _SideLinePart extends ConsumerWidget {
       ...nodes.take(nodes.length - 1).map(
         (node) {
           final moves = [
-            ...moveWithComment(
+            ..._moveWithComment(
               node.children.first,
               parent: node,
               lineInfo: (
@@ -456,7 +456,7 @@ class _MainLinePart extends ConsumerWidget {
               (i, node) {
                 final mainlineNode = node.children.first;
                 final moves = [
-                  moveWithComment(
+                  _moveWithComment(
                     mainlineNode,
                     parent: node,
                     lineInfo: (
@@ -972,7 +972,7 @@ class _MoveContextMenu extends ConsumerWidget {
   }
 }
 
-List<TextSpan> comments(
+List<TextSpan> _comments(
   IList<PgnComment> comments, {
   required TextStyle textStyle,
 }) =>
--- a/lib/src/view/analysis/tree_view.dart
+++ b/lib/src/view/analysis/tree_view.dart
@@ -44,12 +44,12 @@ class AnalysisTreeView extends ConsumerStatefulWidget {
 class _InlineTreeViewState extends ConsumerState<AnalysisTreeView> {
   final currentMoveKey = GlobalKey();
   final _debounce = Debouncer(kFastReplayDebounceDelay);
-  late UciPath currentPath;
+  late UciPath pathToCurrentMove;
 
   @override
   void initState() {
     super.initState();
-    currentPath = ref.read(
+    pathToCurrentMove = ref.read(
       analysisControllerProvider(widget.pgn, widget.options).select(
         (value) => value.currentPath,
       ),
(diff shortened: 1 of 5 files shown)


tom-anders:
(The main trick here that improves the wrapping of comments etc. is to use `Text.rich` to build the lines, instead of the `Wrap` widget)

ijm8710:
This is very cool and agree it looks much better Tom! Thank you! Would you be interested in looking into the two column vertical notation at some point as well? I know it's a different topic but while you're testing notation edge cases, you'll prob be the best man for the job 😀

veloce:
Well thanks for this! I know it is tricky to deal with.

Quick review for now: I like it a lot, especially the way you handle sidelines and nesting. It it much more clear now.

Making it look like the web UI, is a non goal to me, as I remind often. The app should have is own visual identity. Of course when the design is better in web we should copy it, as you did here. But we can also vary on some point where it makes sense.

One thing I'm not sure of, for instance, are the italics. Looks like the rule in lichess website is to have an inside sideline italic. Why not... But I think I actually prefer having the comments in italic and keep the inline sideline as is (with a smaller font and a slight opacity). 

I'll review this carefully after I merge the challenges branch (which should be soon now).

tom-anders:
> Well thanks for this! I know it is tricky to deal with.

Yeah I stopped counting the knots this put into my brain :D

> Making it look like the web UI, is a non goal to me, as I remind often. The app should have is own visual identity. Of course when the design is better in web we should copy it, as you did here. But we can also vary on some point where it makes sense.
> One thing I'm not sure of, for instance, are the italics. Looks like the rule in lichess website is to have an inside sideline italic. Why not... But I think I actually prefer having the comments in italic and keep the inline sideline as is (with a smaller font and a slight opacity).

Yeah I get a that, I can change it. Using the Web UI as a reference, also for font styles etc. was just more convenient while developing/testing, but yeah no strong opinions on that, I'll adjust the style (or maybe it's even easier if you do it after merging the PR)

veloce:
This screen also supports PGN import from random sources. The tree view should look good also in that case.

Could you post a screenshot of the analysis tree when importing this PGN @tom-anders ? https://github.com/lichess-org/dartchess/blob/65fad86b26618785287152e437a7c80efb94b058/data/wcc_2023.pgn#L1-L39

Thanks!

tom-anders:
> This screen also supports PGN import from random sources. The tree view should look good also in that case.
> 
> Could you post a screenshot of the analysis tree when importing this PGN @tom-anders ? https://github.com/lichess-org/dartchess/blob/65fad86b26618785287152e437a7c80efb94b058/data/wcc_2023.pgn#L1-L39
> 
> Thanks!

Huh, fun fact (or not fun), I tried importing them into a study via the web interface for comparison, but parsing failed:

Screenshot_2024-10-02_21-48-51

tom-anders:
Thanks for that PGN, it uncovered a small bug when building sidelines fixed now. I attached a video where I'm scrolling through the first game of the PGN. 

Note that there are some weird looking line break in some comments, but that's not a wrapping issue, those line breaks are actually there in the source PGN you posted. I'd assume that usually a PGN comment will only have intentional line breaks...?

[treeview.webm](https://github.com/user-attachments/assets/e9279a1d-3971-4190-a7f7-f433ebafa86a)

veloce:
Thanks for the vid!
Not sure why there are line breaks in that PGN, let's ignore them.

I find it way more readable in your version, great job!

There's one difference with what I did, the comments are shown in full length, where before only the first 5 lines where shown, and you'd tap on a comment to see the full version in a bottom sheet. I don't know if that change was intentional, but I actually like it, so let's keep it that way!

veloce:
One question, have you put a limit on nesting?

I suppose we should have one. Or at least prevent further indentation when reaching a nesting threshold.

tom-anders:
> There's one difference with what I did, the comments are shown in full length, where before only the first 5 lines where shown, and you'd tap on a comment to see the full version in a bottom sheet. I don't know if that change was intentional, but I actually like it, so let's keep it that way!

Yeah that was intentional, especially for heavily commented PGNs and studies I thought that it would break your flow if you had to keep tapping comments to read them. 

> One question, have you put a limit on nesting?
> 
> I suppose we should have one. Or at least prevent further indentation when reaching a nesting threshold.

Ah yes good point, haven't implemented that yet - what would be a good threshold after which we stop increasing indentation? Something like 4 maybe? 

Also, we could by default hide deeply nested variations and display a + Icon instead like the website does, what do you think about that?

veloce:
> Ah yes good point, haven't implemented that yet - what would be a good threshold after which we stop increasing indentation? Something like 4 maybe?

No more than 4 seems right indeed

> Also, we could by default hide deeply nested variations and display a + Icon instead like the website does, what do you think about that?

👍

tom-anders:
> > Also, we could by default hide deeply nested variations and display a + Icon instead like the website does, what do you think about that?
> 
> 👍

Ok I checked the website, it's actually pretty aggressive when it comes to hiding variations: Only two levels of nesting are displayed by default, anything that's three or more levels is hidden by default.

I guess unless we have a good reason not to, let's do the same for now..?

tom-anders:
Okay, added an indentation limit, collapse nesting>2 by default and added the plus-icon to expand variations. IMO this makes the "expand variations" action in the move context menu redundant, so I removed it.

"Hide variations" works a little bit different now, it used to only hide the variation you selected the move from, but now it hides all other sibling variations as well. I think it's more useful and intuitive this way, but curious to know what others think

[treeview.webm](https://github.com/user-attachments/assets/7c4856ce-b214-4ad4-aa0e-264281da3161)

veloce:
Great work! Yes, I think it should hide all variations. 

I wonder if the "plus" button is easy to tap? Seems a bit small.

Also can you post again the screen recording with the WCC PGN with this new version? thanks!

tom-anders:
Yeah you're right, I made the icon bigger now. New video:

[treeview.webm](https://github.com/user-attachments/assets/7e7a31d6-1880-4c14-b134-83e06408e48e)

lib/src/view/analysis/tree_view.dart
```diff
@@ -162,124 +131,580 @@ class _InlineTreeViewState extends ConsumerState<AnalysisTreeView> {
...
+              notifier: () => ref.read(ctrlProvider.notifier),
             ),
           ),
         ),
       ],
     );
   }
+}
+
+// A group of parameters that are passed through various parts of the tree view
```
veloce, line 151: Even for private members we should use the triple slash comments (with a first short line and empty line and description below). It improves code readability a lot, for instance in vim I press "K" and I see the code doc.

tom-anders: Ah, a fellow vim user I see! 

Yeah that was an oversight, will fix

lib/src/view/analysis/tree_view.dart
```diff
@@ -162,124 +131,580 @@ class _InlineTreeViewState extends ConsumerState<AnalysisTreeView> {
...
+    }
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    final hasHiddenLines = widget.sideLines.any((node) => node.isHidden);
+
+    final visibleSideLines = widget.sideLines.whereNot((node) => node.isHidden);
+
+    _keys = List.generate(
```
veloce, line 650: Doing side effects in `build` is discouraged. Here it looks it can be done in `initState`.

veloce (commented):
There is a lot of changes and this is complex code. I know this part is not much tested yet, but I think it would be nice to add some tests to `analysis_screen_test` relating to what you changed.

Tests don't have to be extra complex first. Using simple PGN with variations, add some tests that check a few basic things:
- when the variation is indented
- when it is inline
- when it is collapsed

It important to start testing this because if later we have bug fixes to make in this tricky screen we'll be happy to have tests. (I shall add more tests for the rest of the analysis feature myself).

One last thing: this is a performance sensitive block. Have you tested the performance using a profile build on a real device? Using large trees to compare with before?

lib/src/view/analysis/tree_view.dart
```diff
@@ -162,124 +131,580 @@ class _InlineTreeViewState extends ConsumerState<AnalysisTreeView> {
...
+}
+
+// A group of parameters that are passed through various parts of the tree view
+// and ultimately evaluated in the InlineMove widget. Grouped in this record to improve readability.
+typedef _PgnTreeViewParams = ({
+  UciPath currentPath,
+  bool shouldShowAnnotations,
+  bool shouldShowComments,
+  GlobalKey currentMoveKey,
+  AnalysisController Function() notifier,
```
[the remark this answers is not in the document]
tom-anders: hmm me neither, not sure why I did this, will remove the function

lib/src/view/analysis/tree_view.dart
tom-anders: Ah, I initially had this line in `initState`, but then the indent lines would get messed up when hiding/collapsing variations - The solution is to update `_keys` in `didUpdateWidget` as well

lib/src/view/analysis/tree_view.dart
```diff
@@ -162,124 +131,580 @@ class _InlineTreeViewState extends ConsumerState<AnalysisTreeView> {
...
+            firstMoveKey: _keys[i],
+            initialPath: widget.initialPath,
+            params: widget.params,
+            nesting: widget.nesting,
+          ),
+        )
+        .toList();
+
+    final padding = widget.nesting < 6 ? 12.0 : 0.0;
+    return RepaintBoundary(
```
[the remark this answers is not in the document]
tom-anders: Without the `RepaintBoundary`, `paint()` is called repeatedly when scrolling (even though it's not necessary to repaint)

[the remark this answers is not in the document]
tom-anders: (added a comment)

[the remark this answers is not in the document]
tom-anders: Or maybe this is not a problem? I thought it's probably better for performance this way

tom-anders:
Thanks for the review!

> There is a lot of changes and this is complex code. I know this part is not much tested yet, but I think it would be nice to add some tests to `analysis_screen_test` relating to what you changed.
> 
> Tests don't have to be extra complex first. Using simple PGN with variations, add some tests that check a few basic things:
> 
>     * when the variation is indented
> 
>     * when it is inline
> 
>     * when it is collapsed
> 
>     * ... etc

Sure, I'll add some more tests :+1: 

> One last thing: this is a performance sensitive block. Have you tested the performance using a profile build on a real device? Using large trees to compare with before?

Will do some before/after tests today, the WCC PGN is probably a good test here again

tom-anders:
I'll plan to reuse the new `_PgnTreeView` for the study feature, so maybe it's better to make it public, move it to a new file and test it in isolation? what do you think?

tom-anders:
Okay, so I did some profiling, by loading the WCC PGN, selecting the first move (e4) and then pressing the "Next" Button, which should rebuild the entire tree view. In general, for both the `main` branch and this PR I observed slight lag, but note that my device has a pretty slow CPU, so I'm seeing lag in a lot of other apps as well. 

Note that I did **not** enable stockfish for these tests (as that would lead to pretty much constant lag anyway)

On the `main` branch, the frame that corresponds to drawing the tree widget takes about 50ms on the UI thread.

With this PR, it instead takes 100-120ms, followed by the frame that draws the indent lines, taking an additional 50-60ms

So yeah, it's a significant performance hit for very large trees like this.

However, I also tested with https://github.com/lichess-org/dartchess/blob/65fad86b26618785287152e437a7c80efb94b058/data/lichess-bullet-game.pgn, which has mostly inline sidelines (but also two indented sidelines due to their length).

With that PGN, `main` takes 15-16ms to render the tree view, while this PR takes 20-22ms, with the next frame taking 6ms to render the indent guides.

tom-anders:
Performance update: I'm working on implementing an optimization that only rebuild the part of the mainline (and it's sidelines) that actually changed. Still need to fix some small bugs, but with that fix in place rendering the tree of the WCC PGN now only takes 15-30ms (and rendering the indents becomes negligible) :tada:

veloce:
> I plan to reuse the new `_PgnTreeView` for the study feature, so maybe it's better to make it public, move it to a new file and test it in isolation? what do you think?

Ideally we'd include the debouncing and auto-scroll logic, but the debouncing relies on analysis controller. We'd need to refactor to have a common interface for analysis and study (and broadcast).
I think it's a lot more work to do this refactoring. Also I see that `_PgnTreeView` has`AnalysisController` in its parameter, so it shouldn't be compatible with study yet?

Perhaps the easiest for now is to add the tests in the analysis_screen_test (so depending on the analysis controller), and we'll move the tests later when the common interface is done.

veloce:
> Performance update: I've implemented an optimization that only rebuilds the subtree of the mainline that actually changed. With this optimization in place rendering the tree of the WCC PGN now only takes 15ms when switching a move within the same mainline subtree, and 30ms when switching the current move from one mainline subtree to another one (since then we need to redraw both the old and the new subtree in that case). Rendering the indents becomes pretty much negligible now. 🎉
> 
> Pretty impressive that with this optimization performance is even better than with the old code

Amazing! Great idea, I hadn't thought about it.

lib/src/view/analysis/tree_view.dart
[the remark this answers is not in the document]
veloce: You mean the `_IndentPainter.paint` specifically right?

From flutter doc, on `RepaintBoundary`:
> This widget creates a separate display list for its child, which can improve performance if the subtree repaints at different times than the surrounding parts of the tree.

I don't know if it applies specifically to the indent line widget here. Looks to me there's no reason the indented line subtree would repaint at different times than the surrounding?

tom-anders:
> > I plan to reuse the new `_PgnTreeView` for the study feature, so maybe it's better to make it public, move it to a new file and test it in isolation? what do you think?
> 
> Ideally we'd include the debouncing and auto-scroll logic, but the debouncing relies on analysis controller. We'd need to refactor to have a common interface for analysis and study (and broadcast). I think it's a lot more work to do this refactoring. Also I see that `_PgnTreeView` has`AnalysisController` in its parameter, so it shouldn't be compatible with study yet?

Oh I forgot that I've already written an interface to make the `_PgnTreeView` independent from `AnalysisController` on my study branch, but not on this branch. For `_PgnTreeView` it's easy, because it only calls notifier methods, but does not have any `ref.watch` in it. But also factoring out the debouncing logic is trickier.

So yeah, let's put it into the `analysis_screen_test.dart` for now and refactor in a follow-up PR.

lib/src/view/analysis/tree_view.dart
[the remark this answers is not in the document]
tom-anders: Hmm, I'm not 100% what's going on, but currently neither `_IndentPainter.paint()` nor `_IndentedSidelinesState.build()` are called when scrolling. But when removing `RepaintBoundary`, `paint()` gets called, but `_IndentedSidelinesState.build()` doesn't. 

I think https://stackoverflow.com/a/66430673 explains it? 

```
By default CustomPainter is in the same layer as every other widget on the same screen so it's paint method will get called if any other widget on the same screen repaint
```

lib/src/view/analysis/tree_view.dart
```diff
@@ -162,124 +131,654 @@ class _InlineTreeViewState extends ConsumerState<AnalysisTreeView> {
...
+  State<_PgnTreeView> createState() => _PgnTreeViewState();
+}
+
+typedef _TreeViewPart = ({
+  List<Widget> lines,
+  bool containsCurrentMove,
+});
+
+class _PgnTreeViewState extends State<_PgnTreeView> {
+  List<List<ViewNode>> mainlineParts = [];
```
veloce, line 201: As I'm reading this for the first time, I'm not sure to understand exactly what `mainlineParts` and `treeParts` are for. The types are not enough to self explain.

Some comments explaining the logic would be welcome where needed imo.

lib/src/view/analysis/tree_view.dart
```diff
@@ -162,124 +131,654 @@ class _InlineTreeViewState extends ConsumerState<AnalysisTreeView> {
...
+
+  void _updateLines({required bool fullRebuild}) {
+    setState(() {
+      if (fullRebuild) {
+        mainlineParts = _mainlineParts(widget.root).toList();
+      }
+
+      var path = UciPath.empty;
+
+      treeParts = mainlineParts.mapIndexed(
```
veloce, line 223: There's a big block of code now in the `setState`; Readability would be improved if the code that generates new `treeParts` was externalised in another function.

lib/src/view/analysis/tree_view.dart
```diff
@@ -162,124 +131,654 @@ class _InlineTreeViewState extends ConsumerState<AnalysisTreeView> {
...
+typedef _LineInfo = ({_LineType type, bool startLine});
+
+List<InlineSpan> _moveWithComment(
+  ViewBranch branch, {
+  required ViewNode parent,
+  required TextStyle textStyle,
+  required _LineInfo lineInfo,
+  required UciPath pathToNode,
+  required _PgnTreeViewParams params,
+  GlobalKey? moveKey,
```
veloce, line 398: Not sure to understand what this `moveKey` is? Is it the current move key? If so it'd be better named `currentMoveKey`.

lib/src/view/analysis/tree_view.dart
```diff
@@ -162,124 +131,654 @@ class _InlineTreeViewState extends ConsumerState<AnalysisTreeView> {
...
+    required this.params,
+  }) : assert(nodes.isNotEmpty);
+
+  final List<ViewBranch> nodes;
+
+  final ViewNode parent;
+
+  final UciPath initialPath;
+
+  final GlobalKey firstMoveKey;
```
veloce, line 433: Why is there a need for a `GlobalKey` for the first move? I don't get it.

lib/src/view/analysis/tree_view.dart
```diff
@@ -162,124 +131,654 @@ class _InlineTreeViewState extends ConsumerState<AnalysisTreeView> {
...
+        textStyle: textStyle,
+        params: params,
       ),
+    ),
+    if (params.shouldShowComments && branch.hasTextComment)
+      ..._comments(branch.comments!, textStyle: textStyle),
+  ];
+}
+
+class _SideLinePart extends ConsumerWidget {
```
veloce, line 418: I'd add a doc for this class to improve readability. What is it? a part of a sideline?

lib/src/view/analysis/tree_view.dart
```diff
@@ -162,124 +131,654 @@ class _InlineTreeViewState extends ConsumerState<AnalysisTreeView> {
...
+
+    return Text.rich(
+      TextSpan(
+        children: moves,
+      ),
+    );
+  }
+}
+
+class _MainLinePart extends ConsumerWidget {
```
veloce, line 497: Class doc here too.

lib/src/view/analysis/tree_view.dart
```diff
@@ -162,124 +131,654 @@ class _InlineTreeViewState extends ConsumerState<AnalysisTreeView> {
...
+              },
+            )
+            .flattened
+            .toList(),
+      ),
+    );
+  }
+}
+
+class _SideLines extends StatelessWidget {
```
veloce, line 559: Class doc here too.

lib/src/view/analysis/tree_view.dart
```diff
@@ -162,124 +131,654 @@ class _InlineTreeViewState extends ConsumerState<AnalysisTreeView> {
...
-        initialPath: newPath,
-      ),
+  @override
+  bool shouldRepaint(_IndentPainter oldDelegate) {
+    return oldDelegate.sideLineStartPositions != sideLineStartPositions ||
+        oldDelegate.color != color;
+  }
+}
+
+class _IndentedSideLines extends StatefulWidget {
```
veloce, line 653: Class doc

lib/src/view/analysis/tree_view.dart
```diff
@@ -162,124 +131,654 @@ class _InlineTreeViewState extends ConsumerState<AnalysisTreeView> {
...
+  final _PgnTreeViewParams params;
+
+  final int nesting;
+
+  @override
+  State<_IndentedSideLines> createState() => _IndentedSideLinesState();
+}
+
+class _IndentedSideLinesState extends State<_IndentedSideLines> {
+  late List<GlobalKey> _keys;
```
veloce, line 677: In flutter we tend to avoid using `GlobalKey` unless having a good reason. Here I'm a little worried about this list of global keys.

I understand you need them to get the position of the sideline to get the proper offset. I think there is a way (no much harder to code I'd say) to do with without all the global keys. The indent width is a fixed one, so in theory you'd need to do sth like:
```dart
final indentOffset = initialOffset + Offset(nestingLevel * fixedIndentWidth, 0)`;
```
So you'd need only one global key to get the initial offset of the tree view, but not more.

lib/src/view/analysis/tree_view.dart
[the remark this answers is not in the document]
veloce: Does the removal of `RepaintBoundary` cause a performance issue? If not let's remove it.

lib/src/view/analysis/tree_view.dart
tom-anders: This is needed so that in `_IndentedSideLinesState` we can query the starting positions of the sidelines and draw the indent lines correctly.

lib/src/view/analysis/tree_view.dart
tom-anders: Well, it's only the `currentMoveKey` if this is the current move :D 

but I could simply make this a bool and let the InlineMove assign the key by itself instead, that would probably be more readable

tom-anders: Oh wait, I get confused as well, yeah this definitely needs a comment. This is `firstMoveKey` if it's the start of the sideline (so that we can draw the indent guides) and null otherwise

lib/src/view/analysis/tree_view.dart
tom-anders: The problem is that this only works if the sidelines do not wrap. As soon as a sidelines wraps, we don't know the vertical positions of all following sidelines.

(Or maybe I did not understand your suggestions correctly)

tom-anders:
Sorry for the general lack of comments, with complex pieces of code like this that had many iterations until it ended up in its final state, I often find it hard to determine which parts need comments and which don't, especially for non-public classes (since at that point in your head everything seems obvious). 

Added some comments and also renamed stuff, let me know if anything is unclear still.

lib/src/view/analysis/tree_view.dart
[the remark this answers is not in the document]
tom-anders: I checked, scrolling performance seems to be fine without it, so I removed it

lib/src/view/analysis/tree_view.dart
```diff
@@ -162,124 +131,696 @@ class _InlineTreeViewState extends ConsumerState<AnalysisTreeView> {
...
+                    parent: mainlineNodes.last,
+                    params: widget.params,
+                    initialPath: sidelineInitialPath,
+                    nesting: 1,
+                  )
+                : null,
+            containsCurrentMove: containsCurrentMove,
+          );
+        } else {
+          // Avoid expensive rebuilds by caching parts of the tree that did not change across a path change
```
veloce, line 297: One might get confused here: by "expensive rebuilds" you mean actually the framework's `State.build`, but this comment is inside a method named `_rebuildChangedSubtrees`. I think it is worth the precision in the comment.

lib/src/view/analysis/tree_view.dart
veloce: I suspected something like that, but I had to be sure to understand so pardon me for the question :)

As long as there's no performance issue, this list of global keys shouldn't cause any problem as the scope is limited.

test/view/analysis/analysis_screen_test.dart
```diff
@@ -92,13 +99,160 @@ void main() {
...
+        );
+        expect(mainline, findsOneWidget);
+
+        expectSameLine(tester, ['1. e4', 'e5', '1… d5', '2. exd5', '2. Nf3']);
+      });
+
+      testWidgets('displays long sideline on its own line', (tester) async {
+        await buildTree(
+          tester,
+          '1. e4 e5 (1... d5 2. exd5 Qxd5 3. Nc3 Qd8 4. d4 Nf6) Nc3 *',
```
veloce, line 190: Looks like it's missing the `2. ` before `Nc3`.

veloce (commented):
Much better with the documentation! Now I think I understand the logic :)
And thanks for the tests.

lib/src/view/analysis/tree_view.dart
```diff
@@ -162,124 +131,696 @@ class _InlineTreeViewState extends ConsumerState<AnalysisTreeView> {
...
+  /// This is nullable since the very last mainline part might not have any sidelines.
+  _IndentedSideLines? sidelines,
+  bool containsCurrentMove,
+});
+
+class _PgnTreeViewState extends State<_PgnTreeView> {
+  /// Caches the result of [_mainlineParts] to avoid recalculating it on every rebuild.
+  List<List<ViewNode>> mainlineParts = [];
+
+  /// Caches the top-level subtrees, where each subtree is a [_MainLinePart] and its sidelines.
```
veloce, line 234: We could add a precision here: this cache is a `Widget` cache obtained from the framework `build` method. The framework will skip `build` of each subtree where the widget reference is the same, this is the real perf optimisation.

Since the other cache above (`mainlineParts`) is not a widget cache it would be good to add this precision to avoid confusion.

lib/src/view/analysis/tree_view.dart
tom-anders: Yeah no problem, I went back and forth about this many times in my head as well, but in the end the problem always was "what if the line wraps though?"

test/view/analysis/analysis_screen_test.dart
tom-anders: ah indeed, but looks like our PGN parsing logic is fine with that? But I changed it now

PR #1057, opened 2024-09-29
feat: add detail view to game history screen

### LINKED ISSUE #947: [Suggestion] Add played opening to "Recent Games"
Short anecdote to give some motivation for this feature: A few days ago, I played a few games of Rapid. As always, in one of the games I got completely caught off-guard by the French Defense. The next day, I remembered that I wanted to analyze the French defense game, so I went to the new app and tried to find it in my recent games. In the old app, this was very easy, as it displayed the played opening (and even a thumbnail of the final position) for each played game. The new app currently does not have this information, making it hard to find a specific game sometimes.

So I'd propose to add the opening info to recent games. I can understand that the board thumbnail would probably take away too much space, but I think we could at least add the opening while still keeping the entries rather compact. As always, I'd be happy to have a go at implementing this, if you think it's a feature we should have :)

tom-anders:
Had this lying around for a while, got the time today to include the feedback I got on the original draft I posted in the issue. Closes #947

Screenshot_1727612203


ijm8710:
Hmm excited for the accurscy number to have a way back in but I prefer your old draft in some other other places

I still think the actual variant/time should show and so I preferred the time/game type display of before

And the opponent name seems bigger? Especially if you try to show rating change the huge name with just the opponent and not yourself probably makes that a harder obstacle?

veloce:
It looks quite nice!! thank you. I'd see the accuracy in even smaller font and a little more transparent. I know some people here really want to see it, but it is least important item here, so should have less emphasis while still being here.

Looks like you implemented a grid view as well, judging by the icon. Can we see the screenshot for it as well please?

tom-anders:
> It looks quite nice!! thank you. I'd see the accuracy in even smaller font and a little more transparent. I know some people here really want to see it, but it is least important item here, so should have less emphasis while still being here.

Thanks, will take a look 

> Looks like you also implemented a grid view, judging by the icon. Can we see the screenshot for it as well please?

Oh, that's a bug actually, it should show a different icon. Pressing that icon/button will switch the view to the "old" list view

veloce:
Oh, I thought it would be a grid view like in the puzzle history. Could make sense by the way, it would show more games at once. But not needed in this PR I guess.

I'm not sure about having the switch icon to see the old list. We probably should display the new list only (perhaps keeping the old list preview on home page only).

julien4215:
Would it make sense to keep both views ? Since the list view is more compact, it can be faster to find a game that is far in the history and you can also glance more games at once.

veloce:
I don't know, need to think about it. I don't like adding too much toggles and settings. We're fetching the same number of games per page, so it is just a matter of vertical space. If the new view allows to quickly see the important information (player name, end position), it could be that the new view would allow to find even more quickly the games.

ijm8710:
Just a minor example but one of a few is it like on homepage when scanning an opponents games you can see a day ago, 3 days ago et 

But when you dive into this detailed view its time stamped. 

If all views including the first 10 on homepage were just timestamp you have to think more to be like is October 2nd a day ago, few days ago etc

Actually another idea here:
I prefer if it said 20 hours ago etc (but clicking that turns it into the actual time)!

I still think there is a bit more refinement that can be done before this becomes the official game list.
Still hoping we can see rating change
And I think it's worthwhile info to show what the variant was (ie 5+0)

tom-anders:
A note about the time stamps: I agree that "3 days ago" is better than a full date, but for longer times I actually prefer the full date. "2023-07-24" is more helpful than "3 months" ago IMO. I think there should be some kind of cut off where we switch from "x ago" to the full date.

tom-anders:
Also, I really like how the old app says "Last Sunday" instead of "2 days ago", I think this is more helpful as well, but probably something for another PR

ijm8710:
What are your thoughts Tom on relative date that can be tapped to Change to actual date or with a popup.

I could have sworn GitHub used to do this. I think I've seen this with a few other apps.

tom-anders:
> What are your thoughts Tom on relative date that can be tapped to Change to actual date or with a popup.
> 
> I could have sworn GitHub used to do this. I think I've seen this with a few other apps.

idk, might be too much complexity, but not for me to decide

tom-anders:
Did some style adjustments, removed the old list view, added rating diffs:

Screenshot_1727816559

ijm8710:
Few thoughts,

Thanks for adding rating diff. I think the way opponent and your usernames now show does look better now.

For draw where rating does not change, I think it looks better in old app (rather than show +/- 0 which looks funky, they just omit the diff)

I also will push back once on Veloce regarding accuracy being even more downsized. You said yourself Veloce, a lot of players (including myself) care about it a lot. To be honest, I sometimes even care more about accuracy than win/loss. I think accuracy looked better slightly bigger in toms older mock. Understood this will stay as is unless you change your mind here Veloce but I think it deserves another shout since it's so minimized that it makes it hard to see at glance

As far as old app, I still prefer the extra line with 3+0•standard•rated and then time on line below
I mentioned I'd like to see the variant subtype (ie 3+0) but differentiating between standard vs 360 etc I think is necessary. Old app was able to fit it so I think it can work

Lastly, I would still heavily hope 1 of 2 things happens
- either current minimzie game list on home feed stays
- or have a toggle option to switch between minimized and detailed game view

ijm8710:
I also think I prefer from old all the subsequent line in the result (ie currently is says black resigned; in old app it wold say black resign, white is victorious).

It adds consistency to see what color won at a glance

veloce:
I'm not sure about having the rating diff in that screen. Is that really important?

Thing is now we have both players names displayed, where the only important information is the opponent's name (because we already know the player name). If we were not on constrained space (width) that would not be an issue. Here it works in the screenshot because player names are short, but longer names would be cut.

I think this screen is interesting to have, but since it is a bit cluttered, having a toggle to switch to the other simple list view would be nice indeed (we just need a different icon).

What about having a grid view as well (2-4 columns, board thumbnail with date, opponent's name and result icon)? Seems easy to do (we have it already for puzzles). We'd then need a button to switch between the 3 formats.

ijm8710:
Few things,

Just noticed (I think) in old app white is first, black second. Currently in this mock, it's "you" always first. I think the old app has better logic here

I don't quite often, he is usually spot on but I do disagree about Veloce about rating diff being too much. Is it possible to see what a photo would look like with max-length player names+master title? I feel like only a small shrink if any would be needed to account for all lengths

If all 3 views were shown which sounds fine, I still think this version should be default but with the  current condensed the default on homepage

tom-anders:
> I'm not sure about having the rating diff in that screen. Is that really important?

> Thing is now we have both players names displayed, where the only important information is the opponent's name (because we already know the player name). If we were not on constrained space (width) that would not be an issue. Here it works in the screenshot because player names are short, but longer names would be cut.

I guess only *your* rating diff is important, your opponent's not so much.
Maybe we could go back to only show the opponent name, and then on a separate line the rating diff? Would still be less clutter than the current version.

> I think this screen is interesting to have, but since it is a bit cluttered, having a toggle to switch to the other simple list view would be nice indeed (we just need a different icon).
> 
> What about having a grid view as well (2-4 columns, board thumbnail with date, opponent's name and result icon)? Seems easy to do (we have it already for puzzles). We'd then need a button to switch between the 3 formats.

Sure why not, we should then probably also remember the chosen view in the shared preferences, right?

tom-anders:
> I also think I prefer from old app the subsequent line in the result (ie currently is says black resigned; in old app it would say black resigned, white is victorious).
> 
> It adds consistency to see what color won at a glance for all entries going down

This would add one more line of clutter/noise though

ijm8710:
> > I also think I prefer from old app the subsequent line in the result (ie currently is says black resigned; in old app it would say black resigned, white is victorious).
> 
> > 
> 
> > It adds consistency to see what color won at a glance for all entries going down
> 
> 
> 
> This would add one more line of clutter/noise though

In old app, this seemed to always be one line.

ijm8710:
> Maybe we could go back to only show the opponent name, and then on a separate line the rating diff? Would still be less clutter than the current version.

I still think it looks a lot nicer in your most recent mock than the one before where YOUR name is included too
Especially if it dynamically orders to have white first
Worked fine in old app and if you're giving the choice of 3 views...
I really don't agree with Veloce on this one!

lib/src/view/game/game_list_detail_tile.dart
```diff
@@ -0,0 +1,306 @@
...
+
+    final dateStyle = TextStyle(
+      color: textShade(
+        context,
+        Styles.subtitleOpacity,
+      ),
+      fontSize: 12,
+    );
+
+    return GestureDetector(
```
veloce, line 44: The problem with the gesture detector, is that there is no feedback that the item has been tapped.

It is better replaced with a `PlatformInkWell` to handle `onTap`, and the detector can be used only for the long press gesture.

lib/src/view/game/game_list_detail_tile.dart
```diff
@@ -0,0 +1,306 @@
...
+        );
+      },
+      onTap: () => openGameScreen(item.game, item.pov, context),
+      child: Padding(
+        padding: const EdgeInsets.symmetric(horizontal: 16.0).add(
+          const EdgeInsets.only(bottom: 8.0),
+        ),
+        child: LayoutBuilder(
+          builder: (context, constraints) {
+            return IntrinsicHeight(
```
veloce, line 66: Why is `IntrisicHeight` needed here?

It should be avoided, as flutter doc says it is relatively expensive, so we probably don't want that in an infinite list.

tom-anders: hmm I copied this from https://github.com/lichess-org/mobile/blob/33904fcfb0daa944159bc7be1013cfa0a1db7bf9/lib/src/view/game/game_list_tile.dart#L138

I think it's needed because we set the `BoardThumbnail` size based on a fraction of the screen width?

veloce: TBH I don't remember why it was needed in the context menu 😅

tom-anders: Seems like it was used to make sure that the `Column` containing all the data has the same height as the board - But look like we can get rid of `IntrinsicHeight` and instead wrap the `Column` inside a `SizedBox(height: boardSize)` instead :tada:

tom-anders:
> I'm not sure about having the rating diff in that screen. Is that really important?
> 
> Thing is now we have both players names displayed, where the only important information is the opponent's name (because we already know the player name). If we were not on constrained space (width) that would not be an issue. Here it works in the screenshot because player names are short, but longer names would be cut.
> 
> I think this screen is interesting to have, but since it is a bit cluttered, having a toggle to switch to the other simple list view would be nice indeed (we just need a different icon).
> 
> What about having a grid view as well (2-4 columns, board thumbnail with date, opponent's name and result icon)? Seems easy to do (we have it already for puzzles). We'd then need a button to switch between the 3 formats.

Added the toggle to switch, but for now it's only between the new (detailed) view and the old one. I think grid view would be something for another PR, wdyt?

I kept both user names for now, but the player being white is now displayed on the left always. This way, it actually adds some value, because you can quickly see in which games you were white and in which games you were black (which wouldn't be so obvious if we only display the opponents name).

lichess usernames seem to be capped to 20 characters, so maybe we could simply have some logic that descreases the font size if the length is above 10 characters or so? This way we should be able to always fit both names onto the screen

tom-anders:
> lichess usernames seem to be capped to 20 characters, so maybe we could simply have some logic that descreases the font size if the length is above 10 characters or so? This way we should be able to always fit both names onto the screen

Added some logic for that, here's what it looks like if both players have the maximum name length:

Screenshot_2024-11-01_11-24-21

veloce:
> Added the toggle to switch, but for now it's only between the new (detailed) view and the old one. I think grid view would be something for another PR, wdyt?

sure and we should also not do it for now

veloce:
Sorry for the delay. Going to deal with it asap. I think I'll add the game bookmark feature at the same time because it is a wanted one.

tom-anders:
Fixed merge conflicts - @julien4215 Can you can have a short look that I didn't break anything related to bookmarks during the merge? thanks!

julien4215:
Sure I'll take a look.

By the way do we want to also add bookmarks to this view ? On the compact view you can bookmark either via the draggable scroll sheet or by sliding the tile. I am not sure if the slidable would look good on these bigger tiles.

julien4215:
Oh never mind I see that you already added bookmark to the detailed view. It looks good to me.

Just a small detail I noticed is that the top padding is missing but there is the bottom padding. The first board touches the app bar.

PR #1063, opened 2024-10-01
[study] refactor: make EngineLines independent of AnalysisController

tom-anders:
We'll want to reuse it for the study screen. It's easiest to remove the dependency from the widget for now (alternative would be to use an interface that both `AnalysisController` and `StudyController` derive from, but as discussed we won't do that for now)


veloce:
I have now an example of Notifier mixin: https://github.com/lichess-org/mobile/blob/ef9fc55fdc08a606f26a7069140e73a1308c3e21/lib/src/model/settings/preferences_storage.dart#L15

We might eventually use this approach to share engine logic in different controllers. I'm not sure but I suspect it is not that easy to do.

lib/src/view/engine/engine_lines.dart
```diff
@@ -0,0 +1,174 @@
...
+        ply.isOdd
+            ? '${(ply / 2).ceil()}. $s '
+            : i == 0
+                ? '${(ply / 2).ceil()}... $s '
+                : '$s ',
+      );
+      ply += 1;
+    });
+
+    final brightness = ref.watch(currentBrightnessProvider);
```
veloce, line 118: We should prefer the flutter way of getting brightness, this refactoring is the occasion to do it:
```suggestion
    final brightness = Theme.of(context).brightness;
```

veloce: I shall add that we need this provider because it is using the user preferences, but ideally it should be a private provider that only controls the brightness set in the `MaterialApp` widget.

tom-anders:
> I have now an example of Notifier mixin:
> 
> https://github.com/lichess-org/mobile/blob/ef9fc55fdc08a606f26a7069140e73a1308c3e21/lib/src/model/settings/preferences_storage.dart#L15
> 
> We might eventually use this approach to share engine logic in different controllers. I'm not sure but I suspect it is not that easy to do.

Ah that sounds very useful indeed, I'll see if I can use it in the study branch.

But I think for now it's safer to go with the callbacks, we can always refactor again later.

lib/src/view/engine/engine_lines.dart
```diff
@@ -0,0 +1,174 @@
...
+  const EngineLines({
+    required this.onTapMove,
+    required this.clientEval,
+    required this.isGameOver,
+    required this.isLandscape,
+  });
+  final void Function(NormalMove move) onTapMove;
+  final ClientEval? clientEval;
+  final bool isGameOver;
+  final bool isLandscape;
```
veloce, line 25: Since it is now a public widget, I'd use an enum in place of that boolean. We already have that one: https://github.com/lichess-org/mobile/blob/ef9fc55fdc08a606f26a7069140e73a1308c3e21/lib/src/view/engine/engine_gauge.dart#L20

tom-anders: Hmm, the only thing it affects is the padding, so maybe it's better to simply pass the padding as a parameter instead?

tom-anders: Removed the `isLandscape` and `Padding` entirely, the calling code should decide padding instead

PR #1064, opened 2024-10-01
feat: add option to show variation arrows in analysis

tom-anders:
Mainly for study, but can be helpful for analysis as well (the website also has this option for analysis)

Open questions:

- [ ] Color choice - The mainline should be highlighted more. The website uses a blue shadow for the mainline arrow, but we could also do something different of course
- [ ] What happens if engine evaluation is enabled and the best move is also the mainline move? The website has a special arrow color for this (blue with white shadow)

Screenshot_1727818512

### DIFF
--- a/lib/src/model/analysis/analysis_controller.dart
+++ b/lib/src/model/analysis/analysis_controller.dart
@@ -763,7 +763,7 @@ class AnalysisCurrentNode with _$AnalysisCurrentNode {
 
   const factory AnalysisCurrentNode({
     required Position position,
-    required bool hasChild,
+    required List<Move> children,
     required bool isRoot,
     SanMove? sanMove,
     Opening? opening,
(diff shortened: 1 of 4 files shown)


veloce:
I have to say I don't like this feature, so I wish it would not be on the app 😅 

In my opinion, it adds too much visual noise; at first glance it is really hard to know what are the arrows for... if we include it, at least it must be disabled by default. But in my dream world we would not include it at all.

Can you tell me why is it useful for study thought?

EmmetSchuler:
Studies rely on many branches offshooting from the mainlines, especially when studying opening theory. Since these were introduced to the website, I think people have found them to be very useful. It gives a visual representation to the often large mass of sideline notation.

They also add functionality to the website by using keyboard shortcuts to quickly switch between lines. For a mobile experience you might not need this functionality though, as you can just tap through the notation below.

In my opinion, I don't think these variation arrows need to be in the analysis board. They do add visual clutter, and most people don't need to be bothered by variations during brief game analysis. But in studies, where people more intentionally analyze sidelines, they are helpful.

tom-anders:
> Studies rely on many branches offshooting from the mainlines, especially when studying opening theory. Since these were introduced to the website, I think people have found them to be very useful. It gives a visual representation to the often large mass of sideline notation.

Fully agree with your reasoning, thanks for writing this out :)

@veloce I also think it's mainly useful for studies (maybe disabled by default), totally fine if we don't add it for the "normal" analysis. 

I cherry picked this from my study branch, just to see if you think it would be useful in analysis as well, but no problem, I'll close the PR then and keep it on the study branch only (unless you decide we don't want the feature for studies either)

veloce:
Let's do this yes, keep it for the study for now.

PR #1072, opened 2024-10-06
Disable resizing of app view on Android

### LINKED ISSUE #914: Wrong pieces scaling
### What is the problem?

Non-consistent scaling

https://github.com/user-attachments/assets/4c2a84ac-feaf-4090-aece-8dfaeb2350d0

### App version

v0.9.7

### Device Info

one plus 10 pro, Android 14

### Additional information

_No response_

meenbeese:
Closes #914 and follows up on #221.


veloce:
Thanks! Indeed this is required as long as we don't support it.

We'll probably never want to support it on phones, but maybe some times on tablets.

PR #1076, opened 2024-10-08
refactor: make PgnTreeView independent of the AnalysisController

tom-anders:
Note: Most of the file should be moved to something like `src/lib/view/pgn/tree_view.dart`, but I have not done that yet to make reviewing easier. I'd push a commit that does this once the rest of the code looks good to you


test/view/analysis/analysis_screen_test.dart
```diff
@@ -300,7 +300,7 @@ void main() {
 
         await tester.tap(find.byKey(const Key('goto-previous')));
         // need to wait for current move change debounce delay
-        await tester.pump(const Duration(milliseconds: 200));
+        await tester.pumpAndSettle(const Duration(milliseconds: 200));
```
tom-anders, line 303: not 100% sure why `pump()` worked fine before and then with this commit `pumpAndSettle()` is needed - but with only `pump()` the `Timer` behind the debouncer never calls it callback

veloce (approved):
LGTM

tom-anders:
Two new commits:

1) Move the tree view to a dedicated file as discussed
2) Fix the "show Comments" and "Toggle move annotations" toggles not taking immediate effect (mentioned on discord here: https://discord.com/channels/280713822073913354/807722604478988348/1295982127741145182)

veloce:
Nice! Before I forget, I noticed another bug with the new tree: hiding a variation not anymore works if you open the context menu on any child but the first one.

tom-anders:
Ah good point, let's keep this PR open then, I'll push a fix and a regression test for that

tom-anders:
Fixed :)

PR #1111, opened 2024-10-25
feat: add study screen

tom-anders:
Feature set is the same as the old app for now, I'll add support for interactive studies in another PR.

Chat is also missing for now, I'll have to figure out how to use the websocket for that.

@veloce The build fails because this PR depends on some of the other study PRs, namely https://github.com/lichess-org/mobile/pull/1109 https://github.com/lichess-org/mobile/pull/1077 and https://github.com/lichess-org/mobile/pull/990. If you want to try out the whole feature, it's probably easiest to create a local branch based on this PR and then rebase it onto the other ones.

[study_screen.webm](https://github.com/user-attachments/assets/7722124a-f554-4619-aa3e-7f91fd7d4463)

### DIFF
--- a/lib/src/model/study/study_controller.dart
+++ b/lib/src/model/study/study_controller.dart
@@ -33,6 +33,17 @@ class StudyController extends _$StudyController implements PgnTreeNotifier {
 
   Timer? _startEngineEvalTimer;
 
+  @override
+  Future<StudyState> build(StudyId id) async {
+    final evaluationService = ref.watch(evaluationServiceProvider);
+    ref.onDispose(() {
+      _startEngineEvalTimer?.cancel();
+      _engineEvalDebounce.dispose();
+      evaluationService.disposeEngine();
+    });
+    return _fetchChapter(id);
+  }
+
   Future<void> nextChapter() async {
     if (state.hasValue) {
       final chapters = state.requireValue.study.chapters;
@@ -131,17 +142,6 @@ class StudyController extends _$StudyController implements PgnTreeNotifier {
     }
 
     return studyState;
-  }
-
-  @override
-  Future<StudyState> build(StudyId id) async {
-    final evaluationService = ref.watch(evaluationServiceProvider);
-    ref.onDispose(() {
-      _startEngineEvalTimer?.cancel();
-      _engineEvalDebounce.dispose();
-      evaluationService.disposeEngine();
-    });
-    return _fetchChapter(id);
   }
 
   EvaluationContext _evaluationContext(Variant variant) => EvaluationContext(
--- a/lib/src/model/study/study_controller.dart
+++ b/lib/src/model/study/study_controller.dart
@@ -248,7 +248,11 @@ class StudyController extends _$StudyController implements PgnTreeNotifier {
     if (!state.hasValue) return;
 
     final node = _root.nodeAt(path);
-    for (final child in node.children) {
+
+    final childrenToShow =
+        _root.isOnMainline(path) ? node.children.skip(1) : node.children;
+
+    for (final child in childrenToShow) {
       child.isHidden = false;
       for (final grandChild in child.children) {
         grandChild.isHidden = false;
--- a/lib/src/model/study/study_controller.dart
+++ b/lib/src/model/study/study_controller.dart
@@ -68,6 +68,8 @@ class StudyController extends _$StudyController implements PgnTreeNotifier {
     final variant = study.chapter.setup.variant;
     final orientation = study.chapter.setup.orientation;
 
+    // Some studies have illegal starting positions. This is usually the case for introductory chapters.
+    // We do not treat this as an error, but display a static board instead.
     try {
       _root = Root.fromPgnGame(game);
     } on PositionSetupException {
--- a/lib/src/model/study/study_controller.dart
+++ b/lib/src/model/study/study_controller.dart
@@ -1,0 +1,622 @@
+import 'dart:async';
+
+import 'package:chessground/chessground.dart';
+import 'package:collection/collection.dart';
+import 'package:dartchess/dartchess.dart';
+import 'package:fast_immutable_collections/fast_immutable_collections.dart';
+import 'package:freezed_annotation/freezed_annotation.dart';
+import 'package:lichess_mobile/src/model/analysis/analysis_preferences.dart';
+import 'package:lichess_mobile/src/model/common/chess.dart';
+import 'package:lichess_mobile/src/model/common/eval.dart';
+import 'package:lichess_mobile/src/model/common/id.dart';
+import 'package:lichess_mobile/src/model/common/node.dart';
+import 'package:lichess_mobile/src/model/common/service/move_feedback.dart';
+import 'package:lichess_mobile/src/model/common/service/sound_service.dart';
+import 'package:lichess_mobile/src/model/common/uci.dart';
+import 'package:lichess_mobile/src/model/engine/evaluation_service.dart';
+import 'package:lichess_mobile/src/model/engine/work.dart';
+import 'package:lichess_mobile/src/model/study/study.dart';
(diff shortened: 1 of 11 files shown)


veloce:
Love it! The UI is so simple yet intuitive, that's how it should be done imo.

tom-anders:
(rebased against `main`)

Mansour-J:
This looks great! Having the list of chapters come from a drawer could really improve it, freeing up more space for the chapter names.

The dialogue section feels a bit cramped at the moment— The drawer gives it a bit more room and could make the content look less tight.

Overall, it’s amazing work! 👏

tom-anders:
> This looks great! Having the list of chapters come from a drawer could really improve it, freeing up more space for the chapter names.
> 
> The dialogue section feels a bit cramped at the moment— The drawer gives it a bit more room and could make the content look less tight.
> 
> Overall, it’s amazing work! 👏

Maybe we could simply display a full-screen dialog to free up as much space as possible for the chapter titles?

Mansour-J:
> > Overall, it’s amazing work! 👏
> 
> Maybe we could simply display a full-screen dialog to free up as much space as possible for the chapter titles?

@tom-anders Not sure how a full screen modal looks like. 

is this how it looks? (https://www.youtube.com/watch?v=8BkXiQD-zkU)

veloce:
In case you missed it @tom-anders , I think now you can rebase on main to avoid conflicts and make the build pass.

tom-anders:
Thanks for the heads-up, rebased now :)

Also, study chapter selection is now a separate screen (and the titles wrap, so they'll never be cut off):

Screenshot_1730833542

veloce:
Isn't it a bit weird to have the chapters on a different screen? Pushing a new screen to the stack and tap an item there to make a change on the screen below seems not very intuitive.
If we do that it should at least be a full screen modal (`fullScreenDialog` param of Navigator.push) that closes itself automatically when a chapter is selected.

lib/src/model/study/study_controller.dart
```diff
@@ -0,0 +1,628 @@
...
+
+    final rootComments = IList(game.comments.map((c) => PgnComment.fromPgn(c)));
+
+    final variant = study.chapter.setup.variant;
+    final orientation = study.chapter.setup.orientation;
+
+    try {
+      _root = Root.fromPgnGame(game);
+    } on PositionSetupException {
+      return StudyState(
```
veloce, line 74: How does the web client handle these situations? Should we display a snackbar warning?

lib/src/model/study/study_controller.dart
```diff
@@ -0,0 +1,628 @@
...
+          _startEngineEval();
+        });
+      });
+    }
+
+    return studyState;
+  }
+
+  @override
+  Future<StudyState> build(StudyId id) async {
```
veloce, line 135: `build` should be the first method in a notifier class, because it is its signature.

lib/src/model/study/study_controller.dart
veloce: After seeing the comment in `StudyState`, I realise this the way to handle illegal positions that can normally happen in study chapters. I think it deserves a comment to explain it to future readers.

tom-anders:
> Isn't it a bit weird to have the chapters on a different screen? Pushing a new screen to the stack and tap an item there to make a change on the screen below seems not very intuitive. If we do that it should at least be a full screen modal (`fullScreenDialog` param of Navigator.push) that closes itself automatically when a chapter is selected.
> 
> It could also be a bottom sheet (like we have for settings).

Will try out a bottom sheet later today 👍

lib/src/view/study/study_screen.dart
```diff
@@ -11,6 +40,460 @@ class StudyScreen extends ConsumerWidget {
...
+        return const PlatformScaffold(
+          appBar: PlatformAppBar(
+            title: Text(''),
+          ),
+          body: Center(child: CircularProgressIndicator()),
+        );
+      },
+      error: (error, st) {
+        _logger.severe('Cannot load study: $error', st);
+        Navigator.of(context).pop();
```
veloce, line 81: I never used that pattern in the rest of the app. Not sure it is good UX wise. But for consistency's sake we should display an error instead and let the user come back.

lib/src/view/study/study_screen.dart
```diff
@@ -11,6 +40,460 @@ class StudyScreen extends ConsumerWidget {
...
+        Scrollable.ensureVisible(
+          currentChapterKey.currentContext!,
+          alignment: 0.5,
+        );
+      }
+    });
+
+    return PlatformScaffold(
+      appBar: const PlatformAppBar(
+        // TODO mobile l10n
```
veloce, line 136: You already have `studyNbChapters` that I believe you can use here?

veloce (commented):
Looks very good! Excited to get this in the app :)

tom-anders:
Thanks for the review!

Moved chapters to a bottom sheet now:

[chapters.webm](https://github.com/user-attachments/assets/5512b1c3-cdb6-4404-a249-1de211c764ce)

veloce:
Thanks for the update @tom-anders !

Just quickly tested study feature and it seems collapsing variation is not working properly anymore:


https://github.com/user-attachments/assets/a8231c87-b8f8-48c7-9cb8-66bfea86c196

tom-anders:
> Thanks for the update @tom-anders !
> 
> Just quickly tested study feature and it seems collapsing variation is not working properly anymore:
> collapse_bug.mov

Oh, good catch, sorry about that!

That's an interesting chain reaction actually :D I wanted to port https://github.com/lichess-org/mobile/pull/1137/commits/a4463522a45b6dd44fda4ca4a088bc4199e7792e to this PR, but in that commit I called the variable `childrentToHide` (instead of `childrenToShow`. So when merging the changes into `StudyController` in this branch, I added them to `collapseVariations()` instead of `expandVariations()`, messing up both collapsing and expanding lol.

veloce:
Thanks for the fix!

Now that I'm playing with the collapsing/expanding, I find it a little disturbing (I know I merge the PR that submitted the choice, in video it looked fine but I hadn't actually tested it).

The "plus" and "minus" buttons are disappearing once they are tapped, which is not the best UX experience. In order to work, only one button is needed and it should stay at the same place, changing its appearance with the collapsed state.

We can compare with a file explorer to have a good idea of what we should achieve:

https://github.com/user-attachments/assets/1ce06e1a-37be-4ac8-909f-534ae1bc4e46

It would be great to have animation when expanding/collapsing. But the most important first is to get the buttons right. I wonder if I should revert the PR that introduces the minus button, or if we can solve this quickly? 
Would it work if the "+" button was at the end of the line (same as minus) instead of being in the indented line? And we could show the first node with a "..." in the indented line in place of the +.

tom-anders:
Hmm, might be not be a small fix, we'd have to change some logic of the `TreeView` widget again for that. I'd say let's revert the change that introduced the minus and I'll have another look at it in a new PR

tom-anders:
Maybe we could even go full file Explorer mode? Have a `v` icon **in front** of the line when it's expanded and a `>` icon if the line is collapsed?

Anyways, I'll revert the commit in this PR and keep you updated about this idea

tom-anders:
(rebased to fix conflicts)

tom-anders:
Also, cherry-picked the bug fix `https://github.com/lichess-org/mobile/pull/1111/commits/f291adbdd0902ebfd465869b0adfe30e3a737cae` from https://github.com/lichess-org/mobile/pull/1137 (which was reverted, but the bugfix was unrelated)

veloce:
> Also, cherry-picked the bug fix [f291adb](https://github.com/lichess-org/mobile/commit/f291adbdd0902ebfd465869b0adfe30e3a737cae) from #1137 (which was reverted, but the bugfix was unrelated)

Thanks! It should be cherry-picked on main too, probably. But no big deal, we'll merge this eventually.

PR #1126, opened 2024-11-01
Increase move list height in tablet/landscape view

### LINKED ISSUE #585: Make board larger on landscape tablets by moving buttons from under the board
Compared to the regular lichess app the new one uses the screen size less efficiently. The menu/navigation buttons could be moved to the right of the board, as before.

Regular app:
old

New app:
new

Maybe do all you can generally to maximise the size of the board. Even if you can't/won't use the full screen mode you often see watching movies there is space above the board on both apps which could be used.

tom-anders:
Partially fixes #585, the bottom bar is still at the very bottom, but we at least use the space on the right hand side more efficiently now.

Before:

Screenshot_1730458251

After:

Screenshot_1730458271

This also affects all other screens that use `BoardTable`, but I think it looks nice, for example puzzle streak:

Before:

Screenshot_1730458340

After:

Screenshot_1730458306

### DIFF
--- a/lib/src/widgets/board_table.dart
+++ b/lib/src/widgets/board_table.dart
@@ -264,7 +264,7 @@ class _BoardTableState extends ConsumerState<BoardTable> {
                         crossAxisAlignment: CrossAxisAlignment.stretch,
                         mainAxisAlignment: MainAxisAlignment.spaceAround,
                         children: [
-                          Flexible(child: widget.topTable),
+                          widget.topTable,
                           if (!widget.zenMode && slicedMoves != null)
                             Expanded(
                               child: Padding(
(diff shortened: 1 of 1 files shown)


veloce:
Looks better, definitely! Perhaps more vertical padding around the move list would help. Here the player widgets look squeezed to the edge.

tom-anders:
> Looks better, definitely! Perhaps more vertical padding around the move list would help. Here the player widgets look squeezed to the edge.

Agreed 👍 or maybe add spacing around the top and bottom player widget instead

tom-anders:
Added `kTabletBoardTableSidePadding` around the top and bottom widget, I think this looks very nice:

Screenshot_1731143843

veloce:
Can you also share a screen recording of a new game showing the move list @tom-anders ? Thanks.

I want to make sure how the move list look like when it is empty (or when it has too few moves to fill all the space).

tom-anders:
Sure, here's a video of the start of the game, and another one later in the game where the move list starts scrolling

[moves1.webm](https://github.com/user-attachments/assets/5b0da810-2152-440e-8ff2-54a14d712c17)

[moves2.webm](https://github.com/user-attachments/assets/66d3642d-db35-410b-9ac0-2cf9b71914b8)

veloce:
Perfect! Thank you!

veloce (approved):

Issue #1098, opened 2024-10-19
[Feature] change clock position (left/right) 


tom-anders:
Scrolled through Google Play reviews recently, and at least two people mentioned this feature. The argument is that your right thumb will obscure the clock while playing. Having an option to move the clock to the other side would fix this.

Apparently the old app has this feature as well.

JozefSlawinski:
Hi, I can take on this one. 
![image](https://github.com/user-attachments/assets/3ced3f46-cbec-4020-a060-94f94f77db02)

Please take a look

Jimima:
Added in #1136, I think this can now be closed

PR #1136, opened 2024-11-05
Added choice picker in settings and game settings sheet to switch the clock position to the left side

Jimima:
Related to issue https://github.com/lichess-org/mobile/issues/1098

New developer here, I have had a crack at this. If there are omissions or errors or anything I have done incorrectly please just let me know. I am happy to learn more 😄


veloce (commented):
Thanks for this PR. Just a couple of comments, otherwise it looks good.

lib/src/model/settings/board_preferences.dart
```diff
@@ -106,6 +112,7 @@ class BoardPrefs with _$BoardPrefs implements Serializable {
     required bool coordinates,
     required bool pieceAnimation,
     required bool showMaterialDifference,
+    required bool switchClockPosition,
```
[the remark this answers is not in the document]
Jimima: Hey, I agree this makes more sense. How would you approach the UI in that case? Would you keep a toggle or change to a choice picker similar to the theme selector? Or something else?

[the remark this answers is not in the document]
veloce: Choice picker looks good in that case yes.

lib/src/model/settings/board_preferences.dart
```diff
@@ -132,6 +139,7 @@ class BoardPrefs with _$BoardPrefs implements Serializable {
     coordinates: true,
     pieceAnimation: true,
     showMaterialDifference: true,
+    clockPosition: ClockPosition.left,
```
tom-anders, line 142: Shouldn't `right` be the default to match the current UI?

tom-anders:
Currently, when the clock is set to `left`, there's currently some space on the right of the player widget (see screenshot below).

I think you can fix this by adding `mainAxisAlignment: clockPosition == ClockPosition.right ? MainAxisAlignment.start : MainAxisAlignment.end` to the two `Row` widgets in side the `playeWidget`

Screenshot_1731145252

lib/src/model/settings/board_preferences.dart
Jimima: Yes typo will fix

tom-anders:
Yeah look good! But I think you also need it to the `Row` that displays the material difference below the user name (see my original screenshot, you can capture some pieces to check this)

Jimima:
Yes I seem, makes sense I should have made some captures! Now added.

lib/src/model/settings/board_preferences.dart
Jimima: Fixed

veloce (approved):

Jimima:
😁 thanks!

PR #1137, opened 2024-11-05
feat: add button to expand collapsed variations

tom-anders:
[expand.webm](https://github.com/user-attachments/assets/0d39f689-e8a7-4b5f-81f5-d3e183dcf024)

### DIFF
--- a/lib/src/model/analysis/analysis_controller.dart
+++ b/lib/src/model/analysis/analysis_controller.dart
@@ -283,7 +283,6 @@ class AnalysisController extends _$AnalysisController
   @override
   void collapseVariations(UciPath path) {
     final node = _root.nodeAt(path);
-
     for (final child in node.children) {
       child.isHidden = true;
     }
--- a/lib/src/model/analysis/analysis_controller.dart
+++ b/lib/src/model/analysis/analysis_controller.dart
@@ -271,7 +271,11 @@ class AnalysisController extends _$AnalysisController
   @override
   void expandVariations(UciPath path) {
     final node = _root.nodeAt(path);
-    for (final child in node.children) {
+
+    final childrenToHide =
+        _root.isOnMainline(path) ? node.children.skip(1) : node.children;
+
+    for (final child in childrenToHide) {
       child.isHidden = false;
       for (final grandChild in child.children) {
         grandChild.isHidden = false;
--- a/lib/src/widgets/pgn.dart
+++ b/lib/src/widgets/pgn.dart
@@ -615,11 +615,38 @@ class _SideLinePart extends ConsumerWidget {
           return moves;
         },
       ).flattened,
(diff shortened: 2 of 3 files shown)


ijm8710:
Thanks Tom!❤️

veloce (approved):
Excellent! Thanks for this.

tom-anders:
@ijm8710 FYI after some discussion in https://github.com/lichess-org/mobile/pull/1111#issuecomment-2462771207 we decided to revert this PR for now, due to several reasons:

- Having the [+] and [-] buttons not in the same place is confusing and a bit irritating when expanind/collapsing variations
- We could have the [+] and [-] at the start of an expanded/collapsed sideline instead, but then the icon is right next to a move button and becomes hard to press
- It's not clear whether collapsing variations is really used that often (by the average user) - Even lichess.org does not display a collapse-button, but requires a right-click instead (but it **does** have the expand-button

ijm8710:
> @ijm8710 FYI after some discussion in https://github.com/lichess-org/mobile/pull/1111#issuecomment-2462771207 we decided to revert this PR for now, due to several reasons:
> 
> 
> 
> - Having the [+] and [-] buttons not in the same place is confusing and a bit irritating when expanind/collapsing variations
> 
> - We could have the [+] and [-] at the start of an expanded/collapsed sideline instead, but then the icon is right next to a move button and becomes hard to press
> 
> - It's not clear whether collapsing variations is really used that often (by the average user) - Even lichess.org does not display a collapse-button, but requires a right-click instead (but it **does** have the expand-button

Hey Tom appreciate you sharing the  heads up, my only hope is that it's temporarily paused while it's rethought rather than becomes abandoned fully. Based on scanning the comments in the pr you shared I think that's the case, but hopefully I'm reading that right.

tom-anders:
> > @ijm8710 FYI after some discussion in [#1111 (comment)](https://github.com/lichess-org/mobile/pull/1111#issuecomment-2462771207) we decided to revert this PR for now, due to several reasons:
> > 
> > * Having the [+] and [-] buttons not in the same place is confusing and a bit irritating when expanind/collapsing variations
> > * We could have the [+] and [-] at the start of an expanded/collapsed sideline instead, but then the icon is right next to a move button and becomes hard to press
> > * It's not clear whether collapsing variations is really used that often (by the average user) - Even lichess.org does not display a collapse-button, but requires a right-click instead (but it **does** have the expand-button
> 
> Hey Tom appreciate you sharing the heads up, my only hope is that it's temporarily paused while it's rethought rather than becomes abandoned fully. Based on scanning the comments in the pr you shared I think that's the case, but hopefully I'm reading that right.

We had some more discussion via DM about this, but the bullet points above are pretty much a summary - I think we should at least see if there's more feedback about this in the future.

Issue #468, opened 2024-01-10
Add an option to show captured pieces instead of material imbalance


veloce:

VaibhavKukreti-Kathud:
Can I work on this issue?

veloce:
Yes you can open a PR when you're ready.

arjun1194:
Hey @VaibhavKukreti-Kathud are you working on this ? if not i would like to pick this up :)

arjun1194:
what is the process of developing such features ? is there a PRD or a Proposal ?

HaonRekcef:
No its simlle, just fork the project, code your features, improvements, or bug fixes, and then create a pull request. The maintainer will then take a look at it, provide feedback, and approve it. However, note that the maintainer is currently on vacation, so it might take some time for the PR to be merged.

saurabhffs:
@veloce I am working on this issue

sandeep-adireddi:
Hey @veloce , Where is the best place to add a toggle option for showing or hiding this enhancement?.

veloce:
Well I'm not sure, because there is a conflict with the "Material imbalance" option, which is already a toggle. We can't have both true at the same time.

sandeep-adireddi:
Yeah, I asked this because i noticed this conflict. One option i found is similar to current app .
![Screenshot_20240513-211449](https://github.com/lichess-org/mobile/assets/16081707/0eb8fd01-ea5e-434c-b10a-94618b6a2c2e)

HaonRekcef:
Closed by #1142

PR #1142, opened 2024-11-08
468 show captured pieces instead of material imbalance

Jimima:
Had a go at this, draft at this stage as there is a bit to do but wanted feedback on the overall idea, if possible

Relates https://github.com/lichess-org/mobile/issues/468


veloce:
UI looks fine!

Jimima:
I added a secondary option to the settings for this. I wonder if this is the best approach, it was simple to add it this way but maybe a single choice with 3 options would be better




Was also a bit unclear on why the theme preference implementation did something custom for the choice picker on ios, this seems to work but can take the other approach if there is a reason to do so

ijm8710:
I think it makes more sense to have one setting with 3 options. Rather than 2 separate toggle due to the fact that the settings screen keeps getting more options and to limit it best we can.

Obv Veloce/toms opinion will carry more weight but I think it's best to keep that contained best they can

Jimima:
Alternate option

Jimima:
In-game option

https://github.com/user-attachments/assets/bf1684f8-4640-40fc-9575-d8cc518bbe88

Jimima:
Would appreciate a review @veloce @tom-anders

veloce:
It is better with a single setting yes. Thanks!

lib/src/model/game/material_diff.dart
```diff
@@ -80,8 +104,16 @@ class MaterialDiff with _$MaterialDiff {
     });
 
     return MaterialDiff(
-      black: MaterialDiffSide(pieces: black.toIMap(), score: -score),
-      white: MaterialDiffSide(pieces: white.toIMap(), score: score),
+      black: MaterialDiffSide(
+        pieces: black.toIMap(),
+        score: -score,
+        capturedPieces: blackCapturedPieces,
```
tom-anders, line 110: Can you extend `test/model/game/material_diff_test.dart` with a check for `capturedPieces`?

lib/src/model/game/material_diff.dart
```diff
@@ -35,6 +36,29 @@ class MaterialDiff with _$MaterialDiff {
...
+        Board.standard.materialCount(Side.black);
+    final IMap<Role, int> whiteStartingCount =
+        Board.standard.materialCount(Side.white);
+    // TODO: parameterise starting position maybe so it can be passed in
+
+    IMap<Role, int> subtractPieceCounts(
+      IMap<Role, int> startingCount,
+      IMap<Role, int> subtractCount,
+    ) {
+      IMap<Role, int> capturedPieces = IMap();
```
tom-anders, line 49: This could be written in a pure functional way:

```
return startingCount.map(
    (role, count) => MapEntry(role, count - (subtractCount.get(role) ?? 0)),
);

```

lib/src/model/game/material_diff.dart
```diff
@@ -35,6 +36,29 @@ class MaterialDiff with _$MaterialDiff {
     final IMap<Role, int> blackCount = board.materialCount(Side.black);
     final IMap<Role, int> whiteCount = board.materialCount(Side.white);
 
+    final IMap<Role, int> blackStartingCount =
+        Board.standard.materialCount(Side.black);
+    final IMap<Role, int> whiteStartingCount =
+        Board.standard.materialCount(Side.white);
+    // TODO: parameterise starting position maybe so it can be passed in
```
tom-anders, line 43: Yes I think this is needed before merging - It's already possible to view (but not play) variants that have a different starting position (e.h. Horde). With this PR, `capturedPieces` will display incorrect values currently.

However, I noticed that the web UI does not display material difference at all for Horde, maybe we should do the same? @veloce wdyt?

lib/src/model/settings/board_preferences.dart
tom-anders, line 1: There's some commented-out code here that probably needs to be removed

lib/src/model/settings/board_preferences.dart
```diff
@@ -300,3 +313,17 @@ enum BoardTheme {
...
 }
+
+enum MaterialDifference {
+  materialDifference(label: 'Material difference', visible: true),
+  capturedPieces(label: 'Captured pieces', visible: true),
+  hidden(label: 'Hidden', visible: false);
+
+  const MaterialDifference({
+    required this.label,
+    required this.visible,
```
tom-anders, line 324: is the `visible` field really needed? A property `bool get hidden => this != MaterialDifference.hidden` would also work (and probably same some memory)

lib/src/model/settings/board_preferences.dart
```diff
@@ -300,3 +313,17 @@ enum BoardTheme {
...
+  });
+
+  final String label;
```
tom-anders, line 327: The label will be translated at some point, so it's maybe better to already make this a method along the lines of `String l10n(BuildContext context) => switch (this) { ... }` instead (and add a `//TODO l10n`)

lib/src/model/settings/board_preferences.dart
```diff
@@ -110,7 +119,9 @@ class BoardPrefs with _$BoardPrefs implements Serializable {
     required bool boardHighlights,
     required bool coordinates,
     required bool pieceAnimation,
-    required bool showMaterialDifference,
+    // required bool showMaterialDifference,
+    required MaterialDifference materialDifference,
```
tom-anders, line 123: Calling this `MaterialDifferenceFormat` (like in the commented line below) would be more readable I think. Currently, there's both `MaterialDiff` and `MaterialDifference` which is confusing

lib/src/view/game/game_player.dart
```diff
@@ -140,16 +143,25 @@ class GamePlayer extends StatelessWidget {
...
           Row(
             children: [
-              for (final role in Role.values)
-                for (int i = 0; i < materialDiff!.pieces[role]!; i++)
-                  Icon(
-                    _iconByRole[role],
-                    size: 13,
-                    color: Colors.grey,
-                  ),
+              if (materialDifference == MaterialDifference.capturedPieces)
```
tom-anders, line 149: Maybe it would be more readable to factor out this whole `Row` into a `class MaterialDifferenceDisplay extends StatelessWidget`. (Then you could also precompute the pieces to loop over in a helper variable and use that in the loop, instead of needing to write the loop twice)

lib/src/model/game/material_diff.dart
Jimima: Thanks, I thought this should be possible but I couldn't figure out the right way to do it

lib/src/model/settings/board_preferences.dart
Jimima: Sorry, I am not 100% sure what you mean here (though I understand the broader point). Is there an example somewhere else I could look at?

Jimima:
@tom-anders I pushed up a bunch of changes, would appreciate you taking another look when you have the time 🙂

lib/src/model/settings/board_preferences.dart
tom-anders: Sure, I meant something like this: https://github.com/lichess-org/mobile/blob/ca78193fdd972e9e21ffd282cde7ef2642450d74/lib/src/model/study/study_filter.dart#L16

But actually it's more common in the codebase to do it via a free function, like this: https://github.com/lichess-org/mobile/blob/ca78193fdd972e9e21ffd282cde7ef2642450d74/lib/src/view/settings/theme_screen.dart#L31

tom-anders:
very nice, no more comments from me :)

Jimima:
Oh should have said, I still don't really know the best label to use for the setting. I just changed it to simply "Material" just now but I've changed it a few times and still can't decide what is best. Happy to be told here!

Jimima:
> very nice, no more comments from me :)

Thanks, will take a look at the l10n stuff tomorrow

lib/src/model/settings/board_preferences.dart
Jimima: Hmm yeah I think maybe it sitting in the enum itself makes more sense but I'm not sure. I can implement either approach if there is an opinion either way. For now I might just go with option A, it's easy to change

Jimima:
Just in case anyone was wondering I changed the base branch as the diffs were out of date and this was suggested as a fix. Maybe I misunderstood what I was looking at but it seems to have worked so...

Jimima:
@veloce @tom-anders I think I've gone as far as I can with this. Would appreciate any further feedback you have for me 🙂

lib/src/view/settings/board_settings_screen.dart
```diff
@@ -247,13 +248,42 @@ class _Body extends ConsumerWidget {
...
+                              .read(boardPreferencesProvider.notifier)
+                              .setMaterialDifferenceFormat(
+                                value ??
+                                    MaterialDifferenceFormat.materialDifference,
+                              ),
+                    );
+                  } else {
+                    pushPlatformRoute(
+                      context,
+                      title: 'Clock position',
```
veloce, line 282: Wrong title

lib/src/model/settings/board_preferences.dart
```diff
@@ -120,7 +124,7 @@ class BoardPrefs with _$BoardPrefs implements Serializable {
     required bool boardHighlights,
     required bool coordinates,
     required bool pieceAnimation,
-    required bool showMaterialDifference,
+    required MaterialDifferenceFormat materialDifferenceFormat,
```
veloce, line 127: A new `required` need a `@JsonKey(defaultValue:` annotation otherwise json parsing fail.

veloce:
Thanks for this! I fixed the small issues mentioned above, the rest was perfect.

Jimima:
Ah okay that's great, thanks!

PR #1159, opened 2024-11-15
feat: play sound when new game is started

### LINKED ISSUE #890: No sound played when a game starts

tom-anders:
Fixes #890, superseeds #891

### DIFF
--- a/lib/src/model/game/game_controller.dart
+++ b/lib/src/model/game/game_controller.dart
@@ -113,6 +113,13 @@ class GameController extends _$GameController {
 
         _socketEventVersion = fullEvent.socketEventVersion;
 
+        // Play "dong" sound when this is a new game and we're playing it (not spectating)
+        final isMyGame = game.youAre != null;
+        final noMovePlayed = game.steps.length == 1;
+        if (isMyGame && noMovePlayed && game.status == GameStatus.started) {
+          ref.read(soundServiceProvider).play(Sound.dong);
+        }
+
         if (game.playable) {
           _appLifecycleListener = AppLifecycleListener(
             onResume: () {


lib/src/model/game/game_controller.dart
```diff
@@ -113,6 +114,16 @@ class GameController extends _$GameController {
 
         _socketEventVersion = fullEvent.socketEventVersion;
 
+        // Play "dong" sound when this is a new game and we're playing it (not spectating)
+        final myUserId = ref.watch(authSessionProvider)?.user.id;
```
veloce, line 118: You don't need to watch session, you already have `game.youAre` which tell the side you're playing.

lib/src/model/game/game_controller.dart
```diff
@@ -113,6 +114,16 @@ class GameController extends _$GameController {
...
+        final isMyGame =
+            myUserId != null && game.playerSideOf(myUserId) != null;
+        if (isMyGame &&
+            game.status == GameStatus.started &&
```
veloce, line 122: Here we should use the number of moves of the current side we're playing with:
```
if (youAre == white && game.steps.length = 0 || 
  youAre == black && game.steps.length = 1
) {
  play(sound.dong)
}
```

Meaning: the current side we're playing with should have played 0 moves.

tom-anders: Ah, as I said in the original PR, I wasn't sure when this would actually happen, but I guess it could be a race condition?

1) accept a challenge (or start public seek)
2) app send "accepted challenge" to opponent
3) app requests game from server
4) opponent gets white and already makes the first move
5) server sends game to app, with first move already being made

anyway, I'll add that change :+1:

tom-anders: Note that if I understand the docs correctly, `0 moves` actually means `steps.length==1` and `1 move` means `steps.length==2`, the first step is always the initial position

veloce: Right, my bad here. `steps` is never empty. 

Not sure to understand your comment about the race condition. And my comment is wrong also, I suppose black should get the dong whether white has played its move or not.

veloce: @tom-anders I saw your comment in the original PR. You're right, let's do as web UI and play the dong only if no moves have been made.

tom-anders: Done :+1:

veloce (approved):

PR #1198, opened 2024-11-29
Tweak analysis layout

julien4215:
In orientation landscape, we already know that this is a tablet so we don't need to check for it.

We also are not checking if it is a tablet for the padding so I think this commit makes the code more consistent.

### DIFF
--- a/lib/src/view/analysis/analysis_layout.dart
+++ b/lib/src/view/analysis/analysis_layout.dart
@@ -176,7 +176,7 @@ class AnalysisLayout extends StatelessWidget {
                                   boardBuilder(
                                     context,
                                     boardSize,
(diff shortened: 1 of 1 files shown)


veloce:
> In orientation landscape, we already know that this is a tablet so we don't need to check for it.

It's wrong to assume that. On folded phone screens you can have landscape mode and be still below the tablet threshold.

julien4215:
Maybe I misunderstood these lines then ? https://github.com/lichess-org/mobile/blob/3f72fa5d0e5e637e5aaff1a9f0e7466cd65530c5/lib/src/init.dart#L115-L117

Doesn't it set only portrait mode for devices below tablet threshold ?

PR #1211, opened 2024-12-02
challenge leela odd bots

### LINKED ISSUE #796: Add possibility to challenge Leela featured bots
Following featured bots are not shown for now:

- LeelaKnightOdds
- LeelaQueenOdds
- LeelaQueenForKnight
- LeelaRookOdds

Because they need a special position and time control to accept the challenge.

Since they are verified bots from lichess we can support them directly by having a special treatment in the challenge screen for them:

- We need a position selector (in the form of selectable mini boards) pre configured with the position choices that the bot accepts.
- When relevant, user can only choose one side (black or white) depending on the selected position
- These bots use a more restricted time control range

HaonRekcef:
closes #796 . Implements challenges featured leela bots with piece odds. The bots can now only be challenged with time controls and positions they will accept: 

video of phone: 

https://github.com/user-attachments/assets/e3398388-cf75-432d-bbf4-f59ebed3c307

video of tablet :

https://github.com/user-attachments/assets/0b33d103-9bf9-41e1-9a53-d8f059d77c28

Remarks:
- I'm unsure if the current method for storing the positions of the different bots is the nicest. Suggestions are welcome.
- The new bot, [LeelaPieceOdds](https://lichess.org/@/LeelaPieceOdds), supports a wide range of starting positions (as shown in the phone video). This might be overwhelming; feedback on better ways to handle this would be appreciated.
- I couldn't test it with the real bots since challenging them on Lichess.org requires an account login, which isn't possible in the debug version.  
- Cannot test it on iOS
- The grid layout is directly adapted from the puzzle history screen.

### DIFF
--- a/lib/src/view/play/challenge_odd_bots_screen.dart
+++ b/lib/src/view/play/challenge_odd_bots_screen.dart
@@ -1,0 +1,302 @@
+import 'package:dartchess/dartchess.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter_layout_grid/flutter_layout_grid.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+import 'package:lichess_mobile/src/model/challenge/challenge.dart';
+import 'package:lichess_mobile/src/model/challenge/challenge_preferences.dart';
+import 'package:lichess_mobile/src/model/common/chess.dart';
+import 'package:lichess_mobile/src/model/common/game.dart';
+import 'package:lichess_mobile/src/model/common/time_increment.dart';
+import 'package:lichess_mobile/src/model/user/user.dart';
+import 'package:lichess_mobile/src/styles/styles.dart';
+import 'package:lichess_mobile/src/utils/l10n_context.dart';
+import 'package:lichess_mobile/src/utils/navigation.dart';
+import 'package:lichess_mobile/src/view/game/game_screen.dart';
+import 'package:lichess_mobile/src/widgets/board_thumbnail.dart';
+import 'package:lichess_mobile/src/widgets/buttons.dart';
+import 'package:lichess_mobile/src/widgets/list.dart';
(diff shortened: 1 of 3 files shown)


HaonRekcef:
> UI is fine by me. The only thing I found weird is that, even if the position was not chosen, you could challenge the opponent nevertheless. I think the button should be disabled as long as the position is not chosen.
> 
I think it should be disabled as long as no board is selected, the challenge button is greyed out and can't be pressed.

> I add a commit to fix the iOS view (needed a safe area), and it completely messed up the history in the PR, sorry about that.
> 
> I could not test really the PR since Leela bots don't accept challenges at the moment.

On the website, you need to select the correct color. For example, if white is missing a knight, you need to select the black pieces for yourself so that the bot accepts. If you tried from the app with my PR and it didn’t accept, I have a bug somewhere.

HaonRekcef:
I rebased it to the newest version and fixed the conflicts

veloce:
Thanks for the update. I tested it again:

- the iOS version is bugged, but I don't know why
- on android choosing a Leela odd bot would mess up the other challenge settings: if I go back to a regular challenge, it will show fromPosition and rated false
- on android the time sliders are stuck to 10+0 and impossible to move

I think the `create_challenge_screen` is over complicated now. 
This feature should be implemented in a completely separate screen for the odd bots. That way it won't mess with regular challenge screen.
I don't think we need to persist the settings, because that feature is quite niche. Just keep the time and inc setting in memory for users who challenge several time in a row maybe?

HaonRekcef:
I refactored it to a different screen. I decided not to persist the setting at all since this feature is indeed niche. Instead, I simply take the time control preferences from the settings.

* I cannot test it on iOS.
* The slider should not be stuck, I think it’s just a bug that sometimes occurs when moving it with the mouse in the emulator.

veloce (approved):

PR #1212, opened 2024-12-03
[WIP] save streak score locally

tom-anders:


HaonRekcef:
Nice work! However, the new streak in the end of the video seems to start with the same first puzzle again. I assume this is not what we want.

tom-anders:
> Nice work! However, the new streak in the end of the video seems to start with the same first puzzle again. I assume this is not what we want.

That's a server issue (or feature?) - If you fail the streak early, you sometimes get the same puzzle sequence again

veloce (approved):
Thank you for the added test!

PR #1213, opened 2024-12-03
Broadcast list screen improvements

julien4215:
- better layout
- uniform pagination scroll behavior


veloce:
It would be nice to have the same tier system as in the web client. When tournaments with higher tier are displayed bigger.

julien4215:
Sure that would be a cool thing but I don't think that's worth it to do it now.

julien4215:
Sorry I didn't think about it, will try to not forget next time.

PR #1218, opened 2024-12-04
Broadcast round improvements

veloce:
- Show location and official website in overview tab
- Display broadcast image in overview tab
- Display round status and date in round selector


ijm8710:
Excited for this one. A title would be nice above the image on overview tab too (perhaps you added, not sure what the final pr will look like)

Ideally I'd like a little title in header bar but assuming that complicates broadcast a bit so I think at least in overview tab itself would suffice

PR #1239, opened 2024-12-09
Added Bluesky to social links

TheOnlyTails:
Added Bluesky to the `LinkSite` enum, hopefully I did everything correctly! 
This should ideally designate bsky.app links as links to Bluesky instead of showing the URL.


veloce:
Havent' tested but I think it should! thanks.

Issue #1090, opened 2024-10-14
Puzzle score fonts


Riahiamirreza:
### What is the problem?

Sometimes when playing tactics, I see the font size of the scores for single digit scores are larger. The below screenshots demonstrate it.

Here the third box, which is a red box with digit 6, its size is not equal to other single digit boxes:
![image](https://github.com/user-attachments/assets/227df81f-a249-4246-9223-6db58272d899)

Here the gray box:
![image](https://github.com/user-attachments/assets/4a2426d2-a06b-465d-bb79-efc65f2f4981)

### Steps to reproduce the bug

I'm not really sure how to reproduce it, I just encountered it a couple of times.

### App version

0.12.0

### Device Info

Android, Samsung

tom-anders:
Hmm, judging from the Screenshots it looks like it only happens for *negative* single-digit numbers, but not for positive or two-digit, right?

Riahiamirreza:
@tom-anders I tried to reproduce it and and I think you're correct. I didn't observed any positive single-digit number with this issue. But every negative single-digit number had this issue.

rhenandias:
I believe this is also happening with positive numbers; however, the size difference is almost imperceptible.

For example, in -18 and +8, there is a size difference in the digit 8, but the most noticeable difference is with the -9 that comes right after.

![Screenshot_2024-10-14-22-48-22-071_org.lichess.mobileV2.jpg](https://github.com/user-attachments/assets/07ddfd50-c357-4b53-a091-3fce2613edca)

![Screenshot_2024-10-14-22-48-22-071_org.lichess.mobileV2~2.jpg](https://github.com/user-attachments/assets/a3b56d66-11b2-4d17-a93f-79d6b90df357)

I think it has more to do with the font size adjusting to the available space in the badge.

Boubou78000:
@tom-anders when do you plan to fix this issue? This issue is still there after 3 months 😅

tom-anders:
It's a small visual inconsistency that does not affect gameplay. So it's a rather low priority issue, and apparently nobody has had the time to work on it yet.

Please keep in mind that (apart from veloce) almost all contributors are working on the app voluntarily and in their free time.

That being said, it's a good issue for newcomers to work on, so I'll add the corresponding label to give it some more visibility

tom-anders:
Well that was quick, guess we should be more liberal with the `good first issue` tag :D

PR #1241, opened 2024-12-09
fix inconsistent font size in puzzle's ratingDiff

rizqicy:
Hi,

not familiar with flutter but I think this small change can close this bug #1090 
please test and if you have suggestion please let me know

before
Screenshot_20241209_161924 before

after
Screenshot_20241209_161858 after


veloce:
Thank you!

PR #1243, opened 2024-12-09
Broadcast players tab

julien4215:
[broadcast_players.webm](https://github.com/user-attachments/assets/a430577e-b537-482d-8a27-aa89984accd6)

Tablet :

Screenshot_1733842417

### DIFF
--- a/lib/src/model/broadcast/broadcast_providers.dart
+++ b/lib/src/model/broadcast/broadcast_providers.dart
@@ -1,3 +1,4 @@
+import 'package:fast_immutable_collections/fast_immutable_collections.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:lichess_mobile/src/model/broadcast/broadcast.dart';
 import 'package:lichess_mobile/src/model/broadcast/broadcast_repository.dart';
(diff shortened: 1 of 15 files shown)


lib/src/model/broadcast/broadcast_providers.dart
```diff
@@ -54,6 +55,50 @@ Future<BroadcastTournament> broadcastTournament(
...
+    BroadcastTournamentId tournamentId,
+  ) async {
+    final players = ref.withClient(
+      (client) => BroadcastRepository(client).getPlayers(tournamentId),
+    );
+
+    return players;
+  }
+
+  void sort(BroadcastPlayersSortingTypes sortingType, [bool reverse = false]) {
```
veloce, line 73: Looks like this function is not used since the sorting is done in the widget state.

veloce (commented):
Nice work! thanks

lib/src/model/broadcast/broadcast_providers.dart
julien4215: Oops I forgot to remove it when moving the sort in the widget

ijm8710:
Very cool Julian. Just making sure you're aware, some tournaments have a teams section as well.

Random example: 

Slovenian Team Championship 2024

PR #1248, opened 2024-12-10
fix: otb in tablet offline mode

HaonRekcef:
Adds the option to play OTB on a tablet while offline.
This option is already available for phones via the floating play button.
Preview:
grafik
I am not sure if it will look good on iOS with the Icon before the text.


veloce:
Thanks for raising this issue! I've fixed it a bit differently, showing the whole section even if offline, as it should simplify things.

PR #1250, opened 2024-12-10
fix: creation of gif api

HaonRekcef:
Fixes two bugs related to the GIF creation process:
- Creating a GIF from studies while the current board theme was set to "system" did not work. It now defaults to brown.
- Sometimes depending on the current theme, the board theme preferences were not respected due to a different naming in the GIF-API . This issue affected GIF creation for games (it defaulted to brown instead of the user prefs) and studies (GIFs were not created at all).

Also, I changed the labels of the board themes to include spaces, e.g., Blue2 → Blue 2.


veloce (approved):
LGTM!

PR #1251, opened 2024-12-10
Custom board colors and theme

veloce:


Boubou78000:
Which themes will you implement? The themes of the old app? Will you add support for custom backgrounds?

Post scriptum: thanks veloce for implementing such cool features

PR #1253, opened 2024-12-11
Broadcast player results screen

julien4215:
[broadcast_results.webm](https://github.com/user-attachments/assets/bb598fbd-460b-4092-b6c9-85d9975d6772)


ijm8710:
Hey Julian, 2 requests.

1) mobile web green/reds wins and losses (the 0 and 1s) so that material results stand out. I like this and we should probably do same.
2) on mobile web, above the results list, they have the player bio (ranks across variants, player age, fide id that links to their fide player page,etc). I'd like to have that too shown outright either above or below the list for this pr.

lib/src/view/broadcast/broadcast_player_results_screen.dart
```diff
@@ -0,0 +1,363 @@
...
+                    );
+                  }
+
+                  final playerResult = playerResults.games[index - 1];
+
+                  return GestureDetector(
+                    onTap: () {
+                      pushPlatformRoute(
+                        context,
+                        builder: (context) => BroadcastGameScreen(
```
veloce, line 227: Since we can now open a screen from another context, this should be refactored:


https://github.com/lichess-org/mobile/blob/9a6a4004237aac6fac86c87555d7ecf9da7d49dc/lib/src/view/broadcast/broadcast_game_screen.dart#L372-L378

lib/src/view/broadcast/broadcast_players_tab.dart
```diff
@@ -118,7 +121,7 @@ class _PlayersListState extends ConsumerState<PlayersList> {
   @override
   Widget build(BuildContext context) {
     final double eloWidth = max(MediaQuery.sizeOf(context).width * 0.2, 100);
-    final double scoreWidth = max(MediaQuery.sizeOf(context).width * 0.15, 70);
+    final double scoreWidth = max(MediaQuery.sizeOf(context).width * 0.15, 90);
```
veloce, line 124: Why this change? Have you noticed a case where it was too small?

lib/src/view/broadcast/broadcast_players_tab.dart
```diff
@@ -178,59 +181,78 @@ class _PlayersListState extends ConsumerState<PlayersList> {
...
+                      : player.name,
+                  player.title,
+                  player.name,
+                ),
+              );
+            },
+            child: ColoredBox(
               color: Theme.of(context).platform == TargetPlatform.iOS
-                  ? index.isEven
+                  ? (index - 1).isEven
```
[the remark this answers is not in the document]
julien4215: This was to have the same color order as in the opening explorer

lib/src/view/broadcast/broadcast_players_tab.dart
julien4215: This is because I changed the score from 'x/y' to 'x / y'.

lib/src/view/broadcast/broadcast_player_results_screen.dart
julien4215: I don't understand why this would not work with deep link. The two controllers are independent, they just listen to the same route because there is only one for broadcast.

veloce: If you can load a game by itself you don't want to load the whole round, right? In a tournament where you have 100 board you'd load 99 boards (and listen to their updates) for nothing.

lib/src/view/broadcast/broadcast_boards_tab.dart
```diff
@@ -60,19 +60,16 @@ class BroadcastBoardsTab extends ConsumerWidget {
...
             child: Center(
               child: Text('Could not load broadcast: $error'),
             ),
           ),
-        _ => const SliverFillRemaining(
-            child: Center(
-              child: CircularProgressIndicator.adaptive(),
-            ),
-          ),
+        _ => const BroadcastPreview.loading()
```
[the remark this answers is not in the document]
veloce: If there is a way to know in advance the number of boards, then we can use the shimmer. Otherwise I think it is ugly to see a flash of 10 shimmer boards where in fact only one is show eventually.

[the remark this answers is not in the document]
julien4215: Having only one board in a round is pretty rare. In the majority of cases I think the board shimmer looks nicer than the circular indicator.

lib/src/view/broadcast/broadcast_player_results_screen.dart
julien4215: I agree with that but I have done it this way because in the API there are no endpoints to get the player information for a specific board. 

Maybe this refactoring can be done in an other branch ?

lib/src/view/broadcast/broadcast_game_screen.dart
```diff
@@ -79,11 +81,15 @@ class _BroadcastGameScreenState extends ConsumerState<BroadcastGameScreen>
   Widget build(BuildContext context) {
     final broadcastGameState = ref
         .watch(broadcastGameControllerProvider(widget.roundId, widget.gameId));
+    final broadcastRoundState =
+        ref.watch(broadcastRoundControllerProvider(widget.roundId));
```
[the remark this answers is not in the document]
julien4215: I used `select` instead because `selectAsync` returns a `Future` which makes unpractical to use inside a widget.

lib/src/view/broadcast/broadcast_boards_tab.dart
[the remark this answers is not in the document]
veloce: Unfortunately the world championship is the most watch broadcast... even if it is not the majority it will be seen much more times than other tournaments with more boards.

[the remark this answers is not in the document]
veloce: It does unless there is only one or two board and you see a flash of shimmer if the request completes very quickly.

lib/src/view/broadcast/broadcast_game_screen.dart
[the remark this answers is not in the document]
veloce: In order to use `selectAsync` you need to create a dedicated provider to retrieve the title. This is the way to do it with riverpod.

lib/src/view/broadcast/broadcast_player_results_screen.dart
veloce: Yes let's do it in another PR.

veloce: This code is also problematic because you assume you assume the round controller is loaded: `select((round) => round.requireValue.games[gameId]!)`

I'm not sure it is a good practice to mix `select` and `requireValue` in riverpod. So far we know it is loaded because the only way to access the game screen was through the round screen.

Now that you can access the game screen from the player result screen directly, can you be sure the round screen will be loaded? If you can access a game from another round it won't be the case. This is the main reason why it must be refactored.

julien4215: I removed the `requireValue`

lib/src/view/broadcast/broadcast_game_screen.dart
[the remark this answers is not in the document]
julien4215: But what is the advantage of creating a new provider instead of using `select` like this
```
final title = ref.watch(
      broadcastRoundControllerProvider(widget.roundId)
          .select((round) => round.value?.round.name),
    );
```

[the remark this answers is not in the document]
veloce: It's mostly a problem of typing. Here you're mixing the async value which represent an asynchronous value which may be not here yet, with a nullable value that you derives from another asynchronous value.

By creating another provider that returns an AsyncValue<String> for the title you're making it clear of what you expect. It is ok to declare a new provider for that. It is like a cached computed value, but async. Doing that is recommended in the riverpod doc (but I don't remember where).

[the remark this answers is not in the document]
veloce: Note also that using `.value?` directly you're not handling the error, it will be rethrown in the widget build method making it crash.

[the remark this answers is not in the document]
julien4215: Ok thanks for the detailed explanation. I'll make a title provider. 

I believe I have already done this in previous PR but I'll make sure to create a provider in the future

lib/src/view/broadcast/broadcast_player_results_screen.dart
veloce: After talking with Thibault, it's decided that it is ok to load the full round to display a single game. So no refactoring after all.

But we need to ensure the round controller is loaded before the game controller in the game screen. (So we can do a `requireValue`)

lib/src/view/broadcast/broadcast_game_screen.dart
```diff
@@ -82,17 +82,21 @@ class _BroadcastGameScreenState extends ConsumerState<BroadcastGameScreen>
...
-          .select((round) => round.value?.round.name),
+    final broadcastGameState = ref.watch(
+      broadcastGameProvider(widget.roundId, widget.gameId),
     );
+    final broadcastGamePgn = ref
+        .watch(broadcastGameControllerProvider(widget.roundId, widget.gameId));
+    final title = widget.title ??
+        (switch (ref.watch(broadcastGameScreenTitleProvider(widget.roundId))) {
+          AsyncData(value: final title) => title,
+          _ => 'Broadcast Game',
```
veloce, line 93: I'd rather have a `SizedBox.shrink()` here than a default title.

julien4215: Done

PR #1254, opened 2024-12-11
feature: share option in the broadcast game screen

HaonRekcef:
Adds share URL, download PGN and GIF Options to the broadcast game screen

**Preview**:
 
[broadcasts-share.webm](https://github.com/user-attachments/assets/51b197dc-2753-43b5-a94c-b488aaa21237)

**Remarks**:
- I refactored the GIF functionality from study into `game_share_service` and reused it for broadcasts.
- The broadcast name and round name are not necessary for the URL, but I believe including them is better, as it helps to humanly read the URL.
- I decided against adding the screenshot position option. I can implement it if you think its needed, but I think this feature is unnecessary on mobile devices.

### DIFF
--- a/lib/src/model/game/game_share_service.dart
+++ b/lib/src/model/game/game_share_service.dart
@@ -69,7 +69,7 @@ class GameShareService {
         .read(defaultClientProvider)
         .get(
           Uri.parse(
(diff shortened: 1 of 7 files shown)


veloce (approved):
Thank you!

PR #1269, opened 2024-12-17
Edit profile page changes, closes The biography resets when reducing the keyboard #1245

### LINKED ISSUE #1245: The biography resets when reducing the keyboard
### What is the problem?

When modifying your biography, if you reduce your keyboard it will revert your changes.

1. Go to your profile
2. Change your biography 
3. Reduce the keyboard

### App version

0.13.2

### Device Info

Android 14

### Additional information

Screenshot_20241209_083133
Screenshot_20241209_083139

nunibye:
Changes made to the edit profile page include:
 * Removed padding from top and bottom outside ListView and added padding inside ListView.
 * Added scroll behavior and GestureDetector to hide keyboard.
 * Refactored to use widgets instead of functions for performance.
     * Each field owns it controller ensuing dispose is called.
     * closes #1245
 * Made screen pop on successful save.
 * trimmed all text fields before upload.
 * added PopScope to reduce accidental data loss.
Some kind of dialog should be displayed before exiting, but that will require changing the appbar and I haven't quite figured out all the custom widgets to maintain native feel.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,3 +38,6 @@ doc/api/
 
 # FVM Version Cache
 .fvm/
+
+# VS Code
+.vscode/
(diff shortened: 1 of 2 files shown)


veloce:
Thanks!

I added the confirm dialog here: https://github.com/lichess-org/mobile/commit/6c43166cf5c0ca40876a3986490bfab383ea6e43

PR #1270, opened 2024-12-18
Fix tab animation not played

### LINKED ISSUE #1232: Swiping away from the Opening Explorer Tab leads to visual bug
### What is the problem?

When swiping off of the opening explorer tab in the analysis board, it doesn’t animate the slide and instead the moves list appears instantly.
This is just a small visual bug, nothing that affects functionality.

Go to the analysis board
Play a few moves for both sides
Swipe to the opening explorer and let it load
Swipe back to the moves list

Expected: the swipe should animate with a sliding transition
Result: no sliding transition is played, and the moves list instantly appears.

### App version

0.13.4

### Device Info

iPhone 13 Mini iOS 18

### Additional information

https://github.com/user-attachments/assets/86af03d9-c01b-4988-bf18-9c90948dfa62

Note that if the position in the opening explorer was previously loaded, the animation does play. But if you start a new analysis session and play some moves, the above issue is reproduceable.

julien4215:
Fix #1232 but creates a new issue

I suggest to remove the automatic scroll to the current move in the analysis screen until a better solution is found.

### DIFF
--- a/lib/src/widgets/pgn.dart
+++ b/lib/src/widgets/pgn.dart
@@ -145,15 +145,20 @@ class _DebouncedPgnTreeViewState extends ConsumerState<DebouncedPgnTreeView> {
     super.initState();
     pathToCurrentMove = widget.currentPath;
     pathToBroadcastLiveMove = widget.broadcastLivePath;
-    WidgetsBinding.instance.addPostFrameCallback((_) {
-      if (currentMoveKey.currentContext != null) {
-        Scrollable.ensureVisible(
-          currentMoveKey.currentContext!,
-          alignment: 0.5,
-          alignmentPolicy: ScrollPositionAlignmentPolicy.keepVisibleAtEnd,
-        );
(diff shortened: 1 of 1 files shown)


veloce:
Well the automatic scroll is more important than the tab animation not working, isn't it?

So this fix is not acceptable. But thanks for finding the issue. There is maybe another way to solve it. Like delay the automatic scroll so it happens after the animation perhaps?

julien4215:
Now that I think about it automatic scroll might be more important.

I don't think delaying the automatic scroll will make things better because the tab animation takes quite a long time to finish (you can see it with the app indicator changing 1 or 2 seconds after having swiped).

The only easy solution I can see is to set the scroll at the bottom but I don't know if a game is always opened on the last move.

veloce:
Tested with a delay of 500ms it works pretty well. And the delay is not really disturbing imo.

Issue #849, opened 2024-07-10
Any chance to bump "new opponent" to the top of the game context menu.


ijm8710:
Much more used than flip board (this is on game completion or revisiting a prior game; you're not even analyzing from that screen so having flip board at top doesn't make as much sense).

Would prefer new opponent first than rematch please. Thoughts?

HaonRekcef:
This can be closed imo. Flip board is not here anymore and now with the new rematch animation rematch should be on top.

tom-anders:
> This can be closed imo. Flip board is not here anymore and now with the new rematch animation rematch should be on top.

Really? Pretty sure I'm on the latest version and I still see "flip board" above "new opponent" - do you have an updated screenshot?

HaonRekcef:
Ah yes my bad. On the pop up there seems to be no more flip board but in the burger menu there still is.

PR #1271, opened 2024-12-18
Tweak analysis screen

julien4215:
Remove flip board from the analysis menu because it is already present in the bottom bar


ijm8710:
Thanks you can prob close my ticket #849 then

ijm8710:
Does/should this remove from top of menu for analysis as well?

PR #1273, opened 2024-12-18
Add broadcast tests

julien4215:
Requires #1313


test/view/broadcast/broadcast_game_screen_test.dart
```diff
@@ -0,0 +1,119 @@
...
+        ],
+      );
+
+      await tester.pumpWidget(app);
+
+      expect(find.byType(CircularProgressIndicator), findsOneWidget);
+
+      await tester.pump();
+
+      expect(find.byType(Chessboard), findsOneWidget);
```
veloce, line 53: You could also check that the player widget with their names is present.

test/view/broadcast/broadcast_game_screen_test.dart
```diff
@@ -0,0 +1,119 @@
...
+
+      await tester.pump();
+
+      expect(find.byKey(const Key('c1-whitebishop')), findsOneWidget);
+
+      fakeSocket.addIncomingMessages([
+        r'''{"v":151,"t":"addNode","d":{"n":{"ply":23,"fen":"rnq2rk1/pp2ppbp/5np1/2P1N3/2N5/4B1P1/PP2PPKP/R2Q1R2 b - - 2 12","id":"%7","uci":"c1e3","san":"Be3","clock":359500},"p":{"chapterId":"G2LUflKg","path":")8aP19YQ(1`Y'*_b.>VF-=F=$3UE3=]O8GOF>EF1)1^]"},"d":"0S 978 TCEJNZ8 WGO 3NV 6xEILQSYZ78 !? UM 5OQZ 2V? XHP","s":false,"relayPath":"!"}}''',
+      ]);
+
+      // Two pump are required to get the move but I don't know why
```
veloce, line 83: I think that is because how the socket is mocked. Since we're using a stream controller, the socket event is received after the next micro task. So 1 pump to wait for the event, and 1 pump to treat the event.

You can add that explanation, or better yet: make a test helper that takes the socket channel and performs the `tester.pump()`. Sth that would allow to write:

```dart
await addSocketMessage(tester, fakeSocket, '<message here>');

// wait for the controller to process the message
await tester.pump();
```

That way writing (and reading) tests will be more clear.

test/view/broadcast/broadcast_round_screen_test.dart
```diff
@@ -0,0 +1,133 @@
...
+          }),
+        ],
+      );
+
+      await tester.pumpWidget(app);
+
+      expect(find.byType(CircularProgressIndicator), findsOneWidget);
+
+      // Two pump are required to load boards tab but I don't know why
+      await tester.pump();
```
veloce, line 55: We wait first for the tournament, then for the round. That's why. The way the screen is loaded, there is no visual difference between the 2 loading states. So you can write that explanation to enlighten future readers.

test/view/broadcast/broadcast_round_screen_test.dart
```diff
@@ -0,0 +1,133 @@
...
+      await tester.pump();
+
+      expect(find.byType(BroadcastBoardsTab), findsOneWidget);
```
veloce, line 58: Would be nice to add a few more expectations. Like the number of boards matches the number of expected (should only return a few boards so they are all visible at once).

Maybe check the first players names too.

test/view/broadcast/broadcast_round_screen_test.dart
```diff
@@ -0,0 +1,133 @@
...
+        await tester.pump();
+
+        // Swipe to the overview tab
+        await tester.flingFrom(const Offset(50, 300), const Offset(500, 0), 1000);
+
+        // Wait for the animation to end
+        await tester.pump(const Duration(seconds: 1));
+
+        expect(find.byType(BroadcastBoardsTab), findsNothing);
+        expect(find.byType(BroadcastOverviewTab), findsOneWidget);
```
veloce, line 88: Add a few checks to verify the content inside the tab is properly loaded.

julien4215:
Something else that I don't understand is that `pumpAndSettle` always time out on both screen despite the fact that the screen is not changing

test/view/broadcast/broadcast_game_screen_test.dart
```diff
@@ -0,0 +1,136 @@
...
+      final app = await makeTestProviderScopeApp(
+        tester,
+        home: const BroadcastGameScreen(
+          tournamentId: _tournamentId,
+          roundId: _roundId,
+          gameId: _gameId,
+        ),
+        overrides: [
+          lichessClientProvider.overrideWith((ref) => LichessClient(client, ref)),
+          webSocketChannelFactoryProvider.overrideWith((ref) {
```
veloce, line 42: It is not necessary to override the fake socket if you're not using it.

test/view/broadcast/broadcast_game_screen_test.dart
```diff
@@ -0,0 +1,136 @@
...
+      );
+
+      await tester.pumpWidget(app);
+
+      expect(find.byType(CircularProgressIndicator), findsOneWidget);
+
+      await tester.pump();
+
+      expect(find.byType(Chessboard), findsOneWidget);
+      expect(find.byType(BroadcastPlayerWidget), findsNWidgets(2));
```
veloce, line 55: Don't check for widget type when you can check text. It is not really interesting to check the type (and if you change the widget name test would fail).

You can use `find.widgetWithText(<widget_type>, text)` if the text might appear twice in different widgets.

test/view/broadcast/broadcast_game_screen_test.dart
```diff
@@ -0,0 +1,136 @@
...
+          webSocketChannelFactoryProvider.overrideWith((ref) {
+            return FakeWebSocketChannelFactory((_) => fakeSocket);
+          }),
+        ],
+      );
+
+      await tester.pumpWidget(app);
+
+      await tester.pump();
+      // await tester.pump(); It might be interesting to know why it breaks the test
```
veloce, line 80: remove the comment, if you have questions or things you don't understand it is better to leave the comment directly in GitHub PR page as long as the review is not done.

test/view/broadcast/broadcast_game_screen_test.dart
```diff
@@ -0,0 +1,136 @@
...
+        overrides: [
+          lichessClientProvider.overrideWith((ref) => LichessClient(client, ref)),
+          webSocketChannelFactoryProvider.overrideWith((ref) {
+            return FakeWebSocketChannelFactory((_) => fakeSocket);
+          }),
+        ],
+      );
+
+      await tester.pumpWidget(app);
+
```
veloce, line 78: It is always interesting to check what is rendered after pumpWidget (usually a loading state).

So when we read the test, we can follow the state transition and understand what is going on.

test/view/broadcast/broadcast_game_screen_test.dart
```diff
@@ -0,0 +1,136 @@
...
+      await tester.pump();
+      // await tester.pump(); It might be interesting to know why it breaks the test
+
+      expect(find.byKey(const Key('c1-whitebishop')), findsOneWidget);
+
+      await addSocketMessage(
```
veloce, line 84: remove the helper, and change to:
```dart
fakeSocket.addIncomingMessage(...);

await pumpEventQueue();
```

Notice the `pumpEventQueue()` this is how we ensure the event is delivered to the listener.

test/view/broadcast/broadcast_game_screen_test.dart
```diff
@@ -0,0 +1,136 @@
...
+
+      // The controller process the message
+      await tester.pump();
+
+      expect(find.byKey(const Key('e3-whitebishop')), findsOneWidget);
+    });
+  });
+}
+
+Future<void> addSocketMessage(
```
veloce, line 98: as mentioned earlier, remove this helper.

test/view/broadcast/broadcast_round_screen_test.dart
```diff
@@ -0,0 +1,253 @@
...
+
+      // Load the tournament
+      await tester.pump();
+
+      expect(find.byType(CircularProgressIndicator), findsOneWidget);
+
+      // Load the round
+      await tester.pump();
+
+      expect(find.byType(BroadcastBoardsTab), findsOneWidget);
```
veloce, line 66: This is useless, in general we don't want to check widget types (as explained earlier). Unless the check is meaningful, like checking the `BoardThumbnail` number matches the expected number of boards.

test/view/broadcast/broadcast_round_screen_test.dart
```diff
@@ -0,0 +1,253 @@
...
+          await tester.pumpWidget(app);
+
+          // Load the tournament
+          await tester.pump();
+
+          // Load the round
+          await tester.pump();
+
+          // Swipe to the overview tab
+          await tester.flingFrom(const Offset(50, 300), const Offset(500, 0), 1000);
```
veloce, line 99: Don't bother to use a gesture when you can use a `tester.tap(find.text('Overview'))`.

test/view/broadcast/broadcast_round_screen_test.dart
```diff
@@ -0,0 +1,253 @@
...
+        // Load the round
+        await tester.pump();
+
+        expect(find.byType(BroadcastOverviewTab), findsOneWidget);
+      },
+    );
+  });
+}
+
+final _broadcast = Broadcast(
```
veloce, line 211: If you have common test data in both files, it is better to put that in another file `test/view/broadcast/example_data.dart` that you can import from the test files.

veloce (commented):
On top of the comments I've made, you should change the broadcast round tests in the following way:

- stay simple, don't test the change of tab so remove both tests that test the change of tab (we don't really care to test the change of tab is working, we should focus first on the feature important logic); instead you can just check the tabs are presents with `find.text`)
- I assume the `pumpAndSettle()` times out because you used a live broadcast source (the clocks are responsible of the timeout)
- to simplify things and to improve the tests you should test content with a not yet started broadcast or a finished broadcast: check that the not yet started loads the overview, and the finished loads the boards
- don't check types but check content (text). Example in the overview tab, check that the dates, the location, the broadcast title are there. We don't care about the type of the widget unless if helps to discriminate text that might be present in several places.
- Once you have all the tests working correctly with the finished and upcoming round, you can add a test to test a live round; keep things simple: only load one board and just checks that the clock is ticking with a few tester.pump()

test/view/broadcast/broadcast_game_screen_test.dart
julien4215: This doesn't work because tests use a different environment for async. 

It needs to be wrapped in `runAsync` but even with this it doesn't work.

veloce: If you replace this function: https://github.com/lichess-org/mobile/blob/84abbae0427e733921fec17d06e8c89089a7c17c/test/network/fake_websocket_channel.dart#L87

with:

```dart
  /// Simulates incoming messages from the server.
  Future<void> addIncomingMessages(Iterable<dynamic> messages) async {
    await _incomingController.addStream(Stream<dynamic>.fromIterable(messages));
  }
```

and then call

```dart
// await for event to be delivered
await tester.runAsync(() => await fakeSocket.addIncomingMessage(...));

// wait for controller to process event and rebuild UI
await tester.pump()
```

does it work?

veloce: In theory we shouldn't use `runAsync` but I'm not sure how to solve this properly actually.

test/view/broadcast/broadcast_game_screen_test.dart
julien4215: I am trying to find the cause of the issue. The reason it breaks the test is that after receiving a move the screen is in loading state again.

I also noticed yesterday watching the world championship that the screen blinks each time a move is received.

julien4215: This seems to be a consequence of using `selectAsync` to get the `broadcastRoundController` state

test/view/broadcast/broadcast_game_screen_test.dart
julien4215: I added a comment to explain what is happening. I don't think we want to test if the screen is loading as the goal of the test is only to see if board is updated when a new move is received.

test/view/broadcast/broadcast_round_screen_test.dart
```diff
@@ -0,0 +1,805 @@
...
+        expect(find.byType(CircularProgressIndicator), findsOneWidget);
+
+        // Load the tournament
+        await tester.pump();
+
+        expect(find.byType(CircularProgressIndicator), findsOneWidget);
+
+        // Load the overview
+        await tester.pump();
+        // await tester.pumpAndSettle();
```
julien4215, line 135: This `pumpAndSettle` still times out even if it is an upcoming broadcast

julien4215:
It is ready for another review

veloce:
Looks like CI is failing, can you look at it before the review please? thanks

julien4215:
This is because it requires #1313 for the tests to pass

veloce:
> This is because it requires #1313 for the tests to pass

Indeed. I just reviewed it @julien4215 , there are requested changes. Thanks.

PR #1285, opened 2024-12-20
feat: add support for the new "hide rating except in game" option

tom-anders:
This adds support for the new "show ratings, except in game" option introduced in https://github.com/lichess-org/lila/pull/16228

One thing that's missing now is the "explanation" for what "hide rating" means/does that was previously displayed as a subtitle - I'm not sure where the best place to display it would be now.

[showRatingsExceptInGame.webm](https://github.com/user-attachments/assets/7f8cfc2a-d290-48e6-b611-62a02e5cc511)


veloce:
A tappable trailing "info" icon that displays the explanation in a tooltip. We once had this (or still have it?) for one setting.

tom-anders:
> A tappable trailing "info" icon that displays the explanation in a tooltip. We once had this (or still have it?) for one setting.

Ah yes, we have it for the move confirmation setting, thanks for the hint!

utterlyrandomuser:
Hi, when selecting this option and then trying to open the preferences in the new beta mobile app it throws an error. Reverting to "Yes" por "No" and it works fine again.

tom-anders:
> Hi, when selecting this option and then trying to open the preferences in the new beta mobile app it throws an error. Reverting to "Yes" por "No" and it works fine again.

That's to be expected, this PR will fix this

tom-anders:
Ok, added the explanation back.

@veloce Since settings are currently broken when selecting the new option (see comment by @utterlyrandomuser above, I can reproduce as well), it would probably make sense to turn this PR into a hotfix release

lib/src/view/game/archived_game_screen.dart
```diff
@@ -185,8 +185,16 @@ class _BoardBody extends ConsumerWidget {
 
     final isBoardTurned = ref.watch(isBoardTurnedProvider);
     final gameCursor = ref.watch(gameCursorProvider(gameData.id));
-    final black = GamePlayer(key: const ValueKey('black-player'), player: gameData.black);
-    final white = GamePlayer(key: const ValueKey('white-player'), player: gameData.white);
+    final black = GamePlayer(
+      key: const ValueKey('black-player'),
+      player: gameData.black,
+      isActiveGameOfCurrentUser: false,
```
veloce, line 191: redundant

lib/src/view/game/archived_game_screen.dart
```diff
@@ -185,8 +185,16 @@ class _BoardBody extends ConsumerWidget {
...
+    );
+    final white = GamePlayer(
+      key: const ValueKey('white-player'),
+      player: gameData.white,
+      isActiveGameOfCurrentUser: false,
```
veloce, line 196: redundant

lib/src/view/game/archived_game_screen.dart
```diff
@@ -205,12 +213,14 @@ class _BoardBody extends ConsumerWidget {
         final black = GamePlayer(
           key: const ValueKey('black-player'),
           player: gameData.black,
+          isActiveGameOfCurrentUser: false,
```
veloce, line 216: redundant

lib/src/view/game/archived_game_screen.dart
```diff
@@ -205,12 +213,14 @@ class _BoardBody extends ConsumerWidget {
...
           clock: blackClock != null ? Clock(timeLeft: blackClock) : null,
           materialDiff: game.materialDiffAt(cursor, Side.black),
         );
         final white = GamePlayer(
           key: const ValueKey('white-player'),
           player: gameData.white,
+          isActiveGameOfCurrentUser: false,
```
veloce, line 223: redundant

lib/src/view/game/game_player.dart
```diff
@@ -40,6 +41,10 @@ class GamePlayer extends StatelessWidget {
   });
 
   final Player player;
+
+  /// Whether we're displaying this player as part of an active game of the current user.
+  /// This might influence whether their rating is shown or not, based on the [ShowRatings] setting.
+  final bool isActiveGameOfCurrentUser;
```
veloce, line 47: I'd rather not have this param and pass directly the `gameState` and the `side` instead of `player`.

Then you can infer everything from the the game state inside the `Player` widget. wdyt?

tom-anders: Yeah that makes sense, will do 👍

tom-anders: Ah it doesn't quite work though, since for the `archived_game_screen.dart` we only have a `LightArchivedGame` (instead of `ArchivedGame`) which does not have the `BaseGame` mixin that we would probably want to use here...?

tom-anders: It's a bit annoying, because the `LightArchivedGame` still has all the info we need, but there's no common interface with `PlayableGame` - Maybe we could have something like `LightBaseGame`, that's like `BaseGame` but with less fields?

veloce: I see the problem. A simple solution would be to delay the player display until we get the `ArchivedGame`, so you can still use the common interface without refactoring. I don't think it is much of an issue.

lib/src/view/game/game_player.dart
```diff
@@ -61,10 +61,14 @@ class GamePlayer extends StatelessWidget {
...
 
   @override
-  Widget build(BuildContext context) {
+  Widget build(BuildContext context, WidgetRef ref) {
     final remaingHeight = estimateRemainingHeightLeftBoard(context);
     final playerFontSize = remaingHeight <= kSmallRemainingHeightLeftBoardThreshold ? 14.0 : 16.0;
 
+    final player = side == Side.white ? game.white : game.black;
+
+    final session = ref.watch(authSessionProvider);
```
veloce, line 70: Instead of watching the session here, I think we should use the `youAre`, `me` and `opponent` props of game.

They are already part of the playable and archived game interface, but the code is duplicated and not in the `BaseGame` interface. It only needs a slight refactoring to fix that.

tom-anders: Done! While I was at it, I also moved `finished`, `aborted` and `playable` fields to `BaseGame`, as these were also duplicated in classes using the `BaseGame` mixin

lib/src/view/settings/account_preferences_screen.dart
```diff
@@ -520,6 +537,67 @@ class _PieceNotationSettingsScreenState extends ConsumerState<PieceNotationSetti
...
+          future: _pendingSetShowRatings,
+          builder: (context, snapshot) {
+            return CupertinoPageScaffold(
+              navigationBar: CupertinoNavigationBar(
+                trailing:
+                    snapshot.connectionState == ConnectionState.waiting
+                        ? const CircularProgressIndicator.adaptive()
+                        : null,
+              ),
+              child: SafeArea(
```
veloce, line 569: This `SafeArea` is useless and prevents the header blur effect so it should be removed.

veloce (commented):
Just a last change and we're good.

lib/src/view/settings/account_preferences_screen.dart
tom-anders: Done :+1:  Maybe it's time to do some refactoring and add a common widget for these setting choice screens? I can have a look in the next few days

veloce: Sure.

PR #1306, opened 2024-12-24
Add flutter lints

julien4215:
The rules that required fixes :
- library_private_types_in_public_api
- unintended_html_in_doc_comment
- use_key_in_widget_constructors
- prefer_const_literals_to_create_immutables 
- prefer_const_constructors

I am not sure if we really want to add this set of rules because some of them seem to especially address to packages like unintended_html_in_doc_comment, use_key_in_widget_constructors, library_private_types_in_public_api. However I think it is still interesting to keep the rules on const.

So I see two options, We can either add manually the two rules on const constructor or add the flutter set of rules and remove the ones that seem specific to Flutter packages.

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -7,8 +7,8 @@
 
 # The following lines activate a set of recommended lints.
 include:
-  - package:lint/strict.yaml
   - package:flutter_lints/flutter.yaml
+  - package:lint/strict.yaml
 
 analyzer:
   language:
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -5,9 +5,10 @@
 # IDEs (https://dart.dev/tools#ides-and-editors). The analyzer can also be
 # invoked from the command line by running `flutter analyze`.
 
-# The following line activates a set of recommended lints for Flutter apps,
-# packages, and plugins designed to encourage good coding practices.
-include: package:lint/strict.yaml
+# The following lines activate a set of recommended lints.
+include:
+  - package:lint/strict.yaml
+  - package:flutter_lints/flutter.yaml
 
 analyzer:
   language:
--- a/lib/src/model/game/archived_game.dart
+++ b/lib/src/model/game/archived_game.dart
@@ -56,7 +56,7 @@ class ArchivedGame with _$ArchivedGame, BaseGame, IndexableSteps implements Base
   /// Create an archived game from the lichess api.
   ///
   /// Currently, those endpoints are supported:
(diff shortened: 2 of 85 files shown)


analysis_options.yaml
```diff
@@ -7,8 +7,8 @@
 
 # The following lines activate a set of recommended lints.
 include:
-  - package:lint/strict.yaml
   - package:flutter_lints/flutter.yaml
+  - package:lint/strict.yaml
```
veloce, line 11: Do we really need to keep both lint packages?

julien4215: It depends on what we want. 

What do you think about these three rules : library_private_types_in_public_api, unintended_html_in_doc_comment, use_key_in_widget_constructors ? Should we add them ? If not I guess that we can keep only the lint package.

veloce: use_key_in_widget_constructors we definitely should have it, and library_private_types_in_public_api looks like a good practice too.

the third one we don't really care.

We should pick the set of rules that best fit our needs, and then adapt by manually enabling/disabling rule individually. Flutter lints is certainly a good choice so we can keep it. Have you found all the differences between flutter_lints and lint/strict?

julien4215: Yes there is google sheets at the end of [this article](https://rydmike.com/blog_flutter_linting.html) that show the differences but this is probably not up to date.

I'll disable the rule about html doc comment.

veloce:
Ok so `lint` the package we're currently using uses 78% of the rules, and flutter_lint uses only 48%.

So we should keep `lint` and add the missing flutter rules we need manually. Another interesting set of rules could be the flutter master repository which also has 78% according to that article.

julien4215:
What's the point of adding it `flutter_lint` rules manually ? 

I guess the best solution would be to set all lint rules manually like they do in the flutter master repository so we have exactly what we want but that would take some time.

veloce:
What's the point of this PR?

I thought that you realised some useful rules were missing. If that is the case, change the PR to manually enable the missing rules while keeping the inclusion of `lint/strict` set of rules.

julien4215:
Well I thought that it would add useful rules because `lint/strict` is only about the Dart language and not Flutter but I realize that was a mistake. 

The rules `prefer_const_literals_to_create_immutables` and `prefer_const_constructors` are already in the set of rules `lint/strict` and the rules `library_private_types_in_public_api`, `use_key_in_widget_constructors` would only be useful if this was a package. So I'll close this PR.

PR #1310, opened 2024-12-25
Play versus computer [ONLINE]

Boubou78000:
Play versus Stockfish level 1-8.
~I started the layout and I am working on making it functional.~ I am waiting for reviews
It's my first time contributing to the repo (and working with Dart and Flutter), so it may not be very good 😅


HaonRekcef:
Cool that you're starting to work on the app! :) 

In the future, I imagine we’d want the 'Play Against Computer' feature to use the offline engine. That said, I think playing against the traditional Lichess Stockfish with levels 1-8 still has its place. 

Maybe it could be better placed at the top of the 'Online Bots' screen? Challenging it could probably reuse much of the logic from the `create_challenge_screen`.

Boubou78000:
> Cool that you're starting to work on the app! :)
> 
> In the future, I imagine we’d want the 'Play Against Computer' feature to use the offline engine. That said, I think playing against the traditional Lichess Stockfish with levels 1-8 still has its place.
> 
> Maybe it could be better placed at the top of the 'Online Bots' screen? Challenging it could probably reuse much of the logic from the `create_challenge_screen`.

Like this? (without the rated button)

[ScreenRecording3.webm](https://github.com/user-attachments/assets/6f20d1a8-1e90-412d-9075-afc5cc52baba)

veloce:
I also like the idea of having Stockfish at the top of the online bots list. Then you reuse the challenge screen and add the level here.

Boubou78000:
> I also like the idea of having Stockfish at the top of the online bots list. Then you reuse the challenge screen and add the level here.

I've done the layout:
[ScreenRecording4.webm](https://github.com/user-attachments/assets/57230696-380f-47db-ac9a-2e00dee1c835)

What should I add at the top of the online bots screen? Maybe a "play with the computer" button with padding?

veloce:
> What should I add at the top of the online bots screen? Maybe a "play with the computer" button with padding?

no you should add "Stockfish <version>" as if stockfish was a bot.

Boubou78000:
Hi @veloce .
I spent many hours on trying to stream the events via `/api/stream/event` but each time I get a 403.
I tried many things but each time it said: `error: missing scope`.
After further investigations, I saw the app uses the `web:mobile` OAuth scope.
Why do I get this 403?

Boubou78000:
Is it possible to get the full id of the created game without using `/api/event/stream` (which returns 403 unauthorized) ?

I looked a bit and it seems that a challenge doesn’t need this because the full id is sent via socket (idem for the lobby) .

Should I:
 - try to use `/api/event/stream` (maybe changing the OAuth part)
 - or am I wrong, and the full id can be received by socket or another API endpoint ?

tom-anders:
You'll probably need a change similar to https://github.com/lichess-org/lila/pull/16023 to get the API endpoint working

Boubou78000:
> You'll probably need a change similar to [lichess-org/lila#16023](https://github.com/lichess-org/lila/pull/16023) to get the API endpoint working

Thanks!

Android Studio is crashing on latest Windows 11 Canary, so I may not progress in this PR for the following days/weeks/months (depending on Microsoft).
I created an issue on lila also.

veloce:
@Boubou78000 don't use the event stream but the socket API. 

Look at `create_game_service.dart` also to see how it works for challenges.

Boubou78000:
> @Boubou78000 don't use the event stream but the socket API. 
> 
> Look at `create_game_service.dart` also to see how it works for challenges.

Can’t look at it for now, but I remember it didn’t have the same response format as a challenge, and it didn’t respond with the same data...

So I don’t think we can listen for these socket events for the AI challenges, as they don’t act as challenges.

I tried in a first place to create a challenge with the data returned, but it didn’t worked.

Boubou78000:
Good news:

[This lila commit](https://github.com/lichess-org/lila/commit/1b4dccb25ce3283f54bc3f90f3ef1d714d3d275e) will make the `/api/challenge/ai` return the full id of the game.

Boubou78000:
> Sorry for interrupting the conversation, I noticed something. In the latest recording, the Play with the Computer and Stockfish is reachable from the Online Bots screen. I think playing with the computer should be available offline as well, just like in the old app. I bet a lot of people (like me) used to play with the computer when there was no internet available for them. I'd miss that feature.

I know.
I'm only implementing the online version for now...

Boubou78000:
Hi @veloce ,
What do you think about it?

veloce:
@Boubou78000 we have to close this, sorry. There is no more plan land this feature on mobile. There are plans to have offline bots instead.

PR #1328, opened 2024-12-31
Add socket batch event

julien4215:
A new [batch event was added to lila-ws](https://github.com/lichess-org/lila-ws/commit/d0f9785fd6fe08a06d453feda626c0532520a0c3) to handle exceptional broadcasts.

[Here](https://github.com/lichess-org/lila/commit/287ba0b2575035f577b8f8334703592deaed75ed) is the change for the client website.

This pull request does the same for the mobile app.

### DIFF
--- a/lib/src/network/socket.dart
+++ b/lib/src/network/socket.dart
@@ -334,16 +334,16 @@ class SocketClient {
         _handlePong(pingDelay);
       case 'ack':
         _onServerAck(event);
-    }
-
-    if (event != SocketEvent.pong && event.topic != 'ack') {
-      if (_streamController.hasListener) {
-        _streamController.add(event);
-      }
-      if (_globalStreamController.hasListener &&
-          _globalSocketStreamAllowedTopics.contains(event.topic)) {
-        _globalStreamController.add(event);
-      }
+      case 'batch':
+        _handleBatch(event);
+      case _:
+        if (_streamController.hasListener) {
+          _streamController.add(event);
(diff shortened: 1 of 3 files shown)


julien4215:
I added the test

lib/src/network/socket.dart
```diff
@@ -334,16 +334,16 @@ class SocketClient {
...
-      if (_streamController.hasListener) {
-        _streamController.add(event);
-      }
-      if (_globalStreamController.hasListener &&
-          _globalSocketStreamAllowedTopics.contains(event.topic)) {
-        _globalStreamController.add(event);
-      }
+      case 'batch':
+        _handleBatch(event);
+      case _:
```
veloce, line 339: You changed the logic here, and looking at the code I think you've introduced a regression:

if the `even.topic` is `n`, you no longer add the event to the stream controllers.

That shows it need a new test case which was missing. A 'should emit events' test with the following expectations:
- should not emit if pong
- should not emit if ack
- should emit in any other case (n, batch, random topic)

julien4215: Oops it would have broken the lobby loading screen. I'll add a test for this regression.

I think we don't want to forward the `batch` event to the stream controller. It is only the events in the batch that we forward.

veloce: yes I meant the batch events themselves

veloce (approved):
lgtm

PR #1347, opened 2025-01-08
1043 add castling method option v2

Jimima:
Adds an option to change castling method. Still needs some work to tidy up but the approach is implemented. Note that it only really works for Standard variants as chess960 always uses king over rook. I think this is okay but maybe the UI could make this clearer.


veloce:
Just for my understanding, why the v1 was closed? thanks :)

Jimima:
Sure, I had attempted an implementation that modified dartchess and chessground which maybe was not the correct approach. I didn't have a great deal of time to look at it at that point so I just closed the PRs for hygeine.  See here for reference https://github.com/lichess-org/dartchess/pull/44

This change is limited to only the mobile app with no changes elsewhere. Not sure if this is the right approach still, happy to be told if you think there is an alternative way to come at it

Jimima:
I think this is ready to be looked at @veloce. Not sure what's going on with the tests but maybe an issue unrelated to this PR?

julien4215:
A fix was pushed on the main branch. The tests need to be rerun manually.

veloce:
@Jimima can you rebase your branch on top of main so the tests pass? thank.

veloce (commented):
Thanks for this. This is looking good but there are a couple of issues to address on top of the comments I've made:

- for now this settings applies only to online games, but it should also apply to puzzles, analysis, study, etc.
- the logic you added to game_controller.dart must be refactored so it can work everywhere.
- this is the time for a refactoring: we need a global `board` wrapper so we can apply settings to the whole app
- a good starting point for this refactoring imo, [is the private _BoardWidget](https://github.com/lichess-org/mobile/blob/daca9cea0890dcb9c9564271498aa365059d75d8/lib/src/widgets/board_table.dart#L321) which should be extracted to a public widget in another file. It should be modified to take only `BoardPrefs` and `BoardSettingsOverride` (and not directly the `ChessboardSettings` from chessground which should be constructed inside the widget from the prefs and the optional override).
- the new castling logic should be tested

Jimima:
Thanks for the feedback, that all makes sense. I also wonder if we are worried about premoves respecting the preference as right now they do not. I think the main site does not either but I would need to check this.

Jimima:
@veloce I would appreciate another review if possible, I have made a few changes:

Created a new BoardWidget component and use this in place of Chessboard where possible
Used translation strings for labels etc.

Some questions:

I wonder if we can support either castling method in the app as it seems like an improvement over the web site. I have left it in for now but can easily remove it if you prefer.

Right now I kept the refactoring light and boardPrefs is passed in as a parameter. Should the board widget be fetching the boardPrefs directly from the shared state rather than taking it as a parameter?

Is there further refactoring to be done? You mentioned taking _only_ BoardPrefs and BoardSettingsOverride but I was a bit unsure about how to make this work. Any further guidance would be gratefully received 🙂

Jimima:
Oh yes, also I will add some tests in the meantime

veloce:
Will review asap.

> I wonder if we can support either castling method in the app as it seems like an improvement over the web site. I have left it in for now but can easily remove it if you prefer.

No here I think we should stick to what the website does. It would be less error prone and easier to maintain.

CloudyDino:
@veloce The website only gives two castling options, but the second castling option (move king onto rook) still allows you to move the king two spaces to castle. So in reality the two options on the website are
- Move king two squares
- Either (move king two squares or move king onto rook)

If we're going to stick to what the website does, we should make sure to do it this way or include all three options since the website options aren't explicit enough.

https://github.com/user-attachments/assets/0dbda333-c97d-4349-b4af-fb226e3d5f7d

veloce:
I wish the website would have more explicit options indeed.

veloce:
Sorry @Jimima for the delay I was focused on fixing the new theme. Could you fix the conflicts and I look at this asap. Thanks!

veloce:
Oh, by the way if we have to support the "either" as the website I'd rather have an "either" choice in the settings as you did initially and asked you to remove... sorry about that.

lib/src/view/settings/board_settings_screen.dart
```diff
@@ -71,6 +71,35 @@ class _Body extends ConsumerWidget {
...
+                          .setCastlingMethod(value ?? CastlingMethod.either);
+                    },
+                  );
+                } else {
+                  Navigator.of(context).push(CastlingMethodSettingsScreen.buildRoute(context));
+                  // pushPlatformRoute(
+                  //   context,
+                  //   title: context.l10n.preferencesCastleByMovingTheKingTwoSquaresOrOntoTheRook,
+                  //   builder: (context) => const CastlingMethodSettingsScreen(),
+                  // );
```
CloudyDino, line 99: Quick question, should this be commented out or deleted?

lib/src/model/settings/board_preferences.dart
```diff
@@ -347,6 +354,19 @@ enum ClockPosition {
...
+enum CastlingMethod {
+  kingOverRook,
+  kingTwoSquares,
+  either;
+
+  String castlingMethodl10n(BuildContext context, CastlingMethod castlingMethod) =>
+      switch (castlingMethod) {
+        CastlingMethod.kingOverRook => context.l10n.preferencesCastleByMovingOntoTheRook,
+        CastlingMethod.kingTwoSquares => context.l10n.preferencesCastleByMovingTwoSquares,
+        CastlingMethod.either => context.l10n.preferencesBothClicksAndDrag,
```
CloudyDino, line 366: Should we be using `preferencesBothClicksAndDrag` here? If the string gets changed in the future from "Either" to "Either click or drag", then we wouldn't want the castling method string to change to that as well. I'm unsure when new strings should be created or if maybe we should rename `preferencesBothClicksAndDrag` to something like `preferencesEither`

lib/src/view/broadcast/broadcast_game_screen.dart
```diff
@@ -354,6 +355,7 @@ class _BroadcastBoardState extends ConsumerState<_BroadcastBoard> {
           newShapeColor: boardPrefs.shapeColor.color,
         ),
       ),
+      boardPrefs: boardPrefs,
```
CloudyDino, line 358: nit: maybe we add this new line right under the size argument to keep argument order the same

lib/src/view/study/study_screen.dart
```diff
@@ -568,6 +569,7 @@ class _StudyBoardState extends ConsumerState<_StudyBoard> {
                 },
               )
               : null,
+      boardPrefs: boardPrefs,
```
CloudyDino, line 572: nit: same here, maybe we add this new line right under the size argument to keep argument order the same

lib/src/model/settings/board_preferences.dart
Jimima: Yeah good point, not sure how to correctly handle this as there is no string currently for this. Maybe I just add a comment?

lib/src/view/settings/board_settings_screen.dart
Jimima: Will remove

Jimima:
@veloce thanks, and no problem at all. I am still not sure of the correct approach here, as I said, happy to take a steer from you if you think there is more to do with refactoring or whatever. I might need some guidance, however :-)

lib/src/model/settings/board_preferences.dart
veloce: Yeah just hardcode it for now and leave a ` // TODO translate`

lib/src/view/settings/board_settings_screen.dart
```diff
@@ -263,6 +287,46 @@ class PieceShiftMethodSettingsScreen extends ConsumerWidget {
   }
 }
 
+class CastlingMethodSettingsScreen extends ConsumerWidget {
+  const CastlingMethodSettingsScreen({super.key});
+
+  static Route<dynamic> buildRoute(BuildContext context) {
+    return buildScreenRoute(
+      context,
+      screen: const BoardClockPositionScreen(),
```
veloce, line 296: Wrong screen here.

lib/src/view/settings/board_settings_screen.dart
```diff
@@ -263,6 +287,46 @@ class PieceShiftMethodSettingsScreen extends ConsumerWidget {
...
+      title: 'Clock position',
```
veloce, line 297: Wrong title.

lib/src/view/settings/board_settings_screen.dart
```diff
@@ -263,6 +287,46 @@ class PieceShiftMethodSettingsScreen extends ConsumerWidget {
...
+      boardPreferencesProvider.select((state) => state.castlingMethod),
+    );
+
+    void onChanged(CastlingMethod? value) {
+      ref.read(boardPreferencesProvider.notifier).setCastlingMethod(value ?? CastlingMethod.either);
+    }
+
+    return CupertinoPageScaffold(
+      navigationBar: const CupertinoNavigationBar(),
+      child: SafeArea(
```
veloce, line 313: No need to have a SafeArea around `ListView`.

lib/src/widgets/board.dart
```diff
@@ -0,0 +1,146 @@
...
+  final ChessboardSettings settings;
+  final String? error;
+  final Widget? boardOverlay;
+  final IMap<Square, Annotation>? annotations;
+  final GlobalKey? boardKey;
+  final Setup setup;
+
+  @override
+  Widget build(BuildContext context) {
+    final Map<Square, Square> castlingMap = {
```
veloce, line 40: This could be a static const.

lib/src/widgets/board.dart
```diff
@@ -0,0 +1,146 @@
...
+  Widget build(BuildContext context) {
+    return SizedBox.square(
+      dimension: boardSize,
+      child: Center(
+        child: Padding(
+          padding: const EdgeInsets.all(16.0),
+          child: Container(
+            decoration: BoxDecoration(
+              color:
+                  Theme.of(context).platform == TargetPlatform.iOS
```
veloce, line 135: Since the new theme we can remove these distinctions between iOS and android. So use `colorScheme.surface` on both platforms.

I should have changed it in the new theme PR, but let's do it here.

veloce (commented):
Good job, I think it's taking the right direction. Made a couple of small comments.

Now, this touches crucial part of the app (board for games and analysis) so this change should be tested.

It should be a widget test for the new `BoardWidget` that test at least the correct behaviour of castling according to all possible settings.

You can look at existing `puzzle_screen_test.dart`, `game_screen_test.dart` or `flutter-chessground` tests for inspiration.

Jimima:
Thanks for the comments I'll get to work on this

Jimima:
@veloce I had a go at the tests. I can check that clicking on the king highlights the relevant squares as valid moves. I wanted to also test that you could/could not perform a castling move in line with the selected preference. I tried (for way too long) to find a way to make this work but couldn't really figure it out. I think I need to learn a bit more about how everything fits together regarding riverpod, testing, mocking, overrides etc. If you have any guidance here I would appreciate it.

In the process I wondered it in any case the widget should be using the shared state directly rather than being passed in. I think this would simplify mocking the state in the tests but I'm not really sure.

Would really appreciate any guidance you can give me on how to progress this. I also appreciate you are likely very busy so don't feel you need to prioritise responding to me!

veloce:
Will look at it asap @Jimima

lib/src/model/settings/board_preferences.dart
```diff
@@ -347,6 +354,19 @@ enum ClockPosition {
   };
 }
 
+enum CastlingMethod {
+  kingOverRook,
+  kingTwoSquares,
+  either;
+
+  String castlingMethodl10n(BuildContext context, CastlingMethod castlingMethod) =>
```
veloce, line 362: The preferred way is:
```dart
String l10n(AppLocalizations l10n) => switch (this) {
         CastlingMethod.kingOverRook => l10n.preferencesCastleByMovingOntoTheRook,
        CastlingMethod.kingTwoSquares => l10n.preferencesCastleByMovingTwoSquares,
        CastlingMethod.either => 'Either', //TODO l10n string
}
```

We don't want to pass `BuildContext` and other enums in the app have the same method named `l10n`

lib/src/view/settings/board_settings_screen.dart
```diff
@@ -71,6 +71,30 @@ class _Body extends ConsumerWidget {
                 }
               },
             ),
+            SettingsListTile(
+              settingsLabel: Text(
+                context.l10n.preferencesCastleByMovingTheKingTwoSquaresOrOntoTheRook,
+              ),
+              settingsValue: boardPrefs.castlingMethod.name,
```
veloce, line 78: You should use the `l10n` method of the `CastlingMethod` enum instead.

lib/src/view/settings/board_settings_screen.dart
```diff
@@ -263,6 +287,44 @@ class PieceShiftMethodSettingsScreen extends ConsumerWidget {
...
 }
 
+class CastlingMethodSettingsScreen extends ConsumerWidget {
+  const CastlingMethodSettingsScreen({super.key});
+
+  static Route<dynamic> buildRoute(BuildContext context) {
+    return buildScreenRoute(
+      context,
+      screen: const CastlingMethodSettingsScreen(),
+      title: 'Castling method',
```
veloce, line 297: Use the translated string: preferencesCastleByMovingTheKingTwoSquaresOrOntoTheRook

test/widgets/interactive_board_test.dart
```diff
@@ -0,0 +1,150 @@
...
+  final setup = Setup.parseFen('r3k2r/8/8/8/8/8/8/R3K2R w KQkq - 0 1');
+  Position pos = Chess.fromSetup(setup);
+
+  final gameData = GameData(
+    playerSide: PlayerSide.white,
+    sideToMove: Side.white,
+    validMoves: makeLegalMoves(pos),
+    promotionMove: null,
+    onMove: (NormalMove move, {bool? isDrop}) {
+      //This doesn't work because the position is not updated in the UI
```
veloce, line 21: For this to work it would need to be part of a StatefulWidget and be called inside `setState`.

Ideally we'd want to have an higher level test on a whole screen, and that would solve this. This is preferable because if we refactor and remove the `InteractiveBoardWidget`, the tests for the whole screen will still work and be relevant.

You could for instance test this on the `AnalysisScreen`, because it is easy to build a test screen with a PGN that gives you a position ready to castle.

You could add a
```dart
group('respect castling preferences', () {
   // add your tests here
});
```

I can take care of the GameScreen tests if you want, since it's a little bit harder to setup (need to mock websocket response). There are already some examples on how to do this though.

Jimima:
Thanks for the review, really helpful comments. I will take a look 👍🏻

Jimima:
Oh and I don't really know why the checks are failing either 👎🏻

veloce:
@Jimima to override the preferences you have the [defaultPreferences](https://github.com/lichess-org/mobile/blob/d285854e238b6d1dfd12caa3eb4f0bc9400b32f2/test/test_provider_scope.dart#L63) parameter of the helper function `makeTestProviderScopeApp`

Here is an example of its usage for the engine test:

https://github.com/lichess-org/mobile/blob/d285854e238b6d1dfd12caa3eb4f0bc9400b32f2/test/view/engine/test_engine_app.dart#L56

Jimima:
@veloce added tests on the Analysis screen, please feel free to review when able.

test/view/analysis/analysis_screen_test.dart
```diff
@@ -488,8 +489,214 @@ void main() {
       expect(find.byType(BoardShapeWidget), findsNothing);
     });
   });
+
+  group('Respect castling preference', () {
+    testWidgets('Castle kingside - king over rook', (tester) async {
+      final castlingMethodVariantUnderTest = castlingMethodVariant.currentValue!;
```
veloce, line 495: Test variants are already used to test the different platforms (android, iOS). So here I'd rather use a for loop:
```dart
for (final variant in CastlingMethod.values) {
   testWidgets('Castle kingside - king over rook with pref: $variant', (tester) async {
       ...
    }
}
```

veloce:
Sorry @Jimima I had to refactor, as I didn't want to parse the fen a second time for nothing in `InteractiveBoardWidget`. I also added tests for game screen.

I realise this PR was much harder to do than it was supposed to, because the castling option was not planned from the beginning.

Jimima:
That's cool, I was struggling to find the time to work on it in any case! I'll review what you did just for my learning, thanks

Issue #1329, opened 2024-12-31
Correspondence confirm move behavior


deathanchor:
### What is the problem?

In a correspondence game with confirm move option. If you make a move, hit back arrow, confirm move dialog stays. If you hit X to decline, it backs out the opponents move. If you hit confirm, it will appear you didn't make a move and now opponent's turn. Exiting the game view, and viewing game again shows correct state.

### Steps to reproduce the bug

In a correspondence game:

1. make a move
2. hit back arrow
3. confirm move dialog stays (unexpected behavior)
4. If you hit X to decline, it backs out the opponents move (unexpected behavior)
5. If you hit confirm, it will appear you didn't make a move and now opponent's turn. (unexpected behavior)
6. Exiting the game view, and viewing game again shows correct state.

Expected behavior:
Pressing back move button should remove move confirmation question.

### App version

v0.12.6

### Device Info

Android 14

### Additional information

Screen capture of replicated issue attached.

https://github.com/user-attachments/assets/4f471089-37f7-4077-b48f-6deeb99a41d2

PR #1348, opened 2025-01-08
Fix: Correspondence confirm move behavior #1329

vimarisoft:
Fixes bug #1329 

Confirm move dialog does not disappeared when back arrow is pressed to visualize previous positions (it disappears correctly in web app).  It's a problem not only related to correspondence chess, but to all time controls  where "Move confirmation" preference is set.

To fix it, an additional condition is added (in GamePlayer widget) to show the ConfirmMove widget:  'canGoForward ==false' (the same condition used to disable the forward button).  To achieve this, an additional parameter 'canGoForward' is added to the GamePlayer widget constructor..


vimarisoft:
Closed because of a dart formatting error in test, submitted new PR with  formatting fixed.

tom-anders:
FYI, it's totally fine (and probably preferred) to push additional commits (or use `git commit --amend) instead of opening a new PR

Issue #457, opened 2024-01-08
Deep linking


veloce:
Add support for deep linking.

We should decide with care what resources would be deep linkable eventually, because we don't want to frustrate users by having a link opening in the app with an incomplete feature.

Non exhaustive list of resources that should eventually have a deep link:
- puzzle
- game
- opening explorer
- tv channel
- streak
- storm
- ...

tom-anders:
@veloce any objections against me starting with an implementation? I think now that we have studies and broadcasts, this feature becomes especially useful

veloce:
@tom-anders no objection. 

This can be a touchy subject though, so we must be careful with what we're doing here. With the old app we've had reports that some users were pissed off with the way the link worked and eventually would deinstall the app.
That's why we should have a conservative approach here, and only implement link per use case (per screen).

Couple of notes:
- flutter recommend using a router package such as go_router, we won't do that;
- we'll keep the existing Navigator based system
- it means that a deep link will just push a root nav screen over the current screen, no matter which one it is

tom-anders:
Got it, thanks for the notes 👍

ijm8710:
Veloce, just curious, were there any prior use report cases cited when the user claimed the need to deinstall in past that seemed reasonable and actionable to slightly keep in mind?

I think a lot of times users can be a little too finicky with their desires and perhaps your precious post already gave Tom the inspiration on what to avoid or how to lean this, but was curious if there were a little bit more of specifics on why some felt so touchy on the redirects as I always felt they were generally fine in the old app and was always curious what was setting them off in past

tom-anders:
> * flutter recommend using a router package such as go_router, we won't do that;
> * we'll keep the existing Navigator based system
> * it means that a deep link will just push a root nav screen over the current screen, no matter which one it is

Ok, looks like https://pub.dev/packages/app_links is what we want to use then (the official docs also recommend it as an alternative)

veloce:
Closing as the app side is done now. We are still missing the apple-app-site-association but this is a server side thing.

Now all deep links are available for now but we'll track the remaining ones in separate issues.

PR #1353, opened 2025-01-12
feat: deep links (android only)

tom-anders:


pubspec.yaml
```diff
@@ -10,6 +10,7 @@ environment:
   flutter: "3.28.0-0.1.pre"
 
 dependencies:
+  app_links: ^6.3.3
```
veloce, line 13: Is an external plugin required to support deep links? I think not, according to flutter documentation. It would be better to do without imo.

tom-anders: Hmm I'm not sure, the flutter docs recommend this package as an alternative to using go_router (see note here https://docs.flutter.dev/cookbook/navigation/set-up-app-links#2-modify-androidmanifest-xml)

I'll check if there's a way to get the URI Stream without a plugin

tom-anders: Yeah I thought app_links is needed to actually get access to the link URL - haven't checked yet if there is a native way do get it, there probably is, I'll have a look

tom-anders: Right, found the way to do it without dependencies.

I only added the implementation for Android (as I'm neither familiar with Swift/iOS, nor do I have the hardware to actually test this)

android/app/src/main/kotlin/org/lichess/mobileV2/MainActivity.kt
```diff
@@ -48,6 +51,28 @@ class MainActivity: FlutterActivity() {
...
+          else -> {
+            result.notImplemented()
+          }
+      }
+    }
+  }
+
+  override fun onCreate(savedInstanceState: Bundle?) {
+    super.onCreate(savedInstanceState)
+  }
```
julien4215, line 70: Does this code is useful ? I don't understand why it is there

lib/src/utils/app_links.dart
```diff
@@ -0,0 +1,76 @@
...
+  @override
+  Future<Uri?> build() async {
+    final initialLink = await _channel.invokeMethod<String?>('getAppLink');
+    _logger.info('Initial app link: $initialLink');
+
+    _channel.setMethodCallHandler((call) async {
+      if (call.method == 'newAppLink') {
+        final uri = Uri.parse(call.arguments as String);
+        _logger.info('Received new app link: $uri');
+        state = AsyncValue.data(uri);
```
julien4215, line 70:
```suggestion
        state = AsyncData(uri);
```

lib/src/navigation.dart
```diff
@@ -127,6 +129,20 @@ class BottomNavScaffold extends ConsumerWidget {
   Widget build(BuildContext context, WidgetRef ref) {
     final currentTab = ref.watch(currentBottomTabProvider);
 
+    ref.listen(appLinksProvider, (_, appLinkUri) {
+      switch (appLinkUri) {
+        case AsyncData(:final Uri value):
```
julien4215, line 134: You can do this if you want to match only non null value
```suggestion
        case AsyncData(:final value?):
```

android/app/src/main/kotlin/org/lichess/mobileV2/MainActivity.kt
julien4215, line 70: Is this override useful ?

lib/src/utils/app_links.dart
julien4215, line 70: I would use the `AsyncData` constructor
```suggestion
        state = AsyncData(uri);
```

lib/src/navigation.dart
```diff
@@ -127,6 +129,20 @@ class BottomNavScaffold extends ConsumerWidget {
...
+          {
```
julien4215, line 135: I believe these brackets can be removed

julien4215 (commented):
Nice work ! I made some small comments on the code.

One small thing I noticed when testing it is that you can see the home screen before the one requested by the app link. Also when you press the back button you always return to the home even if you opened a study via a app link. Maybe that can be avoided by listening to the provider higher in the widget tree but it will make the code more complex.

julien4215:
I tried to use the Flutter's default deeplink handler [here](https://github.com/lichess-org/mobile/compare/main...julien4215:mobile:deeplink-flutter-handler). 

The advantage is that there is no need to write native source code and using Flutter method channels as it is already done by the default Flutter's default deeplink handler. I couldn't find any documentation on how it works, but from what I saw in the [source code](https://github.com/flutter/flutter/blob/bb2d34126cc8161dbe4a1bf23c925e48b732f670/engine/src/flutter/shell/platform/android/io/flutter/embedding/android/FlutterActivityAndFragmentDelegate.java#L935-L949) it does the same as this PR. The route information is pushed in this platform channel [NavigationChannel](https://api.flutter.dev/javadoc/io/flutter/embedding/engine/systemchannels/NavigationChannel.html).

The problem is that I am currently pushing only one route when a app link is used so for example if a user clicks on a puzzle link there isn't the back button to go back to puzzle tab screen as only the puzzle screen route was pushed. Maybe using the go router package would make it easier to push and pop the correct screens when dealing with app links.

lib/src/navigation.dart
tom-anders: huh, are you sure? I though `case AsyncData(:final Uri value)` would match `if and only if` the `uri` is not `null`, while `case AsyncData(:final value?)` would also match the `null` value, so you'd need an additional `if (uri != null)` check in that case

android/app/src/main/kotlin/org/lichess/mobileV2/MainActivity.kt
tom-anders: ah, I initially had some logic in here but then moved it to `configureFlutterEngine()` - so this can be removed, yes

tom-anders:
> I tried to use the Flutter's default deeplink handler [here](https://github.com/lichess-org/mobile/compare/main...julien4215:mobile:deeplink-flutter-handler).
> 
> The advantage is that there is no need to write native source code and using Flutter method channels as it is already done by the default Flutter's default deeplink handler. I couldn't find any documentation on how it works, but from what I saw in the [source code](https://github.com/flutter/flutter/blob/bb2d34126cc8161dbe4a1bf23c925e48b732f670/engine/src/flutter/shell/platform/android/io/flutter/embedding/android/FlutterActivityAndFragmentDelegate.java#L935-L949) it does the same as this PR. The route information is pushed in this platform channel [NavigationChannel](https://api.flutter.dev/javadoc/io/flutter/embedding/engine/systemchannels/NavigationChannel.html).

Ah very cool! Yeah documentation is a bit lacking there, I thought you could only use the default link handler if you also use the `go_router` package, but apparently not! Does it also work for a cold start? i.e. force close the app, and then open it directly via a link?

> The problem is that I am currently pushing only one route when a app link is used so for example if a user clicks on a puzzle link there isn't the back button to go back to puzzle tab screen as only the puzzle screen route was pushed. Maybe using the go router package would make it easier to push and pop the correct screens when dealing with app links.

@veloce Was against using `go_router` [here](https://github.com/lichess-org/mobile/issues/457#issuecomment-2585579294) - Maybe we can just change to the correct tab and only then push the new screen?

lib/src/navigation.dart
tom-anders: ah, my C++ habits, you're right

tom-anders:
> Nice work ! I made some small comments on the code.
> 
> One small thing I noticed when testing it is that you can see the home screen before the one requested by the app link. Also when you press the back button you always return to the home even if you opened a study via a app link. Maybe that can be avoided by listening to the provider higher in the widget tree but it will make the code more complex.

Thanks for the review :)

lib/src/navigation.dart
julien4215: Yes the notation also leaves me perplex but it is documented [here](https://dart.dev/language/pattern-types#null-check).

Both solution works but you don't have to specify the type with the null check pattern type.

julien4215:
Yes the default link handler also works with a cold start.

According to the documentation there are some limitations to use only the `Navigator` described [here](https://docs.flutter.dev/ui/navigation#limitations). From what I understand we will not be able to push multiple screens or change the tab with app links, at least without using some hacks. So for instance, if a user presses an app link that leads to a broadcast game, when he will press the back button he will be brought back to the home tab screen instead of the broadcast round screen.

veloce:
For the record, when I was prototyping the app, I decided against using go_router for 2 reasons:
1. the package was not really ready at the time, still missing a key feature, state persistence across tabs
2. I though it would add much more unnecessary complexity for unclear added value

Now, even with the deep link I think we still should not use go_router. First it would be a very big change, and second I think the limitations are not a deal breaker in our case. If you want to support web, of course then it is mandatory. But for an app, I like the simple straightforward approach of a stack based navigation.

Choosing this approach we have to accept the limitations, and not adding hack to work around them. I think it is fine to push a screen from a deep link on top of the wrong tab. And when you go back, you just go back to where you were. I can be weird in some scenarios, maybe, as I cannot foresee every corner case or usage. But for the most part it will be enough. It will also force us to keep the deep link feature simple (ie: not handling too much cases), which is a good thing in my opinion.

veloce:
> I tried to use the Flutter's default deeplink handler [here](https://github.com/lichess-org/mobile/compare/main...julien4215:mobile:deeplink-flutter-handler).
> 
> The advantage is that there is no need to write native source code and using Flutter method channels as it is already done by the default Flutter's default deeplink handler. I couldn't find any documentation on how it works, but from what I saw in the [source code](https://github.com/flutter/flutter/blob/bb2d34126cc8161dbe4a1bf23c925e48b732f670/engine/src/flutter/shell/platform/android/io/flutter/embedding/android/FlutterActivityAndFragmentDelegate.java#L935-L949) it does the same as this PR. The route information is pushed in this platform channel [NavigationChannel](https://api.flutter.dev/javadoc/io/flutter/embedding/engine/systemchannels/NavigationChannel.html).

Thanks for trying this. I think we definitely should use the default flutter link handler.

> The problem is that I am currently pushing only one route when a app link is used so for example if a user clicks on a puzzle link there isn't the back button to go back to puzzle tab screen as only the puzzle screen route was pushed. Maybe using the go router package would make it easier to push and pop the correct screens when dealing with app links.

What does it mean? If we push the puzzle screen from a deep link there is no back button displayed on the screen? If that is the case it is indeed an issue. We should understand why it is the case and make sure we can go back from any screen pushed from a deep link. Using go_router to solve this problem sounds overkill, and I explained in another comment why I don't think we should use go_router.

julien4215:
Thanks for explaining why you don't want to go with go_router. I would also like to keep things simple but I have too say that was not easy to find how to use the flutter default deeplink handler. The Flutter documentation recommends using the router for most apps and they explain really little about how you should without it. Also the fact that we don't have only static routes but some with ids made it harder like /study/<id>.

Yes you understood the problem correctly but it was only happening with a cold start of the app using a deeplink. I was able to fix the problem by using `onGenerateInitialRoutes`.

tom-anders:
@julien4215 Do we want to open a new PR with your version? I have no problem with you taking over the work :)

julien4215:
I don't know, I did not really intended to work on deep links. I just looked at your PR because I was interested to see how it works and well I did a bit more than I intended.

tom-anders:
> I don't know, I did not really intended to work on deep links. I just looked at your PR because I was interested to see how it works and well I did a bit more than I intended. Feel free to take over my work too if you want.

That's fine as well, just wanted to avoid duplicate work 😅

julien4215:
Go ahead then ! I'll focus on bookmark.

lib/src/navigation.dart
tom-anders: Ok wow, that is really unintuitive, especially when coming from other languages

tom-anders:
Ok, changed the implementation to user Flutter's native deep link handling, thanks again @julien4215 for the hint!

For Android, the server side change has already been done (see https://lichess.org/.well-known/assetlinks.json).

For iOS, the server side change also needs to include the supported paths (see https://docs.flutter.dev/cookbook/navigation/set-up-universal-links#create-and-host-apple-app-site-association-json-file), so this should probably only be done once this PR is merged and we have decided which paths to support.

android/app/src/main/AndroidManifest.xml
```diff
@@ -53,6 +53,22 @@
...
             </intent-filter>
+
+            <intent-filter android:autoVerify="true">
+                <action android:name="android.intent.action.VIEW" />
+                <category android:name="android.intent.category.DEFAULT" />
+                <category android:name="android.intent.category.BROWSABLE" />
+
+                <data android:scheme="https" android:host="lichess.org" />
+
+                <data android:pathPattern="/training/....." />
```
veloce, line 64: Might as well add `streak` and `storm` since they are fully implemented on the app?

lib/src/utils/app_links.dart
veloce, line 1: It is kind of a router for deep links, so I wouldn't put that in the `utils` folder.
I'd either:
- make a `lib/src/navigation/` folder with that file inside and the `lib/src/navigation.dart` renamed to sth like `bottom_tabs.dart`.
- put it directly in `lib/src`

android/app/src/main/AndroidManifest.xml
tom-anders: why not, done :+1:

lib/src/utils/app_links.dart
tom-anders: i went with the 2nd option as it was easier :)

PR #1365, opened 2025-01-16
Fix iOS broadcast test

julien4215:
The test for iOS is failing because the finder in `scrollUntilVisible` needs to match a single element and for iOS there are two text widgets that appears with _Completed_ so I don't know how to fix this test. Maybe we can just remove it or make a separate test for Android and iOS ?

20250116_12h29m31s_grim

### DIFF
--- a/lib/src/view/broadcast/broadcast_list_screen.dart
+++ b/lib/src/view/broadcast/broadcast_list_screen.dart
@@ -197,7 +197,7 @@ class _BodyState extends ConsumerState<_Body> {
         slivers: [
           for (final section in sections)
             SliverMainAxisGroup(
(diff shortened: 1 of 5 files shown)


julien4215:
I solved the issue by using the key of the `SilverMainAxisGroup`. I changed its value because the previous one was quite unpractical. It seemed to not be used so I hope I don't break anything by changing its value.

veloce:
Not sure to understand what is failing here since the tests pass on CI main branch, and on my machine.

julien4215:
You need to trigger the tests action manually. They were not rerun since the fix was merged on the main branch.

veloce (approved):
Thanks

PR #1366, opened 2025-01-18
Fix CI tests 

julien4215:
Fix CI tests by installing SQlite dependencies on Ubuntu that are required for Dart package sqflite_common_ffi


veloce:
I don't get why it suddenly doesn't work anymore. Do you have an explanation?

julien4215:
No I don't have one. But if I had to guess, I would say it's because the Ubuntu version used by the workflow has changed and that in this new version there isn't SQLite by default.

Issue #755, opened 2024-06-07
Add Arrow Button Indication when you are reviewing previous moves during an Active Game


EmmetSchuler:
In the old Mobile app and the Website, when reviewing through the previous moves of an active game, the forward arrow button will blink blue to indicate that you are not viewing the current position of the game.
This functionality is currently missing in the beta Mobile App.

okmsbun:
I believe that the mobile app currently has only two buttons: "Forward" and "Backward." The absence of "Go to Start" and "Go to End" options might be a deliberate choice, likely for space-saving or UI/UX considerations. However, I wonder how adding these options would affect the experience. Could we explore where they might fit into the interface?

### Alternative Suggestions:
- Could these options be integrated into the existing **menu button** (located at the bottom left)? This would add functionality without cluttering the main interface.  
- Alternatively, the "Forward" button could change color or blink to indicate that the user is not at the latest position in the game. This subtle visual cue might make navigation more intuitive.  
- Additionally, we know that long-pressing the "Forward" button already jumps to the latest move. Perhaps this feature could be made more apparent to users.

### Conclusion:
It would be great to discuss how these features can be added or enhanced in a way that maintains the simplicity and effectiveness of the mobile UI. With more feedback, we can create an optimal solution.

tom-anders:
Closed via #1371 I think

PR #1371, opened 2025-01-24
Add blink effect to next button based on game step cursor

okmsbun:
Issue: https://github.com/lichess-org/mobile/issues/755
This PR adds a blink effect to the next button when the user is not at the latest move during game reviews.

| before | after | lichess.org |
| Screenshot_1737730675 | untitled | Screen-Recording-2025-01-24-at-18 22 39 |

### DIFF
--- a/lib/src/view/game/game_body.dart
+++ b/lib/src/view/game/game_body.dart
@@ -603,6 +603,7 @@ class _GameBottomBar extends ConsumerWidget {
                 label: context.l10n.next,
                 icon: CupertinoIcons.chevron_forward,
                 showTooltip: false,
+                blink: gameState.stepCursor != gameState.game.steps.length - 1,
               ),
             ),
           ];
--- a/lib/src/view/game/game_body.dart
+++ b/lib/src/view/game/game_body.dart
@@ -603,7 +603,9 @@ class _GameBottomBar extends ConsumerWidget {
                 label: context.l10n.next,
                 icon: CupertinoIcons.chevron_forward,
                 showTooltip: false,
-                blink: gameState.stepCursor != gameState.game.steps.length - 1,
+                invertBackground:
+                    gameState.stepCursor != gameState.game.steps.length - 1 &&
+                    gameState.game.sideToMove == gameState.game.youAre,
               ),
             ),
           ];
--- a/lib/src/widgets/bottom_bar_button.dart
+++ b/lib/src/widgets/bottom_bar_button.dart
@@ -15,6 +15,7 @@ class BottomBarButton extends StatelessWidget {
     this.showLabel = false,
     this.showTooltip = true,
     this.blink = false,
+    this.invertBackground = false,
     this.tooltip,
     super.key,
   });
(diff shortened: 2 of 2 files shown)


okmsbun:
The grayed-out button provides some indication, but this PR addresses the issue described in #755. In the old mobile app and website, the blinking forward button made it clear when you weren’t on the latest move. Currently, only the countdown timer acts as an alert. If a user is reviewing previous moves and their opponent plays a move, they might think they are still on the current position. The blinking effect helps prevent this confusion and offers a more noticeable visual cue.

I’d love to hear any additional suggestions you might have!

ijm8710:
That's fair. Originally I was thinking not latest move as in you simply rewinded

I didn't quite put enough thought on that a new move has actually been played

okmsbun:
Should we also add the same blinking effect to the TV screen where users watch other people’s games live?

veloce:
I understand the need for a more visible clue that you're not on the last position. I don't really like the blinking though, it is too distracting.

The blinking could be replaced with an inverted background color (with opacity) for instance.

But the best solution to this issue in my opinion, would be, during a live game, to make the button blink if and only if a new move is received when the user is replaying moves. This is the only moment it is useful because you can't know a new move is added when you're not on the last position.

The logic is a little more complex (need to store the move index when starting replaying backwards), but worth it imo.

I don't think we should add this logic on TV, it is important enough only when playing (to avoid losing time).

tom-anders:
> I understand the need for a more visible clue that you're not on the last position. I don't really like the blinking though, it is too distracting.
> 
> The blinking could be replaced with an inverted background color (with opacity) for instance.
> 
> But the best solution to this issue in my opinion, would be, during a live game, to make the button blink if and only if a new move is received when the user is replaying moves. This is the only moment it is useful because you can't know a new move is added when you're not on the last position.
> 
> The logic is a little more complex (need to store the move index when starting replaying backwards), but worth it imo.

Agreed, I think it works on the website like this as well. 

> 
> I don't think we should add this logic on TV, it is important enough only when playing (to avoid losing time).

What about broadcast though? Might be helpful there

okmsbun:
I removed the blinking effect and replaced it with a background change using opacity. You can adjust the background color and transition duration as you like. A video demonstration is provided below.

I modified the logic as follows: If it's my turn and I am reviewing previous moves, the background will flash. In any other case, the background color will not change. Does this approach make sense?

Also, as you suggested, I didn’t store the move index. I believe it wasn’t necessary, and I managed to handle it without keeping track of the index. If I misunderstood your point, please correct me.

Regarding live TV, there's a bit of a dilemma: from whose perspective should we view it—Black or White? 😄 In what situation should the background flash? 😄 Since it's never our turn, we are just watching! 😄

That being said, could the background flash in live TV mode regardless of the perspective?


https://github.com/user-attachments/assets/f68cd069-58ab-4f84-bc98-bc58c86a7a36

veloce:
Thanks for the vid! It is much better now. If you managed to do it without an extra variable that is even better (I didn't really think about it so I guess I missed a simpler way to do it).

veloce:
> Regarding live TV, there's a bit of a dilemma: from whose perspective should we view it—Black or White? 😄 In what situation should the background flash? 😄 Since it's never our turn, we are just watching! 😄

We should not bother with TV imo.

Issue #1372, opened 2025-01-25
FR - add country name when you touch flag mear nickname


Mile-Lile:
When I tap flag under nickname of my opponent I want to see where my opponent is from? can you add this feature similar to web app...

![Image](https://github.com/user-attachments/assets/f45e695e-b05e-47af-a163-f576deed2a10)

ijm8710:
+1

Mile-Lile:
@okmsbun 
hey, one more thing... I see you are improving user expirience, and thank you for your work. Is there a way to implenet like it was on old app to see games against my opponent I played before.... there was number e.g. 3 and when I tap it, app shows 3 games I played against that oponent....

okmsbun:
Thank you for your suggestion! Unfortunately, I'm not familiar with the feature you mentioned. It might be better to open a new issue with more details so it can be addressed properly, as this issue seems to focus on a different topic. 😊

ijm8710:
While I'd also absolutely love this and even ticketed it a while back #820 I don't think it has anything to do with this item

CloudyDino:
@veloce I think this can be closed since we merged in https://github.com/lichess-org/mobile/pull/1373

PR #1373, opened 2025-01-26
Improve user profile location display logic

okmsbun:
Issue
------
[Issue #1372](https://github.com/lichess-org/mobile/issues/1372)

Description
-----------
This pull request introduces a feature that displays country or location information under the flag. It also provides multiple display modes: full (country + location), only country, only location, or none, and these details are always visible.

Comparison
----------
Below is a table comparing the old, new, and web versions:

|                 | Before                 | After                 | Web                 |
|-----------------|---------------------|---------------------|---------------------|
| Full            | old_full | new_full | web_full |
| Only Country    | old_only_country | new_only_counrty | web_only_country |
| Only Location   | old_only_location | new_only_lacation | web_only_lacation |
| None            | old_none | new_none | web_none |

If something here seems off or if there’s a different approach you recommend, please let me know. Your feedback is welcome!

### DIFF
--- a/lib/src/view/user/user_profile.dart
+++ b/lib/src/view/user/user_profile.dart
@@ -160,15 +160,15 @@ class Location extends StatelessWidget {
   Widget build(BuildContext context) {
     return Row(
       children: [
+        if (profile.location != null) ...[Text(profile.location!), const SizedBox(width: 5)],
         if (profile.country != null) ...[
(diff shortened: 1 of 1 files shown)


veloce (approved):

veloce:
Thanks for taking the time to make this comparison table. Very helpful!

PR #1381, opened 2025-01-27
fix: inconsistent hostname in dev docs

tom-anders:


julien4215:
Oops I think it's my bad. Maybe we should keep the `lichess.dev` url otherwise the note below doesn't make a lot of sense.

tom-anders:
Hmm but lichess.dev is already the default, so that would also be a bit weird

Boubou78000:
> Hmm but lichess.dev is already the default, so that would also be a bit weird

Went lichess.dev offline? I can't access it anymore...

PR #1382, opened 2025-01-28
When game ratingDiff is zero (e.g. draw), color should be yellow instead of red

cioccarellia:
In the game UI, when a draw happens between two players, the rating difference is zero.
The game colors the rating diff text with red, but that is inconsistent w.r.t. the home game list, where draw results are shown in yellow.
A more consistent color approach would be to have the zero rating delta in yellow (so that it more accurately represents the rating difference).


veloce:
Your code is not formatted properly, which is why the tests are failing. You can use `dart format` to format the code, or better, setup your editor to do it, thanks.

cioccarellia:
I tried running `dart format` on the file but it looks like maybe the formatter I used has different settings w.r.t. the one in the pipeline (sorry for the trouble, I'm not really familiar with dart/flutter)

veloce:
The formatter is configured for the project here: https://github.com/lichess-org/mobile/blob/de61aa4c1389b6e4a27bc616fe211fcaed091e93/analysis_options.yaml#L29

Your formatter normally would automatically pick these settings. Or if you run `dart format` from the project root.

ItsDang:
Hey @cioccarellia, I formatted the code change and updated context. I tested #1497 and added the pull request [#1](https://github.com/cioccarellia/mobile/pull/1) to your repository.  @veloce just a general question, is there a reason/difference behind using context.lichessColors vs LichessColors? AFAIK there is no difference?

veloce:
> is there a reason/difference behind using context.lichessColors vs LichessColors? AFAIK there is no difference?

yes because in `context.lichessColors`, colors are harmonised with the current color scheme.

veloce (changes requested):
This should use `context.lichessColors` and not directly `LichessColors`. Cf. other comment in the PR.

ItsDang:
@cioccarellia I added a pull request on your repo, reverting the context change.

veloce (approved):

PR #1386, opened 2025-01-28
feat: tournament list screen

tom-anders:
Layout is copied from the old app for now, we can always improve it later of course

[tournamentlist.webm](https://github.com/user-attachments/assets/7518192e-ce39-4f4c-af8f-70fe49f1129d)

### DIFF
--- a/lib/src/model/tournament/tournament.dart
+++ b/lib/src/model/tournament/tournament.dart
@@ -1,0 +1,89 @@
+import 'package:deep_pick/deep_pick.dart';
+import 'package:fast_immutable_collections/fast_immutable_collections.dart';
+import 'package:freezed_annotation/freezed_annotation.dart';
+import 'package:lichess_mobile/src/model/common/chess.dart';
+import 'package:lichess_mobile/src/model/common/id.dart';
+import 'package:lichess_mobile/src/model/common/perf.dart';
+import 'package:lichess_mobile/src/model/common/time_increment.dart';
+import 'package:lichess_mobile/src/model/user/user.dart';
+import 'package:lichess_mobile/src/utils/json.dart';
+
+part 'tournament.freezed.dart';
+
+typedef TournamentSchedule = ({String freq, String speed});
+
+typedef TournamentLists =
+    ({
+      IList<TournamentListItem> started,
+      IList<TournamentListItem> created,
+      IList<TournamentListItem> finished,
+    });
+
+extension TournamentExtension on Pick {
+  TimeIncrement asTimeIncrementOrThrow() {
+    final requiredPick = this.required();
+    return TimeIncrement(
+      requiredPick('limit').asIntOrThrow(),
+      requiredPick('increment').asIntOrThrow(),
+    );
+  }
+
+  TournamentSchedule asTournamentScheduleOrThrow() {
+    final requiredPick = this.required();
+    return (
+      freq: requiredPick('freq').asStringOrThrow(),
+      speed: requiredPick('speed').asStringOrThrow(),
+    );
+  }
+
+  IList<TournamentListItem> asTournamentListOrThrow() =>
+      asListOrThrow((pick) => TournamentListItem.fromServerJson(pick.asMapOrThrow())).toIList();
+}
+
+@freezed
+class TournamentListItem with _$TournamentListItem {
+  const TournamentListItem._();
+
+  const factory TournamentListItem({
+    required TournamentId id,
+    required String createdBy,
+    required String fullName,
+    required TimeIncrement timeIncrement,
+    required bool rated,
+    required Duration? timeToStart,
+    required DateTime startsAt,
+    required DateTime finishesAt,
+    required int? maxRating,
+    required int minutes,
+    required int nbPlayers,
+    required Perf perf,
+    required int position,
+    required TournamentSchedule schedule,
+    required Variant variant,
+    required LightUser? winner,
+  }) = _TournamentListItem;
+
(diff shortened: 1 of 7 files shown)


lib/src/view/tournament/tournament_list_screen.dart
```diff
@@ -0,0 +1,268 @@
...
+import 'package:lichess_mobile/src/widgets/list.dart';
+import 'package:lichess_mobile/src/widgets/platform.dart';
+import 'package:lichess_mobile/src/widgets/platform_scaffold.dart';
+
+class TournamentListScreen extends ConsumerWidget {
+  const TournamentListScreen({super.key});
+
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {
+    return FutureBuilder(
```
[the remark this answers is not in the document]
tom-anders: Oh, sorry for the delay, during rebasing I figured out the reason: It's not actually an `AsyncValue`, but a function returning `Future<TournamentLists>`, that's why it's done with `FutureBuilder`. Should I refactor to use a riverpod provider instead?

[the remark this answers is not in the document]
tom-anders: Refactored to a provider now

lib/src/model/tournament/tournament.dart
```diff
@@ -0,0 +1,89 @@
...
+import 'package:lichess_mobile/src/model/common/chess.dart';
+import 'package:lichess_mobile/src/model/common/id.dart';
+import 'package:lichess_mobile/src/model/common/perf.dart';
+import 'package:lichess_mobile/src/model/common/time_increment.dart';
+import 'package:lichess_mobile/src/model/user/user.dart';
+import 'package:lichess_mobile/src/utils/json.dart';
+
+part 'tournament.freezed.dart';
+
+typedef TournamentSchedule = ({String freq, String speed});
```
veloce, line 13: Is the speed specific to tournament? If not we want to use the `Speed` enum.

(If specific to tournament perhaps an enum is better also).

lib/src/model/tournament/tournament.dart
```diff
@@ -0,0 +1,89 @@
...
+      asListOrThrow((pick) => TournamentListItem.fromServerJson(pick.asMapOrThrow())).toIList();
+}
+
+@freezed
+class TournamentListItem with _$TournamentListItem {
+  const TournamentListItem._();
+
+  const factory TournamentListItem({
+    required TournamentId id,
+    required String createdBy,
```
veloce, line 49: I'd expect a `UserId` here.

lib/src/view/tournament/tournament_list_screen.dart
```diff
@@ -0,0 +1,271 @@
...
+
+  @override
+  void dispose() {
+    _tabController.dispose();
+    super.dispose();
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    return Scaffold(
```
veloce, line 89: Nested `Scaffolds` are better to be avoided.

lib/src/view/tournament/tournament_list_screen.dart
```diff
@@ -0,0 +1,271 @@
...
+    return CupertinoPageScaffold(
+      navigationBar: const CupertinoNavigationBar(),
+      child: SafeArea(
+        child: Column(
+          mainAxisSize: MainAxisSize.max,
+          crossAxisAlignment: CrossAxisAlignment.stretch,
+          children: [
+            Padding(
+              padding: Styles.bodyPadding,
+              child: CupertinoSlidingSegmentedControl<_ViewMode>(
```
veloce, line 138: Now that the `CupertinoPageScaffold` accepts a `bottom` parameter for the app bar, we don't do that anymore, which simplifies greatly the code.

So you can refactor to use a single `PlatformScaffold` and a single `TabBarView` for both platforms. See broadcast round screen or create custom game screen for examples.

veloce (commented):
Nice work, thanks!

lib/src/model/tournament/tournament.dart
tom-anders: Ah yes this seems to be specific to tournaments I think: https://github.com/lichess-org/lila/blob/ede805b16b1e9738c908e177562c742cea5b8865/modules/tournament/src/main/Schedule.scala#L86

But we don't actually use it right now, so let's just remove it? I added an enum for the `freq` field though (see https://github.com/lichess-org/lila/blob/ede805b16b1e9738c908e177562c742cea5b8865/modules/tournament/src/main/Schedule.scala#L60)

lib/src/model/tournament/tournament.dart
tom-anders: It actually contains the username, not the ID. Most notably, for all tournaments of `lichess.org/tournament` it contains `lichess`

lib/src/view/tournament/tournament_list_screen.dart
tom-anders: How is it nested? the parent widget is a `PlatformScaffold` only while loading, but not when it has finished loading (and is displaying the body)

tom-anders: Ah, I guess you mean there should only be one scaffold at the top, not in multiple places? Ok I also saw you other comment now, will do that

lib/src/view/tournament/tournament_list_screen.dart
tom-anders: Ah yes that makes the code much simpler indeed, thanks!

veloce (approved):

PR #1387, opened 2025-01-29
Replace Theme.of(context).someProperty with SomeProperty.of(context)

julien4215:


veloce:
Some code golf :)

Ok, one small conflict to solve and I'll merge this.

Issue #1309, opened 2024-12-25
Issue when restarting the clock tool


teefoox:
### What is the problem?

When you reset the clock when it's the top player turn, and you restart the clock, the clock will show it's the bottom player turn but it's the top player time that will decrease.

### Steps to reproduce the bug

1.  Go to the Clock Tool
2.  Start the clock with the bottom player playing first
3. Play some moves and stop the clock when it's top player turn.
4. Restart the clock
5. See that the colored indicator is for the bottom player while the time is decreasing for the top player

### App version

0.13.12

### Device Info

Android 14

### Additional information

At step 5, you have to press on the bottom side to "cancel" the bug. If you press on the top side nothing will happen.

A video to illustrate :

https://github.com/user-attachments/assets/2e5cc04f-041f-4c1d-8e1a-a2ce2eff06e2

flmng0:
Hey I have also experienced this issue.

I am unable to get Flutter properly installed on my system so I can't fix this myself, but it looks like an issue with this function:
[ClockToolController.reset](https://github.com/lichess-org/mobile/blob/84abbae0427e733921fec17d06e8c89089a7c17c/lib/src/model/clock/clock_tool_controller.dart#L109)

The method does not reset the active side on the `_clock` member of `ClickToolController`.

Again, I'm unfortunately unable to test but @maintainers I hope the above information helps.

PR #1390, opened 2025-01-30
Bugfix for Issue #1309: Always start clock for white

CloudyDino:
Close https://github.com/lichess-org/mobile/issues/1309

When we start the clock, we should always start it for white. When the clock is paused during black's turn and we reset, we reset the `activeSide` for the `state`, but not for the `_clock`

**Recordings**

Before

https://github.com/user-attachments/assets/9c572ec9-c582-48c5-9b4a-2ff322c3de8b

After

https://github.com/user-attachments/assets/8ec20366-d1cb-49fd-b3c8-7594797b2496


CloudyDino:
Hi @veloce! I believe this PR is ready for review, could you take a look?

PR #1394, opened 2025-01-31
[Feat] Load position in board editor

### LINKED ISSUE #991: [Feature] Add a "Load position" button to board editor menu
This button should open a selection popup with positions to choose from.

Positions should include openings, endgames, and a "clear board" option.

For reference, here are the endgame and opening positions that the old app supports: https://github.com/lichess-org/lichobile/tree/master/www/data - We could either load this file as json, or convert it to Dart code.

@veloce I think this is a candidate for the "Good first issue" label

vincendep:
Hi there, this PR contains a raw implementation of the load position feature. 

-----------

closes #991

### DIFF
--- a/lib/src/view/board_editor/board_editor_menu.dart
+++ b/lib/src/view/board_editor/board_editor_menu.dart
@@ -1,10 +1,18 @@
-import 'package:dartchess/dartchess.dart';
+import 'dart:convert';
+
+import 'package:dartchess/dartchess.dart' hide Position;
 import 'package:flutter/material.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
+import 'package:lichess_mobile/src/constants.dart';
 import 'package:lichess_mobile/src/model/board_editor/board_editor_controller.dart';
+import 'package:lichess_mobile/src/model/board_editor/position.dart';
 import 'package:lichess_mobile/src/styles/styles.dart';
 import 'package:lichess_mobile/src/utils/l10n_context.dart';
+import 'package:lichess_mobile/src/utils/navigation.dart';
 import 'package:lichess_mobile/src/widgets/adaptive_bottom_sheet.dart';
+import 'package:lichess_mobile/src/widgets/buttons.dart';
+import 'package:lichess_mobile/src/widgets/list.dart';
+import 'package:lichess_mobile/src/widgets/platform_scaffold.dart';
 
 class BoardEditorMenu extends ConsumerWidget {
   const BoardEditorMenu({required this.initialFen, super.key});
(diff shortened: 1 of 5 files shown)


vincendep:
Hi @veloce I need some advice on how to land this feature. First of all speaking about the UX. imo navigating to a full screen selection list is far way better than using modals when many items are present, let me know if you don t agree. Other than that, I would like to have some indication on how to organize the code for a proper data fetching implementation. Thanks

veloce:
Yes, full screen selection is fine! Will look at the code asap.

lib/src/view/board_editor/board_editor_menu.dart
```diff
@@ -19,6 +27,42 @@ class BoardEditorMenu extends ConsumerWidget {
...
+          padding: Styles.horizontalBodyPadding,
+          child: Wrap(
+            alignment: WrapAlignment.spaceEvenly,
+            children: [
+              SecondaryButton(
+                semanticsLabel: context.l10n.loadPosition,
+                child: Text(context.l10n.loadPosition),
+                onPressed: () {
+                  final notifier = ref.read(editorController.notifier);
+                  pushReplacementPlatformRoute(
```
veloce, line 40: This function doesn't exist anymore. You should update to the last `main` commit. Now we're using directly `Navigator.of(context).push` and we define route as static method of `Screen` type widgets using the `buildScreenRoute` function.

lib/src/view/board_editor/board_editor_menu.dart
```diff
@@ -93,3 +137,129 @@ class BoardEditorMenu extends ConsumerWidget {
...
+class _OpeningsTabState extends State<_OpeningsTab> {
+  // TODO Provider
+  late final Future<List<Position>>? _openings;
+
+  @override
+  void initState() {
+    _openings = DefaultAssetBundle.of(context).loadString('assets/positions.json').then((s) {
+      final List<Position> result = [];
+      for (final opening in jsonDecode(s) as List) {
+        // ignore: avoid_dynamic_calls
```
veloce, line 186: better to remove that and use `as List<Map<String, dynamic>>` instead.

lib/src/view/board_editor/board_editor_menu.dart
```diff
@@ -93,3 +137,129 @@ class BoardEditorMenu extends ConsumerWidget {
...
+  const _EndGamesTab({required this.onPositionSelected});
+
+  final void Function(Position position) onPositionSelected;
+
+  @override
+  State<_EndGamesTab> createState() => _EndGamesTabState();
+}
+
+class _EndGamesTabState extends State<_EndGamesTab> {
+  // TODO Provider
```
veloce, line 229: remove TODO

lib/src/view/board_editor/board_editor_menu.dart
```diff
@@ -93,3 +137,129 @@ class BoardEditorMenu extends ConsumerWidget {
...
+  late final Future<List<Position>>? _endGames;
+
+  @override
+  void initState() {
+    _endGames = DefaultAssetBundle.of(context).loadString('assets/endgames.json').then((s) {
+      final List<Position> result = [];
+      for (final position in jsonDecode(s) as List) {
```
veloce, line 236: cast the explicit type of `List`.

veloce (commented):
I think your implementation if perfectly fine. Using StatefulWidgets instead of providers in that case is valid. You could have used providers, but I don't think we'd gain anything in using providers compared to StatefulWidgets. Also, I don't expect these positions to be loaded elsewhere in the app in the future.

Good job!

vincendep:
Thanks @veloce, my point about the providers was just to easily cache data loaded and parsed from files.

lib/src/widgets/platform_scaffold.dart
```diff
@@ -70,6 +70,7 @@ class PlatformAppBar extends StatelessWidget {
     this.leading,
     this.actions = const [],
     this.androidTitleSpacing,
+    this.bottom,
```
veloce, line 73: Thanks for letting me know about this feature of CupertinoNavigationBar I wasn't aware of ;)

Now putting that in `PlatformScaffold` might lead to some issue. The way the platform scaffold wraps PlatformBottomBar in PreferredSize is not ideal, I will refactor this. Also I tested the `bottom` property of `CupertinoNavigationBar`, and it still has some issues with routes transition, currently being fixed in flutter repo.

I think PlatformScaffold should stay very limited, and don't support too much use cases. For now I'd say lets remove the `bottom` from platform scaffold and use separately `Scaffold` and `CupertinoPageScaffold` in your PR.

PR #1409, opened 2025-02-05
Added emergency sound to the clock tool

sdalvi10:
For #1307.

The `lowTime` sound will be played when the clock reaches 10% of the total duration.


veloce (approved):

veloce:
Thanks!

PR #1413, opened 2025-02-06
Add unbookmark action in context menu on game bookmark screen

julien4215:


lib/src/model/game/game_repository.dart
```diff
@@ -101,7 +101,7 @@ class GameRepository {
             },
           ),
           headers: {'Accept': 'application/x-ndjson'},
-          mapper: (json) => LightArchivedGame.fromServerJson(json),
+          mapper: (json) => LightArchivedGame.fromServerJson(json, isBookmarked: true),
```
veloce, line 104: I don't get this. The server should already return the right value, and we should not have to override it client side.

julien4215: The server only sends bookmark games so it doesn't add the field `bookmarked: true` in the json as it would just wastes bandwidth.

PR #1416, opened 2025-02-09
feat: two-column tree view

tom-anders:
[twoColumn.webm](https://github.com/user-attachments/assets/a730b7cb-b62b-455a-9dce-c56d7862e7a0)

### DIFF
--- a/lib/src/model/analysis/analysis_preferences.dart
+++ b/lib/src/model/analysis/analysis_preferences.dart
@@ -61,6 +61,10 @@ class AnalysisPreferences extends _$AnalysisPreferences with PreferencesStorage<
   Future<void> setEngineSearchTime(Duration engineSearchTime) {
     return save(state.copyWith(engineSearchTime: engineSearchTime));
   }
+
+  Future<void> toggleInlineNotation() {
+    return save(state.copyWith(inlineNotation: !state.inlineNotation));
+  }
 }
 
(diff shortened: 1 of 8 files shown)


veloce:
@tom-anders really cool, thanks!

I'd suggest an improvement (maybe): we could try to have a full width background color for the non-inline moves (including the move index and padding).
Like what the website does actually. I think that way it would look even better.

tom-anders:
> @tom-anders really cool, thanks!
> 
> I'd suggest an improvement (maybe): we could try to have a full width background color for the non-inline moves (including the move index and padding). Like what the website does actually. I think that way it would look even better.

Sure, I’ll try it out. Do we want to use the new background for the old inline notation view as well then? To be consistent?

tom-anders:
> Did your originally analysis notation refactoring from a bit earlier prevent stray ( from sitting at end of line. And if so should that carry over here as well. It's decently minor but still slightly odd to see. Not sure if you have any thoughts to have some logic for it

Not 100% sure if we already have logic for that or not, but I don't think so...? Can you open a separate issue for that?

tom-anders:
Okay, adjusted the background and also added some horizontal dividers:

Dark Theme:
Screenshot_1739224202

Light Theme:
Screenshot_1739224246

Website for comparison:

Screenshot_2025-02-10_22-51-34
Screenshot_2025-02-10_22-51-52

PR #1425, opened 2025-02-10
Enable ongoing games

CloudyDino:
Ongoing games have been fully disabled, was this on purpose @veloce? I see that it's based on [this commit from last week](https://github.com/lichess-org/mobile/commit/41e7ebfca707007a241222123060304150a90f14).
I haven't been able to test ongoing games using my new accounts, but this allows us to see them.

Before

https://github.com/user-attachments/assets/3a167b77-0e6d-4448-95d7-6084080a6083

After

https://github.com/user-attachments/assets/6f52cfb3-b80d-4293-92f5-535493b42b84

### DIFF
--- a/devtools_options.yaml
+++ b/devtools_options.yaml
@@ -1,0 +1,3 @@
+description: This file stores settings for Dart & Flutter DevTools.
+documentation: https://docs.flutter.dev/tools/devtools/extensions#configure-extension-enablement-states
+extensions:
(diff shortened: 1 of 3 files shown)


veloce:
The idea was that we can move the ongoing game widget order, but cannot disable it.

So this is more a data migration issue I believe, even though I tested this.

Instead of relying on stored data we should hardcode the fact ongoing games are always enabled.

PR #1427, opened 2025-02-11
Add a filter to show only playing games in a broadcast round

julien4215:
[broadcast_playing_toggle_3.webm](https://github.com/user-attachments/assets/4910a717-88a6-4734-9c9b-2bb0d7213911)

### DIFF
--- a/lib/src/view/broadcast/broadcast_boards_tab.dart
+++ b/lib/src/view/broadcast/broadcast_boards_tab.dart
@@ -29,11 +29,13 @@ class BroadcastBoardsTab extends ConsumerWidget {
     required this.tournamentId,
     required this.roundId,
     required this.tournamentSlug,
+    required this.showOnlyOngoingGames,
   });
 
   final BroadcastTournamentId tournamentId;
   final BroadcastRoundId roundId;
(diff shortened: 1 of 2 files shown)


julien4215:
I don't think that would look good to add a board count to the scrollable bottom sheet, at least I don't see a good way to do it so I think I will leave it like this for now

veloce:
The problem I see here, is that one of the settings is persisted, while the other is not.

I wish they were part of a different menu (or separate in another way) to mark this distinction. And I wish that any "cog" button were corresponding to persisted settings.

I guess the playing game toggle could be displayed on top of the list. And to make it distinct it could be a checkbox instead of a toggle.

julien4215:
I tried to replicate the choices that were made on the website. There is no distinction also on the website and I think it is fine like this because no one is going to want a persisted setting for playing games.

veloce:
Actually I know how it should be done. With a filter button, like in the study list screen or new one I've added in the broadcast list screen. 

Website has its own UI patterns, we don't have to follow them. This is an app and we need to think of making things consistent within the app.

Screenshot_1739359076

Issue #1408, opened 2025-02-05
Broadcast bug for flip board


ijm8710:
flipping board on live game flips players but not clocks. The clock also gets wonky and if you keep switching back and forth it keeps deducting from itself

PR #1437, opened 2025-02-13
Fix clock when board is flipped

julien4215:
Close https://github.com/lichess-org/mobile/issues/1408

The fix done in the PR #1428 could not work. This time it should really be fixed.

### DIFF
--- a/lib/src/view/analysis/analysis_layout.dart
+++ b/lib/src/view/analysis/analysis_layout.dart
@@ -1,3 +1,4 @@
+import 'package:dartchess/dartchess.dart';
 import 'package:flutter/material.dart';
 import 'package:lichess_mobile/l10n/l10n.dart';
 import 'package:lichess_mobile/src/constants.dart';
@@ -121,6 +122,7 @@ class AnalysisLayout extends StatelessWidget {
     this.tabController,
     required this.boardBuilder,
     required this.children,
+    required this.pov,
     this.boardHeader,
     this.boardFooter,
     this.engineGaugeBuilder,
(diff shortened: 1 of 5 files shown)


veloce:
I'm not sure this fix would work better. `UniqueKey` shouldn't be created in `build()` method, it doesn't really make sense.

You need to store an unique key in a `State` object instead.

lib/src/view/analysis/analysis_layout.dart
```diff
@@ -147,6 +149,11 @@ class AnalysisLayout extends StatelessWidget {
   /// [kAnalysisBoardHeaderOrFooterHeight].
   final Widget? boardFooter;
 
+  /// The side the board is displayed from.
+  ///
+  /// It should only be used when you need the state of the board header or footer to be preserved.
```
veloce, line 154: This comment is misleading. You're not preserving the state, but you're forcing the destruction of the widgets (and recreation) when the `ValueKey` changes.

Since we always have that information available, I'd make that parameter mandatory. It would help to avoid further surprises.

julien4215: If I understand correctly the [documentation](https://api.flutter.dev/flutter/foundation/Key-class.html) it will not recreate the widget if what you mean by widget is the underlying [element](https://api.flutter.dev/flutter/widgets/Element-class.html). The reason I moved the keys in the `AnalysisLayout` was to avoid the element being unmounted. I believe that thanks to the local key, it will simply call [update](https://api.flutter.dev/flutter/widgets/Element/update.html) on the element to update the widget configuration.

julien4215: > Since we always have that information available, I'd make that parameter mandatory. It would help to avoid further surprises.

I didn't want to force this parameter as it is only useful when there is a board footer or header. Should I still make it mandatory and keep only the first line of the doc comment to not make it confusing ?

julien4215: I made the `pov` parameter mandatory and added comments where we use a key to explain why it is needed

PR #1442, opened 2025-02-14
HTTP request logs screen

### LINKED ISSUE #699: Add an http request logs screen
Add a way for the user to see all the http requests made by the app, using the dart logging package.
It might be useful for users who want to know what is going in the background, and for debugging purposes.

The screen should be accessible from the settings screen. Should be a really simple list of text lines in the following form:
`<DateTime> <Http response code> <HTTP METHOD> <Uri>`

- [ ] Color in red response with code > 300.
- [ ] Add a button to clear the logs
- [ ] Use `ListView` with pagination (100 items per page?)

Implementation notes:
- [ ] should not reuse the existing http logger which was added for debugging locally during development
- [ ] should wrap the default client returned in `httpClientFactory` to intercept all requests
- [ ] save records in sqlite with a short retention period (7 days)
- [ ] probably wise to save records by batches (like every 10?); use app lifecycle events to save a pending incomplete batch when app is about to be exited (ie. put in background)

vincendep:
Specs:
- [x] Color in red response with code > 300.
- [x] Add a button to clear the logs
- [x] Use ListView with pagination (100 items per page?)

Implemetation:
- [x] should not reuse the existing http logger which was added for debugging locally during development
- [x] should wrap the default client returned in httpClientFactory to intercept all requests
- [x] save records in sqlite with a short retention period (7 days)
- [ ] probably wise to save records by batches (like every 10?); 
- [ ] use app lifecycle events to save a pending incomplete batch when app is about to be exited (ie. put in background)

This PR closes #699


vincendep:
I @veloce, could I've some feedback if I am on track or not? Following, I'll list what is currently missing in this PR:

- Fix failing tests
- Add Widget test for HttpLogScreen
- Implement batch support in HttpLogStorage
- Flush batch in AppLifecycle events
- Improve throttling/debounce fetching logic
- Polish UI: add empty screen, initial screen etc...

lib/src/network/http.dart
```diff
@@ -135,14 +183,30 @@ String makeUserAgent(PackageInfo info, BaseDeviceInfo deviceInfo, String sri, Li
 
 /// A [Client] that logs all requests.
 class LoggingClient extends BaseClient {
```
veloce, line 185: I think we can now make it private. And rename it since it's no more a `LoggingClient`. Something like `_RegisterCallbackClient`?

veloce (commented):
Thanks for tackling this!

You're definitely on the right track. UI looks neat :)

Haven't looked in detail to the code, but I'm not sure why you want to improve the throttling logic to get the logs? Seems not really necessary but I may miss sth.

vincendep:
Hi @veloce. Regarding the throttling logic, my previous implementation was buggy but now seems to be fixed. I also added a confirmation dialog to delete the logs, then are missing only batching and internationalisation.

veloce:
> Hi @veloce. Regarding the throttling logic, my previous implementation was buggy but now seems to be fixed. I also added a confirmation dialog to delete the logs, then are missing only batching and internationalisation.

Thanks! will review asap.

Batching could be done in another PR I suppose (if we ever need it).

lib/src/model/http_log/http_log_storage.dart
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  Future<void> deleteAll() async {
+    await _db.delete(kHttpLogStorageTable);
+  }
+}
+
+/// Represents an HTTP log entry.
+@Freezed(fromJson: true, toJson: true)
+class HttpLog with _$HttpLog {
```
veloce, line 73: Would name it `HttpLogEntry`.

lib/src/model/http_log/http_log_storage.dart
```diff
@@ -0,0 +1,105 @@
...
+  factory HttpLog.fromJson(Map<String, dynamic> json) => _$HttpLogFromJson(json);
+}
+
+/// A class representing a collection of HTTP logs.
+///
+/// The `HttpLogs` class contains the following properties:
+/// - `items`: A required list of `HttpLog` items.
+/// - `next`: An optional integer representing the next cursor.
+@Freezed(fromJson: true, toJson: true)
+class HttpLogs with _$HttpLogs {
```
veloce, line 101: Would name it `HttpLog`

lib/src/model/http_log/http_log_storage.dart
```diff
@@ -0,0 +1,105 @@
...
+
+  /// Saves an [HttpLog] entry to the database.
+  Future<void> save(HttpLog httpLog) async {
+    await _db.insert(kHttpLogStorageTable, {
+      ...httpLog.toJson(),
+      'lastModified': DateTime.now().toIso8601String(),
+    }, conflictAlgorithm: ConflictAlgorithm.replace);
+  }
+
+  Future<void> update(
```
veloce, line 48: I don't think we need this, do we?

lib/src/model/http_log/http_log_controller.dart
```diff
@@ -0,0 +1,76 @@
...
+/// The number of HTTP logs to fetch per page.
+const _pageSize = 12;
+
+/// A Riverpod controller for managing HTTP logs.
+///
+/// The `HttpLogController` class is responsible for fetching and managing
+/// paginated HTTP log entries from the storage. It uses a throttler to limit
+/// the rate of fetching new pages.
+@riverpod
+class HttpLogController extends _$HttpLogController {
```
veloce, line 18: I would name it `HttpLogPaginator` to be more clear.

veloce (commented):
Nice work!

Internationalisation is not needed yet for this feature. We can leave the hardcoded English strings.

A widget test that test the whole log screen would be nice. Nothing fancy, but perhaps something like a test that loads the whole app (you can find example in `app_test.dart`), then navigates to the log screen and see the startup requests. That would test the whole logging logic at once and having a test that shows the number of requests made during app startup is not bad as well.
No need to test pagination or deletion.

vincendep:
Sure I'll add some tests 👌🏻

lib/src/model/http_log/http_log_storage.dart
vincendep: At the moment http logs are saved when a request is made and then updated with the status code when the response is received.

veloce: No I just thought the log would never need to be updated.

lib/src/db/database.dart
```diff
@@ -76,6 +78,7 @@ Future<Database> openAppDatabase(DatabaseFactory dbFactory, String path) async {
         _createCorrespondenceGameTableV1(batch);
         _createChatReadMessagesTableV1(batch);
         _createGameTableV2(batch);
+        _createHttpLogTableV1(batch);
```
veloce, line 81: It should be _createHttpLogTableV4 imo (as in created in version 4).

lib/src/model/http_log/http_log_storage.dart
vincendep: Yeah maybe was better to save the log directly when receiving the response...only difference is that we cannot save unanswered http requests.

lib/src/network/http.dart
```diff
@@ -48,7 +49,11 @@ Uri lichessUri(String unencodedPath, [Map<String, dynamic>? queryParameters]) =>
 ///
 /// Do not use directly, use [defaultClient] or [lichessClient] instead.
 class HttpClientFactory {
-  Client call() {
+  const HttpClientFactory({this.onCreate});
+
+  final Client Function(Client client)? onCreate;
```
veloce, line 54: Would be better named `wrapClient` or `wrapper`.

veloce:
@vincendep do you wish to continue working on this PR? there are a couple of renaming, and a test missing still.

vincendep:
Hi @veloce, I would like to continue but at the moment I don't know when I will be able to do it.

veloce:
Since it's almost done, if you prefer, I can take it over. I can add the test and merge this.

vincendep:
No problem at all, as you wish.

PR #1452, opened 2025-02-16
Fix broadcast eval bar not working when switching to a different round

julien4215:
When switching to different round, the evals may not be updated until the view is scrolled. This problem was due to the key not being unique.


lib/src/view/broadcast/broadcast_boards_tab.dart
```diff
@@ -159,7 +159,7 @@ class BroadcastPreview extends ConsumerWidget {
           final playingSide = Setup.parseFen(game.fen).turn;
 
           return ObservedBoardThumbnail(
-            boardKey: Key('Board-$index'),
+            boardKey: ValueKey(game.id),
```
veloce, line 162: It should be the concatenation of `game.id` and `board-$index`. The key must be unique within the array as well.

PR #1458, opened 2025-02-17
Add deeplink support for broadcast round

julien4215:


veloce (commented):
Nice!

It would be good to add a test here, to show it loads fine when loading the round directly.

veloce (approved):

PR #1445, opened 2025-02-15
Add unrated puzzles option

CloudyDino:
# Unrated Puzzles
Close https://github.com/lichess-org/mobile/issues/1120
Close https://github.com/lichess-org/mobile/issues/297

## Changes
- Adds a switch to the puzzle settings screen to let users play unrated puzzles that don't affect their rating.

https://github.com/user-attachments/assets/0c7f35d1-528b-4b7e-929f-1a8d1a2566a8

- Logged out users don't see the rating switch:

https://github.com/user-attachments/assets/42ecd261-d12f-48f4-a6ef-9704b4f97f3f

## Testing
Tested using a proxy to spy on what app sends to API.

https://github.com/user-attachments/assets/90a00036-56e5-4273-9f2e-275e8c3c9f7c

1. First puzzle is correctly `win: false, rated: true`
```json
{
	"solutions": [{
		"id": "24pDF",
		"win": false,
		"rated": true
	}]
}
```
2. Second puzzle is correctly `win: true, rated: true`
```json
{
	"solutions": [{
		"id": "7nI7d",
		"win": true,
		"rated": true
	}]
}
```
3. Third puzzle is correctly `win: true, rated: false`
```json
{
	"solutions": [{
		"id": "6YIPJ",
		"win": false,
		"rated": false
	}]
}
```
6.  Rest of the puzzles are played offline and are correctly rated once I turn the internet back on and the batch is submitted.
```json
{
	"solutions": [{
		"id": "8Cm4H",
		"win": true,
		"rated": false
	}, {
		"id": "0haqU",
		"win": true,
		"rated": true
	}, {
		"id": "5T8Mz",
		"win": true,
		"rated": true
	}, {
		"id": "2106v",
		"win": false,
		"rated": true
	}, {
		"id": "2ZCbv",
		"win": false,
		"rated": false
	}, {
		"id": "9h38Q",
		"win": true,
		"rated": false
	}]
}
```

PR #1224, opened 2024-12-05
Added hint button, need to add functionality.

ItsDang:
For #1158

Issue #1158, opened 2024-11-14
Feature Request - Puzzle Hint


NewSupKaizo55:
![Screenshot_2024-11-14-21-46-33-998_com.discord-edit.jpg](https://github.com/user-attachments/assets/02e3cb33-edf4-4ff1-9589-945b894c4df2)

Sorry for not providing the description myself, it's pretty straightforward anyway.

It'd be amazing to have this, that'd be the full Lichess puzzle experience imo.

ItsDang:
Hi, I'd like to work on this!

tom-anders:
> Hi, I'd like to work on this!

Awesome! 

In general, you can open an early draft PR to let people know you're working on something. Don't forget to read through the [contributing guide](https://github.com/lichess-org/mobile/blob/main/CONTRIBUTING.md).

CloudyDino:
I think we need to make sure that when a hint is used, we mark the puzzle as unrated. We should have that functionality as a part of https://github.com/lichess-org/mobile/issues/1120.

PR #1459, opened 2025-02-17
Add Puzzle Hints

CloudyDino:
# Add Puzzle Hints

For https://github.com/lichess-org/mobile/issues/1158
Based on https://github.com/lichess-org/mobile/pull/1224
Merge after https://github.com/lichess-org/mobile/pull/1445

## Changes

- Adds button to puzzle screen bottom bar to show hint
- Once hint is shown, we mark hint shown for the puzzle state
- If hint has been shown for puzzle, we set puzzle to not be rated
- When hint is shown, we use green circle for the piece to move and we use blue circles to show where the piece can move. This is slightly different from what Web does in that it preselects the piece we can move so we can see all the places it moves to. I'm not sure how to do that with the current `BoardTable` widget.

## Screenshots / Recordings

https://github.com/user-attachments/assets/798f7af2-9ba2-44ac-bf0d-450b4179c9af

| Phone | Tablet |
| --- | --- |
| image | image |
| image | image |


veloce:
Looks very good!

I'm not sure why the piece is preselected on the website. Since we're not doing that, I think we can just add the circle around the hinted piece. But not showing the possible destinations. It just adds noise in my opinion. If you're playing puzzle it means you know where the piece can move to ;)

CloudyDino:
Thanks for the review @veloce! Just updated to only circle the square the piece to move is on.

CloudyDino:
Whoops, accidentally closed the PR.
Fixed the merge conflicts and used the `_viewSolutionCompleter.future` in `PuzzleScreen` to show the hint.

lib/src/model/puzzle/puzzle_controller.dart
```diff
@@ -195,12 +197,24 @@ class PuzzleController extends _$PuzzleController {
     });
   }
 
+  NormalMove solutionMove() {
```
veloce, line 200: This can be private. And also a getter instead of a method.

Also I'd name it `nextSolutionMove` since it depends on the current move index.

veloce (commented):
Looks good except a small comment.

Same as with the rated feature, we should add a widget test here. Since we have already the "view solution" puzzle test it should be easy to adapt from it.
The new tests should check:
- that the right square is highlighted
- that the solution is marked as `not rated` if the hint was shown, even though `rated` is enabled in preferences
- (optional) that the hint shown bool is correctly reset in the next puzzle (tested by solving another puzzle without hint and verifying that it is rated).

ijm8710:
Per Veloces last comment if you have it on rated and use a hint do future puzzles stay rated or does it change it?

CloudyDino:
> Per Veloces last comment if you have it on rated and use a hint do future puzzles stay rated or does it change it?

Future puzzles will continue to be rated. I tested that manually and I added that as a part of our test case.

Right now I think the only thing left is to check that the correct square is being circled after the hint button is pressed which is a bit tricky since I don't see a `ValueKey` being set for `ShapeWidget` in [flutter-chessground/](https://github.com/lichess-org/flutter-chessground/).

PR #1461, opened 2025-02-18
Updated puzzle analysis and other eval arrows to match theme arrow.

ItsDang:


veloce (approved):

veloce:
Oops. Sorry I validated this too fast. Actually the theme prefs are only here to change the color of user generated shapes (using 2 fingers). Not the other shapes. So I'll revert.

PR #1464, opened 2025-02-19
Support relative uri reference in broadcast markdown description

julien4215:
Opening a url might fail in the description of a broadcast overview if it is a relative url (e.g. /fide/1503014/-). This PR fixes it by using `Uri.resolve`


veloce:
So we are sure all the broadcast descriptions link to lichess.org?

PR #1466, opened 2025-02-20
Play a notification sound on new messages

### LINKED ISSUE #903: Add a sound when receiving chat messages
I often miss chat messages during or after games, because the small indication icon can be easily missed.

I think a small 'ding' sound would indicate a message nicely, without posing too much of a distraction during games.

It would make sense to disable these in Zen mode (when chat is already disabled)
and have the option to disable them in preferences too.

vincendep:
Make GameBody listen for ChatState changes and play a notification sound when unreadMessages increases.

closes #903


lib/src/model/common/service/sound_service.dart
```diff
@@ -17,7 +17,17 @@ final _soundEffectPlugin = SoundEffect();
...
+enum Sound {
+  move,
+  capture,
+  lowTime,
+  dong,
+  error,
+  confirmation,
+  puzzleStormEnd,
+  clock,
+  genericNotify,
```
veloce, line 29: There's no need to add generic notify, it is the same sound as `dong`.

lib/src/view/game/game_body.dart
```diff
@@ -329,6 +340,18 @@ class GameBody extends ConsumerWidget {
...
 
+  void _chatListener(
+    AsyncValue<ChatState>? prev,
+    AsyncValue<ChatState> state, {
+    required BuildContext context,
+    required WidgetRef ref,
+  }) {
+    if (prev == null || !prev.hasValue || !state.hasValue) return;
+    if (state.requireValue.unreadMessages > prev.requireValue.unreadMessages) {
+      ref.read(soundServiceProvider).play(Sound.genericNotify);
```
veloce, line 351: I don't think we should use that sound. Because it is the same as the game start and game ending sounds it is confusing.

A better one is `confirmation`.

Also in order to make it less disturbing during a game we should lower the volume of this specific sound effect.

You can add another `volume` parameter to [the play method](https://github.com/lichess-org/mobile/blob/main/lib/src/model/common/service/sound_service.dart#L84) to that effect. Should be a number between 0 and 1 to apply to the master volume.

veloce (approved):
Thanks!

PR #1467, opened 2025-02-20
Cloud eval preparation

julien4215:
Prepare cloud eval implementation :
- add websocket eval events in BroadcastAnalysisController
- use ClientEval type instead of LocalEval in AnalysisCurrentNode and engine widgets
- add doc comments and necessary properties to ClientEval
- edit engine depth widget to display a cloud icon when it is a cloud eval (see screenshot below)

### DIFF
--- a/lib/src/model/broadcast/broadcast_analysis_controller.dart
+++ b/lib/src/model/broadcast/broadcast_analysis_controller.dart
@@ -265,7 +265,7 @@ class BroadcastAnalysisController extends _$BroadcastAnalysisController implemen
 
     if (state.requireValue.currentPath != path) return;
 
-    _refreshCurrentNode();
+    _refreshCurrentNode(shouldRecomputeRootView: true);
   }
 
   // ignore: unused_element
--- a/lib/src/model/broadcast/broadcast_analysis_controller.dart
+++ b/lib/src/model/broadcast/broadcast_analysis_controller.dart
@@ -247,8 +247,6 @@ class BroadcastAnalysisController extends _$BroadcastAnalysisController implemen
   void _handleEvalHitEvent(SocketEvent event) {
     final path = pick(event.data, 'path').asUciPathOrThrow();
 
-    if (state.requireValue.currentPath != path) return;
-
     final depth = pick(event.data, 'depth').asIntOrThrow();
     final pvs =
         pick(event.data, 'pvs')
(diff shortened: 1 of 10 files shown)


lib/src/view/engine/engine_depth.dart
```diff
@@ -12,61 +12,88 @@ import 'package:lichess_mobile/src/widgets/list.dart';
...
-                      ? DialogTheme.of(context).backgroundColor ??
-                          ColorScheme.of(context).surfaceContainerHigh
-                      : CupertinoDynamicColor.resolve(
-                        CupertinoColors.tertiarySystemBackground,
-                        context,
-                      ),
-              transitionDuration: Duration.zero,
-              popoverTransitionBuilder: (_, child) => child,
-            );
+            if (eval is LocalEval) {
```
julien4215, line 26: I think we should also display a popover when it is a cloud eval but I didn't find a good way to do it.

veloce:
This is tricky. Will review this thoroughly next week.

veloce:
Also this code is already complex, and with cloud eval will become even more. We need to add tests.
Thing is we need a mock of stockfish that we don't have right now. I'll try to implement a stockfish mock and then we can start adding tests. Mocking cloud evals will be easy after that.

lib/src/model/broadcast/broadcast_analysis_controller.dart
```diff
@@ -240,6 +244,44 @@ class BroadcastAnalysisController extends _$BroadcastAnalysisController implemen
     state = AsyncData(state.requireValue.copyWith(pgnHeaders: pgnHeaders));
   }
 
+  void _handleEvalHitEvent(SocketEvent event) {
+    final path = pick(event.data, 'path').asUciPathOrThrow();
+
```
veloce, line 249: Should update the eval in the corresponding node.

lib/src/model/broadcast/broadcast_analysis_controller.dart
```diff
@@ -240,6 +244,42 @@ class BroadcastAnalysisController extends _$BroadcastAnalysisController implemen
...
+            )
+            .toIList();
+
+    final cloudEval = CloudEval(depth: depth, position: state.requireValue.position, pvs: pvs);
+
+    _root.updateAt(path, (node) => node.eval = cloudEval);
+
+    if (state.requireValue.currentPath != path) return;
+
+    _refreshCurrentNode();
```
veloce, line 268: should pass `true` to `shouldRecomputeRootView` here as we need  to update the move list with the new eval

veloce (commented):
Just one little change and it's good.

veloce (approved):

PR #1469, opened 2025-02-20
Implement cloud eval only for broadcast

julien4215:
It should be only reviewed after #1467 as this a follow-up to this PR (only the second commit is new).

### DIFF
--- a/lib/src/model/engine/evaluation_service.dart
+++ b/lib/src/model/engine/evaluation_service.dart
@@ -237,9 +237,15 @@ class EvaluationOptions with _$EvaluationOptions {
   }) = _EvaluationOptions;
 }
 
+/// A function to choose the eval that should be displayed.
 Eval? pickBestEval({
+  /// The eval from the local engine
   required LocalEval? localEval,
+
+  /// The cached eval which is either a saved eval from the local evaluation or a cloud eval
   required ClientEval? savedEval,
+
+  /// The eval from the server analysis
   required ExternalEval? serverEval,
 }) {
   return switch (savedEval) {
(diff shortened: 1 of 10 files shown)


lib/src/view/engine/engine_gauge.dart
```diff
@@ -14,8 +14,6 @@ enum EngineGaugeDisplayMode { vertical, horizontal }
 
 typedef EngineGaugeParams =
     ({
-      bool isLocalEngineAvailable,
```
julien4215, line 17: I removed this field because it didn't seem useful as `ref.watch(engineEvaluationProvider).eval` will simply have a null value if the local engine is not available and it allows me to use the method `currentEval` that I created to centralize the logic to choose the current eval.

veloce: I don't think we can remove this. Because we want to either:
- always display the gauge and engine lines if the server evals are available
- and if they are not, display the gauge and line only if local eval is enabled

The reason is to avoid showing/hiding the gauge and lines when jumping to another position to avoid any board shift.

lib/src/model/engine/evaluation_service.dart
```diff
@@ -238,3 +236,33 @@ class EvaluationOptions with _$EvaluationOptions {
     required Duration searchTime,
   }) = _EvaluationOptions;
 }
+
+Eval? pickBestEval({
```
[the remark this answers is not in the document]
julien4215: I added comments to explain why there are these functions and what are their parameters but the logic is a bit annoying to explain and not necessary I think

PR #1502, opened 2025-02-28
Use socket versioning for the broadcast round controller

julien4215:


lib/src/model/broadcast/broadcast_round_controller.dart
```diff
@@ -46,9 +46,19 @@ class BroadcastRoundController extends _$BroadcastRoundController {
...
+    );
+
     _socketClient = ref
         .watch(socketPoolProvider)
-        .open(BroadcastRoundController.broadcastSocketUri(broadcastRoundId));
+        .open(
+          BroadcastRoundController.broadcastSocketUri(broadcastRoundId),
+          version: round.socketVersion,
+          onEventGapFailure: () {
+            _syncRound();
```
veloce, line 59: Don't think this is enough; if there is a socket failure we should force reconnect it and `_syncRound()` doesn't do that.

lib/src/view/broadcast/broadcast_round_screen.dart
```diff
@@ -226,6 +226,7 @@ class _BroadcastRoundScreenState extends ConsumerState<BroadcastRoundScreen>
             BroadcastPlayersTab(tournamentId: _selectedTournamentId),
           ],
         ),
+        AsyncError(:final error) => Center(child: Text('Cannot load round data: $error')),
```
veloce, line 229: People don't know about round, better put `Cannot load tournament data`.

lib/src/model/broadcast/broadcast_round_controller.dart
julien4215: Yes I modified `_syncRound` so that it knows reconnect the websocket.

lib/src/model/broadcast/broadcast_round_controller.dart
```diff
@@ -72,20 +62,45 @@ class BroadcastRoundController extends _$BroadcastRoundController {
       },
     );
 
-    final round = await ref.withClient(
-      (client) => BroadcastRepository(client).getRound(broadcastRoundId),
+    return BroadcastRoundState(round: round.round, games: round.games, observedGames: ISet());
+  }
+
+  Future<void> _connectSocket({required int version, bool forceReconnect = false}) async {
+    final socketPool = ref.watch(socketPoolProvider);
```
veloce, line 69: Don't use ref.watch outside of widget.build or provider build() methods. Here using ref.read is fine since the socket pool is a singleton.

lib/src/model/broadcast/broadcast_round_controller.dart
```diff
@@ -72,20 +62,45 @@ class BroadcastRoundController extends _$BroadcastRoundController {
...
 
     final key = _key;
+
     final round = await ref.withClient(
       (client) => BroadcastRepository(client).getRound(broadcastRoundId),
     );
+
+    await _socketOpenSubscription?.cancel();
+    await _socketSubscription?.cancel();
+    await _connectSocket(version: round.socketVersion, forceReconnect: true);
```
veloce, line 102: Do we want to force reconnect socket here? I don't think so, because we hadn't to do it earlier and it was working fine.

Forcing socket reconnection should happen only on event gap failure I think.

julien4215: We have to reconnect to ensure that we receive all socket events after the http request is complete. We could receive a socket event of version `n+1` before the request complete that would need events after `n`.

veloce:
@julien4215 there is a failing test and a conflict to solve here, can you see that?

Issue #819, opened 2024-07-01
Enhancement: display wins/losses on stat pages as a meter


ijm8710:
Benefit of the meter being that it can encapsulate the percentage split visually

![image](https://github.com/user-attachments/assets/18b43298-7ec5-44fd-b503-fae5d05c53a8)

Note: this was originally a two-item ticket. Colored wins losses red/green was captured in an older PR

PR #1504, opened 2025-03-01
Added colors to user performance page.

ItsDang:
For #819

Website

User_Performance_Color_Website

My Changes

User_Performance_Color
User_Performance_Color_1


lib/src/view/user/perf_stats_screen.dart
```diff
@@ -233,15 +234,15 @@ class _Body extends ConsumerWidget {
                         child: _RatingWidget(
                           data.highestRating,
                           data.highestRatingGame,
-                          context.lichessColors.good,
+                          LichessColors.good,
```
veloce, line 237: This was done so for a reason, please don't remove things before understanding why.
Explanation: https://github.com/lichess-org/mobile/blob/main/lib/src/theme.dart#L155

this applies to all the changes here.

veloce (changes requested):

veloce (approved):

PR #1516, opened 2025-03-05
Freezed 3.0 

julien4215:
I had to make `OfflineCorrespondenceGame` and `PlayableGame` abstract instead of sealed because they are implemented in tests. Also when implemented these classes, there is now the warning avoid_implementing_value_types that I just chose to ignore but I am not sure that is the right thing to do.

### DIFF
--- a/lib/src/model/account/ongoing_game.dart
+++ b/lib/src/model/account/ongoing_game.dart
@@ -9,7 +9,7 @@ import 'package:lichess_mobile/src/model/user/user.dart';
 part 'ongoing_game.freezed.dart';
 
 @freezed
-class OngoingGame with _$OngoingGame {
+sealed class OngoingGame with _$OngoingGame {
   factory OngoingGame({
     required GameId id,
     required GameFullId fullId,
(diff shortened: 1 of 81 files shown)


julien4215:
> Yeah I think they should be sealed too. And you need to replace the `Fake` with hand created objects.
> 
> You have examples:
> 
>     * https://github.com/lichess-org/mobile/blob/91186bc7d0f614319f5496b4289ad41ac00d7af9/test/model/correspondence/correspondence_game_storage_test.dart#L49
> 
>     * https://github.com/lichess-org/mobile/blob/91186bc7d0f614319f5496b4289ad41ac00d7af9/test/model/game/game_test.dart#L14
> 
> 
> You can put them in the [example_data.dart](https://github.com/lichess-org/mobile/blob/main/test/example_data.dart) since they are used in several places.

I have done it for the `OfflineCorrespondenceGame`. It looks like the fake was not necessary for the `PlayableGame` so I just removed it in the test.

veloce (approved):
Thanks!

PR #1523, opened 2025-03-06
Refactor broadcast player classes

julien4215:
Refactor broadcast player classes so that they all share a same base class thanks to freezed support of classic classes.

I am not sure this is worth it especially the second commit but I wanted to attempt this now that we can extend freezed classes.


veloce:
Yeah I don't think we should do it. If we can use composition in place of inheritance, we should. And here I see a lot of lint ignore, which tells me sth is wrong.

julien4215:
I don't think there is something wrong with the lint ignore. One is because it seems `freezed_lint` was not updated for freezed 3.0 and the other one is a [freezed issue](https://github.com/rrousselGit/freezed/issues/1176) with classic classes.

julien4215:
Closed because composition is preferred over inheritance

PR #1525, opened 2025-03-06
Fix clock orientation by adding inverted opponent view

okmsbun:
This PR closes https://github.com/lichess-org/mobile/issues/1529

| | | |
| Screenshot_1741303516 | | Screenshot_1741304132 |

### DIFF
--- a/lib/src/view/clock/clock_tool_screen.dart
+++ b/lib/src/view/clock/clock_tool_screen.dart
@@ -170,18 +170,19 @@ class ClockTile extends ConsumerWidget {
               ),
             ),
           ),
-          Positioned(
-            top: 24,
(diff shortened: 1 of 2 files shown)


ijm8710:
Note that if Veloce approves the landscape pr, no need. Just not sure if he feels desire to have landscape anywhere in all

okmsbun:
Even if the PR that checks whether the orientation changes on certain screens is accepted, this PR might still be relevant. This is because the small clock only works in portrait mode; it does not appear in landscape mode. Both PRs can be accepted simultaneously, or only one of them, or even neither :)). They do not affect each other.

veloce (approved):

Issue #1492, opened 2025-02-25
Clock doe not rotate to landscape mode (cant read number of opponent upside down)


malo1972:
### What is the problem?

clock should switch to landsacpe mode so both players can read both numbers
It is a FIDE requirement that both players have to be able to read both times

### Steps to reproduce the bug

open clock.
wonder why.

### App version

0.10.18

### Device Info

Samsung S10

okmsbun:
I've made some code changes to address this issue. Would this be a suitable solution? Does this approach make sense, or is there a better way to handle it?

![Image](https://github.com/user-attachments/assets/4ffbfe95-7270-4ef1-b7cc-d2e8a972a90a)

malo1972:
no, you cannot focus to such a small number during a game. 
Why not turn both numbers 90 degrees so they have the same orientation (landscape)
This may be optional for players who like it as it is.

okmsbun:
The app currently allows landscape orientation only on devices with a larger screen size, such as tablets. On phones with smaller screens, it explicitly forces the orientation to portrait using the following snippet:

https://github.com/lichess-org/mobile/blob/daa8fc2bf003fd500b3d9705c428510faebc1a46/lib/src/init.dart#L148

As seen in the image below, landscape mode already functions like this. I'm unsure whether it would be beneficial to specifically enable both landscape and portrait orientations just for the clock screen. Was your request exactly about this behavior? If I've misunderstood something, please correct me.

![Image](https://github.com/user-attachments/assets/d4592ad5-5ff3-4b96-8156-4c1d29a6d4c9)

malo1972:
yes, thats what it should look like also on phones.

ijm8710:
I like what you did and agree with the logic. Any reason you closed the pr unmerged?

okmsbun:
Yeah, I liked it too. It would definitely be easier to check the opponent’s time in portrait mode. But I closed it since I didn’t think it was directly related to this issue. If someone opens a separate issue for this, I think it’d be great, and we could reopen it.

ijm8710:
> Yeah, I liked it too. It would definitely be easier to check the opponent’s time in portrait mode. But I closed it since I didn’t think it was directly related to this issue. If someone opens a separate issue for this, I think it’d be great, and we could reopen it.

#1529

PR #1531, opened 2025-03-07
Fix Clock Orientation for Better Readability in Landscape Mode

okmsbun:
closes https://github.com/lichess-org/mobile/issues/1492


okmsbun:
https://github.com/user-attachments/assets/0837f48e-6b64-4f4f-8e1e-1e55793a4f44

veloce:
Thanks for tackling this. I am not sure it would work like that though. 

You most likely want to keep the phone in portrait mode at the side of the board on the table, with 2 players facing each other, while having the *display* inside be in landscape mode (hope this makes sense).

So instead of using system chrome and change the orientation through OS, why not having a toggle that switches the orientation?

okmsbun:
@veloce ready for review


https://github.com/user-attachments/assets/b5e52046-8e67-485f-a2a7-e3763ba84919

okmsbun:
Yeah, I think it could definitely be this way. I'll change it accordingly

okmsbun:
Is this what you had in mind? Did I get you correctly? If there's anything else you're unsure about, please feel free to ask!

https://github.com/user-attachments/assets/8c15a2c1-def8-4f46-a41b-98d98a4d67fd

veloce:
Looks good to me! Thank you.

veloce (approved):

PR #1532, opened 2025-03-07
Rework of broadcast player classes

julien4215:
- Add a composition layer to broadcast player classes so that they all share a same base class
- Refactor `BroadcastPlayerWidget`
- A lot of renaming

### DIFF
--- a/lib/src/view/broadcast/broadcast_players_tab.dart
+++ b/lib/src/view/broadcast/broadcast_players_tab.dart
@@ -43,7 +43,7 @@ const _kHeaderTextStyle = TextStyle(fontWeight: FontWeight.bold, overflow: TextO
 class PlayersList extends ConsumerStatefulWidget {
   const PlayersList(this.players, this.tournamentId);
 
-  final IList<BroadcastPlayerExtended> players;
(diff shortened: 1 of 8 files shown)


lib/src/view/broadcast/broadcast_players_tab.dart
```diff
@@ -173,19 +180,17 @@ class _PlayersListState extends ConsumerState<PlayersList> {
             ),
           );
         } else {
-          final player = players[index - 1];
+          final playerWithOverallResult = players[index - 1];
+          final player = playerWithOverallResult.player;
```
veloce, line 184: In dart you can destructure a class instance:
```dart
final PlayerWithOverallResult(:player, :ratingDiff, :score, :played) = playerWithOverallResult;
```

veloce (approved):

lib/src/view/broadcast/broadcast_players_tab.dart
julien4215: Thanks I did not think to use that. I made a PR to use it in several places.

PR #1542, opened 2025-03-09
Tweak broadcast board tab code

julien4215:


lib/src/view/broadcast/broadcast_boards_tab.dart
```diff
@@ -169,8 +168,6 @@ class BroadcastPreview extends ConsumerWidget {
 
 class ObservedBoardThumbnail extends ConsumerStatefulWidget {
   const ObservedBoardThumbnail({
-    super.key,
```
veloce, line 172: It is good practice to leave the `super.key` in public widget classes.

veloce (approved):

lib/src/view/broadcast/broadcast_boards_tab.dart
julien4215: I would understand if it was a package but here we can always add it if we need to use the key.

Issue #1495, opened 2025-02-26
opening explorer and analysis board should show the precise opening name


kwassmuss:
After for example the opening sequence 
1. e4 c5 2. Nf3 Nc6 3. d4 cxd4 4. Nxd4 e6
 the analysis board of the old app and the website identify this opening as 
"B44 Sicilian Defense: Taimanov Variation"
whereas the beta app only identifies this as 
"Sicilian Defense: Open"

It's very helpful if you can name a specific opening line so you can talk about it or look up resources about it so this should be implemented in the new app as well.

ijm8710:
Good post-I agree

PR #1553, opened 2025-03-12
Refactor opening fetching to use FEN instead of moves

okmsbun:
closes https://github.com/lichess-org/mobile/issues/1495

Refactored opening fetching logic to use FEN instead of move sequences. This change ensures precise identification of chess openings on the analysis board and opening explorer, addressing issue 1495.

### DIFF
--- a/lib/src/model/analysis/analysis_controller.dart
+++ b/lib/src/model/analysis/analysis_controller.dart
@@ -162,7 +162,7 @@ class AnalysisController extends _$AnalysisController
           lastMove = branch.sanMove.move;
         }
         if (isMainline && opening == null && branch.position.ply <= 10) {
-          openingFutures.add(_fetchOpening(root, path));
+          openingFutures.add(_fetchOpening(branch.position.fen, path));
         }
       },
     );
@@ -476,7 +476,7 @@ class AnalysisController extends _$AnalysisController
       }
 
       if (currentNode.opening == null && currentNode.position.ply <= 30) {
-        _fetchOpening(_root, path).then((value) {
+        _fetchOpening(currentNode.position.fen, path).then((value) {
           if (value != null) {
             final (path, opening) = value;
             _updateOpening(path, opening);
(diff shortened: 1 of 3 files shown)


okmsbun:
Previously, queries were based on UCI (move sequences), but now queries use the board position directly (FEN). The issue with using UCI is that it's just a move order. You can reach the same position in multiple ways, following different move sequences, each with its own unique UCI notation. If we query by UCI, we only consider how we reached the position, not the final position itself. That means sometimes we won't find the position in the database, even though the resulting board is exactly the same.

That's why using EPD (which is basically FEN without move numbers) gives us better and more accurate results. Because this way, we're checking the actual final position on the board, regardless of the move sequence.

https://github.com/lichess-org/mobile/blob/2c296c41938d4e5ec5d34c393baed12dc57d92f7/lib/src/db/openings_database.dart#L11

https://github.com/lichess-org/chess-openings

https://github.com/lichess-org/chess-openings/blob/8c009155ce4032234536f946dbe4941e167afd77/README.md?plain=1#L12

I didn't check the website's source code directly, but I did inspect the network requests. They send FEN data too.

okmsbun:
Yes, fetching by FEN seems good, but there's something else that doesn't make sense to me. On the opening explorer screen or when using the analysis screen and switching to the opening explorer, each move triggers a separate request like this:

```
https://explorer.lichess.ovh/masters?fen=rnbqkbnr%2Fpppp1ppp%2F4p3%2F8%2F4P3%2F8%2FPPPP1PPP%2FRNBQKBNR+w+KQkq+-+0+2&since=1952
```

The website sends the exact same request, and the response already contains accurate opening information. So why don't we directly use the opening information from this response, instead querying the openings from the database(chess-openings ) again? I'm just trying to understand this.

flutter
Screenshot 2025-03-13 at 21 25 31


web
Screenshot 2025-03-13 at 21 28 46

Screenshot 2025-03-13 at 21 25 13

veloce:
> The website sends the exact same request, and the response already contains accurate opening information. So why don't we directly use the opening information from this response, instead querying the openings from the database(chess-openings ) again? I'm just trying to understand this.

Because it needs to work offline. We should use the online data when available though, as it is regularly updated, contrary to the offline database.

okmsbun:
Thanks for the explanation.

veloce:
Feel free to update the code to use the online response in this PR if you want. Otherwise I'll create an issue for it.

veloce (approved):

PR #1569, opened 2025-03-20
Broadcast players sort

julien4215:
Use the same sorting as the website for broadcast players.

Also fix sorting when the sort is applied to games and not scores.

### DIFF
--- a/lib/src/view/broadcast/broadcast_players_tab.dart
+++ b/lib/src/view/broadcast/broadcast_players_tab.dart
@@ -32,6 +32,8 @@ class BroadcastPlayersTab extends ConsumerWidget {
 
 enum _SortingTypes { player, elo, score }
 
+typedef _BroadcastPlayerPicker<T> = T? Function(BroadcastPlayerWithOverallResult player);
+
 const _kTableRowVerticalPadding = 12.0;
 const _kTableRowHorizontalPadding = 8.0;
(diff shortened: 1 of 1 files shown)


lib/src/view/broadcast/broadcast_players_tab.dart
```diff
@@ -52,61 +54,80 @@ class PlayersList extends ConsumerStatefulWidget {
 
 class _PlayersListState extends ConsumerState<PlayersList> {
   late IList<BroadcastPlayerWithOverallResult> players;
-  late _SortingTypes currentSort;
-  bool reverse = false;
+  late bool reverse;
```
veloce, line 57: better than late, use the default value, it is safer

veloce (commented):
Thanks for this!

I think it should come with new tests showing the sort works correctly in the various situations. We already have broadcast tests so it should not be to hard to test.

lib/src/view/broadcast/broadcast_players_tab.dart
julien4215: It would useless to define it twice

julien4215:
I added some tests. I used `widgetList` as it seems to returns the widgets in the order they are displayed but I don't know if this is the right way to test sorting.

veloce (approved):

PR #1574, opened 2025-03-23
fix: prevent move list from resizing when first move is made

tom-anders:
Problem: If no move has been made yet, `materialDiff` is `null` and its widget won't be displayed. This leads to a layout jump when the first move is made.
Solution: Make the `materialDiff` field inside `MaterialDifferenceDisplay` nullable
Before: 

[before.webm](https://github.com/user-attachments/assets/0cc7c365-a734-40e0-b06f-8af945ced6d2)

After:

[after.webm](https://github.com/user-attachments/assets/5bfe320a-2f41-4b1b-b168-24029522fdcf)


tom-anders:
(I forgot to mention that this only happens in landscape)

veloce (approved):
Thanks!

PR #1580, opened 2025-03-25
Fix: Make opponent's clock red when a player flags (issue #1561)

### LINKED ISSUE #1561: feature request: Make opponent's clock red when a player flags
This feature already exists in the old mobile app, and I think it's a great little QoL improvement. When a player loses by running out of time, simply make their clock red instead of the normal grey/black. This should probably be applied to all game clocks - tournament, otb, quick pair etc

Image

QuartzIII:
Fixes #1561

### DIFF
--- a/lib/src/widgets/clock.dart
+++ b/lib/src/widgets/clock.dart
@@ -69,13 +69,21 @@ class Clock extends StatelessWidget {
           decoration: BoxDecoration(
             borderRadius: const BorderRadius.all(Radius.circular(5.0)),
             color:
-                active
+                timeLeft > Duration.zero
+                    ? active
+                        ? isEmergency
+                            ? effectiveClockStyle.emergencyBackgroundColor
+                            : effectiveClockStyle.activeBackgroundColor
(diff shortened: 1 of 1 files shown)


lib/src/widgets/clock.dart
```diff
@@ -73,6 +73,8 @@ class Clock extends StatelessWidget {
                     ? isEmergency
                         ? effectiveClockStyle.emergencyBackgroundColor
                         : effectiveClockStyle.activeBackgroundColor
+                    : isEmergency
```
veloce, line 76: It adds the red background when flagged, but it also adds the red background when the game is still running and it is not the active side... not sure we want that.

QuartzIII:
<table>
  <tr>
    <th>Before</th>
    <th>After</th>
  </tr>
  <tr>
    <td>Before</td>
    <td>After</td>
  </tr>
</table>

lib/src/widgets/clock.dart
QuartzIII: Fixed

veloce (approved):

PR #1582, opened 2025-03-25
Computer analysis fixes on analysis screen and broadcast game screen

julien4215:
- Fix annotation shown on broadcast game screen when computer analysis is disabled
- Fix best move arrow shown on analysis screen when engine is disabled
- Standardize analysis screen and broadcast game screen code

TODO:
- [x] fix annotations shown on broadcast game screen move list when computer analysis is disabled
- [x] add a test case for annotations shown per settings on analysis screen


veloce:
> Fix best move arrow shown on analysis screen when engine is disabled

This was already fixed (with a test case). So I don't understand. Can you elaborate?

veloce:
Are there test cases covering **every** regression?

veloce (commented):
We've had a lot of bugs and regression in this area, due to the complexity of the settings.

So I think we should also test the broadcast screen. In order to do that we'd need:
- to refactor `makeEngineTestApp` to take a `home` Widget as an optional parameter that would replace the `AnalysisScreen` if provided.
- to move `engine_gauge_test` and `engine_lines_test` to analysis_screen_test
- duplicate all relevant tests from analysis screen test to broadcast screen test using the `makeEngineTestApp` 

Only the `engine_depth_test` should remain in the `view/engine` folder because it doesn't depend on the screen where it is displayed.

julien4215:
> > Fix best move arrow shown on analysis screen when engine is disabled
> 
> I don't reproduce that, or don't understand what's doing wrong. Can you elaborate?

The best moves arrows are still shown on the analysis screen even when the local engine is disabled (see the video below).

[analysis-screenb-bug.webm](https://github.com/user-attachments/assets/a13e6bc2-4a1b-415e-8442-1d444abfc3f9)

julien4215:
> We've had a lot of bugs and regression in this area, due to the complexity of the settings.
> 
> So I think we should also test the broadcast screen. In order to do that we'd need:
> 
>     * to refactor `makeEngineTestApp` to take a `home` Widget as an optional parameter that would replace the `AnalysisScreen` if provided.
> 
>     * to move `engine_gauge_test` and `engine_lines_test` to analysis_screen_test
> 
>     * duplicate all relevant tests from analysis screen test to broadcast screen test using the `makeEngineTestApp`
> 
> 
> Only the `engine_depth_test` should remain in the `view/engine` folder because it doesn't depend on the screen where it is displayed.

Can the broadcast tests be done in a different PR ? I only intended to fix the two bugs I mentioned and use the same code structure between analysis and broadcast so it is easier to see the differences between both.

lib/src/view/analysis/analysis_board.dart
```diff
@@ -114,7 +119,6 @@ class AnalysisBoardState extends ConsumerState<AnalysisBoard> {
       setState(() {
         userShapes = userShapes.remove(shape);
       });
-      return;
```
veloce, line 117: This code change is not tested. Are you sure you can safely remove it?

PR #1583, opened 2025-03-25
feat: add tournament screen

tom-anders:
With this PR, we'd have very basic support for tournaments. However, the game screen is still missing some features, like berserking, displaying rank, etc. That will be a follow-up PR. Same goes for the "podium" screen for finished tournaments.

I did not add widget tests yet, wanted to get feedback on the layout/design first.

[tournament_screen.webm](https://github.com/user-attachments/assets/624e8132-707e-4408-89ae-6fd3fedba34e)

TODO:
- [ ] fix tests
- [ ] fix non started tournament clock display
- [ ] fix bug when non started tournament doesn't refresh when the countdown reaches 0 (it should refresh and start automatically, currently: it doesn't start and the featured game doesn't work)
- [ ]


teefoox:
The tournament screen is great! Love it.
I think it would be useful to also have the players' recent results (at least the last one) to know if one needs to win one or two games to get the win-streak buff. In an end tournament rush, it can help to make strategic decisions.

veloce:
Nice work on the UI! Will review asap.

lib/src/model/tournament/tournament_controller.dart
```diff
@@ -0,0 +1,150 @@
...
+  // so we manually have to set this timer to schedule a reload once we can join again.
+  Timer? _pauseDelayTimer;
+
+  @override
+  Future<TournamentState> build(TournamentId id) async {
+    final state = await _loadTournament(id, standingsPage: 1);
+
+    final socketPool = ref.watch(socketPoolProvider);
+    _socketClient = socketPool.open(Uri(path: '/tournament/$id/socket/v6'));
+    ref.onDispose(() {
```
veloce, line 34: Better to put the `ref.onDispose` at the top of `build` (not after any `await`) to ensure the callback is called.

lib/src/model/tournament/tournament_controller.dart
```diff
@@ -0,0 +1,150 @@
...
+    if (!state.hasValue) {
+      return;
+    }
+
+    final tournament = await ref
+        .read(tournamentRepositoryProvider)
+        .reload(state.requireValue.tournament, standingsPage: standingsPage);
+
+    if (tournament.me?.pauseDelay != null) {
+      _pauseDelayTimer = Timer(tournament.me!.pauseDelay!, () {
```
veloce, line 94: Always cancel the timer before building a new one.

lib/src/view/tournament/tournament_screen.dart
```diff
@@ -0,0 +1,489 @@
...
+  Widget build(BuildContext context) {
+    final timeLeft = state.tournament.timeToStart ?? state.tournament.timeToFinish;
+    return Row(
+      mainAxisAlignment: MainAxisAlignment.spaceBetween,
+      children: [
+        Flexible(child: AutoSizeText(state.tournament.fullName, maxLines: 1, minFontSize: 14.0)),
+        if (timeLeft != null)
+          CountdownClockBuilder(
+            timeLeft: timeLeft,
+            clockUpdatedAt: DateTime.now(),
```
veloce, line 131: It doesn't make sense to pass `DateTime.now()` here. The date time should be part of the state and change each time we receive a new `timeLeft` value from the server.

But here we don't really care to compensate for UI lag, so you could also revert this commit in the PR:
https://github.com/lichess-org/mobile/commit/a8e71c0aac625f8d2c42859538c114a8d9adfed4

lib/src/view/tournament/tournament_screen.dart
```diff
@@ -0,0 +1,489 @@
...
+                      ? LichessColors.good
+                      : LichessColors.error
+                  : null,
+          size: 30,
+        ),
+        const SizedBox(width: 10),
+        Column(
+          crossAxisAlignment: CrossAxisAlignment.start,
+          children: verdicts.list
+              .whereNot((verdict) => verdict.condition == 'Bot players are not allowed')
```
veloce, line 320: Could be refactored with a list comprehension and a `if` inside, which is usually more readable (less indented).

lib/src/view/tournament/tournament_screen.dart
```diff
@@ -0,0 +1,489 @@
...
+                  orientation: featuredGame.orientation,
+                  fen: game.lastPosition.fen,
+                  header: featuredGame.orientation == Side.white ? blackPlayer : whitePlayer,
+                  footer: featuredGame.orientation == Side.white ? whitePlayer : blackPlayer,
+                  lastMove: game.lastMove,
+                ),
+              );
+            }
+          case _:
+            return BoardThumbnail(
```
veloce, line 382: It is more efficient to always return the GestureDetector in the loading state, so only the onTap handler changes.

lib/src/view/tournament/tournament_screen.dart
```diff
@@ -0,0 +1,489 @@
...
+                              .read(tournamentControllerProvider(widget.state.id).notifier)
+                              .joinOrPause();
+                          setState(() {
+                            joinOrLeaveInProgress = true;
+                          });
+                        }
+                        : null,
+              )
+        else
+          BottomBarButton(
```
veloce, line 474: Do we want to be able to sign in directly from here? I'm not sure, but why not.

veloce (commented):
Great work!

lib/src/model/tournament/tournament_controller.dart
```diff
@@ -0,0 +1,150 @@
...
+part 'tournament_controller.freezed.dart';
+part 'tournament_controller.g.dart';
+
+final _logger = Logger('TournamentController');
+
+@riverpod
+class TournamentController extends _$TournamentController {
+  StreamSubscription<SocketEvent>? _socketSubscription;
+
+  late final SocketClient _socketClient;
```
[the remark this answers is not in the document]
tom-anders: ah, I think there are some other controllers that do this, I'll fix them all

lib/src/model/tournament/tournament_controller.dart
tom-anders: ah, coming from C++ it's weird that there are no destructors that take care of this stuff, will try to remember this next time :D

lib/src/view/tournament/tournament_screen.dart
tom-anders: Huh, I just noticed the `GestureDetector` doesn't actually have a `onTap` handler here.  I probably wanted it to open the game screen, but do we really want that? Players might miss their next pairing (as the websocket connection gets closed once you leave the tournament screen) if they open the feature game and enter the game screen, so let's just remove the `GestureDetector` here

lib/src/view/tournament/tournament_screen.dart
tom-anders: Yeah, the website and old app do it, and also right now the bottom bar would be empty otherwise :D

tom-anders:
Thanks for the review! I added two widgets tests now, let me know if you think there's anything else that should be tested

lib/src/view/tournament/tournament_screen.dart
veloce: No I meant when flutter does the tree diff to see what's changed, it is more efficient to always keep the `GestureDetector`, but in the loading state the onTap is null or does nothing.

If you change the tree, first no GestureDetector, then GestureDetector, flutter has more work to do.

That is nitpicking, but it is a good habit to take (even though on some cases we cannot avoid it).

tom-anders: Ah, that makes sense, will keep in mind!

Still, the GestureDetector was actually redundant here, so I removed it entirely

veloce (commented):
Just need to change the mocking of data in tests and it's good!

It would be nice to add some other tests to verify we cannot join the tournament if conditions (tournament conditions or not logged in) are not met.

Once the logic is implemented, we'd want to test the redirect to a game. But that's for another PR.

test/view/tournament/tournament_screen_test.dart
```diff
@@ -0,0 +1,180 @@
...
+void main() {
+  setUpAll(() {
+    registerFallbackValue(kTournament);
+  });
+
+  group('Tournament screen', () {
+    final isOurTestTournament = isA<Tournament>().having((t) => t.id, 'id', kTournamentId);
+
+    testWidgets('Displays tournament data and standings', (WidgetTester tester) async {
+      final mockRepository = MockTournamentRepository();
```
veloce, line 69: I prefer to mock the http client and the server responses instead, in order to test JSON decoding (it can help to track differences between what the app expects and what the server returns).

veloce:
> Maybe I overlooked your comment about this, or you forgot to submit - What needs to be changed?

Oops. Looks like this comment got lost somehow... made another one: https://github.com/lichess-org/mobile/pull/1583#discussion_r2022706754

test/view/tournament/tournament_screen_test.dart
tom-anders: Ah, yeah I found that to be a bit cumbersome, especially when I need to modify the data (e.g. change the gameId, standings, etc.)

Would you be okay if I add tests with real API responses for the `TournamentRepository` instead?

I think for studies we also did this: one test with real data for the repository, and then mock the repository in the screen test

tom-anders:
> > Once the logic is implemented, we'd want to test the redirect to a game. But that's for another PR.
> 
> The redirect is already implemented actually, but not via the socket message, as we could miss the message while we're still in the game. Instead, we're watching the `gameId` here: https://github.com/tom-anders/lichess-mobile/blob/tournamentScreen/lib/src/view/tournament/tournament_screen.dart#L47
> 
> So we could (and should) already test it I think

Hmm, I tried, but I keep getting 

```
══╡ EXCEPTION CAUGHT BY FLUTTER TEST FRAMEWORK ╞════════════════════════════════════════════════════
The following assertion was thrown running a test:
A Timer is still pending even after the widget tree was disposed.
'package:flutter_test/src/binding.dart':
Failed assertion: line 1606 pos 12: '!timersPending'
```

in the test when the `GameScreen` is opened. The stack trace looks related to the websocket client, but I'm not familiar enough with the game screen to figure out why that happens

tom-anders:
> > > Once the logic is implemented, we'd want to test the redirect to a game. But that's for another PR.
> > 
> > 
> > The redirect is already implemented actually, but not via the socket message, as we could miss the message while we're still in the game. Instead, we're watching the `gameId` here: https://github.com/tom-anders/lichess-mobile/blob/tournamentScreen/lib/src/view/tournament/tournament_screen.dart#L47
> > So we could (and should) already test it I think
> 
> Hmm, I tried, but I keep getting
> 
> ```
> ══╡ EXCEPTION CAUGHT BY FLUTTER TEST FRAMEWORK ╞════════════════════════════════════════════════════
> The following assertion was thrown running a test:
> A Timer is still pending even after the widget tree was disposed.
> 'package:flutter_test/src/binding.dart':
> Failed assertion: line 1606 pos 12: '!timersPending'
> ```
> 
> in the test when the `GameScreen` is opened. The stack trace looks related to the websocket client, but I'm not familiar enough with the game screen to figure out why that happens

I pushed a commit with that test if you want to try to fix it. Otherwise, we can also skip that test for now until we figure out how to do it

veloce:
> I pushed a commit with that test if you want to try to fix it. Otherwise, we can also skip that test for now until we figure out how to do it

The reason why timers were still pending is that the game controller `onDispose` callback was not called. I am not sure *exactly* what's going on with riverpod in that case, but that happens if the tests don't give the controller the time to properly load.
Here [this future](https://github.com/lichess-org/mobile/blob/c31f0be6c7177d254c65853ee69c55c8672e3dd8/lib/src/model/game/game_controller.dart#L89) was never resolved, so the test failed.

So we cannot just expect `GameScreen`, we need to load the game.

test/view/tournament/tournament_screen_test.dart
tom-anders: Right, I wasn't sure if using strong interpolation would be considered "cheating" since then you're not really using the real world server responses anymore, but alright, will keep it in mind for the next time :)

lib/src/view/tournament/tournament_screen.dart
```diff
@@ -0,0 +1,506 @@
...
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {
+    return PlatformScaffold(
+      appBarTitle: _Title(state: state),
+      body: Column(
+        children: [
+          Expanded(
+            child: Padding(
+              padding: Styles.bodySectionPadding,
+              child: SingleChildScrollView(
```
veloce, line 84: This should be a `ListView`

lib/src/view/tournament/tournament_screen.dart
```diff
@@ -0,0 +1,506 @@
...
+
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {
+    final standing = state.tournament.standing;
+    if (standing == null) {
+      return const SizedBox.shrink();
+    }
+    return Column(
+      children: [
+        ListView.builder(
```
veloce, line 160: This should not be a list view, since we're showing only 10 items and it's not scrolling.

lib/src/view/tournament/tournament_screen.dart
```diff
@@ -0,0 +1,506 @@
...
+  const _StandingControls({required this.state});
+
+  final TournamentState state;
+
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {
+    return Row(
+      mainAxisAlignment: MainAxisAlignment.center,
+      children: [
+        IconButton(
```
veloce, line 268: Use PlatformIconButton instead.

lib/src/view/tournament/tournament_screen.dart
```diff
@@ -0,0 +1,506 @@
...
+              header: const Shimmer(child: LoadingPlayerWidget()),
+              footer: const Shimmer(child: LoadingPlayerWidget()),
+            );
+        }
+      },
+    );
+  }
+}
+
+class _FeaturedGamePlayer extends StatelessWidget {
```
veloce, line 412: I'd prefer the featured game board be like a broadcast mini board. Using the `header` and `footer` of `BoardThumbnail`.

test/model/tournament/tournament_repository_test.dart
```diff
@@ -0,0 +1,872 @@
...
+                  ),
+                  StandingPlayer(
+                    user: const LightUser(id: UserId('gerogr1231'), name: 'Gerogr1231'),
+                    rating: 1261,
+                    provisional: false,
+                    score: 17,
+                    sheet: (fire: false, scores: [2, 1, 0, 4, 4, 2, 2, 0, 2].toIList()),
+                    withdraw: false,
+                  ),
+                  //{
```
veloce, line 747: Remove the commented code.

veloce (commented):
More comments found after testing.

tom-anders:
Thanks, will take another look!

While I'm at it, I'll also refactor the testing code to be end  to-end as discussed

tom-anders:
> Couple of notes:
> 
>     * the whole screen should probably be refreshable (pull-to-refresh)

The server already sends a `reload` event when something notable changes, so I don't think manual refreshing would have any benefit

>     * the featured game doesn't seem to move to the next one

Ah, it's because currently we only allow one socket at a time: https://github.com/lichess-org/mobile/blob/05d57ced0faf9bf09ac591b64ad4d1a68db89f9a/lib/src/network/socket.dart#L535

But here, we need the tournament socket to stay active. If I delete that linked code, moving to the next game works (but it's probably not the solution we want...?)

test/model/tournament/tournament_repository_test.dart
tom-anders: whoops, that was only there to help copilot :D

lib/src/view/tournament/tournament_screen.dart
tom-anders: Can you elaborate? I'm already using `_FeaturedGamePlayer` as the `footer`/`header` of the `BoardThumbnail`

veloce:
> The server already sends a reload event when something notable changes, so I don't think manual refreshing would have any benefit

Then it needs to register app lifecycle onPause/onResume callbacks to force a reload on resume

> But here, we need the tournament socket to stay active. If I delete that linked code, moving to the next game works (but it's probably not the solution we want...?)

indeed it is by design we don't want more than one socket connection at a time; how does it work on the website? I don't think the tournament page opens 2 sockets. IIRC the old mobile app also do not open 2 sockets at the same time and still has the featured game.

tom-anders:
> > The server already sends a reload event when something notable changes, so I don't think manual refreshing would have any benefit
> 
> Then it needs to register app lifecycle onPause/onResume callbacks to force a reload on resume

Done :+1: 

> > But here, we need the tournament socket to stay active. If I delete that linked code, moving to the next game works (but it's probably not the solution we want...?)
> 
> indeed it is by design we don't want more than one socket connection at a time; how does it work on the website? I don't think the tournament page opens 2 sockets. IIRC the old mobile app also do not open 2 sockets at the same time and still has the featured game.

Ah, I checked with Firefox devtools, the website sends a `startWatching` to the server, so we need to do the same. I guess that means we can't reuse the `TvController` anymore, but that's fine

test/view/tournament/tournament_screen_test.dart
```diff
@@ -474,7 +556,7 @@ void main() {
       ]);
       // wait for socket message
       await tester.pump(const Duration(milliseconds: 10));
-      expect(find.byType(PieceWidget), findsNWidgets(32));
```
tom-anders: @veloce FYI: After adding the feature game in the tests as well, this expectation failed as it seems to still find the PieceWidgets from the FeatureGame board. So I replaced it with a check to find the name of our opponent instead to verify that the game screen is shown.

tom-anders:
Alright, we're using the tournament websocket now for displaying the featured game, now everything works even when there's a new featured game.

I also made the board look more like in broadcasts:

Screenshot_1743714662

lib/src/model/tournament/tournament.dart
```diff
@@ -228,8 +228,15 @@ Tournament _tournamentFromPick(RequiredPick pick) {
 }
 
 Tournament _updateTournamentFromPartialPick(Tournament tournament, RequiredPick pick) {
+  final newFeaturedGameId = pick('featured', 'id').asGameIdOrNull();
   return tournament.copyWith(
-    featuredGame: pick('featured').asFeaturedGameOrNull(),
+    // Sometimes a new FEN comes in via the websocket, but the API reload response
+    // still has the FEN of the previous move, leading to sync issues.
+    // So only copy the whole game here if it's a new ID.
```
tom-anders, line 235: Without this logic, I saw some glitches where moves were taken back due to API and websocket being out of sync

lib/src/view/tournament/tournament_screen.dart
```diff
@@ -366,74 +363,86 @@ class _FeaturedGame extends ConsumerWidget {
...
-              size: boardSize,
-              fen: featuredGame.fen,
-              orientation: featuredGame.orientation,
-              header: const Shimmer(child: LoadingPlayerWidget()),
-              footer: const Shimmer(child: LoadingPlayerWidget()),
-            );
-        }
+
+        // Don't bother parsing the whole FEN here, we just need the turn
+        final turn = featuredGame.fen.endsWith(' b') ? Side.black : Side.white;
```
tom-anders, line 368: @veloce Does chessground have some kind of convenience API for this? I didn't want to waste performance by parsing the whole string here

lib/src/view/tournament/tournament_screen.dart
veloce: I meant the style of the clock be like the ones in broadcasts (smaller clocks, orange when active).

tom-anders: Done :)

lib/src/view/tournament/tournament_screen.dart
veloce: Nope. That would be more a dartchess thing but for now there is only the Position API which parses the full fen. I don't think performance is an issue here though, so I wouldn't bother.

test/view/tournament/tournament_screen_test.dart
veloce: @tom-anders you could refine by using `find.descendant` to check the pieces are descendant of `GameScreen`.

lib/src/model/tournament/tournament.dart
veloce: Have you checked how it works on the website? Given the reload mechanism I find it strange indeed to use both this and the websocket.

lib/src/view/tournament/tournament_screen.dart
tom-anders: Well for bullet games this might be called pretty often

lib/src/model/tournament/tournament.dart
tom-anders: I can check again and/or ask @ornicar, but I think the reload events are not tied to the events of the featured game, but rather to the score/standings changing

lib/src/view/tournament/tournament_screen.dart
veloce: Fen parsing is cheap enough that even for real games (playing), the FEN is parsed each time the position change (it's a chessground parameter).

lib/src/model/tournament/tournament.dart
veloce: No it's fine like that. You need the reload events to know there is a new game. I assume the server sends the `reload` socket message when a new featured game starts.

lib/src/view/tournament/tournament_screen.dart
tom-anders: Alright 👍

lib/src/model/tournament/tournament.dart
tom-anders: Exactly, when there's a new featured game, we get it via `reload` and then send `startWatching` via the socket to get the updates for each move

veloce:
Sorry it messed the PR commit history...

veloce:
Will merge soon I think, still some testing and a couple of UI changes.

PR #1587, opened 2025-03-26
Puzzle streak success flashes times played

### LINKED ISSUE #781: Puzzle streak success flashes times played
When playing a puzzle streak on the app, the rating and amount of plays briefly flash after completion, before rapidly switching to the next puzzle.

This is too quick to be useful to anyone; plus, the rating is already displayed in the bottom right corner so having the information be displayed twice is unnecessary. It would be better to either display this information while the puzzle is being played, as is done in the web app, or get rid of it entirely as it is a bit distracting and wholly unnecessary.

cl-bruant-steineur:
Fixes #781

### DIFF
--- a/lib/src/view/puzzle/puzzle_feedback_widget.dart
+++ b/lib/src/view/puzzle/puzzle_feedback_widget.dart
@@ -6,8 +6,6 @@ import 'package:lichess_mobile/src/model/puzzle/puzzle_controller.dart';
 import 'package:lichess_mobile/src/model/settings/board_preferences.dart';
 import 'package:lichess_mobile/src/styles/styles.dart';
 import 'package:lichess_mobile/src/utils/l10n_context.dart';
-import 'package:lichess_mobile/src/utils/string.dart';
-import 'package:lichess_mobile/src/view/account/rating_pref_aware.dart';
 
 class PuzzleFeedbackWidget extends ConsumerWidget {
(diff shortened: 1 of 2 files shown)


lib/src/widgets/clock.dart
```diff
@@ -73,6 +73,8 @@ class Clock extends StatelessWidget {
                     ? isEmergency
                         ? effectiveClockStyle.emergencyBackgroundColor
                         : effectiveClockStyle.activeBackgroundColor
+                    : isEmergency
```
veloce, line 76: Please make separate PRs. This should not be part of this one.

veloce (commented):
@QuartzIII the tests are failing, can you look into that please?

Also can you please send a video recording showing your fix? (even better would be a "before/after"). I need that to validate the UI. Thanks.

lib/src/view/puzzle/puzzle_feedback_widget.dart
```diff
@@ -48,17 +44,6 @@ class PuzzleFeedbackWidget extends ConsumerWidget {
                         : context.l10n.puzzlePuzzleComplete,
                     overflow: TextOverflow.ellipsis,
                   ),
-          subtitle:
-              onStreak && state.result == PuzzleResult.lose
-                  ? null
-                  : RatingPrefAware(
```
veloce, line 54: You cannot remove that, this widget is also used for regular puzzles.

PR #1588, opened 2025-03-26
Edit analysis controllers build methods

julien4215:
- Returns state.requireValue in the build method of analysis controllers to avoid overriding the current state if it changes before reaching the return statement.
- Tweak the analysis controllers build methods


veloce:
What is the purpose of this PR?

julien4215:
The main purpose was the first point of my comment
> Returns state.requireValue in the build method of analysis controllers to avoid overriding the current state if it changes before reaching the return statement.

but it doesn't seem to be a problem actually.

PR #1589, opened 2025-03-26
Update Flutter version

julien4215:
I was wondering if we want to pin the Dart version since we are doing it for the Flutter version or if we prefer to use `^`.


veloce:
We don't have to pin the dart version imo.

julien4215:
Ok I kept only the Flutter version pinned

Issue #1573, opened 2025-03-23
Clock: Turn button disappears, if time is selected via button within field


malo1972:
### What is the problem?

If you select a time via the small buttons (that allow indiviual setting), the turn button for rotating to landscape disappears

### Steps to reproduce the bug

1. Select a time via individual button
2. watch out for the turn button :-)

### App version

0.14.14

### Device Info

Samsung S10

PR #1590, opened 2025-03-27
Make clockOrientation nullable and update related references for impr…

okmsbun:
closes https://github.com/lichess-org/mobile/issues/1573

This PR addresses an issue caused by commit b13152d. While that commit resolved the disappearing turn button in portrait mode, it inadvertently broke the tablet mode view. Additionally, as discussed in PR [#1531](https://github.com/lichess-org/mobile/pull/1531#issuecomment-2709984560) the use of systemChrome has been complicating the code structure.

### DIFF
--- a/lib/src/model/clock/clock_tool_controller.dart
+++ b/lib/src/model/clock/clock_tool_controller.dart
@@ -44,7 +44,7 @@ class ClockToolController extends _$ClockToolController {
       whiteTime: _clock.whiteTime,
       blackTime: _clock.blackTime,
       activeSide: Side.white,
-      clockOrientation: ClockOrientation.portraitUp,
+      clockOrientation: null,
     );
   }
 
@@ -169,7 +169,7 @@ class ClockToolController extends _$ClockToolController {
     state = state.copyWith(paused: false);
   }
 
-  void toggleOrientation(ClockOrientation clockOrientation) {
+  void toggleOrientation(ClockOrientation? clockOrientation) {
     state = state.copyWith(clockOrientation: clockOrientation);
   }
 }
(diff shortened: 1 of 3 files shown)


okmsbun:
Of course, I don't think it's logical for it to look like this in tablet mode. Also, during dispose, it only checks whether it's a tablet (based on screen size). However, in main, when the app starts, 'DeviceOrientation.portraitUp' is applied only on Android ([#L148](https://github.com/okmsbun/lichess-mobile/blob/4d48816695a7bb4f577e13e27010408ed1f8a7ff/lib/src/init.dart#L148)). This line doesn't run on iOS, and as soon as you exit the clock, iOS is also forced into 'DeviceOrientation.portraitUp'. There's a contradiction here :D

Screenshot 2025-03-27 at 13 34 47

veloce:
Indeed it should not look like that. I tested it only on iOS, but is not working as expected on android. Instead the view should be locked up on portrait mode.

I think we don't want auto-rotate on that specific screen. Using system chrome is the only way. Also `clockOrientation` should not be nullable because we want the same behaviour on phones and tablets.

If you don't get exactly what I mean, I can send a screen recording of what I did on an iPad.

> Also, during dispose, it only checks whether it's a tablet (based on screen size).

Normal since it locks the screen on portrait only for tablets

> However, in main, when the app starts, 'DeviceOrientation.portraitUp' is applied only on Android ([#L148](https://github.com/okmsbun/lichess-mobile/blob/4d48816695a7bb4f577e13e27010408ed1f8a7ff/lib/src/init.dart#L148)).

that is normal, on iOS the lock is done in the Info.plist so there is no need to do it here (there is no such way to do it on android afaik)

PR #1597, opened 2025-03-29
feat: add tournament support to game screen

tom-anders:
[tournamentGame.webm](https://github.com/user-attachments/assets/c9f2cd29-3242-4994-8f7b-a05d17e843eb)

TODO:
- [ ] add berserk.aifc


tom-anders:
(converting to a draft until the other PR is merged. I also need to do some more real-world testing

tom-anders:
@veloce what's the process for adding new sound assets? I found the mp3 for the berserk sound in the old app, but looks like for the new app I need to convert to .aifc as well?

tom-anders:
Did some testing with my local lila instance and fixed some small bugs. I think this is ready to go now :)

tom-anders:
> If you cannot convert easily to aifc I can do it myself no worries. The process to add new sounds is to check if the sound is available in the supported sound themes. Then adding all the required assets.

alright, I added the mp3 from the old app, I couldn't figure out how to get ffmpeg to convert that to aifc

lib/src/model/tournament/tournament.dart
```diff
@@ -355,11 +356,12 @@ FeaturedGame _featuredGameFromPick(RequiredPick pick) {
     orientation: pick('orientation').asSideOrThrow(),
     fen: pick('fen').asStringOrThrow(),
     lastMove: pick('lastMove').asUciMoveOrNull(),
-    finished: pick('finished').asBoolOrNull(),
```
veloce, line 358: Might still want to keep it. Perhaps I'm wrong but if server provides that fields doesn't it mean that absence of clocks is not enough to determine a tournament is finished on some cases?

lib/src/view/game/game_body.dart
```diff
@@ -472,6 +473,16 @@ class _GameBottomBar extends ConsumerWidget {
                 },
                 icon: Icons.info_outline,
               ),
+            if (gameState.canBerserk)
+              BottomBarButton(
+                // TODO l10n
+                label: 'Berserk',
```
veloce, line 479: You now have the arena l10n module.

lib/src/view/game/game_player.dart
```diff
@@ -148,6 +155,16 @@ class GamePlayer extends StatelessWidget {
...
                   ),
                 ),
+              if (player.berserk == true) ...[
+                const SizedBox(width: 5),
+                Icon(
+                  LichessIcons.body_cut,
+                  color: LichessColors.brag,
+                  size: playerFontSize,
+                  // TODO l10n
+                  semanticLabel: 'Berserked',
```
veloce, line 165: available

veloce (commented):
Looks good!

Note I only reviewed the code, and won't be able to test it before a few days.

lib/src/model/tournament/tournament.dart
tom-anders: I'm not sure why I thought so, but the server does not actually provide this field: https://github.com/lichess-org/lila/blob/cc2156805a95b5156e096166db067fd7c8950a07/ui/tournament/src/interfaces.ts#L100

As far as I understand it, you can fully determine the state from the combination of `clocks`, `winner` and `fen`

1) If `clocks` is not null, the game is either started or not started yet
   a) If the FEN is the initial FEN, the game is not started (yes, this breaks for some variants, but lila uses the same heuristic and most tournament games start quickly)
   b) Otherwise, the game is started.
2) If `clocks` is null and `winner` is null, the game is finished and it's a draw
3) Otherwise, the game is finished and `winner` tells us who won

assets/sounds/berserk.aifc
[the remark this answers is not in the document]
tom-anders: ah, `ffmpeg` threw an error when I tried to convert, didn't realize it actually did output something. will remove

lib/src/model/tournament/tournament.dart
veloce: Oh I thought it was for the tournament, but it is for the featured game. Fine then.

veloce:
Not properly merged because it was rebased on main, but effectively merged.

PR #1606, opened 2025-04-06
feat: add badge for current streak score in puzzle tab

tom-anders:
Would be a nice quality of life improvement IMO :)

[streakBadge.webm](https://github.com/user-attachments/assets/1fd434d6-2ede-40d7-ba1b-045f02f9a697)

### DIFF
--- a/lib/src/model/puzzle/puzzle_controller.dart
+++ b/lib/src/model/puzzle/puzzle_controller.dart
@@ -281,12 +281,12 @@ class PuzzleController extends _$PuzzleController with EngineEvaluationMixin {
 
   void _saveCurrentStreakLocally() {
     if (state.streak != null) {
-      ref.read(streakStorageProvider(initialContext.userId)).saveActiveStreak(state.streak!);
+      ref.read(streakStorageProvider(initialContext.userId)).saveActiveStreak(ref, state.streak!);
     }
   }
 
   void _sendStreakResult() {
-    ref.read(streakStorageProvider(initialContext.userId)).clearActiveStreak();
+    ref.read(streakStorageProvider(initialContext.userId)).clearActiveStreak(ref);
 
     if (initialContext.userId != null) {
       final streak = state.streak?.index;
--- a/lib/src/model/puzzle/puzzle_controller.dart
+++ b/lib/src/model/puzzle/puzzle_controller.dart
@@ -281,12 +281,12 @@ class PuzzleController extends _$PuzzleController with EngineEvaluationMixin {
 
   void _saveCurrentStreakLocally() {
     if (state.streak != null) {
-      ref.read(streakStorageProvider(initialContext.userId)).saveActiveStreak(ref, state.streak!);
+      ref.read(streakStorageProvider(initialContext.userId)).saveActiveStreak(state.streak!);
     }
   }
 
   void _sendStreakResult() {
-    ref.read(streakStorageProvider(initialContext.userId)).clearActiveStreak(ref);
+    ref.read(streakStorageProvider(initialContext.userId)).clearActiveStreak();
 
     if (initialContext.userId != null) {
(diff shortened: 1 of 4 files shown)


ijm8710:
Cool idea tom!

lib/src/model/puzzle/puzzle_controller.dart
```diff
@@ -281,12 +281,12 @@ class PuzzleController extends _$PuzzleController with EngineEvaluationMixin {
 
   void _saveCurrentStreakLocally() {
     if (state.streak != null) {
-      ref.read(streakStorageProvider(initialContext.userId)).saveActiveStreak(state.streak!);
+      ref.read(streakStorageProvider(initialContext.userId)).saveActiveStreak(ref, state.streak!);
```
veloce, line 284: That is weird to pass the `ref` like that, especially since the StreakStorage is a provider.

Idiomatic way with riverpod is to pass the `Ref` as a class constructor parameter.

tom-anders: ah thanks for the reminder, fixed :+1:

lib/src/model/puzzle/streak_storage.dart
```diff
@@ -12,12 +13,24 @@ part 'streak_storage.g.dart';
...
+  return StreakStorage(ref, userId);
+}
+
+/// Fetches the current streak score from the local storage if available, returns null otherwise.
+@riverpod
+Future<int?> savedStreakScore(Ref ref) async {
+  final session = ref.watch(authSessionProvider);
+  // cannot use ref.watch because it would create a circular dependency
+  // as we invalidate this provider in the storage saveActiveStreak and clearActiveStreak methods
+  final streakStorage = ref.read(streakStorageProvider(session?.user.id));
```
veloce, line 25: The proper way to deal with this imo, is to invalidate the `streakStorageProvider` directly (better to self invalidate than invalidate another provider to avoid such issues).

tom-anders: Right, FYI I copied this from https://github.com/lichess-org/mobile/blob/a6a8c0e719cdf4069cedb68208e7a28501f10563/lib/src/model/correspondence/correspondence_game_storage.dart#L26, so maybe it should be changed over there as well

tom-anders: Hmm but if I change it to `ref.invalidate(streakStorageProvider);` it breaks the `StreakScreen`:

[streakBroken.webm](https://github.com/user-attachments/assets/ade1aa23-256b-4016-84cd-46afb4efc30e)

tom-anders: If `clearActiveStreak` invalidates the whole provider, the `GAME OVER` text is missing and we go right to the next puzzle.

That's probably also the reason why it's not done that way for offline correspondence?

tom-anders: > hmm actually I noticed some other issues with streak which I'm fixing now and invalidating storage directly is not really possible because it makes reload other providers that we don't want to. So I'll leave it as you did.

Alright! Sorry I didn't see this comment before I wrote the two comments above

tom-anders:
@veloce I noticed the badge is gone in recent versions, probably a regression from adding the new "Learn" tab - or was it intentional?

tom-anders:
Oh nevermind lol, I just forgot how the feature works. I thought it would display my highscore for some reason, but that's not how it works of course, it displays the current streak.

But now that I think about it, maybe we could display the highscore for streak (and maybe puzzle storm) in that Tab somewhere as well?

veloce:
I don't know, because I'm not sure where to put that in the tab.

We could add a top bar button in streak screen to show the score, as we do for Storm. The screen would be rather empty but that is not a problem.

PR #1610, opened 2025-04-08
fix: disable chat if account is in kid mode

### LINKED ISSUE #821: Add Kid Mode Support
Disable all communications for the user when kid mode is enabled.

Implement a password system when setting up kid mode, and have it sync with the website.

tom-anders:
This partially fixes #821

Screenshot_1744142473

### DIFF
--- a/lib/src/model/account/account_repository.dart
+++ b/lib/src/model/account/account_repository.dart
@@ -36,6 +36,11 @@ Future<LightUser?> accountUser(Ref ref) async {
 }
 
 @riverpod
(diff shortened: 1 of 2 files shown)


tom-anders:
This only disabled (and greys out) the chat button, as can be seen in the screenshot above, we might want to completely hide the button instead

veloce:
Yes we should completely hide it in kid mode.

tom-anders:
> Yes we should completely hide it in kid mode.

Done :+1:

veloce (approved):

PR #1612, opened 2025-04-09
Fix broadcast engine gauge

julien4215:
Fix engine gauge shown on broadcast game screen when computer analysis is disabled


veloce:
Test is missing.

PR #1614, opened 2025-04-11
Revert allow using TvController w/o providing a channel

julien4215:
Revert [allow using TvController w/o providing a channel](https://github.com/lichess-org/mobile/commit/cf31c4cb96d732bfa16f50db8d9aa6de0d0d53c5) because it is actually not possible to use the TV controller for the tournament featured game.


tom-anders:
Ah yep I forgot about that, thanks!

Issue #1619, opened 2025-04-15
Change sun icon (☀️) to moon (🌙) in welcome message based on real-time day/night


mdabidhasan00:
Hi Lichess team,

I’ve noticed that in the beta version of the Lichess mobile app, the welcome message says:

☀️ Hello, [username]

This sun icon appears regardless of the time of day, even during night time. I think it would be a nice touch to change this icon dynamically based on the user's local time—using a moon icon (🌙) at night and a sun icon (☀️) during the day. This would make the app feel more personalized and visually responsive to the user's environment.

Sorry if I have used the wrong tag for this post.

veloce:
That would be nice indeed. However we don't want to use geolocation for that, so it would not always be accurate according to the season.

pafmoura:
Hi @veloce and Lichess Team,
I believe it could be implemented based on the user's local time, without geolocation. While this wouldn't account for seasonal variations in sunrise/sunset times, it would still provide a nice touch in a generic way.

In a more refined case, if the user has a country added to their profile, we could use an average latitude/longitude + the date to easily estimate these times (Though this might be overengineering for this use case)

If you think any of these aproaches are relevant, I will be happy to contribute!

veloce:
Sure, go ahead!

PR #1621, opened 2025-04-17
Add time-based greeting icon (☀️/🌙) #1619

pafmoura:
#1619 Adds a dynamic icon to home screen greeting that changes based on the time of day:
- Displays ☀️ between 6:00 AM and 5:59 PM (generic daytime)
- Displays 🌙 between 6:00 PM and 5:59 AM (generic nighttime)

In a more advanced implementation, if the user's country is available in their profile, we could use the average latitude and longitude in combination with the current date to estimate these times more accurately. However, this might be an over-engineered solution for the current use case.
Greeting Icon


veloce (approved):
Thanks!

PR #1634, opened 2025-04-21
Better filter cloud evals

### LINKED ISSUE #1623: Game analysis not trustworthy due to low-quality cloud evals
When I analyze my games, I constantly hit cloud evals that either have very low depth (often <20, as low as 14) or greatly differ from SF 16/17 eval at similar depth (up to more than 100%, presumably due to cloud eval using old SF version). For these reasons, it’s impossible to do trustworthy game analysis in the app.

Potential solutions:
1. Allow to request local eval even if cloud eval exists. I believe this is what the web app does. However, it’s very inconvenient to manually request local eval each time because there are so many low-depth cloud evals and even high-depth cloud evals can’t be trusted.
2. Add setting to ignore cloud evals below a configurable depth. This would be my favorite solution.
3. Improve cloud evals to use a single known SF version and depth >= 35 (discard all other cloud evals).

veloce:
Partially adresses #1623

Changes:
- cloud evals are not requested if ply > 15 (unless for broadcasts)
- cloud eval don't override anymore deeper local eval
- deeper local eval will not override cloud eval
- local eval will always override cloud eval if search time settings is set to 'infinity'
- cloud eval will not be requested if a threefold repetition pattern is detected
-

### DIFF
--- a/lib/src/model/analysis/analysis_controller.dart
+++ b/lib/src/model/analysis/analysis_controller.dart
@@ -710,14 +710,11 @@ class AnalysisState with _$AnalysisState implements EvaluationMixinState {
     IList<PgnComment>? pgnRootComments,
   }) = _AnalysisState;
 
+  @override
+  bool get alwaysRequestCloudEval => currentPosition.ply < 15;
+
   /// Whether the analysis is for a lichess game.
   bool get isLichessGameAnalysis => gameId != null;
-
-  /// Whether to delay the local engine evaluation.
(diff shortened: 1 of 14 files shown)


fcurts:
Any chance to have an “infinite” value to enforce that cloud eval is never used?

veloce:
I reconsidered the use of a settings here, since I realised the logic to filter out unwanted cloud evals could be improved.

PR #1635, opened 2025-04-21
Update edit_profile_screen.dart

TechiePirate:
An warning message while exiting the edit profile page is only displayed when there is a change in the form and the user hasn't saved it yet.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,10 +1,7 @@
 name: Tests
 
 on:
-  push:
-    branches: [ "main" ]
-  pull_request:
-    branches: [ "main" ]
+  on: workflow_dispatch
 
 # Declare default permissions as read only.
 permissions: read-all
--- a/lib/src/view/account/edit_profile_screen.dart
+++ b/lib/src/view/account/edit_profile_screen.dart
@@ -125,6 +125,7 @@ class _EditProfileFormState extends ConsumerState<_EditProfileForm> {
     'links': null,
   };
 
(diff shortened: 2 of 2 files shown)


veloce:
Your code has some formatting issues @TechiePirate (you need to run `dart format .`)

TechiePirate:
Can you check it now?

TechiePirate:
Ignore the comment. I made it mistakenly.

.github/workflows/test.yml
```diff
@@ -1,10 +1,7 @@
 name: Tests
 
 on:
-  push:
-    branches: [ "main" ]
-  pull_request:
-    branches: [ "main" ]
+  on: workflow_dispatch
```
veloce, line 4: You shouldn't change that, really.

lib/src/view/account/edit_profile_screen.dart
```diff
@@ -12,16 +12,12 @@ import 'package:lichess_mobile/src/widgets/buttons.dart';
 import 'package:lichess_mobile/src/widgets/feedback.dart';
 import 'package:lichess_mobile/src/widgets/platform_alert_dialog.dart';
 import 'package:result_extensions/result_extensions.dart';
-
```
veloce, line 15: Don't understand why there is all these lines removed. Something is wrong here.

veloce (changes requested):
Still some issues with formatting it seems. (Unwanted lines removal).

Issue #1681, opened 2025-05-01
Use FadeForwardsPageTransition to better match Material Design 3


TheLastFlame:
FadeForwardsPageTransition corresponds to the standard transition animation for Material Design 3.

In addition, it is devoid of the problem of "sticking" of the button press effect, when returning to the previous screen the button press effect remains visible, which looks strange and unnatural.

https://github.com/user-attachments/assets/bc0249fc-89fa-4410-8103-708097f5fa25

PR #1685, opened 2025-05-02
use FadeForwardsPageTransitionBuilder

TheLastFlame:
close https://github.com/lichess-org/mobile/issues/1681

https://github.com/user-attachments/assets/f96b188e-3416-4698-9e6e-55dac73a0904

### DIFF
--- a/lib/src/theme.dart
+++ b/lib/src/theme.dart
@@ -123,6 +123,12 @@ ThemeData _makeDefaultTheme(
     menuTheme: isIOS ? _kCupertinoMenuThemeData : null,
     bottomSheetTheme: isIOS ? _kCupertinoBottomSheetTheme : null,
     sliderTheme: kSliderTheme,
+    pageTransitionsTheme: const PageTransitionsTheme(
+      builders: {
+        TargetPlatform.android: FadeForwardsPageTransitionsBuilder(),
+        TargetPlatform.iOS: CupertinoPageTransitionsBuilder(),
+      },
+    ),
     extensions: [lichessCustomColors.harmonized(theme.colorScheme)],
   );
 }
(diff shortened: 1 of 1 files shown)


veloce:
Last time I introduced it there were bugs; will check if they are fixed now and if so, will merge. Thanks!

veloce:
Bug is not fixed in the version we're using: https://github.com/flutter/flutter/issues/165681

We'll have to wait for the next stable release.

veloce:
We never used the master channel, but the beta. And indeed there was a bug that impacted us with last beta so I decided to use the stable channel. I found that beta channel was not that useful for us, and flutter team don't even push hot fixes to beta, contrary to the stable one.

julien4215:
I think we should wait for this bug to be fixed too https://github.com/flutter/flutter/issues/164241

TheLastFlame:
@julien4215, I'm not sure if this makes sense. This behavior is relevant for both the currently used ZoomPageTransitionBuilder and CupertinoPageTransitionBuilder, since they snap the page, actually animating the frozen picture instead of the page.

Also, I can't confirm the relevance of this problem right now. It doesn't happen in my application on the stable branch. And I don't remember anything in the builder code that could block interaction.

julien4215:
> @julien4215, I'm not sure if this makes sense. This behavior is relevant for both the currently used ZoomPageTransitionBuilder and CupertinoPageTransitionBuilder, since they snap the page, actually animating the frozen picture instead of the page.

Sorry, I didn't get what you said. What behavior are you referring to and how does it relate to the bug I mentioned?

TheLastFlame:
@julien4215, the bug you're referring to says that you can't interact with the screen below during return animation. But this behavior is just as relevant for other animation builders, not just FadeForwards. And right now there are two builders already in use in the application that have a similar problem.

TheLastFlame:
@julien4215, ZoomPageTransitionsBuilder (which is used in lichess now) has a similar problem. But it's faster, and it has less impact on the user. So I agree, let's wait for a fix. I've already discovered how to fix it.

TheLastFlame:
Yes, both fixes are already in the stable channel

PR #1686, opened 2025-05-02
fix: studies with illegal positions

### LINKED ISSUE #1662: Study containing an empty board fails to load
### What is the problem?

I wanted to open a study and it failed to load. I opened it in the old app and noticed that the first chapter which is just an introduction shows an empty board. I guess the new app fails to open the study because of this. Please mind it's not an empty study, there's just a chapter with an empty board and it seems like the new app fails to load the whole study because of it.

Old Lichobile:

Image

New app:

Image

1. Open a study which contains an empty board (Introduction chapter for example)
2. The app fails to open the study

### App version

0.15.5

### Device Info

Xiaomi Redmi Note 10 5G / Android 13 (MIUI 14)

### Additional information

_No response_

tom-anders:
Making `evaluationContext` nullable might be an alternative solution

Fixes #1662

### DIFF
--- a/lib/src/model/study/study_controller.dart
+++ b/lib/src/model/study/study_controller.dart
@@ -139,7 +139,12 @@ class StudyController extends _$StudyController
         isOnMainline: true,
         root: null,
         currentNode: StudyCurrentNode.illegalPosition(),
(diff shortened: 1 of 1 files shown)


veloce (approved):
Thanks!

PR #1735, opened 2025-05-14
Http log screen only displays path when host is Lichess

julien4215:
Show query parameters and fragment in http log screen when using kLichessHost


lib/src/view/settings/http_log_screen.dart
```diff
@@ -176,7 +176,11 @@ class HttpLogTile extends StatelessWidget {
       ),
       title: Text(
         httpLog.requestUrl.host == kLichessHost
-            ? httpLog.requestUrl.path
+            ? Uri(
+              path: httpLog.requestUrl.path,
+              query: httpLog.requestUrl.query,
+              fragment: httpLog.requestUrl.fragment,
```
veloce, line 182: I see an empty `#` in each entry now. We should only show path and query and remove the fragment (which we do not even use).

PR #1736, opened 2025-05-14
Implement broadcast search feature

### LINKED ISSUE #1321: Request: a search field for broadcasts
...as is available on https://lichess.org/broadcast.

julien4215:
Close #1321


lib/src/model/broadcast/broadcast_providers.dart
```diff
@@ -26,7 +25,7 @@ class BroadcastsPaginator extends _$BroadcastsPaginator {
     final broadcastList = state.requireValue;
     final nextPage = broadcastList.nextPage;
 
-    if (nextPage == null || nextPage > 20) return;
```
veloce, line 29: Why removing the limit? Have you tested that?

lib/src/view/broadcast/broadcast_search_screen.dart
```diff
@@ -0,0 +1,112 @@
...
+  void onSubmitted(String term) {
+    setState(() {
+      searchTerm = term;
+    });
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    final searchBar = PlatformSearchBar(
+      hintText: 'Search broadcasts', // TODO translate
```
veloce, line 34: Simply use the already translated `Search` string.

veloce (commented):
Great work!

A test case is missing. Should be easy to add one since we already have tests for broadcast list and for player search screen.

lib/src/model/broadcast/broadcast_providers.dart
julien4215: I just tested and we need it. I wanted to remove it because if they change the limit on the server, we will also need to change it on the app and it will actually even break the app until we update it.

julien4215:
The tests are failing because Flutter released a new stable version.

lib/src/model/broadcast/broadcast_providers.dart
veloce: Only if the limit is lowered server side. There's little chance for that, and it is not a critical issue if it happens.

veloce:
I updated main with new flutter stable version. The conflicts are because of the new code formatting that is applied with dart 3.8. I think you just need to run dart format to fix conflicts @julien4215

PR #1802, opened 2025-05-24
feat: conditional premoves in correspondence matches

tom-anders:
[conditional_premoves.webm](https://github.com/user-attachments/assets/6899ee4c-e676-4542-b040-46e6dc3e25fb)

### DIFF
--- a/lib/src/model/analysis/analysis_controller.dart
+++ b/lib/src/model/analysis/analysis_controller.dart
@@ -3,11 +3,13 @@ import 'dart:async';
 import 'package:collection/collection.dart';
 import 'package:dartchess/dartchess.dart';
 import 'package:fast_immutable_collections/fast_immutable_collections.dart';
+import 'package:flutter/foundation.dart';
 import 'package:freezed_annotation/freezed_annotation.dart';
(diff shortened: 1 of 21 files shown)


veloce (commented):
Great piece of work!

I didn't realise it was that complex, and I think you've covered all, even the corner cases. Thanks for this!

It is great you added a lot of tests; it would also be nice to have tests for the new methods added in `node.dart` and `uci.dart`, because these utilities are used everywhere.

Can you take care of fixing the conflicts @tom-anders ? I'll merge this after.

tom-anders:
> Great piece of work!
> 
> I didn't realise it was that complex, and I think you've covered all, even the corner cases. Thanks for this!
> 
> It is great you added a lot of tests; it would also be nice to have tests for the new methods added in `node.dart` and `uci.dart`, because these utilities are used everywhere.
> 
> Can you take care of fixing the conflicts @tom-anders ? I'll merge this after.

Sure, I'll fix the conflicts and add also add the additional tests in the next few days 👍

tom-anders:
Done :)

veloce (approved):

PR #1807, opened 2025-05-27
feat: highlight live move in correspondence analysis

tom-anders:
Like web also does, and like we already do for broadcasts

[correspondenceLiveMove.webm](https://github.com/user-attachments/assets/69fddbcc-b9ad-422d-8e56-99dcd950fb56)


veloce (approved):
Nice!

PR #1835, opened 2025-06-04
screen to be on if playing game or watching game

MunibullahShah:


veloce:
There is already a [WakeLock](https://github.com/lichess-org/mobile/blob/cd0d627612195f4e1c25b9ac4df9a1d0a8b4e18c/lib/src/utils/immersive_mode.dart#L37) widget. You should use it, no need to create one.

Also game screen is already keeping screen on (contrary to what the issue says...), so no need to modify game screen.
Happens here: 
https://github.com/lichess-org/mobile/blob/cd0d627612195f4e1c25b9ac4df9a1d0a8b4e18c/lib/src/view/game/game_body.dart#L215

(admitedly it would be better if it was on game screen, it could be changed);

MunibullahShah:
Removed the services created and used the already-made widget.
Also moved the Wakelock from GameBody to GameScreen with the condition if the loadedGame !=null

lib/src/view/game/game_screen.dart
```diff
@@ -181,28 +192,31 @@ class _GameScreenState extends ConsumerState<GameScreen> {
...
-                ? _StandaloneGameTitle(id: loadedGame.gameId, lastMoveAt: widget.lastMoveAt)
-                : widget.seek != null
-                ? _LobbyGameTitle(seek: widget.seek!)
-                : widget.challenge != null
-                ? _ChallengeGameTitle(challenge: widget.challenge!)
-                : const SizedBox.shrink(),
-
-            actions: [if (loadedGame != null) _GameMenu(gameId: loadedGame.gameId)],
+        return WakelockWidget(
+          shouldEnableOnFocusGained: () => loadedGame != null,
```
veloce: The problem is that it is not the condition we want.

Before it was:

```dart
shouldEnableOnFocusGained: () => gameState.game.playable,
```

So I think you should revert the changes on game_body.dart and game_screen.dart.

lib/src/view/watch/tv_screen.dart
```diff
@@ -59,6 +60,16 @@ class _TvScreenState extends ConsumerState<TvScreen> {
   final _whiteClockKey = GlobalKey(debugLabel: 'whiteClockOnTvScreen');
   final _blackClockKey = GlobalKey(debugLabel: 'blackClockOnTvScreen');
 
+  @override
```
veloce: This code does nothing, it can be removed.

MunibullahShah: these two keys are used in the UI in tv_screen in ` CountdownClockBuilder`

lib/src/view/game/game_screen.dart
```diff
@@ -109,6 +109,16 @@ class _GameScreenState extends ConsumerState<GameScreen> {
   final _blackClockKey = GlobalKey(debugLabel: 'blackClockOnGameScreen');
   final _boardKey = GlobalKey(debugLabel: 'boardOnGameScreen');
 
+  @override
```
veloce: please remove that

MunibullahShah: They are being used in GameBody where the comments says that these variables are being used to preserve state if orientation changes.

MunibullahShah:
@veloce any update on this?

lib/src/view/game/game_screen.dart
veloce: I meant the empty `initState` and `dispose`; they are useless here.

lib/src/view/watch/tv_screen.dart
veloce: Sorry if the comment was not placed on the good line.

This is useful:
```dart
final _whiteClockKey = GlobalKey(debugLabel: 'whiteClockOnTvScreen');
  final _blackClockKey = GlobalKey(debugLabel: 'blackClockOnTvScreen');
```

This is useless:
```dart
@override
  void initState() {
    super.initState();
  }

  @override
  void dispose() {
    super.dispose();
  }
```

lib/src/view/game/game_screen.dart
MunibullahShah: Got it, removing it now :)

MunibullahShah:
@veloce removed the code in both files

lib/src/view/watch/tv_screen.dart
```diff
@@ -56,9 +57,6 @@ class _TvScreenState extends ConsumerState<TvScreen> {
     userId: widget.user?.id,
   );
 
-  final _whiteClockKey = GlobalKey(debugLabel: 'whiteClockOnTvScreen');
```
veloce, line 59: Sorry for the delay.
I still cannot accept the PR because these keys are removed where they should be kept.

veloce:
Closing due to inactivity.

PR #1858, opened 2025-06-10
Remove close button parameter from board settings screen

julien4215:
Remove unnecessary close button on board settings screen that was added in commit e5c466b6bc.


veloce:
Are you sure it is safe? What about iOS?

julien4215:
The parameter is never used. I think it was added as a workaround when there was problems with fade forwards page transition.

PR #1859, opened 2025-06-10
Update Social links domains

SergioGlorias:
Trying follow https://github.com/lichess-org/lila/blob/3fe64d0a574770913b8776642c9bd9e21391fdcb/modules/user/src/main/Links.scala#L35-L53


lib/src/model/user/profile.dart
```diff
@@ -80,21 +80,24 @@ enum LinkSite {
...
       'mindly.social',
       'mastodon.world',
-      'masthead.social',
       'techhub.social',
+      'im-in.space',
+      'mastodon.cloud',
     ]),
   ),
-  twitter('Twitter', IListConst(['twitter.com'])),
+  twitter('x', IListConst(['twitter.com', 'x.com'])),
```
veloce, line 100: Isn't it supposed to be a capital letter, 'X'?

SergioGlorias: I did it as it is in lila's code 
When Twitter was abandoned by lichess, it was changed to look like this

SergioGlorias: I did it as it is in lila's code 
When Twitter was abandoned by lichess, it was changed to look like this

PR #1864, opened 2025-06-12
Enable shrink resources on android release builds

julien4215:
I don't know if there was a reason to enable only minify and not shrink resources but the [Android documentation](https://developer.android.com/topic/performance/app-optimization/enable-app-optimization) recommends to enable both.


veloce:
TBH I don't remember, so let's try to shrink.

PR #1869, opened 2025-06-13
Updates Android configuration files

julien4215:
Several updates to Android configuration files to match the defaults used by Flutter in version 3.32:

  - Migrate Gradle scripts to Kotlin DSL

  - Update Gradle, AGP and Kotlin Gradle Plugin versions

  - Replace the gradle.properties file by the default one of Flutter

  - Add a query used by Flutter engine and remove one for url_launcher that is actually not needed


veloce:
Awesome stuff, many thanks for that!

PR #1884, opened 2025-06-15
Modernize Android splash screen

poliver:
Visually, there shouldn't be much difference, but this does allow for the clean up a number of `.png` assets and (now) redundant styles. These changes leverage Android's splash screen [compatibility library](https://developer.android.com/develop/ui/views/launch/splash-screen/migrate#library), which backports Android 12's splash screen features to API 23+.

The splash screen's theme now extends the framework `Theme.SplashScreen` and is applied to the main Activity in essentially the same way it was before. It utilizes a single vector drawable for the logo across all API levels ~(derived from [here](https://commons.wikimedia.org/wiki/File:Lichess_logo_2019.png), so please let me know if there's a better one I should use)~. I referenced [these docs](https://developer.android.com/develop/ui/views/launch/splash-screen#dimensions) for new asset's dimensions. _I just noticed while taking screenshots for this PR that the existing app crops the horse's ear on API 31+._

For reference, the [Flutter samples](https://github.com/flutter/samples/tree/main/android_splash_screen) apply a splash screen in a similar manner to this PR.

API 30 (30 and lower now using compat lib)
|Dark before|Dark after|Light before|Light after|
|-|-|-|-|
|dark_30_before|dark_30_after|light_30_before|light_30_after|

API 36 (31+ mostly unchanged)
|Dark before|Dark after|Light before|Light after|
|-|-|-|-|
|dark_36_before|dark_36_after|light_36_before|light_36_after|


android/app/src/main/res/values-night/styles.xml
```diff
@@ -1,22 +1,18 @@
...
+
     <!-- Theme applied to the Android Window as soon as the process has started.
          This theme determines the color of the Android Window while your
          Flutter UI initializes, as well as behind your Flutter UI while its
          running.
 
          This Theme is only used starting with V2 of Flutter's Android embedding. -->
     <style name="NormalTheme" parent="@android:style/Theme.Black.NoTitleBar">
-        <item name="android:windowBackground">@color/launch_black</item>
+        <item name="android:windowBackground">?android:colorBackground</item>
```
poliver, line 16: `?android:colorBackground` is what was already being used on API 31+, so I used it here as well.

android/app/src/main/res/values/colors.xml
```diff
@@ -1,4 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
-    <color name="launch_black">#131416</color>
```
poliver, line 4: This appears to predate the newer color (`?android:colorBackground`) being used for API 31+, so I removed it

android/app/src/main/res/values-night/styles.xml
veloce: IIRC I put this because I'd rather have an always black screen than a white one while the system itself is in dark mode.

Doest the `?android:colorBackground` follow system dark mode?

veloce:
Great stuff! Thanks for this.

> (derived from [here](https://commons.wikimedia.org/wiki/File:Lichess_logo_2019.png), so please let me know if there's a better one I should use).

You can find logo svg sources here: https://github.com/lichess-org/lila/tree/master/public/logo. (Note the logo looks good in your screenshots).

veloce (commented):
Judging from the screenshots this is better than before, so these changes look good to me.

However you might want to have a look at: https://pub.dev/packages/flutter_native_splash

We're using this package in the application, see:
- https://github.com/lichess-org/mobile/blob/d283f64003de449df2883f601609b35d6cc5fd72/pubspec.yaml#L138

This is needed to continue to show the splash screen while flutter is loading, so the splash screens defined here should be the same as the ones defined on android side.

android/app/src/main/res/values-night/styles.xml
poliver: It should, since it just references a themed value and the current theme is overridden to dark when the device is in dark mode. I'm not completely sure it's necessary (or if it's the desired theme attr to use), but I didn't want to change anything that could potentially introduce a regression. It may be better to just hardcode to black/white if that's the desired behavior (I believe that's what ?colorBackground resolves to anyway).

poliver: I posted in my other comment, but `?android:colorBackground` is coming from `flutter_native_splash`.

poliver:
Oh I see, what I removed is all codegen. Sorry, I'm new to Flutter. From what I can tell, `flutter_native_splash` simply generates the equivalent of my PR (I'm sure you already know this 😅). However, it does not leverage Android's splash screen compat library for API 26+ support of the Android 12 style, nor does it support svg. It's also apparently where that `?android:colorBackground` is coming from.

I you'd like to replace what `flutter_native_splash` generates with this PR, it looks like I can add `android: false` to the package configuration to prevent it overwriting these changes. I don't imagine the splash screen will be changed very often, but I can understand if you'd like to keep the simplicity afforded by the codegen. I will say that `flutter_native_splash`'s implementation of a splash screen on Android is quite strange, even despite its outdated method for doing so.

veloce:
> I you'd like to replace what `flutter_native_splash` generates with this PR, it looks like I can add `android: false` to the package configuration to prevent it overwriting these changes. I don't imagine the splash screen will be changed very often, but I can understand if you'd like to keep the simplicity afforded by the codegen. I will say that `flutter_native_splash`'s implementation of a splash screen on Android is quite strange, even despite its outdated method for doing so.

Actually I don't remember how this package works, as it was a long time ago. From what I remember the reason I used this is to avoid an empty screen during the loading time of Flutter engine.
I don't care about code gen, and I'd rather use the android splash screen and the improvements you did.

As long as the trick of delaying the removal of splash screen, as done here: https://github.com/lichess-org/mobile/blob/c825724de6a4d5619b0736f15b84066a57eef70e/lib/src/app.dart#L30 works I'm fine.

If I understand you correctly, we can use `android: false`, so we can use the native android config while retaining the control to hide the splash screen. So we should definitely go for it.

poliver:
Looking at the [implementation](https://github.com/jonbhanson/flutter_native_splash/blob/8c65117f0a4adc04f93deeb595a699a0d6dac559/lib/flutter_native_splash.dart#L48) for `FlutterNativeSplash.remove()`, it only appears to do anything on web? Again, I'm not entirely sure about the inner workings of Flutter, but on the devices I've tested, the transition from splash to Flutter content looks to be unchanged with this PR.

>If I understand you correctly, we can use `android: false`, so we can use the native android config while retaining the control to hide the splash screen. So we should definitely go for it.

Running `flutter_native_splash` does provide a native Android config, as seen below (as run against this branch), so there should be no change there as far as Flutter is concerned. It's just outdated, unfortunately. I will give `android: false` a shot and report back.
image

poliver:
Applied official logo in df717dbeb6fbfdbf78a1ef0e6b9c73a4beacae84

poliver:
Configured `flutter_native_splash` to only run for iOS in f38d60a0ef55aaae9497de879302862320df8f98

android/app/src/main/AndroidManifest.xml
```diff
@@ -23,19 +23,13 @@
...
             android:resizeableActivity="false"
             android:supportsPictureInPicture="false"
             android:windowSoftInputMode="adjustResize">
-            <!-- Specifies an Android theme to apply to this Activity as soon as
-                 the Android process has started. This theme is visible to the user
-                 while the Flutter UI initializes. After that, this theme continues
-                 to determine the Window background behind the Flutter UI. -->
-            <meta-data
-                android:name="io.flutter.embedding.android.NormalTheme"
-                android:resource="@style/NormalTheme"/>
```
poliver, line 38: This appears to no longer be necessary with Flutter 2.5 based on a note [here](https://docs.flutter.dev/platform-integration/android/splash-screen). This meta data is not applied in the corresponding [sample app](https://github.com/flutter/samples/blob/main/android_splash_screen/android/app/src/main/AndroidManifest.xml#L34).

android/app/src/main/res/values/styles.xml
```diff
@@ -1,22 +1,24 @@
...
+    <!-- Shim in the system UI overrides -->
+    <style name="SplashTheme.NightAdjusted" parent="Theme.SplashScreen">
+        <item name="android:windowLightStatusBar">true</item>
+        <item name="android:windowLightNavigationBar" tools:targetApi="27">true</item>
     </style>
+
+    <style name="SplashTheme" parent="SplashTheme.NightAdjusted">
+        <item name="windowSplashScreenBackground">@color/bg_splash</item>
+        <item name="windowSplashScreenAnimatedIcon">@drawable/ic_splash</item>
+        <item name="postSplashScreenTheme">@style/NormalTheme</item>
```
poliver, line 12: This is what tells Android to switch themes once the app is loaded.

veloce:
Tested on 3 different devices with Android 9, 10, 11, and an emulator with Android 15.

Thanks again!

Issue #1642, opened 2025-04-23
OTB to support "give 15 seconds" and take back"


ijm8710:
Not a huge priority but unless the omission is intentional, would assume it makes sense to eventually support

tom-anders:
takeback is already possible, just press the back button in the bottom bar

ijm8710:
That's fair and I'll admit I didn't think of it that way, but I think of the back arrow is more of a rewind. We can primary this ticket for the other part of it for now and then whomever can  decide if it makes sense to have an actual takeback button as well to me. It makes sense to be uniform with the other modes

veloce:
Let's not add complexity here.

PR #1885, opened 2025-06-15
Give 15 seconds over the board.

dl8sd11:
#1642

### DIFF
--- a/lib/src/model/over_the_board/over_the_board_clock.dart
+++ b/lib/src/model/over_the_board/over_the_board_clock.dart
@@ -93,6 +93,16 @@ class OverTheBoardClock extends _$OverTheBoardClock {
 
     state = state.copyWith(activeClock: newSideToMove);
   }
+
+  void giveTime(Duration amount) {
+    final newTime = state.timeLeft(state.activeClock!)! + amount;
+
+    if (state.activeClock == Side.white) {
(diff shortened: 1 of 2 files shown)


lib/src/model/over_the_board/over_the_board_clock.dart
```diff
@@ -93,6 +93,16 @@ class OverTheBoardClock extends _$OverTheBoardClock {
 
     state = state.copyWith(activeClock: newSideToMove);
   }
+
+  void giveTime(Duration amount) {
+    final newTime = state.timeLeft(state.activeClock!)! + amount;
```
veloce, line 98: I assume the clock can be disabled in which case this method would crash. You should check for that first.

lib/src/view/over_the_board/over_the_board_screen.dart
```diff
@@ -329,6 +329,12 @@ class _BottomBar extends ConsumerWidget {
               );
             },
           ),
+        BottomSheetAction(
```
veloce, line 332: If there is no clock this button should be hidden.

veloce (commented):
Thanks for your contribution. I'll merge it when missing checks are added.

veloce:
No response so closing.

PR #1886, opened 2025-06-15
Remove NDK version from build file to let Android pick the latest one.

julien4215:
Flutter plans to remove the NDK pinning in templates as said in this Github comment https://github.com/flutter/flutter/issues/139427#issuecomment-1989024985.


veloce:
I added this because of a warning when you build the app. Why do you think we should ignore that warning?

PR #1892, opened 2025-06-16
Clean up notification icon

poliver:
- `drawable-anydpi-v24/` is unnecessary now that we are on v26. `drawable-anydpi/` is sufficient
-  `.png` variants are not being used, because a vector drawable variant is available

I opened this as a draft, because I don't have a great way to verify the changes (trigger a notification). I've tried the standard adb commands. Do I just need to trigger a notification from a separate device, or is there a simpler way?


veloce:
I use the script here: https://github.com/lichess-org/mobile/blob/main/scripts/firebase/send-message.js

but if you want to use it you need your own firebase account, and need to replace the `firebase_options.dart` file we use here.

PR #1910, opened 2025-06-23
CI: Upload build artifacts and fix iOS Firebase Swift error

spellr:
This PR updates the CI pipeline to upload Android (.aab) and iOS (.app) build artifacts, and fixes a Swift compiler error when building iOS with Firebase by enabling the required experimental feature flag in the Podfile.
Currently the build CI job is failing, and artifacts are not uploaded - [link](https://github.com/spellr/lichess-mobile/actions/runs/15810161012)
After these 2 changes, the build CI is now fixed, and artifacts are uploaded to GitHub for easy compiling - [link](https://github.com/spellr/lichess-mobile/actions/runs/15827306890)

### DIFF
--- a/ios/Podfile
+++ b/ios/Podfile
@@ -38,6 +38,9 @@ post_install do |installer|
   installer.pods_project.targets.each do |target|
     target.build_configurations.each do |config|
       config.build_settings['IPHONEOS_DEPLOYMENT_TARGET'] = '13.0'
+      if ENV['CI'] == 'true'
+        config.build_settings['OTHER_SWIFT_FLAGS'] ||= ['$(inherited)', '-enable-experimental-feature', 'AccessLevelOnImport']
+      end
     end
     flutter_additional_ios_build_settings(target)
   end


ios/Podfile
```diff
@@ -38,6 +38,7 @@ post_install do |installer|
   installer.pods_project.targets.each do |target|
     target.build_configurations.each do |config|
       config.build_settings['IPHONEOS_DEPLOYMENT_TARGET'] = '13.0'
+      config.build_settings['OTHER_SWIFT_FLAGS'] ||= ['$(inherited)', '-enable-experimental-feature', 'AccessLevelOnImport']
```
veloce: I can build the app without this on my computer. Can you explain why it is necessary for CI? 

I also wish it would be possible to configure this outside of the podfile as I'm reluctant to enable an experimental feature for prod builds where it is needed only on CI debug build.

spellr: Thanks! The error only occurs in CI because the GitHub Actions macOS runners use Xcode 15+ by default, and `FirebaseCoreInternal` now relies on Swift's AccessLevelOnImport feature, which requires the experimental flag unless you're using precompiled Firebase binaries or older Xcode versions.

Also per your suggestion, I've added a condition to only enable the experimental flag in the CI runs.
```java
if ENV['CI'] == 'true'
  config.build_settings['OTHER_SWIFT_FLAGS'] ||= ['$(inherited)', '-enable-experimental-feature', 'AccessLevelOnImport']
end
```

PR #1919, opened 2025-06-26
refactor: make AnalysisOptions a union type

tom-anders:
The `gameId` and `standalone` properties have always been exclusive (exactly one of them had to be non-null), so I think this better expresses the intent.

Additionally, for analysing ongoing correspondence games, I'd introduce a new constructor in https://github.com/lichess-org/mobile/pull/1802 along the lines of

```
AnalysisOptions.ongoingCorrespondenceGame({
    required Side orientation,
    int? initialMoveCursor,
    required GameFullId gameId,
}) = OngoingCorrespondenceGame;
```


veloce (approved):
Nice stuff, thanks!

PR #1921, opened 2025-06-26
Prevent device gestures on the edge of screen to interfere while playing

veloce:


julesjulian:
@veloce @julien4215 this would be great, the edge-of-screen back gesture is infuriating. What can I do to move this forward?

julien4215:
@julesjulian The only two ways I know to prevent edge of the screen interference on Android is to either use [immersive mode](https://developer.android.com/develop/ui/views/layout/immersive) or use [setSystemGestureExclusionRects](https://developer.android.com/reference/android/view/View#setSystemGestureExclusionRects(java.util.List%3Candroid.graphics.Rect%3E)). Each solution has its own pro and cons and they are already used in the codebase but I believe we can improve it.

`setSystemGestureExlucsionRects` is limited to 200dp in vertical extent and on some screens, it is not enough to cover the full height of the board. What we could do is to exclude from system gesture the squares on the first and last column where there are pieces.

Immersive mode has the disavantage of entirely disabling gesture navigation. There is an option in board settings to enable it but it is disabled by default and it only applies to the game screen and puzzle screen. Maybe it could be enabled by default for devices that use gesture navigation but I don't know if there is a reliable way to tell if a device is using gesture navigation.

ijm8710:
> @julesjulian The only two ways I know to prevent edge of the screen interference on Android is to either use [immersive mode](https://developer.android.com/develop/ui/views/layout/immersive) or use [setSystemGestureExclusionRects](https://developer.android.com/reference/android/view/View#setSystemGestureExclusionRects(java.util.List%3Candroid.graphics.Rect%3E)). Each solution has its own pro and cons and they are already used in the codebase but I believe we can improve it.
> 
> 
> 
> `setSystemGestureExlucsionRects` is limited to 200dp in vertical extent and on some screens, it is not enough to cover the full height of the board. What we could do is to exclude from system gesture the squares on the first and last column where there are pieces.
> 
> 
> 
> Immersive mode has the disavantage of entirely disabling gesture navigation. There is an option in board settings to enable it but it is disabled by default and it only applies to the game screen and puzzle screen. Maybe it could be enabled by default for devices that use gesture navigation but I don't know if there is a reliable way to tell if a device is using gesture navigation.

I would absolutely love a way to have a safe space just for where the conflict is present.

I also tried seeing if we can revisit bringing back the small analysis board option #3093.

This would have two benefits:
- one is with the recent addition of analysis screen tab indicators, the analysis window is a bit crammed, a smaller board would allow for it to be larger.
- But second (and more important for this), with a smaller analysis board option, the board would have some padding on the sides which would also lessen the chance of a conflict

Issue #1795, opened 2025-05-23
When challenging leela odds bots, don't reset time control input after changing setup 


allanjoseph98:
Changing the board setup on any of the Leela odds bots challenge screens resets the time control to 1+0 and you have to input the time control all over again

PR #1930, opened 2025-07-01
fix time reseting time if bot position is changed +border radius of thumbnails

HaonRekcef:
Fix: #1795 and correct the border radius of the position thumbnails-

The Rematch button for the Leela Odd Bots isn’t working, they do not get accepted (From the website it works). I haven’t been able to investigate this since I cant play against them from the dev build. Is there a way for me to investigate this further, or could you take a look, @veloce?


veloce:
I can take a look at it @HaonRekcef . Can you open a ticket for that bug? Thanks!

PR #1943, opened 2025-07-06
Don't assume that a broadcast player is not null

### LINKED ISSUE #1856: Broadcast bug
[HttpClient] Could not read JSON object as ({IMap<String, BroadcastGame> games, IList<({String id, String name})>? group, String? groupName, BroadcastRound round, BroadcastTournamentData tournament}): PickException(Expected a non-null value but location picked value "null" using pick(json, "games", 16, "players", 1, "name" (null)) is null. Use asStringOrNull() when the value may be null/absent at some point (String?).)

julien4215:
Close #1856


veloce (approved):
Thanks!

PR #1964, opened 2025-07-13
Analysis board interface

julien4215:
Create analysis board interface to share the code of analysis, broadcast and study boards.


lib/src/model/analysis/abstract_analysis_prefs.dart
```diff
@@ -0,0 +1,8 @@
+/// Interface for Analysis's prefs.
+abstract class AbstractAnalysisPrefs {
```
veloce: I'd rather name it `CommonAnalysisPrefs`, as this is an interface for common prefs found in different contexts.

lib/src/model/analysis/abstract_analysis_state.dart
```diff
@@ -0,0 +1,43 @@
+import 'package:dartchess/dartchess.dart';
+import 'package:fast_immutable_collections/fast_immutable_collections.dart';
+import 'package:lichess_mobile/src/model/common/chess.dart';
+import 'package:lichess_mobile/src/model/common/eval.dart';
+import 'package:lichess_mobile/src/model/engine/evaluation_preferences.dart';
+
+/// Interface for Analysis's State.
+abstract class AbstractAnalysisState {
```
veloce: I'd name it `CommonAnalysisState` for same reason.

lib/src/model/analysis/abstract_analysis_state.dart
```diff
@@ -0,0 +1,43 @@
...
+  /// The variant of the analysis.
+  Variant get variant;
+
+  /// Current position in the position tree. Can be `null` to support illegal position that are
+  /// found in studies.
+  Position? get currentPosition;
+
+  /// The current node in the analysis view.
+  ///
+  /// This is an immutable copy of the actual [Node] at the `currentPath`.
```
veloce: This comment doesn't belong here since it is an interface.

lib/src/model/analysis/abstract_analysis_state.dart
```diff
@@ -0,0 +1,43 @@
...
+
+  /// The side to display the board from.
+  Side get pov;
+
+  /// Possible promotion move to be played.
+  NormalMove? get promotionMove;
+}
+
+/// Interface for Analysis's current node.
+abstract class AbstractAnalysisCurrentNode {
```
veloce: I'd rather name it `AnalysisCurrentNodeInterface`.

lib/src/view/analysis/abstract_analysis_board.dart
```diff
@@ -0,0 +1,140 @@
...
+import 'package:lichess_mobile/src/model/engine/evaluation_preferences.dart';
+import 'package:lichess_mobile/src/model/engine/evaluation_service.dart';
+import 'package:lichess_mobile/src/model/settings/board_preferences.dart';
+import 'package:lichess_mobile/src/widgets/pgn.dart';
+
+/// An abstract widget that provides the common interface for three types of analysis boards:
+/// - Analysis board
+/// - Broadcast board
+/// - Study board
+abstract class AbstractAnalysisBoard extends ConsumerStatefulWidget {
```
veloce: I'd rather avoid naming things as `AbstractX`.

There are many other ways of naming it, but I prefer this one:

- the abstract class is named `AnalysisBoard`
- and so we got 3 concrete classes:
    - `GameAnalysisBoard`
    - `BroadcastAnalysisBoard`
    - `StudyAnalysisBoard`

lib/src/view/analysis/abstract_analysis_board.dart
```diff
@@ -0,0 +1,140 @@
...
+import 'package:lichess_mobile/src/model/analysis/abstract_analysis_state.dart';
+import 'package:lichess_mobile/src/model/common/chess.dart';
+import 'package:lichess_mobile/src/model/common/eval.dart';
+import 'package:lichess_mobile/src/model/engine/evaluation_preferences.dart';
+import 'package:lichess_mobile/src/model/engine/evaluation_service.dart';
+import 'package:lichess_mobile/src/model/settings/board_preferences.dart';
+import 'package:lichess_mobile/src/widgets/pgn.dart';
+
+/// An abstract widget that provides the common interface for three types of analysis boards:
+/// - Analysis board
```
veloce: This comment is enough, so this is nitpicking, but it is even better to directly link to the concrete classes using the square brackets notation.

veloce (commented):
Great work! Thank you, this is really helpful. Only need a couple of renaming to be merged.

lib/src/view/analysis/abstract_analysis_board.dart
julien4215: I had to make the broadcast and study board classes public so I can import them and reference them with the square brackets

lib/src/view/analysis/analysis_board.dart
```diff
@@ -1,46 +1,143 @@
...
+  ISet<Shape> get extraShapes => ISet();
+
+  /// Set of shapes drawn by the user on the board (arrows, circle).
+  ISet<Shape> userShapes = ISet();
 
   @override
-  void onPromotionSelection(Role? role) =>
-      ref.read(analysisControllerProvider(widget.options).notifier).onPromotionSelection(role);
+  Widget build(BuildContext context) {
+    clearShapes();
```
veloce: I missed that in the last review, and I don't think we should do it like that. The name itself is misleading because this function calls `ref.listen`;

What about removing `clearShapes` from `AnalysisBoardState` and do:
```dart
class _StudyAnalysisBoardState extends State<...> {
   ...

   @override
   build() {
          // Clear shapes when switching to a new chapter.
    // This avoids "leftover" shapes from the previous chapter when the engine has not evaluated the new position yet.
    ref.listen(studyControllerProvider(widget.id).select((state) => state.hasValue), (prev, next) {
      if (prev != next) {
        setState(() {
          userShapes = ISet();
        });
      }
    });
    
     return super.build();
   }
}

julien4215: Sure I agree it is less misleading like this

veloce (approved):

Issue #1932, opened 2025-07-02
Rematch Buttton on Odds Bots is not working


HaonRekcef:
### What is the problem?

Asking for a rematch on the Leela Odds Bots doesnt work, and the new game never starts. From the website the rematch button works fine.

### Steps to reproduce the bug

1. Play against a LeelaQueenOdds 
2. Request rematch
3. No new game starts

### App version

0.16.5

### Device Info

-

### Additional information

I suspect it has to do something with the fact that the rematch button on leela odds bots starts a game with the same colors again.

Sanket-Kurle:
I see this is an issue 
I would really like to work on it and solve this ASAP
I will send PR regarding this once i solve this

Sanket-Kurle:
hey i have updated the PR let me know what the issue with my pr and correct me 
@HaonRekcef 
@veloce

Sanket-Kurle:
please someone check the latest PR I pushed.
@HaonRekcef 
@veloce

veloce:
Stop pinging, really.

Sanket-Kurle:
ok

PR #1967, opened 2025-07-16
Fix: Handle rematch for odds bot games

Sanket-Kurle:
This pull request resolves issue #1932. It is a new, clean submission created as requested by veloce in the original PR.

Problem Description
Currently, when a user finishes a game against an "Odds Bot" (e.g., LeelaQueenOdds) or any game started from a custom FEN position (Variant.fromPosition), the "Rematch" button in the game result dialog does not work.

Tapping the button sends a generic rematch-yes socket event, which is only suitable for standard games. For custom position games, the server requires a full, detailed challenge request specifying the variant, FEN, and other parameters. Because this information was missing, the server would ignore the request, and no new game would start, leaving the user on the home screen.

Solution
This PR fixes the issue by making the rematch logic context-aware. The fix is implemented in game_result_dialog.dart:

The onPressed callback for the "Rematch" button now inspects the game.meta.variant of the finished game.

An if/else block has been added:

If the variant is Variant.fromPosition, the code now constructs a full ChallengeRequest object. It uses the details from the completed game (opponent, clock, FEN, etc.) to create a new challenge, mirroring the working logic from the initial challenge screen. It then navigates directly to a new GameScreen with this challenge.

Else (for any standard game), it calls the original proposeOrAcceptRematch() function, which correctly sends the simple rematch-yes event.

This ensures that both standard and custom position rematches are handled correctly, resolving the bug. This PR only contains the single file change required for the fix.

How to Test
Start a game from the "Board editor" to create a Variant.fromPosition game.

Finish the game (by winning, losing, or resigning).

In the game result dialog that appears, tap the "Rematch" button.

Expected Result: A new game with the same settings should start immediately.

### DIFF
--- a/lib/src/model/game/game_controller.dart
+++ b/lib/src/model/game/game_controller.dart
@@ -12,9 +12,11 @@ import 'package:lichess_mobile/src/model/account/account_preferences.dart';
 import 'package:lichess_mobile/src/model/account/account_repository.dart';
 import 'package:lichess_mobile/src/model/account/account_service.dart';
 import 'package:lichess_mobile/src/model/analysis/analysis_controller.dart';
+import 'package:lichess_mobile/src/model/challenge/challenge.dart';
 import 'package:lichess_mobile/src/model/chat/chat_controller.dart';
 import 'package:lichess_mobile/src/model/clock/chess_clock.dart';
 import 'package:lichess_mobile/src/model/common/chess.dart';
+import 'package:lichess_mobile/src/model/common/game.dart';
 import 'package:lichess_mobile/src/model/common/id.dart';
 import 'package:lichess_mobile/src/model/common/service/move_feedback.dart';
 import 'package:lichess_mobile/src/model/common/service/sound_service.dart';
@@ -29,6 +31,7 @@ import 'package:lichess_mobile/src/model/game/game_status.dart';
 import 'package:lichess_mobile/src/model/game/game_storage.dart';
 import 'package:lichess_mobile/src/model/game/material_diff.dart';
 import 'package:lichess_mobile/src/model/game/playable_game.dart';
+import 'package:lichess_mobile/src/model/lobby/create_game_service.dart';
 import 'package:lichess_mobile/src/model/settings/board_preferences.dart';
 import 'package:lichess_mobile/src/network/http.dart';
 import 'package:lichess_mobile/src/network/socket.dart';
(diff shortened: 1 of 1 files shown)


lib/src/view/game/game_result_dialog.dart
```diff
@@ -97,7 +101,34 @@ class _GameResultDialogState extends ConsumerState<GameResultDialog> {
                             gameState.game.opponent?.onGame == true &&
                             gameState.game.opponent?.offeringRematch != true
                         ? () {
-                            ref.read(ctrlProvider.notifier).proposeOrAcceptRematch();
+                            final game = gameState.game;
+
+                            // Check if it was an Odds Bot game
```
veloce: This comment is useless and misleading because `fromPosition` doesn't necessarily mean an odds bot game.

lib/src/view/game/game_result_dialog.dart
```diff
@@ -97,7 +101,34 @@ class _GameResultDialogState extends ConsumerState<GameResultDialog> {
...
+                                  context,
+                                  challenge: ChallengeRequest(
+                                    destUser: game.opponent!.user!,
+                                    variant: Variant.fromPosition,
+                                    timeControl: ChallengeTimeControlType.clock,
+                                    clock: (
+                                      time: game.meta.clock!.initial,
+                                      increment: game.meta.clock!.increment,
+                                    ),
+                                    rated: false,
```
veloce: In theory you can create a rated from position challenge, so you should use the game information to determine if the challenge is rated or not.

lib/src/view/game/game_result_dialog.dart
```diff
@@ -97,7 +101,34 @@ class _GameResultDialogState extends ConsumerState<GameResultDialog> {
...
+                            if (game.meta.variant == Variant.fromPosition) {
+                              // If yes, build the detailed challenge and navigate to the new GameScreen
+                              Navigator.of(context, rootNavigator: true).push(
+                                GameScreen.buildRoute(
+                                  context,
+                                  challenge: ChallengeRequest(
+                                    destUser: game.opponent!.user!,
+                                    variant: Variant.fromPosition,
+                                    timeControl: ChallengeTimeControlType.clock,
+                                    clock: (
```
veloce: We can't assume the clock is always here I believe. So you should check this condition in the above `if`.

Sanket-Kurle:
Hi @veloce,

Thank you very much for your detailed feedback and for guiding me toward the correct solution. I really appreciate you taking the time to explain the proper architecture.

I have closed this pull request and created a new, clean one that implements your suggestions. The new PR contains only the single file change in game_controller.dart and all checks are passing.

You can find the new pull request here #1970 

Thanks again!

PR #1976, opened 2025-07-20
fix: manually update the puzzle context when the puzzle changes

### LINKED ISSUE #1951: Puzzle Streak: Can't reset when failing very first puzzle
### What is the problem?

Puzzle Streak seems to get into an inconsistent state when you immediately fail the very first puzzle 

1) Open Puzzle Streak
2) Immediately play the wrong move on the very first puzzle
3) Press the "Play Button" in the bottom bar

Now the state is not reset correctly, nothing happens when a correct or incorrect move is made. After the player makes a move, it's possible to move the opponent's pieces 

### App version

0.16.5

### Device Info

Moto G 5

### Additional information

_No response_

OldUser101:
Closes #1951

This issue occurred because the `PuzzleController` was stuck in the `PuzzleMode.view` mode after an incorrect move. The  `PuzzleController` never transitions into `PuzzleMode.play`, which allows the puzzle to be played. This is because the `PuzzleController` and the `PuzzleStreakController` aren't properly synchronised. Under some scenarios, like when the very first move of a puzzle streak is incorrect, the `PuzzleController` retains the old state, and the puzzle does not advance properly.

This pull request adds a listener to detect when the state of the `PuzzleStreakController` changes, and updates it in `PuzzleController`, allowing the new puzzle to be played.

### DIFF
--- a/lib/src/view/puzzle/streak_screen.dart
+++ b/lib/src/view/puzzle/streak_screen.dart
@@ -104,6 +104,26 @@ class _BodyState extends ConsumerState<_Body> {
     );
     final puzzleState = ref.watch(ctrlProvider);
 
+    ref.listen(puzzleStreakControllerProvider, (previous, next) {
+      if (previous?.hasValue == true && next.hasValue) {
+        final currentPuzzle = next.requireValue.puzzle;
+        final previousPuzzle = previous!.requireValue.puzzle;
+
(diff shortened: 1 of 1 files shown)


tom-anders:
Cool! This is a way better solution than me simply not failing the very first puzzle lol

PR #1978, opened 2025-07-20
fix: switching pages for large tournaments

tom-anders:
`/tournament/<id>?page=<page>` seems to have a bug where `page>200` will always return page 200. The web uses `/tournament/<id>/standing/<page>` API as well.


veloce (approved):
Thanks!

PR #1983, opened 2025-07-22
Upload libapp.so debug symbols to Play Console

julien4215:


android/app/build.gradle.kts
```diff
@@ -64,9 +64,6 @@ android {
                 "proguard-rules.pro"
             )
             signingConfig = signingConfigs.getByName("release")
-            ndk {
```
veloce, line 67: why this change?

scripts/add-debug-symbols.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+
```
veloce: you can declare a
```sh
DIR=$(cd `dirname $0` && pwd)
```

so the other dirs can be referenced relative to the script dir

scripts/add-debug-symbols.sh
veloce, line 1: Is this script a workaround because there is a current bug in flutter where the debug symbols should be added automatically?

If so let's put that in a comment on top of the script.

veloce (commented):
Thanks for this!

We cannot test the upload itself without making a new version. But have you tested the script? You can run the `fastlane` command locally, and just comment out the upload part to see that the symbols are correctly added.

julien4215:
I didn't test the fastlane file because I don't have fastlane installed but I tested the script I added and from what I can tell it works as expected.

android/app/build.gradle.kts
julien4215: Because I think it is useless in our case. I had added it before because I thought this was the reason we were missing stack trace symbolization.

scripts/add-debug-symbols.sh
julien4215: I am not sure if it is considered as a bug but I added a comment to give some context.

scripts/add-debug-symbols.sh
julien4215: Yes but since this script is meant to be ran only by Fastlane which uses the root project directory maybe we can leave it like this.

veloce: Are you really sure about this?

See the GitHub action for instance: https://github.com/lichess-org/mobile/blob/7e1c385f3256b9fcd321cf8960db2afa604380ce/.github/workflows/deploy_play_store.yml#L47

julien4215: Ah maybe not then. I thought it was the case because we are storing the debug info at this location `./build/app/outputs/bundle/release/symbols` in the Fastfile.

julien4215: According to [fastlane documentation](https://docs.fastlane.tools/advanced/fastlane/#directory-behavior) the script is ran from the fastlane directory so I moved the script into the fastlane directory. I think file paths should be correct now.

android/fastlane/add-debug-symbols.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/sh
+#
+# Because Flutter is not adding the symbols of libapp.so to the App Bundle, this
+# script add the libapp.so symbols to the archive native-debug-symbols.zip in order
+# to upload them to the Play Console with Fastlane.
+#
+# See https://github.com/flutter/flutter/issues/170664.
+
+cd ../..
```
veloce: By experience, I think it is much better to declare paths relative to the directory where the script is, rather than from where it is executed.
You know where the script is but you don't always know in advance from where it is executed.

In any case you should test the `fastline` command from your local computer to see that at least the script is correctly executed. `upload_to_play_store` will fail but that is not what you are testing anyway.

julien4215: I tested the script from fastlane and it works

PR #1984, opened 2025-07-23
Add Unified Push

julien4215:
Firebase messaging is removed in the `remove-firebase.patch`. This PR replaces it with Unified Push.


veloce:
I thought it was better to maintain a patch to *just remove firebase code*.

Now that the patch *is replacing code* I don't think it is a good idea. For a start it is impossible to review correctly.

I see 2 options:
1. firebase free version is maintained in a fork

2. or we keep the patch, but it must be so simple that it is easy to update and contain no logic

If we're going that way, all the code must be pushed normally to the repository, including the UnifiedPush code. The patch would be just used to switch implementations.


I'm open to suggestions if you have better ideas.

julien4215:
I agree it is not ideal but I think it is easier for now to keep a patch. Adding all the code to the main branch would be too much work as it would require an interface and the two packages are quite different. 

I don't plan to add more changes to it and all the logic is contained inside `notifications.dart`. Also it is possible to review it, you can run `git apply` and `git diff` to review the changes in the patch.

If this firebase free version gets more interest I think it could make sense to create a fork.

veloce:
Really, a patch like that is a no. And if I can't review in GitHub, I won't review it.

The notification service has already changed in the ongoing private messaging PR, and it will continue to evolve.

I understand you don't want to make a common interface. Then it must be done in a fork, it will be much easier to maintain.

julien4215:
You can revert the previous commit then. I'll push the changes to my fork.

PR #1986, opened 2025-07-24
Add quick actions

### LINKED ISSUE #1791: Platform Home Screen quick actions
Provide OS Home Screen quick actions.

Image

See: https://pub.dev/packages/quick_actions

According to [apple best practices](https://developer.apple.com/design/human-interface-guidelines/home-screen-quick-actions):
- Create quick actions for compelling, high-value tasks
- Avoid making unpredictable changes to quick actions

Therefor we should limit this to a couple of actions only:
- play a game
- play puzzles

## Play puzzle
Should redirect to the "healthy mix" puzzle run (default puzzle run)

## Play a game
Here it would be nice to launch a game directly but we need to know in advance the preferred time control (and other settings).

I see some potential useful actions here:
- create a game action: it opens the app with the "play bottom sheet" already open
- play action: 
    - it directly creates a game based on the user most played time control
    - since most played time control can change and actions should remain predictable perhaps having 2 actions for the 2 most played time controls if available is a good idea?

Related issue: #1790 (but this one is much easier to implement)

aba2222:
Resolves #1791 

Add a quick action to run "Healthy Mix" puzzle and another quick action to open quick pair screen.

<p float="left">
  
  
</p>

### DIFF
--- a/lib/src/app.dart
+++ b/lib/src/app.dart
@@ -71,6 +71,7 @@ class _AppState extends ConsumerState<Application> {
     ref.read(challengeServiceProvider).start();
     ref.read(accountServiceProvider).start();
     ref.read(correspondenceServiceProvider).start();
+    ref.read(quickActionServiceProvider).start();
 
     // Listen for connectivity changes and perform actions accordingly.
     ref.listenManual(connectivityChangesProvider, (prev, current) async {
@@ -102,13 +103,6 @@ class _AppState extends ConsumerState<Application> {
     });
 
     super.initState();
-
-    WidgetsBinding.instance.addPostFrameCallback((_) {
-      final ctx = navigatorKey.currentContext;
-      if (ctx != null) {
-        ref.read(quickActionServiceProvider).start(ctx);
-      }
-    });
   }
 
   @override
--- a/lib/src/quick_actions.dart
+++ b/lib/src/quick_actions.dart
@@ -2,12 +2,14 @@ import 'package:fast_immutable_collections/fast_immutable_collections.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:lichess_mobile/l10n/l10n.dart';
+import 'package:lichess_mobile/src/localizations.dart';
 import 'package:lichess_mobile/src/model/common/chess.dart';
 import 'package:lichess_mobile/src/model/common/perf.dart';
 import 'package:lichess_mobile/src/model/common/speed.dart';
 import 'package:lichess_mobile/src/model/lobby/game_seek.dart';
 import 'package:lichess_mobile/src/model/puzzle/puzzle_angle.dart';
 import 'package:lichess_mobile/src/model/puzzle/puzzle_theme.dart';
+import 'package:lichess_mobile/src/tab_scaffold.dart';
 import 'package:lichess_mobile/src/view/game/game_screen.dart';
 import 'package:lichess_mobile/src/view/puzzle/puzzle_screen.dart';
 import 'package:quick_actions/quick_actions.dart';
@@ -18,7 +20,6 @@ part 'quick_actions.g.dart';
 @Riverpod(keepAlive: true)
 QuickActionService quickActionService(Ref ref) {
   final service = QuickActionService(ref);
-  //ref.onDispose(service.dispose);
   ref.listen<RecentGameSeekPrefs>(recentGameSeekProvider, (previous, next) {
     service.setQuickActions(next.seeks);
   });
@@ -29,40 +30,37 @@ class QuickActionService {
   QuickActionService(this.ref);
 
   final Ref ref;
+  AppLocalizations get l10n => ref.read(localizationsProvider).strings;
 
   final QuickActions quickActions = const QuickActions();
-  BuildContext? context;
 
-  void start(BuildContext currentlyContext) {
-    context = currentlyContext;
+  void start() {
     final recentSeeks = ref.read(recentGameSeekProvider).seeks;
     quickActions.initialize((String shortcutType) {
+      final context = ref.read(currentNavigatorKeyProvider).currentContext;
+      if (context == null || !context.mounted) return;
+
       if (shortcutType.startsWith('recent_seek_')) {
         final index = int.tryParse(shortcutType.split('_').last);
         if (index != null) {
           Navigator.of(
-            context!,
+            context,
             rootNavigator: true,
-          ).push(GameScreen.buildRoute(context!, seek: recentSeeks[index]));
+          ).push(GameScreen.buildRoute(context, seek: recentSeeks[index]));
         }
       } else if (shortcutType == 'play_puzzles') {
         Navigator.of(
-          context!,
+          context,
           rootNavigator: true,
-        ).push(PuzzleScreen.buildRoute(context!, angle: const PuzzleTheme(PuzzleThemeKey.mix)));
+        ).push(PuzzleScreen.buildRoute(context, angle: const PuzzleTheme(PuzzleThemeKey.mix)));
       }
     });
     setQuickActions(recentSeeks);
   }
 
   void setQuickActions(IList<GameSeek> recentSeeks) {
-    if (context == null) return;
     quickActions.setShortcutItems(<ShortcutItem>[
-      ShortcutItem(
-        type: 'play_puzzles',
-        localizedTitle: AppLocalizations.of(context!).puzzleThemeMix,
-        icon: 'extension',
-      ),
+      ShortcutItem(type: 'play_puzzles', localizedTitle: l10n.puzzleThemeMix, icon: 'extension'),
 
       ...recentSeeks.asMap().entries.map((entry) {
         final index = entry.key;
@@ -70,9 +68,7 @@ class QuickActionService {
 
         final time = seek.clock != null ? seek.clock!.$1.inMinutes.toString() : '-';
         final increment = seek.clock != null ? seek.clock!.$2.inSeconds.toString() : '0';
-        final rated = seek.rated
-            ? AppLocalizations.of(context!).ratedTournament
-            : AppLocalizations.of(context!).casualTournament;
+        final rated = seek.rated ? l10n.ratedTournament : l10n.casualTournament;
 
         final variant = (seek.variant == Variant.standard)
             ? Variant.standard.label
@@ -86,7 +82,7 @@ class QuickActionService {
         return ShortcutItem(
           type: 'recent_seek_$index',
           localizedTitle: '$time+$increment $rated $variant',
-          localizedSubtitle: AppLocalizations.of(context!).recentGames,
+          localizedSubtitle: l10n.recentGames,
           icon: 'add_icon',
         );
       }),


veloce:
Hey! thanks for taking care of this. You should precise this is only for android.

I'm not a fan of a new screen to show the quick game matrix. It does not add much value.

I would much prefer having 2 or 3 "play action" based on the user preferences.

I reckon it is more work, as you need to "record" the setup of a game each time a game is created. But it is not that hard after all.
Some suggestions:
- ~~persist a simple `Map<GameSeek, int>` which holds the number of times a precise game seek was created~~
- this could happen in https://github.com/lichess-org/mobile/blob/inbox/lib/src/model/lobby/create_game_service.dart

This would add much more value imo. If you look at the Spotify app, this is exactly what they do: the last 3 albums you've listened to are in quick actions.

Another feedback: don't put icons, or use the official lichess icons for game kinds; eg: a blitz game should have the blitz (fire) icon, a rapid the rapid (rabbit) icon, etc.

veloce:
UPDATE: don't follow my previous comment actually 😅 

I realise the Spotify approach is better. We should show the last 3 created game types (if we have them).

So it is not a `Map<GameSeek, int` that we need to persist, but only the last 3 created `GameSeek`.

lib/src/model/lobby/game_seek.dart
```diff
@@ -113,4 +116,133 @@ sealed class GameSeek with _$GameSeek {
     if (variant != null) 'variant': variant!.name,
     if (ratingRange != null) 'ratingRange': '${ratingRange!.$1}-${ratingRange!.$2}',
   };
+
+  factory GameSeek.parseGameSeekFromRequestBody(Map<String, String> requestBody) {
```
veloce: There is an easier way to (un)serialize a freezed class. By default the `fromJson and toJson` are disabled to save build time, but you can enable it with:
```dart
@Freezed(fromJson: true, toJson: true)
sealed class GameSeek ...
```

Then you need to add the `fromJson` factory constructor.

One example here: https://github.com/lichess-org/mobile/blob/main/lib/src/model/common/chess.dart#L15

aba2222: The code now uses `fromJson` and `toJson`

aba2222:
It shows the last 3 created game types (if any) and we can click them to play the same type of game.
If the game was rated, there is an 🏅 icon appears next to its name.

**Screenshot**

meenbeese:
iOS has this feature as well. Will this implementation also work on there?

aba2222:
> iOS has this feature as well. Will this implementation also work on there?

It should work on iOS as well, but I currently don't have iOS device to test it.

veloce:
> If the game was rated, there is an 🏅 icon appears next to its name.

That won't do, because we don't use that icon in the app and people would not understand. Also we need a way to indicate this action will create a game.
Also we need to indicate which variant it is. For now we only support chess960, but more variants will come later.
Could be improved, and suggestions are welcome, but I see something like this:

Capture d’écran 2025-07-28 à 10 39 14

In place of the star, there should be the perf icon: bullet, blitz, rapid, classical or variant (960, crazyhouse, etc).

I think like that it is a bit more clear what does the action: the + icon indicates it creates a game, as well as the "Recently played".

aba2222:
> In place of the star, there should be the perf icon: bullet, blitz, rapid, classical or variant (960, crazyhouse, etc).

I think this can't be implemented. Both Android and iOS only support native layout in quick actions. Unless we replace +  icon with perf icon.

veloce:
Oh, right, and I suppose using the LichessIcon font won't work.

Then we can use the `shortTitle` of [Perf](https://github.com/lichess-org/mobile/blob/main/lib/src/model/common/perf.dart).
But only if the variant is different from standard chess. In another bullet point after "rating / casual".

aba2222:
It looks like this picture now. The subtitile "Recently played" may be hidden by some launchers.

translation/source/mobile.xml
```diff
@@ -6,6 +6,7 @@
   <string name="areYouSure" comment="This is shown in confirmation dialogs.">Are you sure?</string>
   <string name="boardSettings" comment="This is shown in the app main settings screen, which opens a screen dedicated to chess board settings.">Board settings</string>
   <string name="cancelTakebackOffer" comment="Menu action that cancels a previous takeback offer">Cancel takeback offer</string>
+  <string name="casualGame" comment="This is shown in quick actions to show the name is casual">(Casual)</string>
```
veloce: There is already a `casual` translation key that we can use. So please remove this one, we want to avoid adding extra work to translators.

translation/source/mobile.xml
```diff
@@ -34,6 +35,8 @@
   <string name="puzzleStreakAbortWarning" comment="Shown as a subtitle in the confirmation dialog for aborting puzzle streak. Prefer to not translate Puzzle Streak">You will lose your current streak and your score will be saved.</string>
   <string name="puzzleThemesSubtitle" comment="Shown as a description for the 'Puzzle themes' mode">Play puzzles from your favorite openings, or choose a theme.</string>
   <string name="puzzlesTab" comment="The tab in bottom navigation bar. Must be very short (10 characters max) to avoid ugly text shifting issues." maxLength="10">Puzzles</string>
+  <string name="ratedGame" comment="This is shown in quick actions to show the name is rated">(Rated)</string>
```
veloce: Same as casual, please remove that one.

translation/source/mobile.xml
```diff
@@ -34,6 +35,8 @@
...
+  <string name="recentlyPlayed" comment="Subtitle shown in quick actions for a recently played game">Recently played</string>
```
veloce: There is already the `arenaRecentlyPlayed` that we can use, so please remove that one.

We should be conservative when adding new keys. For new features it is even recommended to hard code the string at first. Will update the internationalisation doc to make it more clear.

lib/src/app.dart
```diff
@@ -100,6 +112,51 @@ class _AppState extends ConsumerState<Application> {
     });
 
     super.initState();
+
+    WidgetsBinding.instance.addPostFrameCallback((_) {
```
veloce: Why the `addPostFrameCallback`? it doesn't seem necessary.

lib/src/app.dart
veloce, line 1: I would rather see all the quick actions code in its own file. You can create a `lib/src/quick_actions.dart` for that.

If we want to be consistent with the rest of the app, you can create a `QuickActionService` modelled after `MessageService` for instance.

Then in `app.dart` you only need to call `ref.read(quickActionServiceProvider).start()`;

lib/src/model/lobby/game_seek.dart
```diff
@@ -113,4 +117,70 @@ sealed class GameSeek with _$GameSeek {
     if (variant != null) 'variant': variant!.name,
     if (ratingRange != null) 'ratingRange': '${ratingRange!.$1}-${ratingRange!.$2}',
   };
+
+  factory GameSeek.fromJson(Map<String, dynamic> json) => _$GameSeekFromJson(json);
+}
+
+class RecentGameSeekPrefs implements Serializable {
```
veloce: It would be better to use an `IList` and a `@freezed` class here (we do that in the rest of the app).

lib/src/model/lobby/game_seek.dart
```diff
@@ -113,4 +117,70 @@ sealed class GameSeek with _$GameSeek {
...
+    if (updated.length > 3) {
+      updated.removeRange(3, updated.length);
+    }
+    return RecentGameSeekPrefs(seeks: updated);
+  }
+
+  static const empty = RecentGameSeekPrefs(seeks: []);
+}
+
+final recentGameSeekProvider = NotifierProvider<RecentGameSeekNotifier, RecentGameSeekPrefs>(
```
veloce: I don't really mind but for consistency with the rest you could use the `@riverpod` notation.
I also don't like much the generator; hopefully it will be removed with a next version of riverpod.

lib/src/app.dart
```diff
@@ -100,6 +112,51 @@ class _AppState extends ConsumerState<Application> {
...
+      final recentSeeks = ref.watch(recentGameSeekProvider).seeks;
```
veloce: `ref.watch` can't be used in a widget lifecycle method.

But see my above comment about refactoring in quick_actions.dart.

Code will happen in `QuickActionService`. there should be 2 parts:
- first initialisation (in `start` method) where you'd use a `ref.read(recentGameSeekProvider)` to get current value
- listening to game seek provider changes (using ref.listen) and update the quick actions accordingly

veloce:
> > iOS has this feature as well. Will this implementation also work on there?
> 
> It should work on iOS as well, but I currently don't have iOS device to test it.

Will test on iOS myself, and fix what is needed.

lib/src/app.dart
aba2222: It need wait context create.

lib/l10n/app_af.arb
```diff
@@ -69,13 +69,13 @@
...
   "arenaSomeRated": "Sekere toernooie is gegradeer en sal jou gradering beïnvloed.",
   "arenaHowAreScoresCalculated": "Hoe word die puntetelling uitgewerk?",
-  "arenaHowAreScoresCalculatedAnswer": "'n Wen het 'n basis puntetelling van 2 punte, gelykop is 1 punt, en 'n verloor beteken niks nie.\nAs jy twee wedstryde in 'n ry wen, begin jy 'n dubbelpunt reeks, verteenwoordig deur 'n vlam ikoon.\nDie volgende spelle sal voortgaan om dubbelpunte werd te wees totdat jy nie meer wen nie.\nDus is 'n wen 4 punte werd, 'n gelykop 2 punte en 'n verloor is steeds waardeloos.\n\nByvoorbeeld, twee wenne gevolg deur 'n gelykop sal 6 punte werd wees: 2 + 2 + (2 x 1)",
+  "arenaHowAreScoresCalculatedAnswer": "'n Wen het 'n basis puntetelling van 2 punte, gelykop is 1 punt, en 'n verloor beteken niks nie.\r\nAs jy twee wedstryde in 'n ry wen, begin jy 'n dubbelpunt reeks, verteenwoordig deur 'n vlam ikoon.\r\nDie volgende spelle sal voortgaan om dubbelpunte werd te wees totdat jy nie meer wen nie.\r\nDus is 'n wen 4 punte werd, 'n gelykop 2 punte en 'n verloor is steeds waardeloos.\r\n\r\nByvoorbeeld, twee wenne gevolg deur 'n gelykop sal 6 punte werd wees: 2 + 2 + (2 x 1)",
```
veloce: Looks like you're using windows? Ideally the script that run in a windows environment should not change the strings and output LF instead of CRLF.

But that is another matter, not for this PR. Here please revert all the l10n changes, thanks!

lib/src/quick_actions.dart
```diff
@@ -0,0 +1,95 @@
...
+  return service;
+}
+
+class QuickActionService {
+  QuickActionService(this.ref);
+
+  final Ref ref;
+
+  final QuickActions quickActions = const QuickActions();
+  BuildContext? context;
```
veloce: You don't have to pass and store the `BuildContext` in a class instance.

A better way, get the context in the `initialize` callback:
```dart
quickActions.initialize((String shortcutType) {
    final context = ref.read(currentNavigatorKeyProvider).currentContext;
    if (context == null || !context.mounted) return;
    ...
 });
```

This is what is done elsewhere in the app, example: https://github.com/lichess-org/mobile/blob/8cc9fd9eb778e73c16b66b08d8179a07f9230d3f/lib/src/model/message/message_service.dart#L63

lib/src/app.dart
veloce: I'd rather not do that. See the comments in quick_actions.dart file.

lib/src/quick_actions.dart
```diff
@@ -0,0 +1,95 @@
...
+          context!,
+          rootNavigator: true,
+        ).push(PuzzleScreen.buildRoute(context!, angle: const PuzzleTheme(PuzzleThemeKey.mix)));
+      }
+    });
+    setQuickActions(recentSeeks);
+  }
+
+  void setQuickActions(IList<GameSeek> recentSeeks) {
+    if (context == null) return;
```
veloce: Since you're in a provider and have a `Ref`, you don't need the `BuildContext` to access `AppLocalizations`.

You can just declare a
```dart
AppLocalizations get _l10n => _ref.read(localizationsProvider).strings;
```

and use it in QuickActionsService.

You have an example of that here: https://github.com/lichess-org/mobile/blob/8cc9fd9eb778e73c16b66b08d8179a07f9230d3f/lib/src/model/notifications/notification_service.dart#L87

lib/src/quick_actions.dart
```diff
@@ -0,0 +1,95 @@
...
+import 'package:lichess_mobile/src/view/puzzle/puzzle_screen.dart';
+import 'package:quick_actions/quick_actions.dart';
+import 'package:riverpod_annotation/riverpod_annotation.dart';
+
+part 'quick_actions.g.dart';
+
+@Riverpod(keepAlive: true)
+QuickActionService quickActionService(Ref ref) {
+  final service = QuickActionService(ref);
+  //ref.onDispose(service.dispose);
```
veloce: remove this commented code.

PR #1987, opened 2025-07-24
Rewrite the broadcast pagination mechanism

julien4215:
Remove the need to use a `ScrollController` for the two broadcast screens that uses pagination.

I think this is more in line with Riverpod. See this Riverpod example https://github.com/rrousselGit/riverpod/tree/master/examples/pub.


veloce:
Yep, clever way of doing things, with less code. 👍

PR #1994, opened 2025-07-25
Tablebase Explorer

### LINKED ISSUE #1720: Should we have tablebase (explorer for endgame) support like we do on web
image

Seems to be part of opening explorer modal

HaonRekcef:
closes #1720 and implements the Endgame Tablebases.
<details>
  <summary>Screenshot</summary>

  image
</details>

<details>
  <summary>Video</summary>

  https://github.com/user-attachments/assets/2f22f93d-d219-4014-a1bc-902f09e4e4ec
</details>

Notes about the implementation:
- Created a parent widget, `explorer_view.dart`, which handles the logic for displaying either the opening explorer or the tablebase, and shares design constants used by both.
- Renamed the `opening_explorer` folders to `explorer` and included the tablebase code there as well.
- Design is similar to the opening explorer and the tablebases on the website
- Added some basic tests


lib/src/view/explorer/tablebase_view.dart
```diff
@@ -0,0 +1,369 @@
...
+                  color: index.isEven ? context.lichessTheme.rowEven : context.lichessTheme.rowOdd,
+                  onMoveSelected: onMoveSelected,
+                  isWinningForWhite: isWinningForWhite,
+                );
+              }, growable: false),
+            );
+          }
+        }
+
+        addMoveSection(
```
niklasf, line 96: Sorry, the documentation of the categories isn't great.

Here we should aim to order categories such that:
* The best moves are at the top.
* We're "honest" about unknown outcomes: Show unknown moves first, if there's a chance that they are better than known moves.

So the best order of the inverted categories is:

* `win` (we can confidently play this, even if other move results are unknown)
* `unknown` (for all we know an unknown move may be better than any non-`win` move, so prioritize displaying these)
* `syzygy-win` (a win, assuming the user has followed the tablebase recommendation since the last pawn move or capture. otherwise a win or cursed win, but the 7 piece syzygy tablebase does not know which [due to dtz rounding](https://syzygy-tables.info/metrics#rounding))
* `maybe-win` (a win or cursed win, but the upcoming 8 piece dtc tablebase does not know which)
* `cursed-win`
* `draw`
* `blessed-loss`
* `maybe-loss`
* `syzygy-loss`
* `loss`

HaonRekcef: Thanks I have updated the categories as suggested

veloce:
Thank you for this! Going to focus on releasing the next version with correspondence conditional premoves and private messaging. Then I take care of reviewing it.

lib/src/model/explorer/tablebase.dart
```diff
@@ -0,0 +1,107 @@
...
+  maybeWin,
+  cursedWin,
+  draw,
+  blessedLoss,
+  maybeLoss,
+  syzygyLoss,
+  loss,
+}
+
+@Freezed(fromJson: true)
```
veloce: The `fromJson` is not necessary here, since you're implementing it yourself using `fromPick`.

lib/src/model/explorer/tablebase.dart
```diff
@@ -0,0 +1,107 @@
...
+      checkmate: pick('checkmate').asBoolOrThrow(),
+      stalemate: pick('stalemate').asBoolOrThrow(),
+      insufficientMaterial: pick('insufficient_material').asBoolOrThrow(),
+      category: pick('category').letOrThrow((p) => _parseTablebaseCategory(p.asStringOrThrow())),
+      moves: pick('moves').asListOrEmpty((p) => TablebaseMove.fromPick(p.required())).toIList(),
+    );
+  }
+}
+
+@Freezed(fromJson: true)
```
veloce: same

lib/src/model/explorer/tablebase.dart
```diff
@@ -0,0 +1,107 @@
...
+    'unknown' => TablebaseCategory.unknown,
+    'syzygy-win' => TablebaseCategory.syzygyWin,
+    'maybe-win' => TablebaseCategory.maybeWin,
+    'cursed-win' => TablebaseCategory.cursedWin,
+    'draw' => TablebaseCategory.draw,
+    'blessed-loss' => TablebaseCategory.blessedLoss,
+    'maybe-loss' => TablebaseCategory.maybeLoss,
+    'syzygy-loss' => TablebaseCategory.syzygyLoss,
+    'loss' => TablebaseCategory.loss,
+    _ => throw ArgumentError('Unknown tablebase category: $value'),
```
veloce: Not sure about this one, but what about returning `TablebaseCategory.unknown` if parsing fails?

It depends on how important it is to determine the category, but the fact there is already an `unknown` makes me think it would be better not to fail if at some point another category is added server side.

lib/src/view/explorer/explorer_view.dart
```diff
@@ -0,0 +1,69 @@
...
+Color blackBoxColor(BuildContext context) => Theme.of(context).brightness == Brightness.light
+    ? Colors.black.withValues(alpha: 0.7)
+    : Colors.black;
+
+class ExplorerView extends ConsumerWidget {
+  const ExplorerView({
+    required this.position,
+    required this.onMoveSelected,
+    this.opening,
+    this.isComputerAnalysisAllowed = true,
```
veloce: Perhaps it was already a default in the previous implementation, but I'd rather not have one here, as it is better to be sure that computer analysis is not allowed by mistake.

lib/src/view/explorer/explorer_view.dart
```diff
@@ -0,0 +1,69 @@
...
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+import 'package:lichess_mobile/src/model/common/chess.dart';
+import 'package:lichess_mobile/src/utils/l10n_context.dart';
+import 'package:lichess_mobile/src/view/explorer/opening_explorer_view.dart';
+import 'package:lichess_mobile/src/view/explorer/tablebase_view.dart';
+
+/// Unified explorer view that shows either opening explorer or tablebase
+/// based on the position state (opening vs endgame)
+
+const kTableRowVerticalPadding = 10.0;
```
veloce: Since they are now public, I'd prefix all constants here with `kExplorer...`.

veloce (commented):
Outstanding work! thank you very much!!

Only have a couple of small comments.

lib/src/model/explorer/tablebase.dart
HaonRekcef: Good catch, makes sense yes.

lib/src/view/explorer/explorer_view.dart
HaonRekcef: Yes copied it from before, but I agree changing it to be required is better.

HaonRekcef:
Not sure why the pipeline failed, `flutter analyze` works locally

veloce:
> Not sure why the pipeline failed, `flutter analyze` works locally

You need to remove the `part 'tablebase.g.dart';` and run again the build_runner.

PR #2004, opened 2025-07-28
Revert "Upload libapp.so debug symbols to Play Console"

veloce:
Reverts lichess-org/mobile#1983


veloce:
@julien4215 this fails on CI: https://github.com/lichess-org/mobile/actions/runs/16570620751/job/46861570348#step:10:57923

Issue #1620, opened 2025-04-17
Feature request: Opponent popup window with score (me vs. opponent) and his ratings.


dstark1993:
What I liked in the older app is that during a game you could press his account name and a popup window would appear with his variant ratings, and if you ever payed him, also the score. This also worked on LichessTV, where you could see the score between the two players currently playing:

![Image](https://github.com/user-attachments/assets/a2455a30-0a88-4746-863d-e7efe2231c90)

I guess the popup is defined in:  lichobile / src / ui / shared / PlayerPopup.tsx

On the new app, you are just being transferred to opponents account, and the score is no where to be seen.
This feature could be very nice to have. Hopefully this is possible without to many changes.

Issue #820, opened 2024-07-01
Feature request: cross table + head to head record


ijm8710:
Included already on website so see this more as a future must. 

Two separate items
-head to head cross table accessible from opponent/friend profile that also includes mutual game history
-actual match page shows prior h2h record if one exists

![image](https://github.com/lichess-org/mobile/assets/25938701/0a0d54c3-5180-4844-8793-1517c8923bd5)

PR #2014, opened 2025-07-30
add crosstable scores to Userscreen

### LINKED ISSUE #1872: Crosstable of 2 players
https://lichess.org/api#tag/Users/operation/apiCrosstable

HaonRekcef:
closes #1872 
Impelements the h2h. The h2h score is shown only when the logged in user played the clicked user already. Clicking on the H2H score opens the Game History screen with the games against the opponent.
<details>
  <summary>Video</summary>

  [h2h.webm](https://github.com/user-attachments/assets/b142f5aa-664b-4cc2-8970-317815aba419)

</details>

### DIFF
--- a/lib/src/model/user/user.dart
+++ b/lib/src/model/user/user.dart
@@ -429,3 +429,18 @@ class UserRatingHistoryPoint {
 
   const UserRatingHistoryPoint({required this.date, required this.elo});
 }
+
+@freezed
+sealed class Crosstable with _$Crosstable {
+  const factory Crosstable({required Map<UserId, int> users, required int nbGames}) = _Crosstable;
+
+  factory Crosstable.fromJson(Map<String, dynamic> json) {
+    return Crosstable(
+      nbGames: pick(json, 'nbGames').required().asIntOrThrow(),
+      users: pick(
+        json,
+        'users',
+      ).asMapOrThrow<String, int>().map((key, value) => MapEntry(UserId(key), value)),
+    );
+  }
+}
--- a/lib/src/model/user/user.dart
+++ b/lib/src/model/user/user.dart
@@ -431,8 +431,28 @@ class UserRatingHistoryPoint {
 }
 
 @freezed
+sealed class CrosstableMatchup with _$CrosstableMatchup {
+  const factory CrosstableMatchup({required IMap<UserId, int> users, required int nbGames}) =
+      _CrosstableMatchup;
+
+  factory CrosstableMatchup.fromJson(Map<String, dynamic> json) {
+    return CrosstableMatchup(
+      nbGames: pick(json, 'nbGames').required().asIntOrThrow(),
+      users: pick(
+        json,
+        'users',
+      ).asMapOrThrow<String, int>().map((key, value) => MapEntry(UserId(key), value)).toIMap(),
+    );
+  }
+}
+
+@freezed
 sealed class Crosstable with _$Crosstable {
-  const factory Crosstable({required Map<UserId, int> users, required int nbGames}) = _Crosstable;
+  const factory Crosstable({
+    required IMap<UserId, int> users,
+    required int nbGames,
+    CrosstableMatchup? matchup,
+  }) = _Crosstable;
 
   factory Crosstable.fromJson(Map<String, dynamic> json) {
     return Crosstable(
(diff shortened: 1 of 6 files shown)


ijm8710:
Super excited for your work on this :) 

Few comments:

- I feel like it kinda gets caught in the shuffle when mixed with the rest of the options in that block (watch games, compose message etc). Personally I'd love if it was a separate block preferably right above the block we already have. Thoughts?
  - idea: can have a block of games, rated, wins, losses, draws. A 6 row block if you have h2h with opponent and 5 row block if not.
(All those filters seen here and it would show counts and then have easy shortcuts to filter on those result types similar to web; lmk if that's confusing, not sure I explained that well) image
  - carrying this idea even further, for your own profile this could be where we house your "imported games" in the same block
- not sure how best to design it, but I really really hope we can add the actual colored Crosstable that exists on web. It's an actual table of wins and losses for last 10 matches that is an easy summary along with total h2h record as row totals. (Seen in #820)
- in both the old app and on web, when watching chess TV you can see the two players h2h record. While slightly different, I feel it's in the spirit of the work done here and somewhat related (seen in #1620)
- similar to the previous bullet, if you rematch an opponent and or play an opponent you played in past, having the priori matchup h2h record show in the game screen. Personally I prefer how chesscom shows it but if at the very least it can be treated how we handle the chess TV h2h (seen in #820)

HaonRekcef:
> Super excited for your work on this :)
> 
> Few comments:
> 
>     * I feel like it kinda gets caught in the shuffle when mixed with the rest of the options in that block (watch games, compose message etc). Personally I'd love if it was a separate block preferably right above the block we already have. Thoughts?

I see that maybe it should be on top, will let veloce decide on this.

>       
>       * idea: can have a block of games, rated, wins, losses, draws. A 6 row block if you have h2h with opponent and 5 row block if not.>         (All those filters seen here and it would show counts and then have easy shortcuts to filter on those result types similar to web; lmk if that's confusing, not sure I explained that well)
>       * carrying this idea even further, for your own profile this could be where we house your "imported games" in the same block

I think the userscreen is already quite loaded, so we should be careful before we add even more to it.

> 
>     * not sure how best to design it, but I really really hope we can add the actual colored Crosstable that exists on web. It's an actual table of wins and losses for last 10 matches that is an easy summary along with total h2h record as row totals. (Seen in [Feature request: cross table + head to head record #820](https://github.com/lichess-org/mobile/issues/820))
>     * in both the old app and on web, when watching chess TV you can see the two players h2h record. While slightly different, I feel it's in the spirit of the work done here and somewhat related (seen in [Feature request: Opponent popup window with score (me vs. opponent) and his ratings. #1620](https://github.com/lichess-org/mobile/issues/1620))
> 
>     * similar to the previous bullet, if you rematch an opponent and or play an opponent you played in past, having the priori matchup h2h record show in the game screen. Personally I prefer how chesscom shows it but if at the very least it can be treated how we handle the chess TV h2h (seen in [Feature request: cross table + head to head record #820](https://github.com/lichess-org/mobile/issues/820))

I can't think of a suitable space for the crosstable for now, but I agree in the future we should at least add the current matchup score to the game and tv screen if there are rematches. But this probably should be done in a future pull request. I am just not sure yet how to display the scores without it confusing the players.

ijm8710:
Fair enough, just to put a visual to what I was thinking (I didn't change the icons but I think you get the idea)

image

- I think the "games with you" meshes better with these categories than the other block
- would be nice to have shortcuts to see filter game history for these categories like you can on web. ("Draws" would be one last row but I ran out of room in the mockup)
- this would be an opponent's profile view. For YOUR OWN profile, imported games can fall in this block
- notice on the bottom I felt activity and recent games should be tabbed. This would save vertical space to help balance the additional block added.i also think it makes it much easier to navigate rather than an endless scroll. Also, it's already tabbed on web.

lib/src/model/user/user.dart
```diff
@@ -429,3 +429,18 @@ class UserRatingHistoryPoint {
 
   const UserRatingHistoryPoint({required this.date, required this.elo});
 }
+
+@freezed
+sealed class Crosstable with _$Crosstable {
+  const factory Crosstable({required Map<UserId, int> users, required int nbGames}) = _Crosstable;
```
veloce: Since a freezed class is immutable, better use an `IMap` instead of a `Map`.

lib/src/view/user/user_screen.dart
```diff
@@ -186,6 +198,30 @@ class _UserProfileListView extends ConsumerWidget {
...
+              if (crosstable?.valueOrNull != null &&
+                  (crosstable?.valueOrNull?.nbGames ?? 0) > 0) ...[
+                () {
+                  final crosstableData = crosstable!.valueOrNull!;
+                  final currentUserScore = crosstableData.users[session.user.id] ?? 0;
+                  final otherUserScore = crosstableData.users[user.id] ?? 0;
+
+                  return ListTile(
+                    title: Text(context.l10n.yourScore('$currentUserScore - $otherUserScore')),
+                    leading: const Icon(Icons.sports_score),
```
veloce: `Icons.scoreboard` is better here I think.

HaonRekcef:
I addressed your reviews and added a matchup field to the api requests we will need it for the scores in TV later.

veloce (approved):

PR #2030, opened 2025-08-03
Matchup in TVScreen and GameScreen

HaonRekcef:
Adds the matchup scores to the TV screen and the game screen:

<details>
  <summary>Screenshot</summary>

  grafik

</details>
If the score is 0-0 its not displayed, as then matchup is null.

I'm not sure about the design. I think it looks decent, but there's room for improvement.

Also fixes a bug in Crosstable JSON parsing. if an `int` was sent, it couldn't be cast to a `double`.


ijm8710:
Cool! Quick question, does it show   0.5 etc for draws?

HaonRekcef:
Yes it shows 1/2 for draw.
<details>
  <summary>Screenshot</summary>
grafik

</details>
Tests seem to fail because of an unrelated issue.

ijm8710:
Awesome! My only super super nitpick is food for thought what looks cleaner 1 1/2 or 1.5. (I know both are interchangeable for a singular draw but once they stack). 

Not super opinionated but just suggesting the thought :)

veloce:
Thanks for this!

Design is OK. The color could be replaced with green and red for better and lower scores perhaps. But it would not necessarily be nice depending on the background (custom backgrounds). So it is fine like that imo.

lib/src/view/game/game_body.dart
```diff
@@ -100,9 +102,24 @@ class GameBody extends ConsumerWidget {
       case AsyncData(value: final gameState, isRefreshing: false):
         final youAre = gameState.game.youAre ?? Side.white;
 
+        AsyncValue<Crosstable>? crosstable;
+        // If playing against Stockfish, user is null
+        if (gameState.game.white.user != null && gameState.game.black.user != null) {
```
veloce: Using a `final crosstable` would be better; The ternary operator reads quite well, as it is correctly indented by dart.

lib/src/view/watch/tv_screen.dart
```diff
@@ -95,9 +96,19 @@ class _TvScreenState extends ConsumerState<TvScreen> {
                     final game = gameState.game;
                     final position = gameState.game.positionAt(gameState.stepCursor);
 
+                    AsyncValue<Crosstable>? crosstable;
```
veloce: Same remark as in game_body, use a final crosstable.

lib/src/view/game/game_player.dart
```diff
@@ -79,6 +81,24 @@ class GamePlayer extends StatelessWidget {
...
+                  decoration: BoxDecoration(
+                    color: ColorScheme.of(context).primaryContainer,
+                    borderRadius: BorderRadius.circular(4.0),
+                  ),
+                  child: Text(
+                    _scoreDisplay(matchupScore!),
+                    style: TextStyle(
+                      fontSize: 12,
+                      fontWeight: FontWeight.bold,
+                      color: ColorScheme.of(context).primary,
```
veloce: Have you tested this with several backgrounds and color schemes to see if it displays well in each case? (also on light and dark modes).
If I want to be 100% sure the font is visible I'd use `onPrimaryContainer`, but visually it may not be as nice.

HaonRekcef: On flutter docs: https://api.flutter.dev/flutter/material/ColorScheme-class.html it says:
```
 primaryContainer → Color
 A color used for elements needing less emphasis than primary
```
 So I guess they should be different, but still the difference is minimal for the ones checked just now, so lets just go with `onPrimaryContainer`

veloce (approved):

PR #2040, opened 2025-08-04
Reapply #1978 but keep using /api/tournament

tom-anders:


lib/src/model/tournament/tournament_repository.dart
```diff
@@ -61,17 +61,28 @@ class TournamentRepository {
...
+        },
       ),
       headers: {'Accept': 'application/json'},
       mapper: (Map<String, dynamic> json) => tournament.updateFromPartialServerJson(json),
     );
   }
 
+  Future<Tournament> goToPage(Tournament tournament, int page) {
+    return client.readJson(
+      Uri(path: '/tournament/${tournament.id}/standing/$page'),
```
veloce, line 80: I assume this is normal that it doesn't have the `/api` prefix here?

lib/src/model/tournament/tournament_repository.dart
```diff
@@ -61,17 +61,28 @@ class TournamentRepository {
...
+          if (tournament.standing != null) 'page': tournament.standing!.page.toString(),
+          'partial': 'true',
+        },
       ),
       headers: {'Accept': 'application/json'},
       mapper: (Map<String, dynamic> json) => tournament.updateFromPartialServerJson(json),
     );
   }
 
+  Future<Tournament> goToPage(Tournament tournament, int page) {
```
veloce: `goToPage` is a bit weird in that context. Perhaps `loadPage`?

tom-anders: yeah makes sense, it the end we *do* go to that page, but in the context of the repository `load` is more appropriate, agreed

PR #2044, opened 2025-08-04
Try again to upload the debug symbols of libapp.so to the Play Console.

julien4215:
Retry https://github.com/lichess-org/mobile/pull/1983 reverted in https://github.com/lichess-org/mobile/pull/2004.


veloce:
Need to revert again: https://github.com/lichess-org/mobile/actions/runs/17007987680/job/48220164089

Play store doesn't seem to like the file with the symbols.

julien4215:
Ah that's annoying. I think it is just because the zip file contains the folder native-debug-symbols which it shouldn't contain: https://developer.android.com/build/include-native-symbols.

PR #2045, opened 2025-08-04
make promotion dialog dissapear if pressed back

HaonRekcef:
While testing https://github.com/lichess-org/flutter-chessground/pull/79 I noticed that the promotion dialog doesn't disappear when pressing back in the GameScreen, OTB Game Screen, and Offline Correspondence Screen. This can lead to strange behavior see the video below:

[strange-behavior-promotion.webm](https://github.com/user-attachments/assets/e4820c69-6927-47b2-bdc7-b86d059ef172)

This PR fixes the issue in all the affected screens I could find. In the analysis screens this is already handled correctly by the `_setPath` function.


veloce (approved):
Good catch! thanks.

PR #2053, opened 2025-08-06
Add Result to Analysisboard

HaonRekcef:
Improves the UI of the analysis board by adding the result to the analysis board and making the names bold, similar to the broadcast screen.

| Before | After |
|--------|-------|
|  |  |


veloce (approved):
Nice!

PR #2069, opened 2025-08-10
Add: Challenge from custom position (using board editor)

### LINKED ISSUE #1363: Challenge friend from custom position (using board editor)
Right now you can only create a custom position by pasting an FEN. It would be nice if you could click on the position preview to set it up yourself (like you can do on the website).

aba2222:
Resolves #1363

### DIFF
--- a/lib/src/view/board_editor/board_editor_screen.dart
+++ b/lib/src/view/board_editor/board_editor_screen.dart
@@ -5,6 +5,7 @@ import 'package:flutter/material.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:lichess_mobile/src/constants.dart';
 import 'package:lichess_mobile/src/model/analysis/analysis_controller.dart';
+import 'package:lichess_mobile/src/model/auth/auth_session.dart';
 import 'package:lichess_mobile/src/model/board_editor/board_editor_controller.dart';
 import 'package:lichess_mobile/src/model/common/chess.dart';
 import 'package:lichess_mobile/src/model/settings/board_preferences.dart';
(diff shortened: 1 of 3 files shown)


lib/src/view/board_editor/board_editor_screen.dart
```diff
@@ -289,6 +293,46 @@ class _BottomBar extends ConsumerWidget {
...
+                  final session = ref.read(authSessionProvider);
+                  if (session == null) {
+                    showSnackBar(
+                      context,
+                      context.l10n.challengeRegisterToSendChallenges,
+                      type: SnackBarType.error,
+                    );
+                    return;
+                  }
+                  showModalBottomSheet<void>(
```
veloce: Nesting modal bottom sheets is not a good idea generally. Also SearchScreen was not meant to be displayed as a bottom sheet, which means it will probably not display well in all devices (thinking of safe area).
Instead it should be push with `Navigator.push(fullscreenDialog: true)`

veloce (commented):
It is a good idea to send a screenshot or screen recording when you implement a new feature so I can see it in action.

Here I think we should add a title to the search screen: "Challenge from position", otherwise we only see a search bar with "Search" which is a little confusing.

It only need a small refactoring of SearchScreen which can accept an optional title and if provided move the search bar to the `bottom` part.

aba2222:
# Screenshots
Screenshot_20250813_141403_org lichess mobileV2 debug
Screenshot_20250813_141344_org lichess mobileV2 debug
Screenshot_20250813_141332_org lichess mobileV2 debug

ijm8710:
This is cool; minor question, does real time have increment slider as well?

PR #2079, opened 2025-08-11
Add Player Details to the Tournament Screen

### LINKED ISSUE #1707: Tourney v2 item: allow players table to have players dived to their profiles
Intuitive to do but also noted web does this

HaonRekcef:
Add Player Details tournament screen, closes #1707 
<details>
<summary>Screenshots</summary>

grafik

grafik

</details>

Considerations

- I choose a bottom sheet modal, to prevent the user from navigating away from the tournament screen, and making it intuitive that they are still joined in the tournament.
- Uses the undocumented `tournament/$tournament/player/$userId` endpoint, as there is no official API available.
- Navigating to the userprofiles and to the games is possible, except if the user checking the details is currently joined in the tournament to prevent them from navigating away from the tournament (Maybe we can also allow this, as on the website it is possible too. Not sure though if it will work flawlessly to get a new tournament pairing if the user is navigating away) If the player pauses the tournament they can investigate other games or players.
- Includes some basic tests.

### DIFF
--- a/lib/src/view/tournament/tournament_screen.dart
+++ b/lib/src/view/tournament/tournament_screen.dart
@@ -7,10 +7,12 @@ import 'package:flutter/material.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:intl/intl.dart';
 import 'package:lichess_mobile/src/model/account/account_repository.dart';
+import 'package:lichess_mobile/src/model/analysis/analysis_controller.dart';
 import 'package:lichess_mobile/src/model/auth/auth_controller.dart';
 import 'package:lichess_mobile/src/model/auth/auth_session.dart';
 import 'package:lichess_mobile/src/model/common/id.dart';
 import 'package:lichess_mobile/src/model/game/game_history.dart';
+import 'package:lichess_mobile/src/model/game/game_status.dart';
 import 'package:lichess_mobile/src/model/tournament/tournament.dart';
 import 'package:lichess_mobile/src/model/tournament/tournament_controller.dart';
 import 'package:lichess_mobile/src/model/tournament/tournament_repository.dart';
@@ -25,8 +27,11 @@ import 'package:lichess_mobile/src/utils/focus_detector.dart';
 import 'package:lichess_mobile/src/utils/l10n_context.dart';
 import 'package:lichess_mobile/src/utils/navigation.dart';
 import 'package:lichess_mobile/src/utils/share.dart';
+import 'package:lichess_mobile/src/view/analysis/analysis_screen.dart';
 import 'package:lichess_mobile/src/view/chat/chat_screen.dart';
 import 'package:lichess_mobile/src/view/game/game_screen.dart';
+import 'package:lichess_mobile/src/view/user/user_screen.dart';
+import 'package:lichess_mobile/src/view/watch/tv_screen.dart';
 import 'package:lichess_mobile/src/widgets/board_thumbnail.dart';
 import 'package:lichess_mobile/src/widgets/bottom_bar.dart';
 import 'package:lichess_mobile/src/widgets/buttons.dart';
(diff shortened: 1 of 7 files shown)


ijm8710:
I think it's calculating win losses as 0/1 when it should be as. The correct multiplier with consecutive wins and berzerk, right?

HaonRekcef:
I decided to display the score instead of the result now same as in lichobile. Note that on the website its the result. I think the score is better.

grafik

lib/src/view/tournament/tournament_screen.dart
```diff
@@ -447,13 +452,14 @@ class _Standing extends ConsumerWidget {
...
-class _StandingPlayer extends StatelessWidget {
+class _StandingPlayer extends ConsumerWidget {
   const _StandingPlayer({required this.player});
 
   final StandingPlayer player;
 
   @override
-  Widget build(BuildContext context) {
+  Widget build(BuildContext context, WidgetRef ref) {
+    final tournamentId = context.findAncestorWidgetOfExactType<_Body>()?.id;
```
veloce: We don't do that in the app, so I looked at the doc; and it is not really recommended. It is O(N) in the depth of the tree, and not usually useful in a build method since it does not trigger a rebuild if the dependency changes.
Nothing of the above is a problem here, but for the sake of it, I'd rather not use this since it is pretty easy to pass the whole `TournamentState` to `_StandingPlayer`;

lib/src/view/tournament/tournament_screen.dart
```diff
@@ -956,3 +967,278 @@ class _BottomBarState extends ConsumerState<_BottomBar> {
...
+void _showPlayerDetails(
+  BuildContext context,
+  WidgetRef ref,
+  TournamentId tournamentId,
+  UserId userId,
+) {
+  showModalBottomSheet<void>(
+    context: context,
+    isScrollControlled: true,
+    shape: const RoundedRectangleBorder(
```
veloce: For the visual consistency within the app we should not customise the shape per invocation and leave what is defined in the theme.

lib/src/view/tournament/tournament_screen.dart
```diff
@@ -956,3 +967,278 @@ class _BottomBarState extends ConsumerState<_BottomBar> {
...
+      borderRadius: BorderRadius.vertical(top: Radius.circular(16)),
+    ),
+    builder: (context) {
+      return DraggableScrollableSheet(
+        initialChildSize: 0.6,
+        minChildSize: 0.3,
+        maxChildSize: 0.9,
+        expand: false,
+        builder: (context, scrollController) {
+          return FutureBuilder<TournamentPlayer>(
```
veloce: Do we want to cache this for a little while? like 10s or so. Hard to predict, but this is the place where users might rapidly switch over several player details in a row to compare. So a cache would not be bad I guess.
It is easy to cache with a provider and `ref.withClientCacheFor`.

lib/src/view/tournament/tournament_screen.dart
```diff
@@ -956,3 +967,278 @@ class _BottomBarState extends ConsumerState<_BottomBar> {
...
+    required this.nbGames,
+    required this.tournamentId,
+  });
+
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {
+    final tournamentState = ref.watch(tournamentControllerProvider(tournamentId));
+
+    final resultColor = pairing.score != null && pairing.score! >= 4
+        ? LichessColors.brag
```
veloce: Using `context.lichessColor.brag` you have the color harmonised with the current color scheme. Most other screen use it so it'd be nice to use is here for consistency.

lib/src/view/tournament/tournament_screen.dart
HaonRekcef: Agreed. Changed it.

lib/src/view/tournament/tournament_screen.dart
HaonRekcef: I added a 10s cache now, not really sure how many request we are going to avoid with this, but why not?!

lib/src/view/tournament/tournament_screen.dart
HaonRekcef: Changed it in everywhere in the Tournament Screen

veloce:
Merged manually

PR #2087, opened 2025-08-13
fix studies with illegal positions not loading

### LINKED ISSUE #1727: Study containing an illegal position fails to load
### What is the problem?

‌

‌

### App version

‌

### Device Info

‌

### Additional information

Relates to #1662

HaonRekcef:
Fixes #1727

The problem was that when encountering illegal positions in `_fetchChapter()`, we weren't immediately updating the state . This could lead to access to stale data, which caused some code to fail without giving an error.

After this PR, the state is updated correctly before continuing with the rest of the code.

I think the reason why issue #1662 appeared fixed at some point was that the offending chapter was chapter 1, so there was no prior state.

I tested the change with:
- https://lichess.org/study/ieCOdimW/1QYacAkq
- https://lichess.org/study/ygVnJzbX/q45se9Z6

and all chapters load correctly.

### DIFF
--- a/lib/src/model/study/study_controller.dart
+++ b/lib/src/model/study/study_controller.dart
@@ -133,7 +133,7 @@ class StudyController extends _$StudyController
     try {
       _root = Root.fromPgnGame(game);
     } on PositionSetupException {
-      return StudyState(
+      final illegalPositionState = StudyState(
         variant: variant,
         study: study,
         currentPath: UciPath.empty,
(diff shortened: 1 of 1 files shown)


tom-anders:
Very nice! IIRC we already have a test for illegal positions in studies, can we extend it so that it would have caught this bug as well?

HaonRekcef:
> Very nice! IIRC we already have a test for illegal positions in studies, can we extend it so that it would have caught this bug as well?

I added a test for studies with illegal positions now, which fails without the changes.

veloce (approved):
Thanks for the fix and the test.

PR #2090, opened 2025-08-15
Make it easier to report in the app

### LINKED ISSUE #1806: Make it easier to report in the app
Flag chat messages ⚠  - similar to the website would help alot. Good example here is arena chat - where you don't even have easy access to opens someones profile ( ie. by long press or swipe ). So you would currently need to search up the account. 

Currently, the report button on a profile opens a webview. Would be nice if that was not needed ( or only there as a fallback option )

aba2222:
Resolves #1806 

**Screenshot**
Screenshot_20250815_200657_org lichess mobileV2 debug


veloce:
Thanks for this. Could you solve the conflicts please?

veloce:
Thanks for this. But I'm closing in favour of #2096 which is more in line with what server does and also allow to copy messages.

PR #2101, opened 2025-08-16
Upload the debug symbols of libapp.so to the Play Console

julien4215:
Retry https://github.com/lichess-org/mobile/pull/2044


veloce:
Will revert again because now I see this in the console:

Capture d’écran 2025-08-24 à 07 42 19

julien4215:
It looks like it doesn't pick up anymore the ReTrace mapping file from the App Bundle since we are uploading the native debug symbols with a zip file instead of using the ones present in the App Bundle.

PR #2107, opened 2025-08-17
Add study chat room

### LINKED ISSUE #1818: Study chat room
Add study chat room.

- [ ] check that chat room is enabled in study to display it

aba2222:
Resolves #1818 

**Screenshots**
Screenshot_20250817_122536_org lichess mobileV2 debug
Screenshot_20250817_122547_org lichess mobileV2 debug

### DIFF
--- a/lib/src/model/chat/chat_controller.dart
+++ b/lib/src/model/chat/chat_controller.dart
@@ -12,6 +12,7 @@ import 'package:lichess_mobile/src/model/common/id.dart';
 import 'package:lichess_mobile/src/model/common/service/sound_service.dart';
 import 'package:lichess_mobile/src/model/common/socket.dart';
 import 'package:lichess_mobile/src/model/game/game_controller.dart';
+import 'package:lichess_mobile/src/model/study/study_controller.dart';
 import 'package:lichess_mobile/src/model/tournament/tournament_controller.dart';
 import 'package:lichess_mobile/src/model/user/user.dart';
 import 'package:lichess_mobile/src/network/http.dart';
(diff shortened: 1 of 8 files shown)


ijm8710:
The only thing I'll add is I did love the simplicity of studies with so few buttons

Especially when chat is mostly nonsensical spamming

I'm not sure a better way to do this, and this 100% was a Veloce request so that's good but I wonder if there's another place to have chat besides where it is 🤔

Would sticking it in top right overflow menu be crazy?

aba2222:
> The only thing I'll add is I did love the simplicity of studies with so few buttons
> 
> Especially when chat is mostly nonsensical spamming
> 
> I'm not sure a better way to do this, and this 100% was a Veloce request so that's good but I wonder if there's another place to have chat besides where it is 🤔
> 
> Would sticking it in top right overflow menu be crazy?

Maybe I could add a setting to toggle where it is

aba2222:
It has a setting to set it display in  in top right overflow menu now.

Screenshot_20250817_192341_org lichess mobileV2 debug

ijm8710:
Appreciate you giving options for Veloce to consider aba. Hopefully didn't take you too much time :) 

I'll admit I don't know which one works best 


Another idea is to only show chat room only on the first study screen you see, but I think it's best to wait for Veloce for putting in more time

veloce:
One sure thing: we don't want a settings here. Either we put it on the bottom bar, or in the top right menu. But we don't want it to be configurable.

Since the chat room in study is not really important, putting it in the top right menu is fine by me.

veloce (approved):

PR #2110, opened 2025-08-18
Update gen-arb.mjs

odovacrius:
Adding lb-LU to locales, as Luxembourgish is available on the old app and desktop, but as of now not on the new app, even though the mobile specific messages are translated at 100 %.

### DIFF
--- a/scripts/gen-arb.mjs
+++ b/scripts/gen-arb.mjs
@@ -84,6 +84,7 @@ const locales = [
   'ja-JP',
   'kk-KZ',
   'ko-KR',
+  'lb-LU',
   'lt-LT',
   'lv-LV',
   'mk-MK',


veloce:
I had to revert.

We actually can't have Luxembourgish right now because it is not supported by flutter out of the box. We'd need to translate the flutter strings to support it.

odovacrius:
So, if I understand correctly, I would first have to translate the cupertino, material and widget ARB files [here](https://github.com/flutter/flutter/tree/master/packages/flutter_localizations/lib/src/l10n) and then find a way to integrate them into the flutter project?

odovacrius:
In any case, I translated the three mentioned ARB files to Luxembourgish and opened a [flutter issue](https://github.com/flutter/flutter/issues/174068).

Issue #2, opened 2022-11-06
Desktop Support


ozbonus:
Would it be worthwhile to add MacOS/Windows/Linux as supported platforms? It would allow for these platforms to play even without an internet connection, and Flutter would make it simple to build anyway.

ffxsam:
It's a cool idea, but the website works pretty well as is. I can't see any particular advantages of having a dedicated desktop app (except maybe easier/better DGT e-board integration?).

semog:
I've had a certain amount of experience with compiling native Flutter app for Windows/Linux/macOS as well as web app, so this is something that I will be taking a close look at contributing to the project. Having a nice standalone native application with the stockfish engine running would be great.
The biggest current drawback to Flutter cross-platform support is the lack of cross-compilation. You have to build on the platform you are targeting. Luckily, I have access to all three platforms.

ffxsam:
What would you see as the advantages of having a native desktop app, compared to just using the website in a browser?

schlawg:
The flutter app will be heavily geared towards touch.  I don't think the mac, windows, and linux targets that flutter puts in the starter project would even survive the dependencies we'll be adding, and I'm not sure Vincent wants to worry about supporting mouse & keyboard.  

If you can't use a browser because you don't have internet on your computer, there is always catalyst and android compatibility layers if you must run lichess mobile on desktop to get at the puzzles.

ozbonus:
> The flutter app will be heavily geared towards touch. I don't think the mac, windows, and linux targets that flutter puts in the starter project would even survive the dependencies we'll be adding, and I'm not sure Vincent wants to worry about supporting mouse & keyboard.
> 
> If you can't use a browser because you don't have internet on your computer, there is always catalyst and android compatibility layers if you must run lichess mobile on desktop to get at the puzzles.

That's reasonable. I guess after the mobile app is stable someone can take a shot at a desktop fork if they really want.

M7rc7ry:
The main dartchess package used in this app can not have 64bit Int types published for web via flutter due to current javascripts ^54 limitation. The package have would to be converted to use fixnum package or strings before you could begin converting this app for web. If anyone has done this conversion please let us know.

PunchCakee:
> The main dartchess package used in this app can not have 64bit Int types published for web via flutter due to current javascripts ^54 limitation. The package have would to be converted to use fixnum package or strings before you could begin converting this app for web. If anyone has done this conversion please let us know.

[...]

PR #2114, opened 2025-08-18
Add initial Linux desktop support

christianfl:
PR related to https://github.com/lichess-org/mobile/issues/2

Hey @veloce

Concerning your comment on the issue above, I created a small PR to add initial Linux desktop support. In case you have any comments or suggested changes to make merging more likely, feel free to reach out.

## Info

The Linux platform must be created before. I thought you might want to do that by yourself via: `flutter create --platforms=linux .`

- Lichess starts and most things should work, however It's not battle tested
- Sound does not work, a separate PR for https://github.com/lichess-org/flutter-sound-effect is needed
- There are uncaught errors related to Firebase which can be fixed later (I think there was an idea to create an optional build without it?)
- For the `flutter_secure_storage_linux` plugin I forked the exact version which is used in Lichess and only added three necessary changes to let it compile. This is an open upstream issue (see https://github.com/juliansteenbakker/flutter_secure_storage/issues/920)

### DIFF
--- a/lib/src/db/database.dart
+++ b/lib/src/db/database.dart
@@ -32,13 +32,11 @@ Future<Database> database(Ref ref) async {
 
 /// Returns the database path including filename.
 Future<String> get _databasePath async {
-  // Use specific db path on Linux
   if (Platform.isLinux) {
     final directory = await getApplicationSupportDirectory();
     return join(directory.path, kLichessDatabaseName);
   }
 
-  // Use default db path
   return join(await getDatabasesPath(), kLichessDatabaseName);
 }
 


lib/src/widgets/platform.dart
```diff
@@ -18,6 +18,8 @@ class PlatformWidget extends StatelessWidget {
         return androidBuilder(context);
       case TargetPlatform.iOS:
         return iosBuilder(context);
+      case TargetPlatform.linux:
```
veloce: Here we can even have a 
```dart
if (Theme.of(context).platform == TargetPlatform.iOS) return iosBuilder(context);
else return androidBuilder(context);
```

lib/src/db/database.dart
```diff
@@ -23,10 +23,25 @@ final _logger = Logger('Database');
...
+  if (Platform.isLinux) {
+    databaseFactory = databaseFactoryFfi;
+  }
+  final dbPath = await _databasePath;
   return openAppDatabase(databaseFactory, dbPath);
 }
 
+/// Returns the database path including filename.
+Future<String> get _databasePath async {
+  // Use specific db path on Linux
```
veloce: unecessary comments in that function

veloce (commented):
Changes are indeed contained, so we can have that linux platform for dev purposes.

lib/src/widgets/platform.dart
christianfl: Fixed!

lib/src/db/database.dart
christianfl: Also fixed! Hope you meant the comments within the function as I thought it makes sense to mention that the functions returns path + filename.

christianfl:
Thanks for your review :slightly_smiling_face:

veloce:
I don't have a linux computer right now to test the platform so I cannot add it. Will do later or feel free to open a pr with the platform files.

christianfl:
Ok! Thanks for merging @veloce

PR #2117, opened 2025-08-19
Fix another study loading issue

HaonRekcef:
Fixes a bug reported on discord. Problem appeared due to the variant of an illegal position being set to `fromPosition` and therefore having no initial position. 
Bugfix tested on this study: https://lichess.org/study/UUSBv8yN

### DIFF
--- a/lib/src/model/study/study_controller.dart
+++ b/lib/src/model/study/study_controller.dart
@@ -144,7 +144,7 @@ class StudyController extends _$StudyController
         // since the position is illegal and `isComputerAnalysisAllowed` is false anyway.
         evaluationContext: EvaluationContext(
           variant: variant,
-          initialPosition: study.chapter.setup.variant.initialPosition,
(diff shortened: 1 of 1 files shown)


lib/src/model/study/study_controller.dart
```diff
@@ -144,7 +144,7 @@ class StudyController extends _$StudyController
         // since the position is illegal and `isComputerAnalysisAllowed` is false anyway.
         evaluationContext: EvaluationContext(
           variant: variant,
-          initialPosition: study.chapter.setup.variant.initialPosition,
+          initialPosition: Variant.standard.initialPosition,
```
veloce, line 147: Theoretically both standard and 960 are supported in evaluation contexts (and more variants later).

So we need to take that into account I think.

veloce: Oh never mind, it is for the illegal position.

veloce (approved):

PR #2122, opened 2025-08-20
only request 10 games for recentGames

HaonRekcef:
Currently we were requesting 20 recent games from the server, but only displaying 10.


veloce:
Even on tablet?

HaonRekcef:
> Even on tablet?

Yes, all `RecentGamesWidget` do not have a maxGamesToShow set so they all used 10. 

I changed this constant now to be dependent on the already defined constant `kNumberOfRecentGames `in `game_history`, but its still 10.

PR #2142, opened 2025-08-24
refactor: use @freezed unions to improve readability of GameScreen code

tom-anders:
Tried to understand the code because I wanted to implement the "Play with a friend feature" (aka generating challenge links), but I found the code hard to follow in some places.

I think the union types help to improve readability and better express some of the invariants.

### DIFF
--- a/lib/src/view/game/game_screen_providers.dart
+++ b/lib/src/view/game/game_screen_providers.dart
@@ -1,5 +1,6 @@
-import 'package:dartchess/dartchess.dart' show Move, Side;
+import 'package:dartchess/dartchess.dart' show Side;
 import 'package:flutter_riverpod/flutter_riverpod.dart';
+import 'package:freezed_annotation/freezed_annotation.dart';
 import 'package:lichess_mobile/src/model/challenge/challenge.dart';
 import 'package:lichess_mobile/src/model/common/id.dart';
 import 'package:lichess_mobile/src/model/common/speed.dart';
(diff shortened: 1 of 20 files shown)


veloce:
Yeah this definitely needed a refactoring; thanks for tackling it!

While freezed union types are nice, they are not really standard dart code. There is a bit of "magic" in how they are defined.

I wonder if we could use standard dart code:
```dart
sealed class GameScreenSource {}

@freezed
sealed class LoadedGameSource with _$LoadedGameSource implements GameScreenSource {
  ...
}

@freezed
sealed class LobbySource...
```

instead?

lib/src/view/game/game_screen_providers.dart
```diff
@@ -10,13 +11,41 @@ import 'package:lichess_mobile/src/model/lobby/game_seek.dart';
 import 'package:riverpod_annotation/riverpod_annotation.dart';
 
 part 'game_screen_providers.g.dart';
+part 'game_screen_providers.freezed.dart';
+
+@freezed
+sealed class LoadedGame with _$LoadedGame {
```
veloce: The name `LoadedGame` is misleading (I know it is me that wrote it in the first place ;)) because at that point the game is not loaded, it's only we have the ID.

`CreatedGame` would be more accurate.

lib/src/view/game/game_screen_providers.dart
```diff
@@ -10,13 +11,41 @@ import 'package:lichess_mobile/src/model/lobby/game_seek.dart';
...
+  const factory LoadedGame({
+    required GameFullId gameId,
+    String? lastFen,
+    Move? lastMove,
+    Side? side,
+  }) = _LoadedGame;
+}
+
+@freezed
+sealed class CurrentGameState with _$CurrentGameState {
```
veloce: Perhaps a comment is welcome here. It is not obvious why the only 2 states are "loadedGame" or "declinedChallenge"....

tom-anders:
> Yeah this definitely needed a refactoring; thanks for tackling it!
> 
> While freezed union types are nice, they are not really standard dart code. There is a bit of "magic" in how they are defined.
> 
> I wonder if we could use standard dart code:
> 
> ```dart
> sealed class GameScreenSource {}
> 
> @freezed
> sealed class LoadedGameSource with _$LoadedGameSource implements GameScreenSource {
>   ...
> }
> 
> @freezed
> sealed class LobbySource...
> ```
> 
> instead?

Hmm, if the concern is that it's not really obvious where the subclasses are defined, maybe https://pub.dev/packages/freezed#ejecting-an-individual-union-case is a compromise?

veloce:
> Hmm, if the concern is that it's not really obvious where the subclasses are defined, maybe https://pub.dev/packages/freezed#ejecting-an-individual-union-case is a compromise?

That yes, and also that I'd rather not use advanced freezed features that rely on code generation where the same thing can be achieved through normal dart code. 
This feature was introduced while dart did not support sealed classes. When you read the code and you don't know the freezed feature, it is not immediately obvious what it does, especially with shared properties. Yet another thing to learn.

veloce:
Not sure it is worth using the ejecting thing... if it's too much of a hassle to extend a sealed class with freezed let's keep it the way you did then

tom-anders:
> Not sure it is worth using the ejecting thing... if it's too much of a hassle to extend a sealed class with freezed let's keep it the way you did then

Turns out it's not that much of a hassle, and it even gets less verbose on the call site :partying_face:

lib/src/view/game/game_screen_providers.dart
tom-anders: I was able to entirely factor out the fen/lastMove/orientation fields from this provider (it was never really used and can be kept entirely in the `GameScreen` class instead), so the `LoadedGame` is removed entirely. But I renamed the LoadedGameSource and LoadedGameState to ExistingGameSource and GameCreatedState, respectivelt

PR #2145, opened 2025-08-25
feat: add change password and 2FA buttons to account settings

### LINKED ISSUE #1873: Add 2FA button to settings

### LINKED ISSUE #1874: Add change password button to settings

OldUser101:
Add change password and two-factor authentication buttons to the account settings page, linking to `/account/passwd` and `/account/twofactor` respectively.

Closes #1873
Closes #1874

### DIFF
--- a/lib/l10n/app_af.arb
+++ b/lib/l10n/app_af.arb
@@ -1579,5 +1579,6 @@
   "timeagoNbMonthsAgo": "{count, plural, =1{{count} maand gelede} other{{count} maande gelede}}",
   "timeagoNbYearsAgo": "{count, plural, =1{{count} jaar gelede} other{{count} jare gelede}}",
   "timeagoNbMinutesRemaining": "{count, plural, =1{nog {count} minuut oor} other{nog {count} minute oor}}",
-  "timeagoNbHoursRemaining": "{count, plural, =1{nog {count} uur oor} other{nog {count} ure oor}}"
+  "timeagoNbHoursRemaining": "{count, plural, =1{nog {count} uur oor} other{nog {count} ure oor}}",
+  "tfaTwoFactorAuth": "Tweeledige verifikasie"
 }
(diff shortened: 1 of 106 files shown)


veloce:
Have you tested the whole process to ensure it works? I'm worried about the redirect not working if you are not connected (which is the default) on website.

Screenshots showing the working process welcome.

OldUser101:
@veloce 

I have tested it, yes. When offline, the account settings menu is inaccessible. If you go offline while in the menu, you can still select the option, but it is up to the web browser what happens then. This is the same behaviour as the "Close Account" button on the same page.

Screenshots for reference:

When offline:
Screenshot_2025-08-25_18:01:49

Accessing the menu:
Screenshot_2025-08-25_18:02:28

After selecting "change password" when online:
Screenshot_2025-08-25_18:05:04

After selecting "change password" when offline:
Screenshot_2025-08-25_18:16:19

OldUser101:
I should also add that I've been trying to use the contextual translations, as the new strings are currently hardcoded. The strings `security` and `changePassword` are already included, however, `twoFactorAuth` from the `tfa` source isn't.

I've added `tfa` to the `modules` list in `scripts/gen-arb.mjs`, and run the following, as in the docs:

```sh
./scripts/gen-arb.mjs
flutter gen-l10n
```

This generates all files without issue, except `lib/l10n/l10n.dart`, which needs to be changed for the `twoFactorAuth` string to actually be usable.

Is this something that is auto-generated elsewhere, or something I'm not supposed to interfere with?

veloce:
Ok thanks for the testing. What happens if you tap "Change password" (or "Two factor authentication") and you are redirected to the browser while not yet signed in (on browser side)?
It should prompt you to connect, and once connected it should automatically redirect to the requested screen, hence "change password". Do you confirm it works like that?

About the translations: `lib/l10n/l10n.dart` doesn't need to be re-generated. You can check it only contains imports for all the locales.
If you only need the `twoFactorAuth` string you could use a whitelist in `gen-arb.mjs` to avoid adding all the unnecessary strings.

OldUser101:
@veloce 

It would appear that if you aren't signed in on the browser after redirection, you get taken to the registration page. I'm guessing this is something that needs to be changed in the server, rather than the app?

Screenshot_2025-08-25_21:28:01

veloce:
> It would appear that if you aren't signed in on the browser after redirection, you get taken to the registration page. I'm guessing this is something that needs to be changed in the server, rather than the app?

Indeed

OldUser101:
@veloce 

I believe I found the source of this problem. I've opened an issue: lichess-org/lila#18070, I'll get to work submitting a PR for this.

Edit: PR at lichess-org/lila#18072

veloce:
Thanks!

PR #2155, opened 2025-08-27
feat: show a list of study members

### LINKED ISSUE #1817: Study members list
Add study members list.

For now we show only the chapters list. Study menu probably needs to be rethought to include members.

OldUser101:
Closes #1817


OldUser101:
@veloce 

Can you check over this?

I've added the list through the context menu on the study page, which then opens a sheet. Screenshots illustrate.

image
image

I experimented with putting the list on the same sheet as the chapters, but it made it feel more cluttered.

The names in the list are interactive, and navigate to the user's profile page when pressed.

veloce (approved):
Looks good!

PR #2164, opened 2025-09-02
fix: reset scroll when navigating study chapters

OldUser101:
The scroll position in the study comment persists even after the text is changed. 

This results in behaviour that results in the user having to manually scroll to the top of the comment if they want to read the text.

This PR resets the scroll position when the widget changes, to save the user having to manually scroll.

### DIFF
--- a/lib/src/view/study/study_gamebook.dart
+++ b/lib/src/view/study/study_gamebook.dart
@@ -26,14 +26,26 @@ class StudyGamebook extends StatelessWidget {
   }
 }
 
-class _Comment extends ConsumerWidget {
+class _Comment extends ConsumerStatefulWidget {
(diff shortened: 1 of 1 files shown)


veloce:
Hi! Not sure to understand the problem here; can you post a before/after screen recording please?

OldUser101:
Before the change, when moving to the second page, I have to manually scroll upwards, since the scroll position doesn't reset.

https://github.com/user-attachments/assets/5ed1e4fe-fc62-47af-8836-765d661911e1

After the change, this is automatic when any content is changed.


https://github.com/user-attachments/assets/52cc76d6-1fa3-4fa4-9e11-e3e4c8a870cb

PR #2169, opened 2025-09-05
feat: add haptic feedback for page refreshes

### LINKED ISSUE #2025: Can a haptic be added for the pull to refresh?
Makes it feel more full.

OldUser101:
Add haptic feedback when refreshing pages. 

This is done by creating a wrapper, `HapticRefreshIndicator`, around the normal `RefreshIndicator`. This wrapper then triggers haptic feedback when the refresh takes place.

This wrapper can be used as an (almost) drop-in replacement for `RefreshIndicator`.

Closes #2025

### DIFF
--- a/lib/src/widgets/haptic_refresh_indicator.dart
+++ b/lib/src/widgets/haptic_refresh_indicator.dart
@@ -2,6 +2,7 @@ import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter/services.dart';
 
+/// A wrapper widget over RefreshIndicator to provide haptic feedback on iOS
 class HapticRefreshIndicator extends StatelessWidget {
   final Widget child;
   final double edgeOffset;
--- a/lib/src/widgets/haptic_refresh_indicator.dart
+++ b/lib/src/widgets/haptic_refresh_indicator.dart
@@ -1,0 +1,31 @@
+import 'package:flutter/material.dart';
+import 'package:flutter/services.dart';
+
+class HapticRefreshIndicator extends StatelessWidget {
+  final Key? refreshIndicatorKey;
+  final Widget child;
+  final double edgeOffset;
+  final RefreshCallback onRefresh;
+
+  const HapticRefreshIndicator({
+    this.refreshIndicatorKey,
+    required this.child,
+    this.edgeOffset = 0.0,
+    required this.onRefresh,
+  });
+
+  Future<void> _onRefreshWithHaptics() async {
+    HapticFeedback.lightImpact();
+    await onRefresh();
+  }
+
+  @override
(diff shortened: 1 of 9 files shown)


lib/src/widgets/haptic_refresh_indicator.dart
```diff
@@ -0,0 +1,31 @@
+import 'package:flutter/material.dart';
+import 'package:flutter/services.dart';
+
+class HapticRefreshIndicator extends StatelessWidget {
+  final Key? refreshIndicatorKey;
```
veloce: This can probably just be named `key`.

veloce (commented):
Looks good! There is one change needed though: the haptic feedback should only happen on iOS, as this is not an android thing. So just adding that extra platform check should be ok.
And in the widget doc comment, just precise the haptic is for iOS only.

ijm8710:
@OldUser101 if it's an easy tweak, any chance to make the haptic and spinner more similar to how it works in most apps

- The haptic is traditionally felt when you fully reach the pull down point (the point where it's most pulled down). Here I think we feel it after the release when it's fully rebounded back 
- the spinner itself has two different animation parts normally. When it's pulled you see the spinner begin to fill in more and more of a full circle as it's pulled down. Then once the refresh is activated it has the regular spinning animation. Currently we only have the second part of this (fully know your original PR only worked on the first bullet, but not sure if it's easy to fix both at same time; especially since just looking for the normal iOS refresh that we typically see in most apps)

Is that possible :) ?

OldUser101:
@ijm8710 

> The haptic is traditionally felt when you fully reach the pull down point (the point where it's most pulled down). Here I think we feel it after the release when it's fully rebounded back

Yes, it looks like this should be possible with the `onStatusChange` callback. I'll see if I can get it to work.

> the spinner itself has two different animation parts normally. When it's pulled you see the spinner begin to fill in more and more of a full circle as it's pulled down. Then once the refresh is activated it has the regular spinning animation. Currently we only have the second part of this (fully know your original PR only worked on the first bullet, but not sure if it's easy to fix both at same time; especially since just looking for the normal iOS refresh that we typically see in most apps)

Unfortunately, the animation is provided by the Flutter Material package, and I don't think there's anyway to get it to behave this way (if there is, I can't find it in the docs).

ijm8710:
Thanks! Awesome on first.

OldUser101:
@ijm8710 

> Thanks! Awesome on first.

I've attempted to get this working but, unfortunately, it looks like the `onStatusChange` callback which would be required for this, isn't exposed for the adaptive `RefreshIndicator` we are using. The alternative would be to reimplement our own refresh indicator mechanism for iOS, which, given the work involved, probably isn't worth it for the small change.

PR #2200, opened 2025-09-17
bring back streak rating on tablet

HaonRekcef:
Adds rating back to the streak screen in landscape view:

grafik


veloce:
Oops I guess my bad from a previous fix. Thanks!

PR #2201, opened 2025-09-17
Fix game loading in Landscape mode

HaonRekcef:
Prevents the board from resizing in landscape mode after getting paired. This PR puts the cancel button to the right where the move list will be:
grafik

### DIFF
--- a/lib/src/view/game/game_loading_board.dart
+++ b/lib/src/view/game/game_loading_board.dart
@@ -84,43 +84,43 @@ class _LobbyScreenLoadingContentState extends State<LobbyScreenLoadingContent> {
                   ),
                 ),
               ),
+              userActionsBar: BottomBar(
+                children: [
+                  FutureBuilder(
+                    future: _cancelGameCreationFuture,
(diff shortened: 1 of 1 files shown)


lib/src/view/game/game_loading_board.dart
```diff
@@ -84,44 +84,44 @@ class _LobbyScreenLoadingContentState extends State<LobbyScreenLoadingContent> {
                   ),
                 ),
               ),
+              userActionsBar: BottomBar(
```
veloce: Good catch!

I think the column can be removed, and only keep the SafeArea + GameLayout.

Probably unnecessary but could you also post a screenshot of portrait mode so I am sure it does not break on that side? Thanks.

HaonRekcef: Yes you are right, they were unnecessary.
I also fixed the challenge screen now.
Portrait mode is still normal:
grafik

veloce (approved):
thanks!

PR #2202, opened 2025-09-17
handle crazyhouse deeplinks

HaonRekcef:
Currently crazyhouse games can be opened via deeplinks (f.ex. https://lichess.org/KVB5HJjl/black) However, this leads to errors when replaying the game. As far as I can tell we cannot determine from the game ID alone whether the variant is supported, so the app itself must handle the error.
My suggestion is this: 
grafik

The “Open game in your browser” option would strip the color after the slash so it wont be deeplinked again. The other option would be to open it in `LaunchMode.inAppWebView `.

### DIFF
--- a/test/view/analysis/analysis_screen_test.dart
+++ b/test/view/analysis/analysis_screen_test.dart
@@ -961,6 +961,7 @@ void main() {
       );
 
       await tester.pumpWidget(app);
+      await tester.pumpAndSettle();
 
       await switchToPremoveTab(tester);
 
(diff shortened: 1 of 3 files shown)


test/view/analysis/analysis_screen_test.dart
```diff
@@ -961,6 +961,7 @@ void main() {
       );
 
       await tester.pumpWidget(app);
+      await tester.pumpAndSettle();
```
veloce: Why is this suddenly needed? If a `test.pump()` can do the job it is better; in any case it is good to explain why it is needed in a comment so when we go back to the test months later we understand things better.

HaonRekcef: yes indeed tester.pump() is better. We need it now because we have to wait until the async analysisControllerProvider(options) is completed.

veloce (approved):

PR #2207, opened 2025-09-17
Fix: add manual check highlight in spectator mode

Sanket-Kurle:
The red check indicator was not visible when watching other players' games.
This is because gameData is null in spectator mode, and the Chessboard widget was not receiving any check information. As a result, spectators had no feedback that the king was under attack.
So I Updated Board.dart to manually detect check state when gameData == null.
It is using circular shape, not the exact internal check highlighting as we are not passing gameData.

image

### DIFF
--- a/lib/src/widgets/board.dart
+++ b/lib/src/widgets/board.dart
@@ -30,13 +30,14 @@ class BoardWidget extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
-    final Chess position = Chess.fromSetup(Setup.parseFen(fen));
+    final safeFen = fen.trim().isNotEmpty ? fen : Setup.standard.fen;
+    final Chess position = Chess.fromSetup(Setup.parseFen(safeFen));
     // ✅ Safe fallback for spectator mode
     final effectiveGameData =
         gameData ??
         GameData(
           playerSide: PlayerSide.none, // spectator cannot move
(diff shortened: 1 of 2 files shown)


HaonRekcef:
Hi @Sanket-Kurle 
Thanks for looking into this issue. I think, however this approach might not be the right one.

> This is because gameData is null in spectator mode, and the Chessboard widget was not receiving any check information. As a result, spectators had no feedback that the king was under attack.

As you pointed out `gameData` of the board Widget in the TV Screen is null, so instead of adding a workaround in `board.dart`  I think the better approach is to pass `gameData` to the board Widget. To do this I think you need to pass `interactiveBoardParams` to the `gameLayout `in `tv-screen.dart` .

Sanket-Kurle:
@HaonRekcef i tried way passing gameData to board and doing this i got correct highlight effect to but got some error in test files
Screenshot 2025-09-18 at 2 01 02 PM
Screenshot 2025-09-18 at 3 59 55 PM

you can see where the PR got failed

HaonRekcef:
@Sanket-Kurle As I said, you should not change `board.dart` at all. The change should be in `tv_screen.dart` where you will pass `interactiveBoardParams `to the `gameLayout `.

Sanket-Kurle:
ooh got it 
thanks

Sanket-Kurle:
@HaonRekcef check #2208 
i made changes as you suggested

PR #2213, opened 2025-09-19
Discard piece after tapping square again in Editor

HaonRekcef:
Small UX improvement for the editor: 
Empties the square after tapping it again with the same piece in hand. This is possible on the website too.
Tests failing is expected, needs chessground changes.


veloce:
@HaonRekcef now the change in flutter-chessground is published. You should be able to use it in this PR.

PR #2224, opened 2025-09-25
Fixing Clock Misorientation in tablets

Sanket-Kurle:
https://github.com/user-attachments/assets/0403d2c6-05a9-4e97-a31a-014104f817ed

here you can see both behaviour
iPad A16 is what we have now
iPad 8th generation is what we will get after updating


Sanket-Kurle:
@veloce thanks 
looking forward to contribute more

PR #2227, opened 2025-09-26
Add patron colored wings

veloce:


lib/src/model/user/leaderboard.dart
```diff
@@ -36,13 +36,20 @@ sealed class LeaderboardUser with _$LeaderboardUser {
...
   }) = _LeaderboardUser;
 
-  LightUser get lightUser =>
-      LightUser(id: id, name: username, title: title, flair: flair, isPatron: patron);
+  LightUser get lightUser => LightUser(
+    id: id,
+    name: username,
+    title: title,
+    flair: flair,
+    isPatron: patron,
```
ornicar, line 52: this field is only here for BC. You don't need it, and should only use `patronTierColor`.

`isPatron == patronColor.isDefined`

veloce: ok will change that 👍

veloce: @ornicar another missing one is chat api: https://github.com/veloce/lila/blob/b65600921a925205a20e4a713c6a5f20ca5f848c/modules/chat/src/main/JsonView.scala#L80

veloce: Called it `pc` on mobile side: https://github.com/lichess-org/mobile/pull/2238/files#diff-9bfae106b2e39b5384d3cdc8f29c5f2719dc214849fa1c749c52073054815a82R51

veloce: Also `Lila-ws` has not been deployed it seems, cause `patronColors` is missing in online friends WS API.

ornicar: Indeed and it might be another week before it's deployed.

ornicar: > Called it `pc` on mobile side: https://github.com/lichess-org/mobile/pull/2238/files#diff-9bfae106b2e39b5384d3cdc8f29c5f2719dc214849fa1c749c52073054815a82R51

:+1: https://github.com/lichess-org/lila/blob/master/modules/chat/src/main/ChatJsonView.scala#L71-L71

PR #2229, opened 2025-09-27
feat: learn from your mistakes

tom-anders:
[retro_final.webm](https://github.com/user-attachments/assets/a29c94f1-eba4-432e-b2e5-a48e1988d189)

### DIFF
--- a/lib/src/model/analysis/retro_controller.dart
+++ b/lib/src/model/analysis/retro_controller.dart
@@ -406,11 +406,13 @@ class RetroController extends _$RetroController with EngineEvaluationMixin {
 
         if (diff > _kCorrectMovePovDiffThreshold) {
           _onCorrectMove();
+
+          // We used unlimited search time during Feedback.evalMove,
+          // now restart evaluation with normal search time
+          requestEval();
         } else {
           _onIncorrectMove();
         }
-
-        stopEval();
       }
     }
   }
--- a/lib/src/model/engine/evaluation_mixin.dart
+++ b/lib/src/model/engine/evaluation_mixin.dart
@@ -129,13 +129,6 @@ mixin EngineEvaluationMixin {
   }
 
   @mustCallSuper
-  Future<void> stopEval() async {
-    if (evaluationState.isEngineAvailable(evaluationPrefs)) {
-      _evaluationService?.stop();
-    }
-  }
-
-  @mustCallSuper
   void setNumEvalLines(int numEvalLines) {
     // clear all saved evals since the number of eval lines has changed
     positionTree.updateAll((node) => node.eval = null);


veloce:
the screen recording just shows a static screen @tom-anders , perhaps wrong one?

tom-anders:
Oh that may be related to my recent emulator problems, I'll make a new one today

lib/src/model/analysis/retro_controller.dart
```diff
@@ -0,0 +1,514 @@
...
+      node.position == serverBranch.position ||
+      node.position.isCheckmate ||
+      // Any move that was found in the master's database is also considered a solution.
+      openingExplorerSolutions.contains(node.sanMove!.move.uci);
+}
+
+/// Depth needed to evaluate alternative solution moves.
+const double _kEvalDepthThreshold = kDebugMode ? 12 : 18;
+
+/// When checking a move that is not the server solution,
```
veloce, line 70: Nitpicking but dart doc comments should have this format:

```dart
/// Short title on on line
///
/// Extended description that can take several lines.
const double _myVar = 'bluk';
```

If you do this, it is best rendered in IDE tools which expect this format.

lib/src/model/analysis/retro_controller.dart
```diff
@@ -0,0 +1,514 @@
...
+/// consider it a correct move if the eval difference is above this threshold,
+/// i.e. the move does not make the position significantly worse.
+const double _kCorrectMovePovDiffThreshold = -0.04;
+
+/// Logic and magic numbers are based on the `RetroCtrl` class in the lila frontend.
```
veloce: Better put this comment where the magic numbers are defined, and with a link to Lila code where they were taken.

lib/src/model/analysis/retro_controller.dart
```diff
@@ -0,0 +1,514 @@
...
+
+        final hasSolution = branch.children.length > 1;
+
+        final isMistake = (bigEvalSwing || lostEasyMate) && hasSolution;
+        if (!isMistake) return null;
+
+        var openingExplorerSolutions = const IList<UCIMove>.empty();
+        final middlegame = _game.meta.division?.middlegame;
+        if (middlegame == null || branch.position.ply + 1 < middlegame) {
+          final entry = await ref
```
veloce: Would be nice to handle network errors here. 
In the future we may want to have full off-line support for this feature. In general if we can't get the opening explorer solutions we should still be able to review mistakes (with a slightly degraded experience).

lib/src/model/analysis/retro_controller.dart
```diff
@@ -0,0 +1,514 @@
...
+      ),
+      root: _root.view,
+      currentPath: mistakes.isNotEmpty
+          ? _root.mainlinePath.truncate(mistakes[0].branch.position.ply)
+          : UciPath.empty,
+    );
+  }
+
+  void onUserMove(NormalMove move) {
+    if (!state.requireValue.currentPosition!.isLegal(move)) return;
```
veloce: Why `currentPosition` is nullable here? Seems it shouldn't.

lib/src/model/analysis/retro_controller.dart
```diff
@@ -0,0 +1,514 @@
...
+      isNavigating: true,
+    );
+  }
+
+  void userPrevious() {
+    _setPath(state.requireValue.currentPath.penultimate, isNavigating: true);
+  }
+
+  void viewSolution() {
+    onUserMove(state.requireValue.currentMistake!.serverMove);
```
veloce: this null check operator could be avoided I guess.

lib/src/model/analysis/retro_controller.dart
```diff
@@ -0,0 +1,514 @@
...
+        );
+
+        if (diff > _kCorrectMovePovDiffThreshold) {
+          _onCorrectMove();
+        } else {
+          _onIncorrectMove();
+        }
+
+        // We used unlimited search time during Feedback.evalMove,
+        // now restart evaluation with normal search time (effectively stopping it for this move)
```
veloce: unless the effective search time is less than configured search time I guess?

we could add a new `stopEval()` method to handle this case I think.

lib/src/model/analysis/retro_controller.dart
```diff
@@ -0,0 +1,514 @@
...
+
+  Duration? get evalTime =>
+      evalRequestedAt != null ? DateTime.now().difference(evalRequestedAt!) : null;
+
+  double get evalProgress => feedback == RetroFeedback.evalMove && currentNode.eval != null
+      ? min(1.0, currentNode.eval!.depth / _kEvalDepthThreshold)
+      : 0.0;
+
+  @override
+  Position? get currentPosition => currentNode.position;
```
veloce: This can be non nullable.

lib/src/view/analysis/server_analysis.dart
```diff
@@ -63,7 +65,21 @@ class ServerAnalysisSummary extends ConsumerWidget {
                 const Padding(
                   padding: EdgeInsets.only(top: 16.0),
                   child: WaitingForServerAnalysis(),
+                )
+              else
+                Padding(
```
veloce: I'd add vertical padding here too. And perhaps make the button not full width.

lib/src/view/puzzle/puzzle_feedback_widget.dart
```diff
@@ -119,8 +86,52 @@ class PuzzleFeedbackWidget extends ConsumerWidget {
   }
 }
 
-class _FeedbackTile extends StatelessWidget {
-  const _FeedbackTile({this.leading, required this.title, this.subtitle});
+class SideToPlayPiece extends ConsumerWidget {
```
veloce, line 90: A comment title would be nice cause the name of the class is not really self explanatory.

test/view/analysis/retro_screen_test.dart
```diff
@@ -0,0 +1,271 @@
...
+import 'package:mocktail/mocktail.dart';
+
+import '../../example_data.dart';
+import '../../network/fake_websocket_channel.dart';
+import '../../test_helpers.dart';
+import '../../test_provider_scope.dart';
+
+const testId = GameId('abcdefgh');
+
+class MockGameRepository extends Mock implements GameRepository {}
```
veloce: It is really better not to mock the repository imo. Almost all the other tests mock the network response using a `MockClient` instead which has a lot of advantages. And I believe this doesn't take more time or complexity to do it.

lib/src/view/analysis/retro_screen.dart
```diff
@@ -0,0 +1,391 @@
...
+                savedEval: asyncState.valueOrNull?.currentNode.eval,
+                goDeeper: () =>
+                    ref.read(retroControllerProvider(options).notifier).requestEval(goDeeper: true),
+              ),
+            _RetroMenu(options: options),
+          ]
+        : <Widget>[];
+
+    return Scaffold(
+      appBar: AppBar(title: Text(context.l10n.learnFromYourMistakes), actions: appBarActions),
```
veloce: You can use `AppBarTitleText` to auto size the title so it doesn't appear truncated.

veloce: Maybe we can even append the error index after the title, like `Learn from your mistakes 2/4`.

veloce:
To improve the discoverability of the feature, I'd suggest to add a "Learn from your mistakes" button on game screen if the server analysis is available. What do you think? The button could be added in the game screen end pop-up, and game menu.

lib/src/model/analysis/retro_controller.dart
tom-anders: added a try-catch block :+1:

lib/src/model/analysis/retro_controller.dart
tom-anders: It's nullable because it's part of `CommonAnalysisState` which allows `null` for illegal positions in studies

lib/src/model/analysis/retro_controller.dart
tom-anders: Ah I figured if currentMistake is `null`, the `viewSolution` button shouldn't be there in the first place - But you're right, if it still is possible to enter this code path without any mistakes, we should just do nothing instead of crashing

lib/src/model/analysis/retro_controller.dart
tom-anders: So basically something like this..?

```
if (state.requireValue.evalTime!.inSeconds > ref.read(engineEvaluationPreferencesProvider).engineSearchTime) {
   stopEval()
else
   requestEval() // Restart with configured search time
```

Maybe we actually don'y need to do anything though:
- If the move was incorrect, it's taken back and `requestEval()` will be called anyway due to the change in position, resetting to the default search time
- If the move was correct, it's not that big of a deal to keep the engine running, is it? Most likely the user will press "next mistake" anyway, which will reset the search time. And if they want to keep analysing the current position, making any move will reset the search time to the configured one as well.

lib/src/model/analysis/retro_controller.dart
tom-anders: see above, it's nullable in the mixin because of illegal study positions

lib/src/view/analysis/server_analysis.dart
tom-anders: Sure, I wasn't really sure which button to use here anyway, feel free to tweak after merging if needed

lib/src/view/puzzle/puzzle_feedback_widget.dart
tom-anders: Agreed, I think I wanted to find a better name, but failed :D

tom-anders:
Thanks for the review!

> To improve the discoverability of the feature, I'd suggest to add a "Learn from your mistakes" button on game screen if the server analysis is available. What do you think? The button could be added in the game screen end pop-up, and game menu.

Good idea! Maybe directly after the game, we could also add a "learn from mistakes" button to the menu that would request server analysis, display some kind of progress dialog, and then take you directly to the retro screen...?
IIRC we used to have a "request analysis" button there, but it was removed at some point...?

tom-anders:
Think I addressed all the comments now. Also added a new video that hopefully works this time

lib/src/model/analysis/retro_controller.dart
veloce: Yes but you can override it and make non nullable. As it is done in analysis controller.

lib/src/model/analysis/retro_controller.dart
veloce: I think I misunderstood then. I thought based on your comment we'd want to stop analysis whatever the move was.
In that case a simple `stopEval()` will be enough, I don't see why we'd need to check the eval time.

veloce:
> Good idea! Maybe directly after the game, we could also add a "learn from mistakes" button to the menu that would request server analysis, display some kind of progress dialog, and then take you directly to the retro screen...?

yes we could do that; this adds complexity and will likely not work from time to time, but not a big deal.

if we do that though, I'd handle the logic in the retro controller:
- if the game was not analysed yet, request server analysis in retro `build` method
- while server analysis is pending, simply display a circular progress with the text "Calculating moves..."
- probably need a timeout like 1 minute, if for some reason the server analysis never finishes

lib/src/model/analysis/retro_controller.dart
```diff
@@ -0,0 +1,537 @@
...
+        }
+
+        final eval = branch.serverEval;
+        final newEval = branch.children.first.serverEval;
+
+        if (eval == null || newEval == null) {
+          return null;
+        }
+
+        // https://github.com/lichess-org/lila/blob/d29f27d8cbb0e0dac38308c23c63b828028c085f/ui/analyse/src/nodeFinder.ts#L26
```
veloce: This is nitpicking again, but it is better to make a constant of the `0.1` here. So you can name it for better clarity and put the ref to lila code here.

lib/src/model/analysis/retro_controller.dart
```diff
@@ -0,0 +1,537 @@
...
+
+        var openingExplorerSolutions = const IList<UCIMove>.empty();
+        final middlegame = _game.meta.division?.middlegame;
+        if (middlegame == null || branch.position.ply + 1 < middlegame) {
+          try {
+            final entry = await ref
+                .read(openingExplorerRepositoryProvider)
+                .getMasterDatabase(branch.position.fen, since: MasterDb.kEarliestYear);
+
+            // https://github.com/lichess-org/lila/blob/d28f27d8cbb0e0dac38308c23c63b828028c085f/ui/analyse/src/retrospect/retroCtrl.ts#L108
```
veloce: I think we can remove this comment; this is sufficiently self explanatory. The link will most probably be out of date at some point so better not having it.

lib/src/model/analysis/retro_controller.dart
veloce: Also better just put the file where the logic is found, without the line number; lines number will most likely change.

lib/src/model/analysis/retro_controller.dart
```diff
@@ -0,0 +1,537 @@
...
+
+/// Threshold for considering a move a correct alternative solution.
+///
+/// When checking a move that is not the server solution,
+/// consider it a correct move if the eval difference is above this threshold,
+/// i.e. the move does not make the position significantly worse.
+/// https://github.com/lichess-org/lila/blob/d29f27d8cbb0e0dac38308c23c63b828028c085f/ui/analyse/src/retrospect/retroCtrl.ts#L161
+const double _kCorrectMovePovDiffThreshold = -0.04;
+
+/// Logic is based on the `RetroCtrl` class in the lila frontend.
```
veloce: Not really useful, and not relevant for a doc comment; should be removed.

lib/src/model/analysis/retro_controller.dart
```diff
@@ -0,0 +1,537 @@
...
+      node.position == serverBranch.position ||
+      node.position.isCheckmate ||
+      // Any move that was found in the master's database is also considered a solution.
+      openingExplorerSolutions.contains(node.sanMove!.move.uci);
+}
+
+/// Depth needed to evaluate alternative solution moves.
+///
+/// https://github.com/lichess-org/lila/blob/d29f27d8cbb0e0dac38308c23c63b828028c085f/ui/analyse/src/retrospect/retroCtrl.ts#L151
+const double _kEvalDepthThreshold = kDebugMode ? 12 : 18;
```
veloce, line 61: In lila code they also compare the millis with depth >= 14.

lib/src/model/analysis/retro_controller.dart
```diff
@@ -0,0 +1,537 @@
...
+  bool isSolution(RetroCurrentNode node) =>
+      node.position == serverBranch.position ||
+      node.position.isCheckmate ||
+      // Any move that was found in the master's database is also considered a solution.
+      openingExplorerSolutions.contains(node.sanMove!.move.uci);
+}
+
+/// Depth needed to evaluate alternative solution moves.
+///
+/// https://github.com/lichess-org/lila/blob/d29f27d8cbb0e0dac38308c23c63b828028c085f/ui/analyse/src/retrospect/retroCtrl.ts#L151
```
veloce, line 60: Better not put the link to Lila code actually; line number will most probably change; or just put the link to the whole rectroCtrl file.

tom-anders:
> > Good idea! Maybe directly after the game, we could also add a "learn from mistakes" button to the menu that would request server analysis, display some kind of progress dialog, and then take you directly to the retro screen...?
> 
> yes we could do that; this adds complexity and will likely not work from time to time, but not a big deal.
> 
> if we do that though, I'd handle the logic in the retro controller:
> 
>     * if the game was not analysed yet, request server analysis in retro `build` method
> 
>     * while server analysis is pending, simply display a circular progress with the text "Calculating moves..."
> 
>     * probably need a timeout like 1 minute, if for some reason the server analysis never finishes

Sounds good, I moved it to a new issue

lib/src/model/analysis/retro_controller.dart
tom-anders: As I understand it, since I used the "Copy permalink" button, these links refer to a specific commit, not to the `master` branch. So they will always point to the same line/code

lib/src/model/analysis/retro_controller.dart
tom-anders: Yeah we do that already, but I haven't factored it out to a constant yet, done now.

lib/src/model/engine/evaluation_mixin.dart
```diff
@@ -128,6 +128,13 @@ mixin EngineEvaluationMixin {
     }
   }
 
+  @mustCallSuper
+  Future<void> stopEval() async {
+    if (evaluationState.isEngineAvailable(evaluationPrefs)) {
+      toggleEngine();
```
veloce: We don't want toggleEngine() here. It will effectively stop the engine if it is running, but also kill it.

Here we only want to stop the running eval, not kill the engine.

You want to call `EvaluationService.stop()` instead.

test/view/analysis/retro_screen_test.dart
```diff
@@ -1,88 +1,131 @@
...
-  final mockOpeningExplorerRepository = MockOpeningExplorerRepository();
-  when(
-    () => mockOpeningExplorerRepository.getMasterDatabase(any(), since: any(named: 'since')),
-  ).thenAnswer(
-    (invocation) => Future.value(
-      openingExplorerEntries.get(invocation.positionalArguments.first as String) ??
-          OpeningExplorerEntry.empty(),
-    ),
-  );
+  final mockClient = MockClient((request) {
```
veloce, line 30: 👍

lib/src/model/engine/evaluation_preferences.dart
```diff
@@ -36,6 +36,11 @@ class EngineEvaluationPreferences extends _$EngineEvaluationPreferences
     return save(state.copyWith(isEnabled: !state.isEnabled));
   }
 
+  Future<void> disable() {
```
veloce: Looks like it is not used? at least cannot find the usage within these changes.

lib/src/model/engine/evaluation_mixin.dart
tom-anders: makes sense, fixed

tom-anders: Oh that broke a test now, will have a look later

tom-anders: Hmm I couldn't really figure out why exactly it broke. I went back to the original version now, but calling `requestEval()` only when the move is correct, to go back to the original user-configured search time. If the move was incorrect, we restart the eval anyway after taking it back

veloce:
rebasing your changes now with main @tom-anders

PR #2243, opened 2025-10-02
feat: "show treat" mode in analysis

tom-anders:
Some notes/questions:

- The icon should probably be the same as in the web UI, but it doesn't seem to be available in our `LichessIcons` currently.
- I only added the button to toggle this feature to the Analysis screen for now, because I wasn't sure where the ideal place to put it would be. In this PR, it's in the top bar, but maybe there's a better place. Eventually the button should be added to study, retro, puzzle and broadcast as well.
- Do you think it's feasible to add tests for this? It looks like (almost?) all engine related tests rely on cloud evaluation. I think it would also be difficult to verify that the engine is indeed in threat mode instead of normal mode..? Do you have any ideas?

[threat.webm](https://github.com/user-attachments/assets/4efc8ab1-a431-4f81-b6d5-840f42ea9b67)


veloce:
I don't think it should be on top bar, for sure. I guess the best place is in the sandwich bottom menu ("Show threat"). The menu item should only be displayed is an analysis is currently running.

tom-anders:
> I don't think it should be on top bar, for sure. I guess the best place is in the sandwich bottom menu ("Show threat"). The menu item should only be displayed is an analysis is currently running.

Hmm, at least on the Analysis Screen, there would also be space in the bottom bar. Since "show treat" resets after each move, I imagine people will use this button rather frequently, so it would be nice to have it accessible with a single click, like in web

veloce:
> I imagine people will use this button rather frequently

Are you sure about that? I never use this feature for instance. I cannot tell how many % of user use that feature obviously, but I bet this is fairly low (even on web where the button is accessible).

A bottom bar button makes sense of course, but at the same time a bottom bar button in that context is like a super important button imo. Too important for the threat mode. I can be wrong of course but for instance "Switch sides" button makes more sense. The problem of being in bottom bar is also that the button is inactive when engine is not active, taking space for nothing for people not using engine.

Could be worth to check what other chess app do here? like stockfish app, etc.

tom-anders:
> > I imagine people will use this button rather frequently
> 
> Are you sure about that? I never use this feature for instance. I cannot tell how many % of user use that feature obviously, but I bet this is fairly low (even on web where the button is accessible).

Ah, sorry, I should have been more clear: I don't think that a large percentage of users will use the button. I meant that those who **do** use the button will use it frequently (since threat mode resets when you make a move)

> A bottom bar button makes sense of course, but at the same time a bottom bar button in that context is like a super important button imo. Too important for the threat mode. I can be wrong of course but for instance "Switch sides" button makes more sense. The problem of being in bottom bar is also that the button is inactive when engine is not active, taking space for nothing for people not using engine.

Agreed, probably not a fit for the bottom bar then. Do you think the top bar would be too crowded?

tom-anders:
Anyway, then let's just put it into the menu for now and wait if people complain that it's too cumbersome to use

ijm8710:
I have a somewhat wacky roundabout idea, curious y'all's thoughts

Threats can be clicked from the hamburger menu. But once it is clicked, it replaces the engine toggle button on bottom bar just for that single game (or adds a 5th icon temporarily).

The benefit is that way it allows for easier consecutive access to threats if you're analyzing a series of them

But it only temporarily takes over the premier space.

Thoughts?

veloce:
> Ah, sorry, I should have been more clear: I don't think that a large percentage of users will use the button. I meant that those who do use the button will use it frequently (since threat mode resets when you make a move)

This is how it works on website, and I guess it makes sense because the button is accessible. But we could consider *not* reseting the threat mode when making a move or changing position. 

What is the normal usage of threat mode? I don't know cause I never use that feature. I suppose you want to show threats only for *some* moves (I guess very few); in that case it makes sense that it resets. And also, if my assumption is correct, it doesn't seem a big deal on mobile to have to open a menu again to show threats after one move cause one probably not want to see threats in the next position.

In conclusion what we have now is good imo; then again I don't use the feature so we should wait for chess players advices/feedback here and the good way to obtain that is to publish the feature as is.

tom-anders:
> > Ah, sorry, I should have been more clear: I don't think that a large percentage of users will use the button. I meant that those who do use the button will use it frequently (since threat mode resets when you make a move)
> 
> This is how it works on website, and I guess it makes sense because the button is accessible. But we could consider _not_ reseting the threat mode when making a move or changing position.
> 
> What is the normal usage of threat mode? I don't know cause I never use that feature. I suppose you want to show threats only for _some_ moves (I guess very few); in that case it makes sense that it resets. And also, if my assumption is correct, it doesn't seem a big deal on mobile to have to open a menu again to show threats after one move cause one probably not want to see threats in the next position.
> 
> In conclusion what we have now is good imo; then again I don't use the feature so we should wait for chess players advices/feedback here and the good way to obtain that is to publish the feature as is.

Agreed, I also have use the feature that much yet tbh. So let's publish it like this and wait for feedback

ijm8710:
Did yall have a chance to glance at my idea? No worries if you just  prefer publishing as is but I (biased) thought my idea was pretty decent :)

> I have a somewhat wacky roundabout idea, curious y'all's thoughts
> 
> Threats can be clicked from the hamburger menu. But once it is clicked, it replaces the engine toggle button on bottom bar just for that single game (or adds a 5th icon temporarily).
> 
> The benefit is that way it allows for easier consecutive access to threats if you're analyzing a series of them
> 
> But it only temporarily takes over the premier space.
> 
> Thoughts?

tom-anders:
> Did yall have a chance to glance at my idea? No worries if you just prefer publishing as is but I (biased) thought my idea was pretty decent :)
> 
> > I have a somewhat wacky roundabout idea, curious y'all's thoughts
> > Threats can be clicked from the hamburger menu. But once it is clicked, it replaces the engine toggle button on bottom bar just for that single game (or adds a 5th icon temporarily).
> > The benefit is that way it allows for easier consecutive access to threats if you're analyzing a series of them
> > But it only temporarily takes over the premier space.
> > Thoughts?

Ah sorry I forgot about that - I think it's a good idea, but not sure if it's worth the complexity and maintenance cost

veloce:
I rebased your branch to main fyi @tom-anders

lib/src/model/analysis/analysis_controller.dart
```diff
@@ -663,6 +663,15 @@ class AnalysisController extends _$AnalysisController
     }
   }
 
+  Future<void> toggleEngineThreatMode() async {
+    if (state.hasValue) {
+      state = AsyncData(
+        state.requireValue.copyWith(engineInThreatMode: !state.requireValue.engineInThreatMode),
+      );
+      requestEval();
```
veloce, line 671: why is `requestEval()` called here and not in broadcast and retro controllers?

lib/src/model/common/eval.dart
```diff
@@ -242,11 +243,41 @@ sealed class PvData with _$PvData {
...
+};
+
 ISet<Shape> computeBestMoveShapes(
   IList<MoveWithWinningChances> moves,
   Side sideToMove,
-  PieceAssets pieceAssets,
-) {
+  PieceAssets pieceAssets, {
+  required Color bestMoveColor,
+  required Color otherMovesColor,
```
veloce: What would be the "other moves"? it is not very clear. Since this is about threat mode and the only other color is for threats I'd name it threatColor.

lib/src/model/engine/evaluation_service.dart
```diff
@@ -216,16 +218,30 @@ class EvaluationService {
       currentWork: null,
     );
 
+    Position initialPosition() {
```
veloce: Not sure to understand the logic behind this. To me the initial position should not be dependent on the state of `threatMode`; it doesn't feel right.

lib/src/model/engine/evaluation_service.dart
```diff
@@ -216,16 +218,30 @@ class EvaluationService {
...
       hashSize: maxMemory,
+      threatMode: threatMode,
       searchTime: goDeeper ? kMaxEngineSearchTime : options.searchTime,
       isDeeper: goDeeper,
       multiPv: options.multiPv,
       path: path,
-      initialPosition: context.initialPosition,
-      steps: IList(steps),
+      initialPosition: initialPosition(),
+      steps: threatMode ? const IList.empty() : IList(steps),
```
veloce: Same with `steps`. I feel it is weird to do it like that. I'd rather not change the `work` parameters since we already have that boolean which indicates the threat mode.

That logic should be move elsewhere imo, probably in uci_protocol?

lib/src/view/analysis/analysis_board.dart
```diff
@@ -90,7 +92,28 @@ abstract class AnalysisBoardState<
...
 
-    return computeBestMoveShapes(eval.bestMoves, currentPosition.turn, pieceAssets);
+    final bestMoveShapes = computeBestMoveShapes(
+      eval.bestMoves,
+      currentPosition.turn,
+      pieceAssets,
+      // Same colors as in the Web UI with a slightly different opacity
+      // The best move has a different color than the other moves
+      bestMoveColor: const Color(0x66003088),
+      otherMovesColor: const Color(0x664A4A4A),
```
veloce: Ok now I understand this is to distinguish from the other lines, so forget my previous comment. Well the name is still a bit confusing, but I cannot find a good name atm ;)

lib/src/view/engine/engine_gauge.dart
```diff
@@ -61,11 +61,13 @@ class EngineGauge extends ConsumerWidget {
     final localEval = params.isLocalEngineAvailable
         ? ref.watch(engineEvaluationProvider).eval
         : null;
-    final eval = pickBestEval(
-      localEval: localEval,
-      savedEval: params.savedEval,
-      serverEval: params.serverEval,
-    );
+    final eval = localEval?.threatMode == true
```
veloce: Seems this logic is used at several places. What about factoring it out in `pickBestEval`?

veloce (commented):
Currently crashing on the iOS simulator (haven't checked why yet).

lib/src/view/engine/engine_gauge.dart
tom-anders: IIRC I considered it but thought it was not possible, because in `analysis_board.dart` the logic is a bit different than in the other places. But after thinking about it some more it's not actually a problem. I added an `if (localEval?.threatMode == true) { return localEval; }` to the `pickBestEval` function now

lib/src/view/analysis/analysis_board.dart
tom-anders: Maybe `secondBestMovesColor`? idk

lib/src/model/analysis/analysis_controller.dart
tom-anders: I think I forgot to add it there, fixed now

lib/src/model/engine/evaluation_service.dart
tom-anders: I initially followed lila here (https://github.com/lichess-org/lila/blob/9993424e794088f3c28fd412cba30c8d56fd5ced/ui/lib/src/ceval/ctrl.ts#L24) but we of course can do it differently. I moved it to `uci_protocol.dart` now

lib/src/model/engine/work.dart
```diff
@@ -28,10 +28,18 @@ sealed class Work with _$Work {
     required IList<Step> steps,
   }) = _Work;
 
-  Position get position => steps.lastOrNull?.position ?? initialPosition;
-
-  /// The work ply.
-  int get ply => steps.lastOrNull?.position.ply ?? initialPosition.ply;
+  Position get position {
```
veloce: Not sure about having such logic in the getter. Imo `position` should be the expected position.

Another getter called `threatPosition` maybe?

Can you explain to me why is it necessary to change the position in such a way for threat mode by the way? I am not sure to understand why.
If there is another way to do it, I'd rather explore that. It feels weird to change the position like that.

veloce:
Tried to fix conflicts through web and it messed up the formatting... my bad.

veloce:
will rebase branch and fix lint @tom-anders

veloce:
@tom-anders I have tested it and I see a regression in your branch. Don't know the cause but here is the steps to reproduce:

- go to a move, wait for the analysis to complete (go idle)
- tap the top engine depth widget and tap the "+" (go deeper)
- wait for the engine to compute
- switch move again (go backward for instance)

The evaluation is now in a bad state as the did not switch the computation to the last move; engine depth widget is showing "nullkn/s". I can reproduce all the time in this branch and I do not reproduce on main.

tom-anders:
> @tom-anders I have tested it and I see a regression in your branch. Don't know the cause but here is the steps to reproduce:
> 
> * go to a move, wait for the analysis to complete (go idle)
> * tap the top engine depth widget and tap the "+" (go deeper)
> * wait for the engine to compute
> * switch move again (go backward for instance)
> 
> The evaluation is now in a bad state as the did not switch the computation to the last move; engine depth widget is showing "nullkn/s". I can reproduce all the time in this branch and I do not reproduce on main.

Will look into it in the next few days 👍

tom-anders:
(Un)fortunately, I think I can reproduce the issue on `main` as well, or am I seeing a different issue here? If so, can you post a quick recording as well?

[godeeperbug.webm](https://github.com/user-attachments/assets/574374b7-5a10-4978-83a3-3d06ec71f536)

veloce:
I could reproduce the bug on main (with slightly different steps, thanks for showing me). It is fixed now and I think this is good too on this branch.

PR #2248, opened 2025-10-05
Fix retro evaluation and simplify retro interface

veloce:


lib/src/model/analysis/retro_controller.dart
```diff
@@ -85,16 +80,12 @@ class RetroController extends _$RetroController with EngineEvaluationMixin {
...
   @protected
   EngineEvaluationPrefState get evaluationPrefs =>
-      state.valueOrNull?.feedback == RetroFeedback.evalMove
-      // Increase search time to make sure we reach the limit for deciding whether the move
-      // is good enough to be considered an alternative solution.
-      ? ref
-            .read(engineEvaluationPreferencesProvider)
-            .copyWith(engineSearchTime: kMaxEngineSearchTime)
-      : ref.read(engineEvaluationPreferencesProvider);
+      ref.read(engineEvaluationPreferencesProvider).copyWith(isEnabled: true, numEvalLines: 1);
```
veloce, line 88: These settings should not change based on some logic, as this will effectively kill and restart the engine.

lib/src/model/analysis/retro_controller.dart
```diff
@@ -163,14 +152,13 @@ class RetroController extends _$RetroController with EngineEvaluationMixin {
         feedback: RetroFeedback.findMove,
         mainlinePath: _root.mainlinePath,
         pov: options.initialSide,
-        currentNode: RetroCurrentNode.fromNode(_root.view),
```
veloce, line 166: Avoid using root.view when possible.

lib/src/model/analysis/retro_controller.dart
```diff
@@ -492,7 +459,8 @@ class RetroController extends _$RetroController with EngineEvaluationMixin {
             this.state = AsyncValue.data(
               state.copyWith(feedback: RetroFeedback.evalMove, evalRequestedAt: DateTime.now()),
             );
-            requestEval();
+            // Be sure to get enough depth to evaluate the move properly
```
tom-anders, line 462: Ahh that’s a much better solution of course, nice!

lib/src/view/analysis/retro_screen.dart
```diff
@@ -214,10 +184,10 @@ class _RetroAnalysisBoardState
   AnalysisPrefs get analysisPrefs => ref.watch(analysisPreferencesProvider);
 
   @override
-  bool get showAnnotations => analysisPrefs.showAnnotations;
+  bool get showAnnotations => false;
 
   @override
-  bool get hideBestMoveArrow => analysisState.isSolving;
+  bool get hideBestMoveArrow => true;
```
tom-anders, line 190: Hmm, don't we want to show the arrows after the solution has been found, in case the user wants to analyse further?

lib/src/model/analysis/retro_controller.dart
tom-anders: Makes sense, I think the missing conversion from view back to mutable node was the reason why I used views here

lib/src/model/analysis/retro_controller.dart
tom-anders: Makes sense, maybe we can add a warning comment about this somewhere in the code?

lib/src/view/analysis/retro_screen.dart
veloce: I think it is better to keep it separate from analysis, and then we don't want to use analysis settings. So I found it simpler to remove the lines and arrows for now.

test/view/analysis/retro_screen_test.dart
```diff
@@ -161,13 +161,13 @@ void main() {
       await tester.pumpWidget(await makeTestApp(tester, moves: 'e4 e5'));
 
       // Wait for retro to load
-      await tester.pumpAndSettle();
```
veloce, line 164: Should always use pump instead of pumpOfSettle. The latter is normally only needed when you need to wait for an animation to finish.

test/view/analysis/retro_screen_test.dart
```diff
@@ -189,36 +189,35 @@ void main() {
...
       // Wait for failure message to appear and move to be taken back
-      await tester.pumpAndSettle(const Duration(milliseconds: 500));
+      await tester.pump(const Duration(milliseconds: 500));
       expect(find.byKey(const ValueKey('g1-whiteknight')), findsOneWidget);
 
       expect(find.text('You can do better'), findsOneWidget);
       expect(find.text('Try another move for white'), findsOneWidget);
 
       await playMove(tester, 'd2', 'd4');
-      await tester.pumpAndSettle(const Duration(milliseconds: 500));
```
veloce, line 207: This was causing the test to fail.

PR #2255, opened 2025-10-07
fix: don't try to account for alternative castling notation in annotations

### LINKED ISSUE #2231: Blunder symbol in the analysis board is incorrectly displayed
### What is the problem?

Some times the blunder mark is displayed in the wrong square in the analysis board screen

https://github.com/user-attachments/assets/4f225a7c-166d-4a7d-aa36-e1eb7223e08e

1. Open this game in the mobile app: https://lichess.org/7YMnLtkA/black
2. Go to the analysis board
3. Navigate to move 32 and observe how the 33 blunder move is displayed

### App version

0.17.8

### Device Info

Samsung S23 android 15

### Additional information

_No response_

tom-anders:
This leads annotations being displayed on the wrong square e.g. when moving the rook from e1 to h1 (as its incorecctly interpreted as a castling move).

Luckily, we already convert alternative castling notation when constructing the node tree [here](https://github.com/lichess-org/mobile/blob/88e67e4bc7a347dfb6de41391a904ef718e8bb76/lib/src/model/common/node.dart#L223), so instead of extending the logic in the AnalysisBoard, we can just get rid of this logic entirely.

Fixes #2231


veloce:
Thanks for this! A test would be nice here, to avoid any future regression.

PR #2275, opened 2025-10-19
Made it so that Wakelock disables when a game ends - DC

### LINKED ISSUE #2271: Suggestion: don't keep the screen on if the game is over (wake lock)
I know very well about the app keeping the screen on, but I always forget about this after the game ends. Which means I end a game, leave the phone, and it happily shows the game over screen until I pick it up or remember about this feature.

ChopinDavid:
This would close #2271.

Please, let me know if there are other places I should have added this and I'll be sure to make updates.

### DIFF
--- a/lib/src/view/game/game_body.dart
+++ b/lib/src/view/game/game_body.dart
@@ -107,11 +107,6 @@ class GameBody extends ConsumerWidget {
         debugPrint('SEVERE: [GameBody] could not load game data; $e\n$s');
         return const LoadGameError('Sorry, we could not load the game. Please try again later.');
       case AsyncData(value: final gameState, isRefreshing: false):
-        if (gameState.game.status != GameStatus.created &&
-            gameState.game.status != GameStatus.started) {
-          WakelockPlus.disable();
-        }
-
         final youAre = gameState.game.youAre ?? Side.white;
 
         // If playing against Stockfish, user is null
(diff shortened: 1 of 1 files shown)


lib/src/view/game/game_body.dart
```diff
@@ -105,6 +107,11 @@ class GameBody extends ConsumerWidget {
         debugPrint('SEVERE: [GameBody] could not load game data; $e\n$s');
         return const LoadGameError('Sorry, we could not load the game. Please try again later.');
       case AsyncData(value: final gameState, isRefreshing: false):
+        if (gameState.game.status != GameStatus.created &&
```
veloce: Would be better in the `_stateListener` I think, instead of making a side effect in the `build` method.

test/view/game/game_screen_test.dart
```diff
@@ -789,6 +792,51 @@ void main() {
       });
     });
   });
+
+  group('Wakelock', () {
```
veloce, line 796: 👍 thanks for the tests

lib/src/view/game/game_body.dart
ChopinDavid: Done 👍

ChopinDavid:
@veloce went ahead and made that change and rebased upstream/main. Thanks!

PR #2277, opened 2025-10-19
Fix(Clock Tool): Support for 0+X Time Controls

cappylot:
Previously, in OTB mode, time controls with zero starting time (e.g., 0+6) caused an immediate flag since the active player's clock started ticking instantly. Now, the clock only starts counting down for each player after they make their first move, allowing 0+X controls to function as expected.

### DIFF
--- a/lib/src/model/clock/clock_tool_controller.dart
+++ b/lib/src/model/clock/clock_tool_controller.dart
@@ -101,8 +101,9 @@ class ClockToolController extends _$ClockToolController {
     );
     // Start the countdown only if either this is not a zero-start clock
     // or the new active side has already made at least one move.
-    final Duration initialOfNewActive =
-    playerType.opposite == ClockSide.top ? state.options.topTime : state.options.bottomTime;
+    final Duration initialOfNewActive = playerType.opposite == ClockSide.top
+        ? state.options.topTime
+        : state.options.bottomTime;
     final bool hasNewActiveMoved =
         (playerType.opposite == ClockSide.top ? state.topMoves : state.bottomMoves) > 0;
     if (initialOfNewActive.inMilliseconds != 0 || hasNewActiveMoved) {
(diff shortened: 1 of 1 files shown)


veloce:
Thanks for this fix. 

I am not that familiar with OTB clocks and how they should behave really. So I wonder:
- should we have this logic to start timer immediately only if initial time > 0
- or should we always delay the timer after move 1

Seems that the first option (what you did in the PR) adds more complexity. In lichess games (online) the timer do start after the first move, so I wonder, why would it be different in OTB?

ijm8710:
I may be completely wrong, but I also was trying to test this and I "think" this applies moreso to the clock tool rather than OTB

In the clock tool, you are met with a flag instantaneously when the game is started despite having increment time. It seems impossible to use the clock tool with 0+x.

For otb, it appears to me you are given the increment time to start which allows time for one to play their first turns. 

Is there a chance this patch is for the clock tool and not otb?

cappylot:
@veloce  
grafik

In online lichess games the timer does **not** start automatically. (This is a web screenshot but the same goes for the mobile app)
If the first player (white) makes their move, black also has a little buffer to make their first move and only then does the timer go down while waiting for future moves. 

I figured the Chess Tool for OTB should mimic a real over the board clock as closely as possible (they start counting down immediately) except for 0+X time controls. Since the lichess app offers setting the starting time to 0 seconds, it has to give the players time to make their first moves but otherwise I would keep it as is, to stay close to real OTB chess clocks. 

I guess this way adds "more complexity" as you said, but I think this is the best way to do it if we want to have the best of both worlds, namely making 0+X possible and keeping it "realistic" at the same time.

cappylot:
@veloce  just checking in — is there anything I can adjust or improve to help get this pull request accepted?

veloce:
> I figured the Chess Tool for OTB should mimic a real over the board clock as closely as possible (they start counting down immediately) except for 0+X time controls. Since the lichess app offers setting the starting time to 0 seconds, it has to give the players time to make their first moves but otherwise I would keep it as is, to stay close to real OTB chess clocks.

Agreed, we should stay as close as possible to real OTB clocks. Thanks for the explanation.

PR #2284, opened 2025-10-22
Made it so that an empty state is shown instead of the perpetual spinner when navigating to Friends page and following 0 users - DC

### LINKED ISSUE #2283: Endless spinner when viewing friends pages and you are not following anybody
### What is the problem?

When you are following 0 users and you navigate to your friends page, either from the "Home" tab or the "More" tab, there is a perpetual spinner instead of an empty state.

Before:
https://github.com/user-attachments/assets/0bcba04d-9b47-444d-8bd6-37110bb96dce
https://github.com/user-attachments/assets/87b9b405-be9b-4841-a563-73f74a82cc45

After:
https://github.com/user-attachments/assets/6e77a949-4905-461f-abc6-4af045740312
https://github.com/user-attachments/assets/e5789b0a-e19b-431e-9e12-59b9aae1a17c

1. Launch the app
2. Ensure you are logged in
3. From the "Home" tab, click on the friends icon in the top-right
4. Click on "More >"

or

1. Launch the app
2. Ensure you are logged in
3. Click on the "More" tab
4. Click on "Friends"

### Expected behavior
The circular spinner eventually goes away and the "n friends online"/"n following" tabs are shown with empty states.

### Actual behavior
The circular progress indicator never disappears

### App version

0.17.12+001712

### Device Info

Pixel 8 Pro Emulator, Android 16

### Additional information

_No response_

ChopinDavid:
This would close #2283 

I am wondering a couple of things:

- Should there be an empty state text under the "n friends online" tab, similar to the "You are not following any user." text we see under the "n following" tab? If so, would it just be the same text?

- Should the tabs be shown at all if the logged in user is following 0 other users? Or should we just display a more generic empty state message?

I am also soliciting feedback in a couple of areas:

- I noticed that "You are not following any user." would probably be better as "You are not following any users." I attempted to make the appropriate l10n update but am not seeing the change in the UI. Did I not update localization correctly, or does this all update after commits to main?
- I have worked almost exclusively with bloc throughout my career. I am wondering if the `friends_screen_test` tests look good, especially around how riverpod functionality was mocked.

### DIFF
--- a/lib/src/view/relation/friend_screen.dart
+++ b/lib/src/view/relation/friend_screen.dart
@@ -176,6 +176,9 @@ class _Online extends ConsumerWidget {
 
     switch (onlineFriends) {
       case AsyncData(:final value):
+        if (value.isEmpty) {
+          return Center(child: Text(context.l10n.nbFriendsOnline(0)));
+        }
         return ListView.separated(
           itemCount: value.length,
           separatorBuilder: (context, index) => Theme.of(context).platform == TargetPlatform.iOS
--- a/test/view/relation/friend_screen_test.dart
+++ b/test/view/relation/friend_screen_test.dart
@@ -143,7 +143,7 @@ void main() {
       expect(find.byType(CenterLoadingIndicator), findsNothing);
 
       expect(find.byType(Tab), findsNWidgets(2));
-      expect(find.text('0 friends online'), findsOneWidget);
+      expect(find.text('0 friends online'), findsNWidgets(2));
       expect(find.text('0 following'), findsOneWidget);
     });
 
@@ -235,7 +235,7 @@ void main() {
         // We pump and settle because the switch between tabs is animated
         await tester.pumpAndSettle();
 
-        expect(find.text('You are not following any user.'), findsOneWidget);
+        expect(find.text('You are not following any users.'), findsOneWidget);
       },
     );
   });


veloce:
> Should there be an empty state text under the "n friends online" tab, similar to the "You are not following any user." text we see under the "n following" tab? If so, would it just be the same text?

Yes and we should use the "0 friends online" which is already translated.

> Should the tabs be shown at all if the logged in user is following 0 other users? Or should we just display a more generic empty state message?

I would say yes.

> I noticed that "You are not following any user." would probably be better as "You are not following any users." I attempted to make the appropriate l10n update but am not seeing the change in the UI. Did I not update localization correctly, or does this all update after commits to main?

I assume you made the change on crowdin, thank you for that. The UI is not updated automatically, changes from crowdin are pulled automatically each end of week.

lib/src/view/relation/friend_screen.dart
```diff
@@ -192,12 +201,13 @@ class _Following extends ConsumerWidget {
...
   @override
   Widget build(BuildContext context, WidgetRef ref) {
-    final followingAndOnlines = ref.watch(_followingStatusesProvider);
+    final followingAndOnlines = ref.watch(followingStatusesProvider);
 
     switch (followingAndOnlines) {
       case AsyncData(:final value):
         IList<User> following = value.$1;
         return StatefulBuilder(
+          key: const Key('following_tab_view'),
```
veloce: There's nothing wrong with having keys for testing purposes, but I like to avoid it if possible. Widget matchers work pretty well in flutter, and usually it is not cumbersome at all to test a match with a widget directly (`find.byType`) or text within a widget.

test/view/relation/friend_screen_test.dart
```diff
@@ -0,0 +1,251 @@
...
+            httpClientFactoryProvider.overrideWith((ref) {
+              return FakeHttpClientFactory(() => mockClient);
+            }),
+            onlineFriendsProvider.overrideWith(() => _MockOnlineFriends(const IList.empty())),
+          ],
+        );
+
+        await tester.pumpWidget(app);
+
+        await tester.pumpAndSettle();
```
veloce, line 227: I find it a good practice to indicate with a one-liner comment the reason of the `pumpAndSettle`. It helps the future reader to understand the sometime complex asynchronous calls. And sometimes we're lazy and use a pumpAndSettle where it is not needed, which should be avoided.

test/view/relation/friend_screen_test.dart
veloce, line 1: Thanks for adding tests. Always appreciated :)

I usually don't bother to test the provider as it is (partially) tested with the widget test. But I won't complain you added one!

translation/source/mobile.xml
```diff
@@ -20,7 +20,7 @@
   <string name="mustBeLoggedIn" comment="This is shown when the user tries to access a feature that requires login.">You must be logged in to view this page.</string>
   <string name="noSearchResults" comment="This is shown when the search did not return any results.">No results</string>
   <string name="notAllFeaturesAreAvailable" comment="This is shown in the welcome message when the app is first installed">Please note that not all features from the old app or the website are currently available, but we are adding features all the time.</string>
-  <string name="notFollowingAnyUser" comment="This is shown in the list that contains user followed by the user. Displayed if the user does not follow any users.">You are not following any user.</string>
+  <string name="notFollowingAnyUser" comment="This is shown in the list that contains user followed by the user. Displayed if the user does not follow any users.">You are not following any users.</string>
```
veloce, line 23: Oh I see now, the source was wrong. Thanks.

To update the UI after changing the source or pulling new changes from crowdin you need to:

```sh
./scripts/gen-arb.mjs && flutter gen-l10n
```

(should be documented in the l10n doc)

ChopinDavid:
@veloce ok, this should be good for re-review!

translation/dest/mobile/en-US.xml
```diff
@@ -20,7 +20,7 @@
   <string name="mustBeLoggedIn" comment="This is shown when the user tries to access a feature that requires login.">You must be logged in to view this page.</string>
   <string name="noSearchResults" comment="This is shown when the search did not return any results.">No results</string>
   <string name="notAllFeaturesAreAvailable" comment="This is shown in the welcome message when the app is first installed">Please note that not all features from the old app or the website are currently available, but we are adding features all the time.</string>
-  <string name="notFollowingAnyUser" comment="This is shown in the list that contains user followed by the user. Displayed if the user does not follow any users.">You are not following any user.</string>
+  <string name="notFollowingAnyUser" comment="This is shown in the list that contains user followed by the user. Displayed if the user does not follow any users.">You are not following any users.</string>
```
veloce, line 23: This file should not be changed by hand because it will be overwritten fyi.

PR #2287, opened 2025-10-23
Tablebase: Prepare 8 piece DTC support, tweak metric display, relax response parser

niklasf:
* Relax `TablebaseEntry` factory: Currently all JSON keys are always present, but that was undocumented and only intended to be guaranteed for `moves`, `category`, `uci` and `san`.
* Code golf category split: Just filter the collection multiple times in `addMoveSection`, for slightly more concise code.
* Query with 9 pieces to prepare for 8-piece tablebase: There may soon be information about 9-piece positions that have a capture leading to 8 pieces.
* Fix display of zeroing move if only DTZ known: Due to small typo.
* Order tablebase metric display: DTZ < DTC < DTM, because soon all three may be available for some positions.


veloce:
Thanks!

Issue #2220, opened 2025-09-23
Move app settings to the "more" menu tab


allanjoseph98:
Clicking the one tiny initial of your username to open a side-menu that opens settings isn't intuitive. I'd also move the about section there as well.

stephanhuebsch:
You can also swipe from the left border of the screen towards the right in order to open the menu. 

If you already know that there is such a menu, then swiping is a natural way to open it in my opinion. However, at first glance it might not be obvious that there even is a menu. Maybe replacing the user flair on the top left of the "start" screen with a hamburger menu icon or something else would help?

allanjoseph98:
Swiping is not discoverable/virtually impossible on androids with gesture navigation enabled

ijm8710:
I agree it's slightly unintuitive but personally I really like the side menu and having profile inbox and settings all there together.

I think the 2 important things are


1) I've seen a bunch of posts where users can't find a specific menu setting such as how to enable the "play" matrix. But I never see posts where they can't find settings altogether

2) nowadays I think users expect/know settings will be in an app. So they're willing to hunt for it. So perhaps it may take you an extra 15 seconds to find, but I do really believe almost all users will find it!

veloce:
I don't mind putting the settings in 2 places.

> Clicking the one tiny initial of your username to open a side-menu that opens settings isn't intuitive. I'd also move the about section there as well.

Signal app does that though. Spotify app too.

allanjoseph98:
#3065 makes settings more discoverable by putting it in an overflow menu

PR #2288, opened 2025-10-24
Adds settings to the more menu.

manzor99:
Adds a link to the settings activity from the more menu as requested in the following issue: https://github.com/lichess-org/mobile/issues/2220

### DIFF
--- a/lib/src/model/explorer/tablebase.dart
+++ b/lib/src/model/explorer/tablebase.dart
@@ -23,7 +23,6 @@ sealed class TablebaseEntry with _$TablebaseEntry {
 
   const factory TablebaseEntry({
     required int? dtz,
-    required int? preciseDtz,
     required int? dtc,
     required int? dtm,
     required bool checkmate,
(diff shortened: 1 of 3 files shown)


veloce:
It was requested, but I'm not sure about it. I don't feel it is necessary.

tom-anders:
IMO it's a good thing for discoverability, it's not obvious that there's a menu that you can open by sliding or tapping the profile icon. Using a hamburger icon instead of the profile Pic would be another way to make settings more discoverable I guess, like it's done in the old app 🤔

veloce:
I hear you and I can't deny there are obviously people struggling with finding the settings. But at the same time I don't want to overreact to lichess users feedback, and especially old app users. This is not because it was done so in the old app (or in the website) that we *have* to do it in the new one.

As I said elsewhere, I know at least 2 very popular apps that use the profile pics to open a menu with settings: Signal and Spotify. At least on iOS, but I assume it is also the case on android, isn't it?

If they do it, I don't see why we can't. Unless there is something that I don't see here?

I'm leaving this open for now, waiting to gather more reactions. For now they are not overwhelming so I'm not yet convinced.

tom-anders:
> I hear you and I can't deny there are obviously people struggling with finding the settings. But at the same time I don't want to overreact to lichess users feedback, and especially old app users. This is not because it was done so in the old app (or in the website) that we _have_ to do it in the new one.
> 
> As I said elsewhere, I know at least 2 very popular apps that use the profile pics to open a menu with settings: Signal and Spotify. At least on iOS, but I assume it is also the case on android, isn't it?
> 
> If they do it, I don't see why we can't. Unless there is something that I don't see here?
> 
> I'm leaving this open for now, waiting to gather more reactions. For now they are not overwhelming so I'm not yet convinced.

True, Spotify does it on Android as well. And WhatsApp too btw. I can see both sides of the arguments as well, but now I'm also a bit more convinced that's it's fine as it is now 😅

veloce:
Decide not to do that for now.

PR #2298, opened 2025-10-26
fix: hide study hint when changing chapter or correct move has been played

### LINKED ISSUE #2212: Hints in studies
### What is the problem?

When you take a hint in the studies, it doesn't disappear in subsequent chapters.

1. Enter the study with an interactive mode and a hint.
2. Go to another chapter with a hint (or make a move).
3. See that the hint is not hidden.

### App version

?

### Device Info

?

### Additional information

_No response_

tom-anders:
Fixes #2212

### DIFF
--- a/lib/src/view/study/study_gamebook.dart
+++ b/lib/src/view/study/study_gamebook.dart
@@ -90,8 +90,32 @@ class _Hint extends ConsumerStatefulWidget {
 class _HintState extends ConsumerState<_Hint> {
   bool showHint = false;
 
+  void _hideHint() {
+    setState(() {
+      showHint = false;
+    });
+  }
+
   @override
   Widget build(BuildContext context) {
+    ref.listen(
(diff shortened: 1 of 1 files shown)


lib/src/view/study/study_gamebook.dart
```diff
@@ -90,8 +90,32 @@ class _Hint extends ConsumerStatefulWidget {
...
+    ref.listen(
+      studyControllerProvider(widget.id).select((state) => state.valueOrNull?.gamebookState),
+      (prev, next) {
+        if (prev == GamebookState.correctMove && next == GamebookState.findTheMove) {
+          _hideHint();
+        }
+      },
+    );
+
+    ref.listen(
```
veloce, line 110: Not sure having 2 `ref.listen` is better than a single one with no select. I'd tend to say no, but it's no big deal, let's keep it that way.

veloce (approved):

PR #2320, opened 2025-10-30
fix: otb config bottom sheet will show up at otb game start

### LINKED ISSUE #2317: Choosing a time from the play button does not apply it to your OTB match. It'll always just start with 5 minutes
As reported by another user on discord

nirajk14:
Closes #2317

### DIFF
--- a/test/view/over_the_board/over_the_board_screen_test.dart
+++ b/test/view/over_the_board/over_the_board_screen_test.dart
@@ -179,8 +179,16 @@ void main() {
 Future<Rect> initOverTheBoardGame(WidgetTester tester, TimeIncrement timeIncrement) async {
   final app = await makeTestProviderScopeApp(tester, home: const OverTheBoardScreen());
   await tester.pumpWidget(app);
+
+  // Wait for bottom sheet to show up
   await tester.pumpAndSettle();
+
+  // User taps Play to start game now
   await tester.tap(find.text('Play'));
+
+  // Wait for bottom sheet to disappear
+  await tester.pumpAndSettle();
+
   await tester.tap(find.byIcon(CupertinoIcons.play));
 
   final container = ProviderScope.containerOf(tester.element(find.byType(Chessboard)));


veloce:
Nothing wrong in the code of this PR. But:

- I do not reproduce #2317 , do you?
- this PR is about showing automatically the setup bottom sheet; I don't see how this could fix any issue reported in the above issue?

nirajk14:
Yes, I am able to reproduce it. I decided to show the new game bottom sheet so user can decide the time control and variant to play before game start.
I chose this approach to fix the issue as over the board is separated from the widget that contains the create a game button and the buttons are also far apart. This also doesn't cause any coupling between create a game and over the board.

veloce:
Closing as the issue is not an issue (cf #2317 )

HaonRekcef:
Maybe we want to show the bottom sheet anyways, as it improves the UX. 
Currently the players need to open the a bit hidden menu to change the timecontrol except if they want to play 5+3. (Its not even obvious that its +3 at the beginning of the game)

veloce:
Yes the UX could be improved for sure. I am open to new suggestions, or PRs, if you feel like it. Note that I also don't like the bottom sheet automatically open if it doesn't allow to start the game in one tap. Which was the case if I'm not mistaken. So I'd accept an UX change, but it'd need to be a little better than just this opening of bottom sheet.

HaonRekcef:
It seems like the bottom sheet allows to start the game in one tap:

grafik

veloce:
> It seems like the bottom sheet allows to start the game in one tap:

It doesn't, as you still need to hit the "play" button after. But I agree we could bring back this PR with that change (play on bottom sheet effectively starts the game).
I have other things to do but if someone is willing to reopen this PR with the aforementioned change I'll accept it.

tom-anders:
> > It seems like the bottom sheet allows to start the game in one tap:
> 
> It doesn't, as you still need to hit the "play" button after. But I agree we could bring back this PR with that change (play on bottom sheet effectively starts the game). I have other things to do but if someone is willing to reopen this PR with the aforementioned change I'll accept it.

Hmm, that was intended behavior though, the clock is not supposed to start until you make the first move, right? You don't actually have to press the play button, making the first move starts the clock/game as well

nirajk14:
I am open to make any changes necessary if the PR is opened just need to know what the actual change required is. I thought my change was a quick smart fix that is also intuitive to end users as they will be using same bottom sheet to start game and start new games from that point and beyond. 

Also, I was also thinking of adding unlimited time option as that's how I used to play with my pals at restaurants while waiting for the food to arrive. Can I work on that on a different PR of course? Just want to get an yes from someone who can merge.

veloce:
I'm sold! Feel free to make another PR with unlimited time if you want @nirajk14

ijm8710:
Hopefully it’s okay to ask 2 questions please

1.) with the bottom sheet showing by default. Can we save the last used time control. Currently it always resets back to 5+3. The same way that online bots can save time control preferences separate from your regular games (I can create a pr if you agree) 
2. Any chance to have immersive mode off until the game is started? (Happy to pr if it makes sense)

PR #2322, opened 2025-10-30
Add Takeback button to Over-the-Board screen (#1642)

### LINKED ISSUE #1642: OTB to support "give 15 seconds" and take back"

overcharged-coder:
This PR implements a "Takeback" button for Over-the-Board (OTB) games.

- Adds a new BottomBarButton labeled "Takeback"
- Uses the existing `goBack()` logic from OTB controller
- Clock state is updated consistently if active
- Uses `Icons.undo` for clear UX

Fixes #1642

### DIFF
--- a/lib/l10n/l10n.dart
+++ b/lib/l10n/l10n.dart
@@ -130,12 +130,13 @@ abstract class AppLocalizations {
   /// Additional delegates can be added by appending to this list in
   /// MaterialApp. This list does not have to be used at all if a custom list
   /// of delegates is preferred or required.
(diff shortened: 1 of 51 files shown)


lib/main.dart
```diff
@@ -37,5 +37,10 @@ Future<void> main() async {
     await androidDisplayInitialization(widgetsBinding);
   }
 
-  runApp(ProviderScope(observers: [ProviderLogger()], child: const AppInitializationScreen()));
+  runApp(
```
veloce: You have a problem with your formatter. It should not reformat the files. It should be set to 100.

Issue #2150, opened 2025-08-25
Online status indicator incorrectly marks user as offline after short inactivity


ijm8710:
https://streamable.com/3mv11m

Description:
I noticed that the online status indicator seems to go offline too quickly. To test this, I used two accounts:

1.	I logged into my main account (Account A: kupggs using the app) — i did not play a game. But I stayed in for 6 seconds on multiple screens which I hope is enough to be considered "online".
2.	Then I switched to my test account (Account B: kupggstest using the mobile web) about 40 seconds later.
3.	From Account B, I opened the friends list to send a real-time challenge to Account A.

At this point, Account A appeared offline, so the challenge request couldn’t be sent. When I switched back to Account A, the challenge never appeared — which is expected since Account B thought I was offline.

However, what seems unexpected is that simply leaving the app for around 40 seconds was enough for my main account to be marked offline, even though I was in the app moments before.

Expected behavior:
The account should remain online for a reasonable duration (e.g., several minutes) of brief inactivity or app switching, not go offline after just 40 seconds.

Steps to reproduce:
	1.	Log in to Account A on Lichess mobile app.
	2.	Wait 30–40 seconds while the app remains in the background.
	3.	Log in to Account B and view the friends list.
	4.	Observe that Account A shows as offline.

Actual result:
Account A appears offline after only ~40 seconds of inactivity.

Expected result:
Account A should remain online for a longer duration when briefly switching apps.

veloce:
I don't understand, neither the text, nor the video. Sorry.

ijm8710:
> I don't understand, neither the text, nor the video. Sorry.

That is my fault @veloce . I rewrote both this and #2149 to try to make them clearer. Hopefully they are easier to understand.

I think #2149 is more likely to matter more so if you're able to view that one first please (when you get a chance of course)

veloce:
Actually this is normal, iOS only allows 30s of background activity.

Issue #2149, opened 2025-08-25
Real-time challenge lost within one minute when app is backgrounded (connection timeout too short)


ijm8710:
https://streamable.com/cscybm

Description:
This issue stems from comments in #1798, though that report primarily concerns behavior when the app is force-quit or killed.

In this case, the problem occurs even when the Lichess app is simply backgrounded, not closed. Real-time challenge requests appear to be lost within less than one minute after backgrounding, suggesting the connection times out prematurely.

Steps to reproduce:
	1.	Log into Account A (main account: kupggs in the app) — confirm it shows online.
	2.	From Account B (test account; kupggs in mobile web), view friends list and confirm Account A is showing online.
	3.	Immediately send a real-time challenge to Account A.
	4.	Switch back to Account A less than one minute later. Meaning we intentionally waited to grab the challenge request, but waited less than one minute.

Observed behavior:
The real-time challenge does not appear in Account A (even after a refresh), as if the session dropped almost immediately after backgrounding.

Expected behavior:
Account A should remain available to receive real-time challenge requests for at least several minutes after being backgrounded.

Notes:
Developer Veloce previously stated that if the app is backgrounded briefly (e.g., 15 seconds), it should still receive notifications and maintain connection for about 5 minutes in the background.
However, in my testing, the connection seems to drop in under one minute, far sooner than expected.

PR #2341, opened 2025-11-05
Improved Focus handling on seeks

HaonRekcef:
Fixes issue reported on lichess: https://lichess.org/forum/team-lichess-beta-testers/mobile-app-alpha?page=16#156
The issue was that while seeking a game the screen could lock or the user could switch to another app without cancelling the seek, which led to game aborts.
Implemented fix: 
- Cancel seek when focus is lost and navigate back (the option would be to try to recreate the challenge or seek instead)
- Added a Wakelock to prevent the screen from timing out during a seek that take a long time.

### DIFF
--- a/lib/src/view/game/game_loading_board.dart
+++ b/lib/src/view/game/game_loading_board.dart
@@ -287,9 +287,10 @@ class LoadingPlayerWidget extends StatelessWidget {
 }
 
 class LoadGameError extends StatelessWidget {
-  const LoadGameError(this.errorMessage);
+  const LoadGameError(this.errorMessage, {this.showBackButton = false});
 
   final String errorMessage;
+  final bool showBackButton;
 
   @override
   Widget build(BuildContext context) {
@@ -311,7 +312,7 @@ class LoadGameError extends StatelessWidget {
           children: [
             BottomBarButton(
               onTap: () => Navigator.of(context).pop(),
-              label: context.l10n.cancel,
+              label: showBackButton ? 'Back' : context.l10n.cancel,
               icon: CupertinoIcons.xmark,
               showLabel: true,
             ),
--- a/lib/src/view/game/game_loading_board.dart
+++ b/lib/src/view/game/game_loading_board.dart
@@ -287,10 +287,10 @@ class LoadingPlayerWidget extends StatelessWidget {
 }
 
 class LoadGameError extends StatelessWidget {
-  const LoadGameError(this.errorMessage, {this.showBackButton = false});
+  const LoadGameError(this.errorMessage, {this.showBottomBar = true});
 
   final String errorMessage;
-  final bool showBackButton;
+  final bool showBottomBar;
 
   @override
   Widget build(BuildContext context) {
@@ -310,12 +310,13 @@ class LoadGameError extends StatelessWidget {
         ),
         BottomBar(
           children: [
-            BottomBarButton(
-              onTap: () => Navigator.of(context).pop(),
-              label: showBackButton ? 'Back' : context.l10n.cancel,
-              icon: CupertinoIcons.xmark,
-              showLabel: true,
-            ),
+            if (showBottomBar)
+              BottomBarButton(
+                onTap: () => Navigator.of(context).pop(),
+                label: context.l10n.cancel,
+                icon: CupertinoIcons.xmark,
+                showLabel: true,
+              ),
           ],
         ),
       ],
--- a/lib/src/view/game/game_screen.dart
+++ b/lib/src/view/game/game_screen.dart
@@ -14,6 +14,7 @@ import 'package:lichess_mobile/src/model/settings/board_preferences.dart';
 import 'package:lichess_mobile/src/model/settings/general_preferences.dart';
 import 'package:lichess_mobile/src/network/http.dart';
 import 'package:lichess_mobile/src/utils/duration.dart';
+import 'package:lichess_mobile/src/utils/focus_detector.dart';
 import 'package:lichess_mobile/src/utils/gestures_exclusion.dart';
 import 'package:lichess_mobile/src/utils/immersive_mode.dart';
 import 'package:lichess_mobile/src/utils/l10n_context.dart';
@@ -207,7 +208,27 @@ class _GameScreenState extends ConsumerState<GameScreen> {
               _ => const SizedBox.shrink(),
             },
           ),
-          body: PopScope(canPop: false, child: WakelockWidget(child: loadingBoard)),
+          body: PopScope(
+            canPop: false,
+            child: FocusDetector(
+              onFocusLost: () {
+                if (!mounted) return;
+                // Cancel the seek or challenge if the user leaves the screen while waiting, otherwise the game will be created but the user won't see it.
+                switch (widget.source) {
+                  case LobbySource():
+                    ref.read(createGameServiceProvider).cancelSeek();
+                    Navigator.of(context).pop();
+                  case UserChallengeSource():
+                    ref.read(createGameServiceProvider).cancelChallenge();
+                    Navigator.of(context).pop();
+                  case ExistingGameSource():
+                    break;
+                }
+              },
+              onFocusGained: () => {},
+              child: WakelockWidget(child: loadingBoard),
+            ),
+          ),
         );
     }
   }
--- a/lib/src/view/game/game_screen.dart
+++ b/lib/src/view/game/game_screen.dart
@@ -15,6 +15,7 @@ import 'package:lichess_mobile/src/model/settings/general_preferences.dart';
 import 'package:lichess_mobile/src/network/http.dart';
 import 'package:lichess_mobile/src/utils/duration.dart';
 import 'package:lichess_mobile/src/utils/gestures_exclusion.dart';
+import 'package:lichess_mobile/src/utils/immersive_mode.dart';
 import 'package:lichess_mobile/src/utils/l10n_context.dart';
 import 'package:lichess_mobile/src/utils/navigation.dart';
 import 'package:lichess_mobile/src/view/game/game_body.dart';
@@ -206,7 +207,7 @@ class _GameScreenState extends ConsumerState<GameScreen> {
               _ => const SizedBox.shrink(),
             },
           ),
-          body: PopScope(canPop: false, child: loadingBoard),
+          body: PopScope(canPop: false, child: WakelockWidget(child: loadingBoard)),
         );
     }
   }
--- a/lib/src/view/game/game_screen.dart
+++ b/lib/src/view/game/game_screen.dart
@@ -94,11 +94,9 @@ class _GameScreenState extends ConsumerState<GameScreen> {
     }
     switch (widget.source) {
       case LobbySource():
-        ref.read(gameScreenLoaderProvider(widget.source).notifier).cancelSeek();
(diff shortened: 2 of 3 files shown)


veloce:
Thanks for this. But I don't think we should do that.

When the app goes into background the socket connection remains active for 5 minutes, which means in theory the user is seeking a game for that long. Canceling the seek is certainly better than having aborted games, but it is counterintuitive as how an app should work.
So the real solution would be to send a notification when a game starts, as suggested in the forum. If I'm not mistaken I think it shouldn't be hard as we have already all the code to do that.
For the challenges I think we need as well https://github.com/lichess-org/mobile/pull/2244 (I need to finish that one).
About the Wakelock, I don't think it is needed if we properly send notifications.

ijm8710:
Hey guys,

I very much agree with veloces approach on this one, but I’m a bit confused.

Most times  i start to look for a game and multitask out of app quickly (say to answer a text) I am almost always greeted with the following error message 
IMG_0018
So essentially this sounds great, but I’m confused  why it’s assumed we maintain background connection for 5 minutes when I lose it within seconds.

perhaps this relates to when I send challenge requests (not correspondence but real time games) to myself and I lose the request within seconds as well. I really never see the connection maintained in the background. Am I doing something wrong? Or is it different on iOS?

HaonRekcef:
> When the app goes into background the socket connection remains active for 5 minutes, which means in theory the user is seeking a game for that long. 

I am not sure at all that this is the case. I think both OS pause background processes really quickly. For example: https://developer.apple.com/forums/thread/685525/ . So I dont think we can assume the user continues listening to the websocket once the app is backgrounded. 

I think a solution would be to get a push notification from lila when a game starts, which we will then only display if the user is currently not in the app, but I guess serverside changes would be needed too for this.

Or we could restart the seek automatically when the user returns the app to the foreground.

veloce:
Maintaining the websocket connection is not the same as executing code in background. It is true that platforms are somewhat conservative in allowing that. Especially on iOS.
You can test by yourself that the connection is really maintained: it is easy if you have 2 lichess accounts that follow each other. Just watch online friends from an account and put the app in background with the other, you'll see that your friend stays online for 5 minutes after.

It was important to allow the connection to remain active for playing games, otherwise lichess thinks you are disconnected and you may lose games quickly (for longer time controls it is useful).

It is not because the connection is maintained that you can do what you want while challenging real time or seeking a game. The game was *never* designed to allow this, so this is why you just can't do it. It is not a bug, that is just never been possible.

Now, ideally I think we should allow users to go in the background while searching for a game, as I said above. And if we want to do that, we'll probably have issues with the background execution restrictions, so it is not that simple.
There are ways to do it normally. Platforms authorise background code execution for some specific task, and ideally they should not run for a long time. In our case 5m doesn't seem too long.

This plugin is probably what we want: https://pub.dev/packages/workmanager

veloce:
> I think a solution would be to get a push notification from lila when a game starts, which we will then only display if the user is currently not in the app, but I guess serverside changes would be needed too for this.

I don't think push notifications are the good choice for that. Not reliable enough. Also Thibault probably don't want to do that change server side.

veloce:
> Or we could restart the seek automatically when the user returns the app to the foreground.

We can't do that as people will think the app is looking for a game while in background where actually it is not the case. Very confusing.

HaonRekcef:
> Maintaining the websocket connection is not the same as executing code in background. It is true that platforms are somewhat conservative in allowing that. Especially on iOS. You can test by yourself that the connection is really maintained: it is easy if you have 2 lichess accounts that follow each other. Just watch online friends from an account and put the app in background with the other, you'll see that your friend stays online for 5 minutes after.

I can not replicate. If I watch lichess tv on my Android 15 Phone and check with a browser on my laptop if I am online or not, it switches to offline within 20s of the app being backgrounded. Same for my iPad, there it is even within a few seconds.

veloce:
> I can not replicate. If I watch lichess tv on my Android 15 Phone and check with a browser on my laptop if I am online or not, it switches to offline within 20s of the app being backgrounded. Same for my iPad, there it is even within a few seconds.

that is not the way to do it; try with friends as I suggested. TV socket is closed automatically when the app is backgrounded.
or try locally with your emulator: start a game, put in background, watch in web from the other side you'll see that you are still connected;

ijm8710:
Veloce I had two related tickets with repro videos on these comments

Should friend status be preserved for several minutes if they’re not in active game?

https://github.com/lichess-org/mobile/issues/2150

should realtime challenge connection be sent if one has been in the app in past few minutes?

https://github.com/lichess-org/mobile/issues/2149

not sure if I’m misunderstanding how you are saying the websocket connection should work

HaonRekcef:
> that is not the way to do it; try with friends as I suggested. TV socket is closed automatically when the app is backgrounded. or try locally with your emulator: start a game, put in background, watch in web from the other side you'll see that you are still connected;

I still can not replicate, see the attached video:

https://github.com/user-attachments/assets/e7fcf886-3244-496a-85a0-270d164ccabd

Where the white account was me playing on my phone account and immediately after playing Qxf1 I switched to another app.

veloce:
Ok, that is interesting. Possibly a regression, thanks for reporting. That thing is tested at the socket level but it definitely should be tested at the game level.

Note that I have tested it with the friend online status and it works (on iOS).

HaonRekcef:
> Ok, that is interesting. Possibly a regression, thanks for reporting. That thing is tested at the socket level but it definitely should be tested at the game level.
> 
> Note that I have tested it with the friend online status and it works (on iOS).

Is there any version you want me to test? I tried with my iPad too and while I do not replicate the socket staying alive for 5min it is significantly longer(more than 1 min normally) than on my Android device.

veloce:
I have no clue about a version sorry. I'm also surprised that there is a difference between android and iOS here, especially since it stays longer on iOS as I would expect the opposite.
I'll also do tests myself because these things are tricky. Using dev tools and watch logs at the FINE level would definitely help to see what is happening in the socket.
You should be able to replicate the socket closing after 5m when in background and see the corresponding log message (so not talking about in game socket here).

veloce:
@HaonRekcef I confirm as well that on android the websocket shuts down only after a few seconds when the app is put into background. I have no idea why tbh; I am pretty sure that at the time when I first implemented this I had tested that and it was working on android, unless I overlooked android... To investigate...

I don't know what is shutting the connection. I doubt that is at the OS level, because in the previous lichess app socket was definitely staying alive in background on android.

ijm8710:
> especially since it stays longer on iOS as I would expect the opposite.


Veloce, are you sure it's not an issue on iOS as well

I had created 2 separate issue tickets with videos for each scenario

Would you have expected web socket to apply for either of these where connection was lost in one minute or less 

https://github.com/lichess-org/mobile/pull/2341#issuecomment-3503256705

HaonRekcef:
Updated the PR, needs a thorough review as it is quite critical
- I switched to a `AppLifecycleListener` instead of the visibility detector, as this seems more in line with what we’re trying to achieve. I think `onPause `is what we need (`onHide`  is the other option) and it worked as expected on my Android 15 Phone and on the Android 16 Emulator, but this should be tested on iOS and on Android Versions <15 too.
- Displays that the seek/challenge was cancelled on the screen to give feedback to the user 
- Keeps screen active while seeking with a wakelock

lib/src/view/game/game_screen.dart
```diff
@@ -68,12 +69,69 @@ class _GameScreenState extends ConsumerState<GameScreen> {
...
+    if (!mounted) return;
+    final loader = ref.read(gameScreenLoaderProvider(widget.source));
+    // Only cancel if we are still seeking or waiting for challenge to be accepted
+    if (loader is! AsyncLoading<GameScreenState>) {
+      return;
+    }
+    switch (widget.source) {
+      case LobbySource():
+        ref.read(gameScreenLoaderProvider(widget.source).notifier).cancelSeek();
+        await ref.read(createGameServiceProvider).cancelSeek();
```
veloce: Better to call the service from the `GameScreenLoader.cancelSeek`.

lib/src/view/game/game_loading_board.dart
```diff
@@ -311,7 +312,7 @@ class LoadGameError extends StatelessWidget {
           children: [
             BottomBarButton(
               onTap: () => Navigator.of(context).pop(),
-              label: context.l10n.cancel,
+              label: showBackButton ? 'Back' : context.l10n.cancel,
```
veloce: The label shows `Back` but the icon is still a cross, this is confusing. Better remove this change and see the other comment.

lib/src/view/game/game_screen.dart
```diff
@@ -68,12 +69,69 @@ class _GameScreenState extends ConsumerState<GameScreen> {
...
   @override
   Widget build(BuildContext context) {
     final boardPreferences = ref.watch(boardPreferencesProvider);
 
     switch (ref.watch(gameScreenLoaderProvider(widget.source))) {
+      case AsyncData(value: SeekCancelledState()):
+        return Scaffold(
+          resizeToAvoidBottomInset: false,
+          appBar: AppBar(
+            leading: const SocketPingRatingIcon(),
```
veloce: Instead of showing the back button in the bottom bar, remove this widget to show it in the top bar as it should be.

lib/src/view/game/game_screen.dart
```diff
@@ -68,12 +69,69 @@ class _GameScreenState extends ConsumerState<GameScreen> {
...
+            title: switch (widget.source) {
+              LobbySource(:final seek) => _LobbyGameTitle(seek: seek),
+              _ => const SizedBox.shrink(),
+            },
+          ),
+          body: const LoadGameError('The game search was cancelled.', showBackButton: true),
```
veloce: Remove the `showBackButton` parameter.

HaonRekcef:
Addressed the review comments.
I dont think we should have a cancel button when the text says `game search has been cancelled` so I removed the button.

lib/src/view/game/game_screen_providers.dart
```diff
@@ -132,12 +132,14 @@ class GameScreenLoader extends _$GameScreenLoader {
     state = AsyncValue.data(GameCreatedState(id));
   }
 
-  void cancelSeek() {
+  Future<void> cancelSeek() async {
     state = const AsyncValue.data(SeekCancelledState());
+    await ref.read(createGameServiceProvider).cancelSeek();
```
veloce: Here we have to deal with the loading state. Unfortunately we cannot use Mutations yet since we are not using riverpod 3.0.

I overlooked this when I said to move all logic in the GameScreenLoader... now I think it would be awkward to add a new state for this in the GameScreenState.
I don't see a better way to move the loading logic as a local state of `_GameScreenState` (outside riverpod), which means you need to call again the `createGameServiceProvider` from the game screen.
Sorry for the confusion.

lib/src/view/game/game_screen.dart
```diff
@@ -113,24 +111,24 @@ class _GameScreenState extends ConsumerState<GameScreen> {
         return Scaffold(
           resizeToAvoidBottomInset: false,
           appBar: AppBar(
-            leading: const SocketPingRatingIcon(),
+            leading: const BackButton(),
```
veloce: It is better to let the framework handle the leading widget automatically by just removing it. Or am I missing something that makes the `BackButton()` required here?

PR #2344, opened 2025-11-07
added: unlimited switch in otb config bottom sheet for unlimited time game

### LINKED ISSUE #2315: Support OTB unlimited time control option

nirajk14:
Closes #2315

### DIFF
--- a/lib/src/model/common/time_increment.dart
+++ b/lib/src/model/common/time_increment.dart
@@ -14,7 +14,9 @@ class TimeIncrement implements Comparable<TimeIncrement> {
 
   const TimeIncrement.infinite() : time = 0, increment = 0;
 
-  TimeIncrement.blitzDefault() : time = const Duration(minutes: 5).inSeconds, increment = const Duration(seconds: 3).inSeconds;
+  TimeIncrement.blitzDefault()
+    : time = const Duration(minutes: 5).inSeconds,
+      increment = const Duration(seconds: 3).inSeconds;
 
   /// Clock initial time in seconds
   final int time;
--- a/lib/src/model/over_the_board/over_the_board_clock.dart
+++ b/lib/src/model/over_the_board/over_the_board_clock.dart
@@ -39,9 +39,7 @@ class OverTheBoardClock extends _$OverTheBoardClock {
       }
     });
 
-    return OverTheBoardClockState.fromTimeIncrement(
-      TimeIncrement.blitzDefault(),
-    );
+    return OverTheBoardClockState.fromTimeIncrement(TimeIncrement.blitzDefault());
   }
 
   void setupClock(TimeIncrement timeIncrement) {


nirajk14:
Couldn't quite decide if I wanted to put the "Unlimited --- Switch" layout at bottom or top of minutes per side, but decided to put at bottom. Also the text should I make "Unlimited time" or is "Unlimited" enough? I am open to any suggestions and feedback. Thank you.

ijm8710:
Should it follow the same layout as “challenge a friend”

I would probably mimic it completely imo? (Except for the rated toggle and correspondence control option )

- uses “Unlimited”
- Variant is below time control 

IMG_0042
IMG_0043
IMG_0044

nirajk14:
> Should it follow the same layout as “challenge a friend”
> 
> I would probably mimic it completely imo? (Except for the rated toggle and correspondence control option )
> 
> * uses “Unlimited”
> * Variant is below time control
> 
> IMG_0042 IMG_0043 IMG_0044

This seems like a great idea and I think it'd be more consistent too. I think I'd just not have the Rated part and remove Correspondence Time Control option and From Position option for Variant. Those features if to be developed for otb should be separate PR. I will work on it and keep a branch. I'd also like to know what others think.

nirajk14:
> > Should it follow the same layout as “challenge a friend”
> > I would probably mimic it completely imo? (Except for the rated toggle and correspondence control option )
> > 
> > * uses “Unlimited”
> > * Variant is below time control
> > 
> > IMG_0042 IMG_0043 IMG_0044
> 
> This seems like a great idea and I think it'd be more consistent too. I think I'd just not have the Rated part and remove Correspondence Time Control option and From Position option for Variant. Those features if to be developed for otb should be separate PR. I will work on it and keep a branch. I'd also like to know what others think.

After further thought, I found out if I were to make this change there will be 1 more click involved as the user will have to open a whole new dialog just to select unlimited or clock. As for my current approach a single switch is enough as it's only 2 option. I also think correspondence in otb isn't going to happen so unless there are more than 2 options i.e clock and unlimited. I think the current approach of simple switch is optimal. While I do think it'd be more consistent in a pure layout point of view but at the cost of UX isn't something that feels right. So I think this PR should function well without those changes.

ijm8710:
My only retort would be that generally a user will probably have consistent behavior.
Meaning someone who often plays otb will usually lean specifically clock or specifically unlimited rather than switch back and forth. Meaning that even though it's one extra click, you usually don't have to make any clicks to change it once you choose your preference.

This would mean it would be helpful though if it is able to save your prior state of time control specifics if possible (which I was already thinking of ticketing separately anyway as it's a little unnecessary that if someone always plays 3+0 that it keeps jumping to 5+3 rather than retains).

But obviously it would be nice to hear what Veloce thinks for the overall layout for how to handle switching between unlimited and clock:

- a menu prompt similar to how the same screen exists when challenging a friend 
- a unique toggle to switch

nirajk14:
Screenshot_20251120_212141
Screenshot_20251120_212149
Screenshot_20251107_213313
Screenshot_20251120_212120


Here are some screenshots of the changes.

veloce:
Sorry for the long delay on this. I kept your PR but changed the switch to a time control selector.

nirajk14:
Thanks for the merge.

PR #2355, opened 2025-11-09
fix: false positive for alternative castling notation in puzzles

### LINKED ISSUE #2354: Puzzle mode stuck waiting for CPU move
I've just gotten the android version in a state where you can't advance in a puzzle.

Here is the starting board:

It's white to play. I moved qa8+, which takes us to this board:

It says keep going, but I can't, because the app hasn't taken black's move.

If I tap "get a hint" it circles the black queen, (presumably with qb7 in mind to block my check). But that's black's move to make, not mine!

This is the November 7th build from Google play on a Samsung a34 5g.

The bug survives app restarts. Could be unique to this puzzle. Is there a way to get an ID of the puzzle to show you?

Thanks

tom-anders:
Plus some refactoring for `SanMove`

Fixes #2354

### DIFF
--- a/lib/src/model/puzzle/puzzle.dart
+++ b/lib/src/model/puzzle/puzzle.dart
@@ -31,7 +31,7 @@ sealed class Puzzle with _$Puzzle {
       if (sanMove.isCheckmate) {
         return true;
       }
-      if (uci != solutionUci && (!altCastles.containsKey(uci) || altCastles[uci] != solutionUci)) {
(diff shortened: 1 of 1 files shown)


tom-anders:
This also fixes a (very rare) bug in study and analysis board where alt castling notation would not be correctly applied for `O-O+` or `O-O#`, since we'd look for an exact string match with `O-O` - changed it to use `startsWith` now

PR #2370, opened 2025-11-16
upstream julien's f-droid branch

tom-anders:
We agreed that it's better to have this branch in the main repository. 

@veloce can you create a `fdroid` branch so that I can use that as the target of this PR?

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -7,7 +7,7 @@ version: 0.18.1+001801 # See README.md for details about versioning
 environment:
   # this should be updated with every new flutter stable release
   sdk: ^3.9.0
-  flutter: ^3.35.0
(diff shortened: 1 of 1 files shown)


tom-anders:
f-droid inclusion MR: https://gitlab.com/fdroid/fdroiddata/-/merge_requests/29372

Once we have the `froid` branch and push the first tag to it, the f-droid MR's CI should pass

tom-anders:
From f-droid code review: f-droid wants fastlane in one of 3 specific locations: https://gitlab.com/-/snippets/1895688#locations

We currently have `android/fastlane`, would there be a problem with moving it to `fastlane/metadata/android/` instead? If not, should probably be done on the main branch as well to be consistent

PR #2392, opened 2025-11-24
feat: add engine eval display in analysis for tablet landscape mode

lumiknit:
Hello.

I enjoy using Lichess on my tablet. However, I was unable to check the engine evaluation when using the tablet in landscape mode during analysis.

| Portrait | Landscape |
| :-: | :-: |
|  |  |

So, I modified it to look like the following.

| White | Black |
| :-: | :-: |
|  |  |

The rotated text might be difficult to read slightly, but I think it is a decent choice because there are cases where long words (e.g., Checkmate) appear.

Thank you.


veloce:
Thanks for this, but it is now longer needed with https://github.com/lichess-org/mobile/pull/2400

PR #2398, opened 2025-11-27
Team Tournament l10n changes

HaonRekcef:


veloce:
There are strange changes here. Why do I see `\n` replaced by `\r\n`?

I'd like to see only the team tournament changes if possible. Ideally the script should produce the same output regardless of the environment.

veloce:
Also how many strings do you use from the `team` Lila module? If only a few, please use the a white list: https://github.com/lichess-org/mobile/blob/4a91c903fe8324b3809a51d7d810a014bb946a8a/scripts/gen-arb.mjs#L45

HaonRekcef:
> There are strange changes here. Why do I see `\n` replaced by `\r\n`?
> 
> I'd like to see only the team tournament changes if possible. Ideally the script should produce the same output regardless of the environment.

Indeed, did it in on Windows and it screwed up the line endings, I added a simple workaround now.


> Also how many strings do you use from the `team` Lila module? If only a few, please use the a white list:
> 
> https://github.com/lichess-org/mobile/blob/4a91c903fe8324b3809a51d7d810a014bb946a8a/scripts/gen-arb.mjs#L45

did that already 👍

PR #2419, opened 2025-12-02
Add a workflow to build and upload F-Droid APKs as a Github artifact

julien4215:
I tried to add this workflow on the fdroid branch but it didn't appear in the Github interface so I think we have to add it on the main branch.

This was asked by the F-Droid team (https://gitlab.com/fdroid/fdroiddata/-/merge_requests/29372#note_2927842381) to check if it is easy to have reproducible builds.

### DIFF
--- a/.github/workflows/upload_fdroid_github.yml
+++ b/.github/workflows/upload_fdroid_github.yml
@@ -8,6 +8,7 @@ permissions: read-all
 jobs:
   github-upload:
     runs-on: ubuntu-latest
+    if: github.ref == 'refs/heads/fdroid'  # only runs if on the 'fdroid' branch
     steps:
       - name: Set up JDK 17
         uses: actions/setup-java@v5


tom-anders:
Hmm that build would include the firebase dependencies though, so I think it wouldn't be helpful for fdroid. 

does [this](https://stackoverflow.com/questions/61891328/trigger-github-action-only-on-new-tags) do what we want instead? Maybe veloce also has to enable workflow runs for branches other than main somewhere in the repo settings

tom-anders:
Ahh, apparently workflows on other branches only work if the main branch also has the same workflow (https://stackoverflow.com/questions/67707244/how-to-test-github-workflow-without-merging-into-master-main-branch#comment138662509_77882618), nevermind my other comment then

julien4215:
This workflow uses `workflow_dispatch` so it is triggered manually. It should be possible to trigger it on tags.

veloce:
What is the goal of this? I mean for f-droid? Because I don't see why we need to upload the f-droid APK to GitHub.

PR #2451, opened 2025-12-19
potential fix for notification issues on iOS

tom-anders:
getAPNSToken() is the only iOS specific thing that happens in this file, so I think there's a high probabillity that the issues are related to that.

According to
https://firebase.google.com/docs/cloud-messaging/flutter/get-started we need to get permission *before* calling getAPNSToken(). Currently it looks like there could be a race condition where the connectivity change listener fires before we call requestPermission(). Fix this by moving the ref.listen() below the requestPermission() call.


veloce:
thanks!

PR #2452, opened 2025-12-21
Fix: reset puzzle streak correctly when failing the first puzzle

sam-baumann:
#1951 had a previous fix but issue has returned.

Previous fix checked if the puzzle id is the same or has changed, but there seems to be a separate issue where the same puzzle streak re-plays sometimes if you fail early. I logged that as a separate issue at #2455 

I replaced the puzzle id check with a check for if the value of `finished` is going from true to false. Also wrote a unit test to prevent further regression

### DIFF
--- a/lib/src/view/puzzle/streak_screen.dart
+++ b/lib/src/view/puzzle/streak_screen.dart
@@ -537,7 +537,6 @@ class _BottomBar extends ConsumerWidget {
             highlighted: true,
             label: context.l10n.puzzleNewStreak,
             icon: CupertinoIcons.play_arrow_solid,
-            key: const Key('puzzle-new-streak-button'), //added for quick access in unit test
           ),
       ],
     );
--- a/test/view/puzzle/streak_screen_test.dart
+++ b/test/view/puzzle/streak_screen_test.dart
@@ -339,7 +339,7 @@ void main() {
       //game over correctly appears
       expect(find.text('GAME OVER'), findsOneWidget);
 
-      final button = find.byKey(const Key('puzzle-new-streak-button'));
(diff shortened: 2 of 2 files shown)


lib/src/view/puzzle/streak_screen.dart
```diff
@@ -537,6 +537,7 @@ class _BottomBar extends ConsumerWidget {
             highlighted: true,
             label: context.l10n.puzzleNewStreak,
             icon: CupertinoIcons.play_arrow_solid,
+            key: const Key('puzzle-new-streak-button'), //added for quick access in unit test
```
veloce: You should be able to find the button just fine with the label, so you don't need the key (all the other parts of the app tend to avoid using key for test only purposes)

sam-baumann: removed the key

sam-baumann:
> I'm not sure to understand why there was a regression, and the fix. Could you explain more? thank.

@veloce 
The previous fix only applied when the previous puzzle is different from the new one being loaded. It did a check for `previous.puzzle.id != next.puzzle.id` in a listener on `puzzleStreakControllerProvider`

It seems that if you fail a puzzle streak very soon after starting, the lichess API will send back the exact same puzzle streak. This behavior is reproducible on https://lichess.org/streak 

When you fail on the first puzzle of the streak, and the server sends back the exact same streak, it causes the previous fix for this issue not to work, because the puzzle you failed on is the same as the one it's trying to load.

My fix replaces the `puzzle.id` equivalence check with a check for `previous.streak.finished == true && next.streak.finished == false`, as this condition will always be true after hitting the 'new streak' button.

PR #2463, opened 2025-12-27
feat: add log screen

tom-anders:
[logs.webm](https://github.com/user-attachments/assets/b59679ad-cb82-4754-ad51-219e8a8878a6)

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -8,7 +8,6 @@ targets:
       freezed:
         generate_for:
           - lib/src/model/**/*.dart
-          - lib/src/log.dart
           - lib/src/**/*_models.dart
           - lib/src/**/*_providers.dart
         options:
--- a/lib/src/log.dart
+++ b/lib/src/log.dart
@@ -1,29 +1,32 @@
 import 'dart:developer' as developer;
 
-import 'package:fast_immutable_collections/fast_immutable_collections.dart';
 import 'package:flutter/foundation.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:lichess_mobile/src/model/settings/log_preferences.dart';
+import 'package:lichess_mobile/src/utils/lru_list.dart';
 import 'package:logging/logging.dart';
 
 const _loggersToShowInTerminal = {'HttpClient', 'Socket', 'EvaluationService'};
 
-typedef LogState = ({IList<LogRecord> logs});
-
-/// Provides an instance of [AppLogStorage] using Riverpod.
-final appLogStorageProvider = NotifierProvider<AppLogStorage, LogState>(
-  AppLogStorage.new,
-  name: 'AppLogStorageProvider',
+/// Provides an instance of [AppLogStorageService] using Riverpod.
+final appLogStorageServiceProvider = Provider<AppLogStorageService>(
+  (Ref ref) => AppLogStorageService(ref),
+  name: 'AppLogStorageServiceProvider',
 );
 
 /// Manages log entries created via [Logger] instances
 ///
-/// Currently, simply saves log entries in memory, so they do not persists across app restarts.
-class AppLogStorage extends Notifier<LogState> {
-  @override
-  LogState build() => (logs: const IList.empty());
+/// Currently, simply saves the most recent log entries in memory, so they do not persists across app restarts.
+class AppLogStorageService {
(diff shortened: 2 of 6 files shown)


lib/src/view/settings/app_log_settings_screen.dart
```diff
@@ -0,0 +1,96 @@
...
+              ),
+            ],
+          ),
+          if (logs.isNotEmpty)
+            Expanded(
+              child: Padding(
+                padding: Styles.bodySectionPadding,
+                child: Card(
+                  margin: EdgeInsets.zero,
+                  child: SingleChildScrollView(
```
veloce: This is very inefficient, especially since the logs can be large. I know the scroll is not that important as it is meant to be exported but still.

You should use a `ListView.builder` here and each log entry could be formatted a little with the time and the log level and logger name stand out somehow (with colors for instance).

lib/src/log.dart
```diff
@@ -1,28 +1,63 @@
...
+
+      _save(record);
+    });
+  }
+
+  void clear() {
+    state = (logs: const IList.empty());
+  }
+
+  void _save(LogRecord record) {
```
veloce: Ok, let's not use a Provider here to avoid this.

I don't think we need to watch changes here, we just want to display the last snapshot. So you can convert the notifier to a service that can expose its current state through a method.

build.yaml
```diff
@@ -8,6 +8,7 @@ targets:
       freezed:
         generate_for:
           - lib/src/model/**/*.dart
+          - lib/src/log.dart
```
veloce: You don't need that, there's nothing to generate here.

lib/src/log.dart
```diff
@@ -1,28 +1,63 @@
...
+  AppLogStorage.new,
+  name: 'AppLogStorageProvider',
+);
+
+/// Manages log entries created via [Logger] instances
+///
+/// Currently, simply saves log entries in memory, so they do not persists across app restarts.
+class AppLogStorage extends Notifier<LogState> {
+  @override
+  LogState build() => (logs: const IList.empty());
```
veloce: I am a little worried about memory usage here. This list can grow fast and it could be an issue for long app sessions, especially on cheap devices.

We should use a list with a capped size and LRU mechanism. Should be easy to implement one with a [LinkedList](https://api.dart.dev/dart-collection/LinkedList-class.html).

Something like that:
```dart
final class LRUListEntry<T> extends LinkedListEntry<LRUListEntry<T>> {
  final T value;

  LRUListEntry(this.value);
}

class LRUList<T> {
  final int capacity;
  final LinkedList<LRUListEntry<T>> _list = LinkedList<LRUListEntry<T>>();

  LRUList({required this.capacity});

  void put(T value) {
    if (_list.length >= capacity) {
      _list.remove(_list.first);
    }
    _list.add(LRUListEntry<T>(value));
  }

  Iterable<T> get values => _list.map((entry) => entry.value);
}
```

lib/src/model/settings/log_preferences.dart
```diff
@@ -0,0 +1,65 @@
...
+
+  @override
+  LogPrefs build() => fetch();
+
+  Future<void> setLogLevel(Level level) {
+    return save(state.copyWith(level: level));
+  }
+}
+
+const _kDefaultLevel = kDebugMode ? Level.FINE : Level.OFF;
```
veloce: Not sure about this. But default log level could be sth else than OFF?

lib/src/view/settings/app_log_settings_screen.dart
tom-anders: Done, I based it on the formatting from the http log screen: 

Screenshot_1766920543

build.yaml
tom-anders: Ah yes, I had a @freezed in there at some point, but then forgot to adapt build.yaml after removing it

lib/src/log.dart
tom-anders: Done, added it to `lru_list.dart` :+1: capacity for the log screen is hardcoded to 1024 now

lib/src/model/settings/log_preferences.dart
tom-anders: Changed it to `WARNING` now

veloce (approved):

PR #2470, opened 2025-12-29
feat: persist ongoing OTB games

### LINKED ISSUE #2466: Game state is lost in "over the board" mode after iOS app is suspended too long
### What is the problem?

Observed on iOS in Lichess v0.18.2.

If you start an over the board game, then suspend the app (e.g. lock your phone for some time or switch apps), after an hour or so when the app is resumed, it returns to the home screen and the game state is lost.

This seems to happen pretty consistently and I've observed it a few times now.

Please let me know if you need any more information. Thank you!

1. Start an "over the board" game
2. Lock the screen or switch apps for a couple hours
3. Return to lichess
4. Game state is lost, app reloads at the home screen

### App version

v0.18.2

### Device Info

iOS 26.1

### Additional information

_No response_

tom-anders:
Closes #2466

[otb.webm](https://github.com/user-attachments/assets/6fe33a5a-0785-42a2-b3e8-19e5c9a3606d)

### DIFF
--- a/lib/src/view/over_the_board/over_the_board_screen.dart
+++ b/lib/src/view/over_the_board/over_the_board_screen.dart
@@ -148,8 +148,18 @@ class _BodyState extends ConsumerState<_Body> {
       }
     });
 
-    final clockState = ref.read(overTheBoardClockProvider);
-    final otbGameStorage = ref.read(overTheBoardGameStorageProvider);
+    void saveGameState() {
+      if (!context.mounted) return;
+      final clockState = ref.read(overTheBoardClockProvider);
+      ref
+          .read(overTheBoardGameStorageProvider)
+          .save(
+            gameState.game,
+            timeIncrement: clockState.timeIncrement,
+            whiteTimeLeft: clockState.whiteTimeLeft,
+            blackTimeLeft: clockState.blackTimeLeft,
+          );
+    }
 
     return WakelockWidget(
       child: PopScope(
@@ -176,7 +186,10 @@ class _BodyState extends ConsumerState<_Body> {
                 title: Text(context.l10n.mobileAreYouSure),
                 content: const Text('No worries, your game will be saved.'),
                 onNo: () => Navigator.of(context).pop(false),
-                onYes: () => Navigator.of(context).pop(true),
+                onYes: () {
+                  saveGameState();
+                  Navigator.of(context).pop(true);
+                },
               );
             },
           );
@@ -187,14 +200,7 @@ class _BodyState extends ConsumerState<_Body> {
           }
         },
         child: FocusDetector(
-          onFocusLost: () {
-            otbGameStorage.save(
-              gameState.game,
-              timeIncrement: clockState.timeIncrement,
-              whiteTimeLeft: clockState.whiteTimeLeft,
-              blackTimeLeft: clockState.blackTimeLeft,
-            );
-          },
+          onForegroundLost: saveGameState,
           child: Column(
             children: [
               Expanded(
--- a/test/view/over_the_board/over_the_board_screen_test.dart
+++ b/test/view/over_the_board/over_the_board_screen_test.dart
@@ -4,6 +4,7 @@ import 'package:chessground/chessground.dart';
 import 'package:dartchess/dartchess.dart';
 import 'package:fast_immutable_collections/fast_immutable_collections.dart';
 import 'package:flutter/cupertino.dart';
+import 'package:flutter/material.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:flutter_test/flutter_test.dart';
 import 'package:lichess_mobile/src/model/common/chess.dart';
@@ -16,6 +17,7 @@ import 'package:lichess_mobile/src/model/game/over_the_board_game.dart';
 import 'package:lichess_mobile/src/model/over_the_board/over_the_board_clock.dart';
 import 'package:lichess_mobile/src/model/over_the_board/over_the_board_game_controller.dart';
 import 'package:lichess_mobile/src/model/over_the_board/over_the_board_game_storage.dart';
+import 'package:lichess_mobile/src/utils/navigation.dart';
 import 'package:lichess_mobile/src/view/over_the_board/over_the_board_screen.dart';
 import 'package:lichess_mobile/src/widgets/clock.dart';
 import 'package:mocktail/mocktail.dart';
@@ -216,6 +218,13 @@ void main() {
                 ),
                 sanMove: SanMove('e4', Move.parse('e2e4')!),
               ),
+              GameStep(
+                position: Position.setupPosition(
+                  Rule.chess,
+                  Setup.parseFen('rnbqkbnr/pppp1ppp/8/4p3/4P3/8/PPPP1PPP/RNBQKBNR w KQkq - 0 2'),
+                ),
+                sanMove: SanMove('e5', Move.parse('e7e5')!),
+              ),
             ].lock,
             meta: GameMeta(
               createdAt: DateTime.now(),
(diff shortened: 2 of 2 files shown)


ijm8710:
This is cool, does it persist if say battery dies mid game

tom-anders:
> This is cool, does it persist if say battery dies mid game

Currently it only saves when the OTB screen loses focus (either by closing the screen or by closing the app entirely). I think if we run out of battery there would be no time to save (phone would be dead immediately).

Could be fixed by saving the game at every move instead, but I decided against this because it's a lot of unnecessary I/O to write to the file at every move, especially with faster time controls.

lib/src/view/over_the_board/over_the_board_screen.dart
```diff
@@ -166,69 +186,79 @@ class _BodyState extends ConsumerState<_Body> {
...
-                      });
-                    },
+        child: FocusDetector(
+          onFocusLost: () {
+            otbGameStorage.save(
+              gameState.game,
+              timeIncrement: clockState.timeIncrement,
+              whiteTimeLeft: clockState.whiteTimeLeft,
+              blackTimeLeft: clockState.blackTimeLeft,
+            );
```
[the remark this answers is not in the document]
tom-anders: Not really neccessary, because when loading a saved game we use it only if it's not already finished

lib/src/view/over_the_board/over_the_board_screen.dart
```diff
@@ -66,8 +68,23 @@ class _BodyState extends ConsumerState<_Body> {
...
+      if (ongoingGame != null && ongoingGame.game.steps.length > 1 && !ongoingGame.game.finished) {
+        ref.read(overTheBoardGameControllerProvider.notifier).loadOngoingGame(ongoingGame.game);
+
+        ref
+            .read(overTheBoardClockProvider.notifier)
+            .setupClock(
+              ongoingGame.timeIncrement,
+              whiteTimeLeft: ongoingGame.whiteTimeLeft,
+              blackTimeLeft: ongoingGame.blackTimeLeft,
+            );
```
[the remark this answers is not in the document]
tom-anders: the `mounted` check is only necessary when using `context`

lib/src/view/over_the_board/over_the_board_screen.dart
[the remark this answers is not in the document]
tom-anders: Not possible, because using `ref` in `onFocusLost` throws a riverpod exception, so we have to live with this limitation I think

lib/src/model/over_the_board/over_the_board_game_storage.dart
```diff
@@ -0,0 +1,92 @@
...
+}, name: 'OverTheBoardGameStorageProvider');
+
+const kOtbGameFileName = 'otb_game.json';
+
+class OverTheBoardGameStorage {
+  const OverTheBoardGameStorage(this.ref);
+  final Ref ref;
+
+  Future<File> _getFile() async {
+    final dir = await getApplicationCacheDirectory();
```
veloce: It would be better in the `getApplicationSupportDirectory` IMO.

lib/src/view/over_the_board/over_the_board_screen.dart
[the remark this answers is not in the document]
veloce: Are you sure about that? I don't see any reason why it would throw an exception in `onFocusLost`; the latter is not a riverpod lifecycle callback, and the stateful widget should not be unmounted while the focus is lost.

lib/src/view/over_the_board/over_the_board_screen.dart
```diff
@@ -131,6 +148,9 @@ class _BodyState extends ConsumerState<_Body> {
       }
     });
 
+    final clockState = ref.read(overTheBoardClockProvider);
```
veloce: Don't use ref.read in `build()`;

If this is meant to be used in a callback, then use `ref.read` directly in the callback.

tom-anders: As mentioned in my response to Copilot, this throws an exception in tests (In the emulator it only happens ~50% of the time):

```
00:07 +0: Playing over the board (offline) Checkmate and Rematch
══╡ EXCEPTION CAUGHT BY SCHEDULER LIBRARY ╞═════════════════════════════════════════════════════════
The following StateError was thrown during a scheduler callback:
Bad state: Using "ref" when a widget is about to or has been unmounted is unsafe.
Ref relies on BuildContext, and BuildContext is unsafe to use when the widget is deactivated.
To safely refer to the state of providers inside State.dispose(), save the provider state in a field
of your State class.

When the exception was thrown, this was the stack:
#0      ConsumerStatefulElement._assertNotDisposed (package:flutter_riverpod/src/core/consumer.dart:432:7)
#1      ConsumerStatefulElement.read (package:flutter_riverpod/src/core/consumer.dart:511:5)
#2      _BodyState.build.<anonymous closure> (package:lichess_mobile/src/view/over_the_board/over_the_board_screen.dart:189:28)
#3      _FocusDetectorState._notifyFocusLoss (package:lichess_mobile/src/utils/focus_detector.dart:142:25)
#4      _FocusDetectorState._notifyVisibilityStatusChange (package:lichess_mobile/src/utils/focus_detector.dart:129:7)
#5      _FocusDetectorState.build.<anonymous closure> (package:lichess_mobile/src/utils/focus_detector.dart:106:7)
#6      RenderVisibilityDetectorBase._fireCallback (package:visibility_detector/src/render_visibility_detector.dart:87:26)
#7      RenderVisibilityDetectorBase._scheduleUpdate.<anonymous closure> (package:visibility_detector/src/render_visibility_detector.dart:150:7)
#8      RenderVisibilityDetectorBase._processCallbacks (package:visibility_detector/src/render_visibility_detector.dart:62:15)
#9      RenderVisibilityDetectorBase._scheduleUpdate.<anonymous closure> (package:visibility_detector/src/render_visibility_detector.dart:162:11)
#10     SchedulerBinding._invokeFrameCallback (package:flutter/src/scheduler/binding.dart:1434:15)
#11     SchedulerBinding.handleDrawFrame (package:flutter/src/scheduler/binding.dart:1361:11)
#12     AutomatedTestWidgetsFlutterBinding.scheduleWarmUpFrame (package:flutter_test/src/binding.dart:1432:5)
#13     _runWidget (package:flutter/src/widgets/binding.dart:1643:7)
#14     runApp (package:flutter/src/widgets/binding.dart:1576:3)
#15     TestWidgetsFlutterBinding._runTestBody (package:flutter_test/src/binding.dart:1066:7)
<asynchronous suspension>
<asynchronous suspension>
(elided one frame from package:stack_trace)
```

lib/src/view/over_the_board/over_the_board_screen.dart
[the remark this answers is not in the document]
tom-anders: (see my other comment containing a stack trace)

lib/src/view/over_the_board/over_the_board_screen.dart
veloce: you need to check if the widget is still mounted before using `ref`

veloce: if the widget is unmounted during `onFocusLost` and it is not normal you should try to find the reason why

tom-anders: I couldn't really figure out why sometimes this would happen, Claude suggested it's probably a race condition between the flutter framework and the `visibility_detector` package we're using.

It only happens when I exit the game via the back-button, but not if I put the app into the background.

I now changed the callback from `onFocusLost` to `onForegroundLost` (to handle the case where the user puts the app into the background, and at some point the app would be killed because of low RAM) and added another callback instead when the user presses the `Yes` Button of the confirmation dialog

veloce: iirc `onFocusLost` also handle the case where the user puts the app into background; will try to debug it myself when I have the time

veloce:
I rebased this to test it @tom-anders

PR #2496, opened 2026-01-08
Fix multiple analysis session issues

HaonRekcef:
Fixes:
- Opening a standalone analysis session from puzzle or load position works now even when there is a saved analysis. Previously it just loaded the saved analysis instead.
- Jumps to the same path that was previously being analyzed
- No orange box highlighting the last move of the analysis in standalone analysis

### DIFF
--- a/lib/src/model/analysis/analysis_controller.dart
+++ b/lib/src/model/analysis/analysis_controller.dart
@@ -244,8 +244,7 @@ class AnalysisController extends AsyncNotifier<AnalysisState>
     final List<Future<(UciPath, FullOpening)?>> openingFutures = [];
 
     _root = switch (options) {
-      Standalone(:final pgn) when _savedStandalone != null && pgn.isEmpty =>
-        _savedStandalone!.root,
+      Standalone(:final pgn) when _savedStandalone != null && pgn.isEmpty => _savedStandalone!.root,
       _ => Root.fromPgnGame(
         game,
         isLichessAnalysis: options.isLichessGameAnalysis,
@@ -285,8 +284,7 @@ class AnalysisController extends AsyncNotifier<AnalysisState>
         });
 
     final currentPath = switch (options) {
-      Standalone(:final pgn) when _savedStandalone != null && pgn.isEmpty =>
-        _savedStandalone!.path,
+      Standalone(:final pgn) when _savedStandalone != null && pgn.isEmpty => _savedStandalone!.path,
       _ => options.initialMoveCursor == null ? _root.mainlinePath : path,
     };
     final currentNode = _root.nodeAt(currentPath);


lib/src/model/analysis/analysis_controller.dart
```diff
@@ -111,9 +111,11 @@ final analysisControllerProvider = AsyncNotifierProvider.autoDispose
     );
 
 Root? _savedStandaloneRoot;
+UciPath? _savedStandalonePath;
```
veloce: Let's use a `({ Root root, UciPath path })? _savedStandalone;` instead.

lib/src/model/analysis/analysis_controller.dart
```diff
@@ -243,7 +247,8 @@ class AnalysisController extends AsyncNotifier<AnalysisState>
     final List<Future<(UciPath, FullOpening)?>> openingFutures = [];
 
     _root = switch (options) {
-      Standalone() when _savedStandaloneRoot != null => _savedStandaloneRoot!,
+      Standalone(:final pgn) when _savedStandaloneRoot != null && pgn.isEmpty =>
```
veloce: It would be easy to support restoring a standalone analysis with any PGN.

Is it useful though? Probably not, but what do you think?

veloce (commented):
Thanks for the fix. Having some tests would be nice here.

lib/src/model/analysis/analysis_controller.dart
HaonRekcef: The goal of this line is to use the saved analysis only when the Analysis Board is opened directly. In that case the PGN is empty.
If the Analysis Board is opened from puzzles, the board editor, or via “load position,” the PGN is not empty, so the saved standalone analysis is not loaded.

lib/src/model/analysis/analysis_controller.dart
HaonRekcef: done

HaonRekcef:
> Thanks for the fix. Having some tests would be nice here.

added extensive tests

lib/src/model/analysis/analysis_controller.dart
veloce: What I meant is that instead of checking if the PGN is not empty, you could also save the initial PGN used to open the analysis and then compare it with the currently opened PGN to restore a standalone analysis opened with a PGN.
Then we'd be able to:
* restore a game analysis loaded from a PGN
* go back to the last puzzle analysis, etc.

Probably not useful but since I don't use standalone analysis I wanted to have your opinion.

veloce: Perhaps also confusing

PR #2497, opened 2026-01-08
remove duplicate button

HaonRekcef:
There is already a button in the bottombar that is redirecting to the analysis board, no need for it to be in the menu too.


veloce (commented):
Don't necessarily agree that redundancy is an issue here... but ok.

PR #2511, opened 2026-01-11
add Team to Player Screen in Broadcasts

HaonRekcef:
Adds the team field to the Player Screen for team tournaments.

There is no l10n for team, the website also uses the hardcoded `Team`

grafik


veloce (approved):
Thanks!

Issue #2133, opened 2025-08-22
Deep links to game analysis should support linking a specific ply


tom-anders:
For example https://lichess.org/sOFJLM0J/white#43

tom-anders:
See https://github.com/lichess-org/mobile/blob/main/CONTRIBUTING.md#what-to-work-on:

> We don't assign issues to contributors, so feel free to work on any issue you like, but check the existing pull requests to avoid duplicated work. Once you start working on an issue, submit a pull request as soon as possible (in draft mode if it's not ready yet) to let others know that you're working on it.

Quick tip, for testing https://developer.android.com/training/app-links/deep-linking?hl=de will be helpful

PR #2518, opened 2026-01-12
Added support for linking a specific ply in game analysis

AC-BOSS:
https://github.com/user-attachments/assets/ca438d9d-83b6-400d-a3ba-edc028c29b64

### DIFF
--- a/lib/src/widgets/misc.dart
+++ b/lib/src/widgets/misc.dart
@@ -108,9 +108,11 @@ class _LichessMessageState extends State<LichessMessage> {
       spans.add(TextSpan(text: trans));
     }
 
-    return Text.rich(
-      TextSpan(style: widget.style, children: spans),
-      textAlign: widget.textAlign,
+    return MergeSemantics(
+      child: Text.rich(
+        TextSpan(style: widget.style, children: spans),
+        textAlign: widget.textAlign,
+      ),
     );
   }
 }


veloce:
Thanks for your contribution!

So now it allows sharing a link with a specific ply.

But looking at #2133 I am not entirely sure that this issue is about sharing link. I think it is about opening an analysis screen in the app from a deep link with ply support. Do you confirm @tom-anders ?

tom-anders:
> Thanks for your contribution!
> 
> So now it allows sharing a link with a specific ply.
> 
> But looking at #2133 I am not entirely sure that this issue is about sharing link. I think it is about opening an analysis screen in the app from a deep link with ply support. Do you confirm @tom-anders ?

Yep! So #2133 should remain open if we merge this PR

veloce:
Also the button says "Share Game Url", it doesn't say that it will make a link at the current ply. Isn't it a bit confusing / unexpected?

We could add another button "Share game at current position".

AC-BOSS:
> Also the button says "Share Game Url", it doesn't say that it will make a link at the current ply. Isn't it a bit confusing / unexpected?
> 
> We could add another button "Share game at current position".

In my opinion adding another button specifically to share game at current position feels a bit unnecessary. If a user shares link to move 22 while wanting to share of move 0, I don't think it will be an issue.
On the other hand, users may very often want to share a specific move of a game, and in that case just sharing a link is more convenient than sharing a link and move number.
Do tell me what you think is better though and I would like to implement that!

ijm8710:
I agree. But if there was one place to add it, I actually think it would
Make the most sense within the analysis board from
The bottom left button "share and export menu"
Not next to the "share game url button"

veloce:
In a way, it could make more sense to put this in the "Share and Export" submenu from the bottom menu.

I know it is not obvious, but the top menu is a menu common for game and analysis screen: it should behave the same for both screens. Your change here makes it work differently for game because game link doesn't link to the ply (that could be fixed).

The bottom share and export menu is only in analysis screen and is a contextual menu. Here you are sharing the current state of the analysis: current position, current PGN (with your edits, contrary to the share game PGN from top menu which is always the same, coming from server).

So 2 possibilities:
1. add the ply on game screen to share the game url with ply to be consistent with analysis screen
2. don't share game url with ply in top menu, and add another button in bottom menu in analysis screen only

I don't really have a preference for 1 or 2.

AC-BOSS:
Sure, understood. I agree with both veloce and ijm8710. Will make the required changes to this PR

PR #2520, opened 2026-01-13
Fix(a11y): Merge semantics for welcome message to prevent TalkBack skipping text

Septic-H:
**Problem:** On the welcome screen (signed out), TalkBack focused directly on the second half ("libre, no-adds..."), skipping the beginning of the sentence entirely ("Lichess is a free...")

**Fix:** Wrapped the text in MergeSemantics. This forces screen readers to group the text spans and read the full sentence.

**Verified:** Tested manually with TalkBack on Android. The entire sentence is now read correctly.


veloce:
Thanks!

Issue #2521, opened 2026-01-14
Back to tournament not working


vaibhavsection:
### What is the problem?

https://github.com/user-attachments/assets/119a597e-d77a-455c-8608-6416c3bc4995

### Steps to reproduce the bug

recent game
tap on game which your player recently (tournament is still going on or running)
tap on back to tournament 

### App version

0.19.4

### Device Info

One Plus 13R/Android 16

Septic-H:
I can look into this.

PR #2523, opened 2026-01-14
Fix: Enable 'Back to Tournament' navigation during ongoing tournaments (#2521)

Septic-H:
**In game_body.dart:** Changed the condition from isFinished to tournament != null. This ensures "Back to Tournament" is accessible during ongoing tournaments.

**In game_result_dialog.dart:** Decoupled navigation logic from pause logic.
- "Back to Tournament" is now visible for all tournament games.
- "Pause" remains restricted to isOngoing tournaments only.


veloce:
This does not fix #2521

veloce:
Closing as no activity there, and this is not the proper fix.

PR #2531, opened 2026-01-16
Broadcasts deeplinks improved

### LINKED ISSUE #2508: Push Broadcast Round Screen for Broadcast Game Deep Links
Deep links for a broadcast game should open both the broadcast round tab and the specific broadcast game. This allows users to view other games in the tournament.

HaonRekcef:
closes #2508

For cold starts and in-app links: `onGenerateInitialRoutes` pushes both the round screen and board screen.

For warm starts (app already running): `onGenerateRoute` pushes only the board screen with a `PopScope` that replaces itself with the round screen on back navigation.

I decided to use two approaches because if we push both screens avoids the back animation that would occur if we do  `PopScope` + `pushReplacement` 

This can in the future be applied to other deep links (profile stats, etc.) that need similar navigation stack handling.

Also:
- Added logger to `app_links.dart` 

Videos:
Cold start:

https://github.com/user-attachments/assets/168191d8-d3d9-4dd0-9ff8-b6df7301bd79

warm start:

https://github.com/user-attachments/assets/429216e4-c5d7-4c03-a178-dcf52d5980db

in app:

https://github.com/user-attachments/assets/2a5cd1ea-121d-4225-a998-aa14c20ec37e

### DIFF
--- a/lib/src/app.dart
+++ b/lib/src/app.dart
@@ -129,14 +129,16 @@ class _AppState extends ConsumerState<Application> {
                 context,
               ).copyWith(height: isShortVerticalScreen(context) ? 60 : null),
       ),
-      onGenerateRoute: (settings) =>
-          settings.name != null ? resolveAppLinkUri(context, Uri.parse(settings.name!)) : null,
+      onGenerateRoute: (settings) => settings.name != null
+          ? resolveAppRoutes(context, Uri.parse(settings.name!), deepLinkWarmStart: true)?.last
+          : null,
       onGenerateInitialRoutes: (initialRoute) {
         final homeRoute = buildScreenRoute<void>(context, screen: const MainTabScaffold());
-        return <Route<dynamic>?>[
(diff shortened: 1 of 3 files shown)


veloce:
Nice! need to test this on iOS too.

HaonRekcef:
> Nice! need to test this on iOS too.

For now there is no specific code, I think, that would need to be tested on iOS, as we dont support deeplinks there yet.

veloce:
After thinking about it, I'd like to handle this differently.

There is a way to opt out from flutter deep link handling: https://docs.flutter.dev/ui/navigation/deep-linking#migrating-from-plugin-based-deep-linking

We should do that and use our own logic. There is a pre-made plugin for that: https://pub.dev/packages/app_links

We could also have our own logic to not depend on a 3rd party, it does not seem that complicated: https://medium.com/flutter-community/deep-links-and-flutter-applications-how-to-handle-them-properly-8c9865af9283

PR #2544, opened 2026-01-18
add password protected tournaments

### LINKED ISSUE #1935: Joining user created(private) tournaments in the lichess mobile app
I'am currently unable to join user created private tournaments (shared with me via a link) in the lichess mobile app, i think it will be a nice feature for the app since it is already possible on the website

HaonRekcef:
closes #1935 

video: 

https://github.com/user-attachments/assets/1370248c-a990-47bb-bdbc-99b8e7121d72

includes testcases

### DIFF
--- a/test/view/tournament/tournament_screen_test.dart
+++ b/test/view/tournament/tournament_screen_test.dart
@@ -1420,7 +1420,7 @@ void main() {
       await tester.pump();
 
       // Verify error message appears
-      expect(find.text('Incorrect entry code'), findsOneWidget);
+      expect(find.text('Incorrect entry code.'), findsOneWidget);
 
       // Verify still not joined
       expect(find.text('Join'), findsOneWidget);


TBestLittleHelper:
I think both the code and the UI should consistently call it an "entry code" or similar and avoid naming it password. Since password comes with expectations around encryption etc. Which is not what the code is intended for, it's for giving to others so they can join. It's why the website renamed it back in 2021 https://github.com/lichess-org/lila/issues/9467

HaonRekcef:
> I think both the code and the UI should consistently call it an "entry code" or similar and avoid naming it password. Since password comes with expectations around encryption etc. Which is not what the code is intended for, it's for giving to others so they can join. It's why the website renamed it back in 2021 [lichess-org/lila#9467](https://github.com/lichess-org/lila/issues/9467)

Good point, makes sense!
I will adjust it later today.

veloce:
Thanks for the PR. Will review it when I have dealt with older ones.

lib/src/view/tournament/tournament_screen.dart
```diff
@@ -1150,6 +1151,39 @@ class _BottomBarState extends ConsumerState<_BottomBar> {
...
+                            setState(() {
+                              joinOrLeaveInProgress = true;
+                            });
+
+                            try {
+                              await ref
+                                  .read(tournamentControllerProvider(widget.state.id).notifier)
+                                  .joinOrPause(entryCode: entryCode);
+                            } catch (e) {
+                              setState(() {
```
veloce, line 1172: Better put this in a `finally` clause.

HaonRekcef: If the join request is successful line 1089 will update the joinOrLeaveProgress (which is better than updating it immediately so that it keeps spinning until the green bar appears) so we only need it for the unsuccessful case.

Issue #2292, opened 2025-10-25
[L10n] Inconsistent placeholder value for puzzleFromGameLink ("From game %s") string between Web and Mobile


carotinoid:
### What is the problem?

The same string key `fromGameLink` (which translates to `From game %s` in English) is populated with two different types of data depending on the platform.

1. On Web
    * The `%s` placeholder is filled with the game's time control and type.
    * Example: `From game 10+3 • Rapid`
    * <img width="1125" height="734" alt="Image" src="https://github.com/user-attachments/assets/11441aea-0b1f-4104-9586-ab1471dc0a01" />

2. On Mobile App 
    * The `%s` placeholder is filled with the Game ID.
    * Example: `From game GICIW2b2`
    * <img width="357" height="740" alt="Image" src="https://github.com/user-attachments/assets/22f9191f-cfd8-48e5-9ba2-2af60df0f904" />
    * This logic is found in the `lib/src/view/puzzle/puzzle_screen.dart` file.

Since web can provide more information about the game below (like player names and colors), it makes sense to show the time control and type there. However, on mobile app, since there is not enough space to show that information, showing the Game ID is more practical.

However, this inconsistency makes it impossible to provide a single, natural-sounding translation for this string.

**Impact on Localization**

Because my native language is Korean, I can illustrate the issue with Korean translations. Now, "From game %s" is translated as `게임 %s에서` (where '게임' means 'game' and the suffix '...에서' means 'from...').

Current translation examples:
* For Web: `게임 10+3 • Rapid에서` (natural-sounding)
    * <img width="1093" height="735" alt="Image" src="https://github.com/user-attachments/assets/505630be-4425-4239-94f2-fa9874f01661" />

* For Mobile App: `게임 GICIW2b2에서` (unnatural-sounding)
    * <img width="346" height="739" alt="Image" src="https://github.com/user-attachments/assets/a44d1dd3-389d-4173-b6fb-4a3924cd6726" />

The problem arises because in Korean, the phrase `게임 10+3 • Rapid에서` sounds natural, but `게임 GICIW2b2에서` does not. This is because in Korean, placing a technical ID directly after the noun `게임` (Game) is grammatically awkward; the ID `GICIW2b2` is not a descriptive name or type. It feels like a required label (like "ID:") is missing. The latter would be better translated as `게임에서 (ID: GICIW2b2)` or `게임(ID: GICIW2b2)에서` to make sense. 

Similar issues may arise in other languages as well.

[...]

PR #2559, opened 2026-01-21
Fix localization for puzzle game ID display

DejoeNithin:
closes #2278

### DIFF
--- a/lib/l10n/app_en.arb
+++ b/lib/l10n/app_en.arb
@@ -102,6 +102,7 @@
   "mobileWaitingForOpponentToJoin": "Waiting for opponent to join...",
   "mobileWatchTab": "Watch",
   "mobileWelcomeToLichessApp": "Welcome to Lichess app!",
+  "mobileViewGame": "View Game",
   "activityActivity": "Activity",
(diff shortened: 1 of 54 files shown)


DejoeNithin:
@carotinoid @tom-anders @veloce can you check this?

DejoeNithin:
@tom-anders @veloce I pushed the generated files lib/l10n (*.dart and *.arb) because tests were failing without them.
Is this the right way to do it?

veloce:
@DejoeNithin this is indeed the correct way to add a new translation.

However I'd rather do things differently here. See: https://github.com/lichess-org/mobile/issues/2292#issuecomment-3789500048

DejoeNithin:
@veloce I have updated the PR. can you check now?
image

DejoeNithin:
@veloce CI is failing with `Could not find package 'custom_lint'`. Looks like this commit removed it but CI still tries to run it: https://github.com/lichess-org/mobile/commit/483326654b02b4aa3722689a5aa7e39ce8142653

veloce:
I rebased your branch to fix that @DejoeNithin

Issue #2534, opened 2026-01-16
Piece animations not working if system menu animations disabled in Android


cristian64:
### What is the problem?

If user has disabled menu animations in the Android system (e.g. **Accessibility > Text and display > Remove animations** in Android 12), piece animations in Lichess do not work, regardless of the option being enabled in the game settings.

### Steps to reproduce the bug

1. Disable menu animations in Android (may depend on the version).
2. Enable **Piece animations** in Lichess (which is on by default).
3. Start a game.
4. Piece animations will not work.

### App version

0.18.2

### Device Info

Android 12

### Additional information

The issue does not manifest in the legacy mobile app; only in the Flutter-based version.

Ahmed-M-Hany:
check this pull request @tom-anders 
I fixed it in the flutter-chessground package
[https://github.com/lichess-org/flutter-chessground/pull/86](https://github.com/lichess-org/flutter-chessground/pull/86)

veloce:
As I answered in the pull request, I am not sure we want to "fix" that. If you disable the the animation system wide, why wouldn't it apply to the piece animations? Piece animation is not an essential feature, you can still play without, so this is not a bug imo.

cristian64:
I see the system-wide preference as something that should apply to menus only, but not to the game itself.

Lichess is the only game-like application I've seen that applies the system-wide setting to the in-game screen. If every game out there would disable animations, they would be unplayable. Lichess happens to remain playable but, as a user, that's not the behaviour I expect. Certainly one of the reasons why I'm still on the legacy version, which isn't affected.

**Piece animation** is an "on"/"off" preference. You could argue it could be changed to being a "System"/"off" preference, and then it would at least be accurate.

veloce:
Fair point, if all the game apps don't respect that system setting, then we should do the same.

PR #2565, opened 2026-01-22
Fix for android "Remove animations" system setting not reflected by board settings > piece animation

Sriram-L:
_Fixes_ #2534 

**Fix description:**
The fix basically fetches current value of "Remove animations" in Android, and the board settings UI refreshes so that after fetching the value from system settings, the toggle option will be disabled if the system animation is disabled, and enabled when system animation is enabled.

Attaching a recording of how the fix works.

https://github.com/user-attachments/assets/0d319b1e-ac78-4f78-b571-5c1e19b1efa8

### DIFF
--- a/android/app/src/main/kotlin/org/lichess/mobileV2/MainActivity.kt
+++ b/android/app/src/main/kotlin/org/lichess/mobileV2/MainActivity.kt
@@ -4,6 +4,7 @@ import android.app.ActivityManager
 import android.content.Context
 import android.graphics.Rect
 import android.os.Bundle
+import android.provider.Settings
 import androidx.core.splashscreen.SplashScreen.Companion.installSplashScreen
 import androidx.core.view.ViewCompat
 import io.flutter.embedding.android.FlutterActivity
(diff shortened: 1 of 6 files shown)


tom-anders:
Thanks for the contribution!

However, I think this does not fix the issue in #2534. The reporter of this issue suggested that the piece animation toggle should work, *regardless of whether animations are disabled on the system level or not*.

Sriram-L:
> Thanks for the contribution!
> 
> However, I think this does not fix the issue in #2534. The reporter of this issue suggested that the piece animation toggle should work, _regardless of whether animations are disabled on the system level or not_.

Thanks for the reply!

But I checked when "Remove animations" is enabled, piece animations do not work no matter what. So how can we satisfy this requirement regardless of this setting?

tom-anders:
> > Thanks for the contribution!
> > However, I think this does not fix the issue in #2534. The reporter of this issue suggested that the piece animation toggle should work, _regardless of whether animations are disabled on the system level or not_.
> 
> Thanks for the reply!
> 
> But I checked when "Remove animations" is enabled, piece animations do not work no matter what. So how can we satisfy this requirement regardless of this setting?

No idea - you'd have to do some research if flutter somehow allows overriding/ignoring the system's "remove animations" setting.

veloce:
Do we really want to follow the android system "Remove animations" toggle for our piece animations? I don't think this is a good idea, it looks like 2 very different things to me.

tom-anders:
> Do we really want to follow the android system "Remove animations" toggle for our piece animations? I don't think this is a good idea, it looks like 2 very different things to me.

Yes, the problem is that currently we *do* follow this toggle in the android system (see the original issue): When "remove animations" is active in the Android systems, them piece animations currently are disabled as well. I assume that flutter somehow internally notices this and overrides the animation duration. The OP in the original issue argues that piece animations should ignore that setting. So we'd need to find some way to stop flutter respecting that system toggle.

veloce:
I had no idea that Flutter did this. But it makes sense in a way, this can be used to save battery for instance.

Not sure we can do sth about it, it needs an investigation.

veloce:
Fixed by https://github.com/lichess-org/flutter-chessground/pull/86

PR #2571, opened 2026-01-25
feat: add tip to customize home screen widgets

tom-anders:
(Plus two small fixes to the "edit home" screen when not logged it.)

Main intention is to make the quick game matrix more discoverable, especially for users coming from the website or the old app. But in general this feature is too cool for it to be hidden in the settings, so I think it's a good thing to make it more discoverable :)

If the user does not interact with this widget for more than 3 cold app starts, we do not show it anymore.

Otherwise, pressing the "Settings" or "Dismiss" button hides it forever as well.

[customize.webm](https://github.com/user-attachments/assets/6076ae21-6835-4954-bceb-03ef7800968d)

### DIFF
--- a/test/binding.dart
+++ b/test/binding.dart
@@ -75,7 +75,7 @@ class TestLichessBinding extends LichessBinding {
   FakeSharedPreferences? _sharedPreferences;
 
   @override
-  int get numAppStarts => 0;
+  int numAppStarts = 1;
 
   @override
   FakeSharedPreferences get sharedPreferences {
@@ -88,6 +88,7 @@ class TestLichessBinding extends LichessBinding {
   void reset() {
     _firebaseMessaging = null;
     _sharedPreferences = null;
+    numAppStarts = 1;
   }
 
   FirebaseCrashlytics? _firebaseCrashlytics;
--- a/test/view/home/home_tab_screen_test.dart
+++ b/test/view/home/home_tab_screen_test.dart
@@ -11,6 +11,7 @@ import 'package:lichess_mobile/src/view/home/games_carousel.dart';
 import 'package:lichess_mobile/src/widgets/buttons.dart';
 import 'package:lichess_mobile/src/widgets/feedback.dart';
 
+import '../../binding.dart';
 import '../../example_data.dart';
 import '../../mock_server_responses.dart';
 import '../../model/auth/fake_auth_storage.dart';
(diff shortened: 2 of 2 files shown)


lib/src/view/home/home_tab_screen.dart
```diff
@@ -989,3 +1001,68 @@ class _ChallengeScreenButton extends ConsumerWidget {
...
 }
+
+class _HomeCustomizationTip extends StatelessWidget {
+  const _HomeCustomizationTip();
+
+  bool _shouldDisplayHomeWidgetCustomizationTip() {
+    final prefs = LichessBinding.instance.sharedPreferences;
+
+    return prefs.getBool(kHideHomeWidgetCustomizationTip) != true &&
+        LichessBinding.instance.numAppStarts <= 3;
```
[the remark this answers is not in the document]
tom-anders: first boot starts counting app starts from 1, so it's correct

veloce:
I rebased the branch to main to fix tests @tom-anders

lib/src/view/home/home_tab_screen.dart
```diff
@@ -227,36 +230,43 @@ class _HomeScreenState extends ConsumerState<HomeTabScreen> {
 
         if (shouldShowWelcomeScreen) {
           final welcomeWidgets = [
-            const _GreetingWidget(),
```
veloce, line 230: Don't get the reason of the changes in this block, can you explain? thx

tom-anders:
> This is nice, thank you!
> 
> I'd like to have tests for this, since it should be easy to write them (you can override the number of cold app start in testBindings easily). Test scenarios to check:
> 
>     * in case cold app starts < 3:
>       
>       * should be shown when not logged in
>       * should be shown even if the welcome message is not shown anymore (games were played or logged in)
>       * should be shown when logged in
>       * should be dismissed after going to settings
> 
>     * when cold app start >= 3
>       
>       * should not be shown

Ah sorry, should've added these right away. I think I covered all scenarios you mentioned now :+1:

lib/src/view/home/home_tab_screen.dart
tom-anders: These are two separate bugfixes/commits:

- a4b7560c00c65e188322aafb733100247695b75d -> The "Hello XY" widget is editable when logged in, but not if you're logged out. The commit makes it editable when logged out as well
- d1c90cb47f564ade8743bc433c540481734c9458 -> The welcome widgets ("lichess is a free...", sign-in-button, about-button) are currently shown when the home screen is in edit mode. I think this is a bit confusing, especially when new users enter this screen via the new tip widget. So this commits hides them in edit mode.

lib/src/view/home/home_tab_screen.dart
```diff
@@ -989,3 +1001,73 @@ class _ChallengeScreenButton extends ConsumerWidget {
...
+  @override
+  State<_HomeCustomizationTip> createState() => _HomeCustomizationTipState();
+}
+
+class _HomeCustomizationTipState extends State<_HomeCustomizationTip> {
+  bool _shouldDisplayHomeWidgetCustomizationTip() {
+    final prefs = LichessBinding.instance.sharedPreferences;
+
+    return prefs.getBool(kHideHomeWidgetCustomizationTip) != true &&
+        LichessBinding.instance.numAppStarts <= 3;
```
veloce, line 1017: Can you extract this to a `kColdAppStartsHideCustomizationTipThreshold` (or better name if you have) constant?

Then you can use this constant in tests and we can change it. Perhaps we should set it to 5, wdyt?

tom-anders: ah looks like I wasn't fast enough, thanks for taking care of it :D

PR #2581, opened 2026-01-28
Replace shouldExcludeGesturesOnFocusGained callaback by a boolean

julien4215:
The goal is to simplify the code by replacing a callback by a boolean in the `AndroidGesturesExclusionWidget`.

This was part of the PR https://github.com/lichess-org/mobile/pull/1921. I only took this change for now because I am not sure about the rest.

### DIFF
--- a/lib/src/utils/gestures_exclusion.dart
+++ b/lib/src/utils/gestures_exclusion.dart
@@ -33,13 +33,10 @@ class AndroidGesturesExclusionWidget extends StatefulWidget {
   /// Global key of the board widget.
   final GlobalKey boardKey;
 
-  /// Optional callback called when focus is gained, to check whether
-  /// to exclude gestures or not.
+  /// Whether to exclude gestures when focus is gained.
   ///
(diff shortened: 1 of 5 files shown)


veloce:
Please provide a description in your PR. What are you trying to solve? How? Anything that can help me understand. I don't want to figure things out just by reading the code. Thanks.

julien4215:
I have added a description to the PR.

veloce (commented):
Code looks good. Have you tested it to be sure it still works as expected?

julien4215:
I have tested it. The behavior is the same as before so I think it can be merged.

However it made me realize that the gesture exclusion is broken on the game screen. The gesture exclusion happens below the board.

veloce:
How so? I may be, but I don't understand why since this code hasn't changed.

julien4215:
I don't know. It works correctly on the puzzle board (well only on half of the board because of the 200 dp limit on `setAndroidBoardGesturesExclusion`) but it doesn't work at all for the game screen.

PR #2590, opened 2026-01-30
Fix gesture exclusion for game screen

julien4215:
When `findRenderObject` is used it find the render object of `GameLayout` I think when it should be the board widget.


veloce (approved):
Good catch!

PR #2609, opened 2026-02-02
allow split-screen on older android versions

### LINKED ISSUE #1938: Android split screen support missing on new Lichess app
Old Lichess app allowed Android split screen where you can put two apps on the screen at the same time. It is particularly useful on tablets for studying chess books with the Lichess app analyser board on the side.

When trying to perform same task with new Lichess app, option to split screen does not appear when holding the app on Android task manager. If instead you select another app and try to choose to split screen with the new Lichess app you also won't be able to because the new Lichess app will not appear in the list of possible apps to split screen.

Sample screenshot showcasing how the old Lichess app allowed split screen:
Image

HaonRekcef:
closes #1938 
Removes `android:resizeableActivity="false"` to re-enable split-screen (multi-window) support on older Android versions.

On newer android versions (Android 12L +), the system allows split-views anyways even with it set to false.

We support split screens on iPadOs and newer android versions so I don't see why we would disallow it on older android versions.

The issues that lead to `android:resizeableActivity="false"` are fixed now: https://github.com/lichess-org/mobile/issues/914

Tested on multiple older android versions and could not find any issues. Screenshot on Android 8: 
grafik

Note: This also enaboles split screen on phones, but some non stock Android version enabled that anyways.

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -5,7 +5,6 @@
         android:name="${applicationName}"
         android:icon="@mipmap/ic_launcher"
         android:roundIcon="@mipmap/ic_launcher_round"
-        android:resizeableActivity="false"
         android:enableOnBackInvokedCallback="true"
         android:fullBackupContent="@xml/backup_rules">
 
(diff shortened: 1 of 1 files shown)


veloce:
I understand this is useful in tablets. However on phones there may be layout issues (which is why I disabled it in the first place). 
that being said I’m not against merging that even if we have some issues on phones as long as they are not too important. Have you tested this with a phone?

HaonRekcef:
Yes tested it on the phone and for reasonable layouts I could not find any issues. For very small app sizes there are some overflow issues, but every user that uses such small layout should understand that not everything can be displayed correctly in a small view.

PR #2615, opened 2026-02-04
feat: Add settings button and auto-next toggle to puzzle streak (#1296)

### LINKED ISSUE #1296: Add the "immediately go to the next puzzle" toggle in normal puzzle + puzzle streak
This feature exists in the website. You can choosd whenever to analyze the puzzle to understand some variants, etc. or to immediatly go to the next puzzle, staying concentrated.

For now, this toggle is set to false on normal puzzles and true on puzzle streak... 

Post scriptum: Maybe also implement the like/dislike button with it? Would be great.

DejoeNithin:
closes #1296 

https://github.com/user-attachments/assets/17121630-1d5c-433f-8148-aa1603556d53

### DIFF
--- a/lib/src/view/puzzle/streak_screen.dart
+++ b/lib/src/view/puzzle/streak_screen.dart
@@ -35,8 +35,8 @@ import 'package:lichess_mobile/src/widgets/bottom_bar.dart';
 import 'package:lichess_mobile/src/widgets/buttons.dart';
 import 'package:lichess_mobile/src/widgets/list.dart';
 import 'package:lichess_mobile/src/widgets/pgn.dart';
-import 'package:lichess_mobile/src/widgets/settings.dart';
 import 'package:lichess_mobile/src/widgets/platform_alert_dialog.dart';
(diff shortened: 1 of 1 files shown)


DejoeNithin:
@tom-anders @veloce @ijm8710 @ghost can you check demo video ?

DejoeNithin:
@veloce @tom-anders streak score only increments after clicking "Next Puzzle" when autoNext is off, but the test expects it to go up immediately when solved. Which should it be? This is causing the test `test/view/puzzle/streak_screen_test.dart:"StreakScreen Score is saved when exiting screen"` to fail.

veloce:
> @veloce @tom-anders streak score only increments after clicking "Next Puzzle" when autoNext is off, but the test expects it to go up immediately when solved. Which should it be? This is causing the test `test/view/puzzle/streak_screen_test.dart:"StreakScreen Score is saved when exiting screen"` to fail.

We should maintain the logic that the score is updated on the next screen I think, visually at least. Because, of course, if you exit the screen before tapping "next" the good score should be saved.

Issue #2468, opened 2025-12-29
Activity tab to "link to correspondence game/opponent" and show "started following player" (like it does on web)


ijm8710:
![image](https://github.com/user-attachments/assets/6bf622b8-4774-4d67-9671-6ad1721eaed3)
![image](https://github.com/user-attachments/assets/7f4752a7-bf3b-433d-9062-d94ab2300c31)

See the differences between web and the app for these 2 items

Few other things it shows:
- Joined a team
- Links to tournament 
- Gained a follower
- links to a practice

PR #2618, opened 2026-02-06
User Activity Widget Improvements

r3econ:
## Changes

 This PR:

- Adds opening recently solved puzzles by tapping on a puzzle activity item
- Makes it possible to open recent games by tapping on a recent game activity item

###### Demo

https://github.com/user-attachments/assets/4d331e29-7c8a-42cd-a877-309d2b09271a

### DIFF
--- a/lib/src/view/user/user_activity.dart
+++ b/lib/src/view/user/user_activity.dart
@@ -122,12 +122,8 @@ class UserActivityEntry extends ConsumerWidget {
                 loss: gameEntry.value.loss,
               ),
               onTap: () => Navigator.of(context).push(
-                    GameHistoryScreen.buildRoute(
-                      context,
-                      user: user,
-                      isOnline: user.isOnline == true,
-                    ),
-                  ),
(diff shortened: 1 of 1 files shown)


ijm8710:
Any interest in anything from #2468 which is inspired by how the site is handled on web?

r3econ:
@ijm8710 definitely, I'll have a look into it 👍🏾  I didn't know https://github.com/lichess-org/mobile/issues/2468 existed. 
Can we merge this PR? I'd prefer to keep the change set small and keep delivering small testable increments separately

veloce (approved):

veloce:
Thanks!

ijm8710:
Hey @r3econ what are your thoughts if possible to when you tap and it opens the archive game menu it opens filtered to the respective variant

r3econ:
Yes this would be even better. I had been thinking about it but at the end I left it for later. I'll have a look, this could probably be added very easily.

I'm new to the codebase and although I'm familiarizing myself with it fast, currently more time is spent learning where the things are,  looking for things to reuse, etc.

Issue #2574, opened 2026-01-25
Screen times out and turns off when in Analysis board / Conditional premoves


KnightAzul:
### What is the problem?

Hi,
I have the latest Android lichess version 0.19.4 installed on my Samsung S23 Android 16.
When playing correspondence chess, I go into Analysis board / Conditional premoves for the game. The screen times out and turns off according to Android system settings (after 1 minute). This means I have to keep touching the screen to keep the screen alive to allow me to analyse the game. The screen should not turn off here.
In the old app the screen would not timeout when in Analysis board / Conditional premoves, which from a usability perspective is much better.
Please let me know if you need more information.
Thanks,
KnightAzul

### Steps to reproduce the bug

1. Start a correspondence chess game.
2. Go into Analysis board / Conditional premoves for the game.
3. The screen times out and turns off according to Android system settings (after 1 minute).. This means I have to keep touching the screen to keep the screen alive to allow me to analyse the game. The screen should not turn off here.

### App version

Android lichess version 0.19.4

### Device Info

Samsung S23 Android 16

### Additional information

In the old app the screen would not timeout when in Analysis board / Conditional premoves, which from a usability perspective is much better.

PR #2619, opened 2026-02-07
Enable wakelock in analysis and opening explorer screens (#2574)

karansuthardev:
Import **wakelock_plus** and keep the device awake while analysis and opening explorer screens are active. **AnalysisScreen** now enables wakelock in initState and disables it in dispose.

**OpeningExplorerScreen** was converted from a ConsumerWidget to a ConsumerStatefulWidget with a _OpeningExplorerScreenState to manage wakelock in initState/dispose; references to options were updated to use widget.options and the constructor now forwards the key.

_**This ensures the UI stays awake during long-running analysis/explorer sessions.**_


veloce:
It is better to use the `WakelockWidget` since it handles the focus.

Issue #2611, opened 2026-02-02
Grab fen from opening explorer


ijm8710:
Should this be possible? Haven't needed it often but desired it today and it is available on web

tom-anders:
The share button in top right corner does that

ijm8710:
> The share button in top right corner does that 

Do you see something I don't ![image](https://github.com/user-attachments/assets/4192f96b-c41d-4c87-af1d-51ce42085dbe)

tom-anders:
Whoops, somehow I read "board editor" instead of "opening explorer", my bad

ijm8710:
> Whoops, somehow I read "board editor" instead of "opening explorer", my bad 

All good, benefit is I think you inspired the best way to add it in which would be to just do the same as the editor

PR #2632, opened 2026-02-13
Make it possible to share FEN of the position in opening explorer

r3econ:
- Resolves https://github.com/lichess-org/mobile/issues/2611
- Make it possible to share FEN of the position in opening explorer (same solution as in the board editor)

#### Testing

✅  I tested it using the iOS Simulator and copied the FEN of the position into an online tool to verify its correctness

#### Demo

https://github.com/user-attachments/assets/9af7505e-11d6-4d0d-b28b-ee8f30c59271

### DIFF
--- a/lib/src/view/explorer/opening_explorer_screen.dart
+++ b/lib/src/view/explorer/opening_explorer_screen.dart
@@ -11,6 +11,7 @@ import 'package:lichess_mobile/src/styles/styles.dart';
 import 'package:lichess_mobile/src/utils/l10n_context.dart';
 import 'package:lichess_mobile/src/utils/navigation.dart';
 import 'package:lichess_mobile/src/utils/screen.dart';
+import 'package:lichess_mobile/src/utils/share.dart';
 import 'package:lichess_mobile/src/view/analysis/game_analysis_board.dart';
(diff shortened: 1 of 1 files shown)


veloce:
Thanks!

ijm8710:
Hey @r3econ thanks for fulfilling my issue ticket :) 

Thinking about it some more what are your thoughts of the share button for opening explorer and editor to [launch to this screen](https://github.com/user-attachments/assets/5afd89c3-4726-4864-8b95-d1d67767cdda) similar to how it does with analysis board.

Are there any reasons you can think of that we do fen and not pgn for these two? It appears web does both.

And less similar but still quasi-related any idea broadcasts would do pgn but not fen? Web appears to do both as well so feel having either option across the board makes sense

r3econ:
@ijm8710, I think it's a good idea to align the functionality with the web (and the rest of the app) 👍🏾 

I've just opened a pull request adding sharing of the PGN from the openings explorer https://github.com/lichess-org/mobile/pull/2664. I'll have a look into the broadcasts 👀 🛠️

r3econ:
@ijm8710, just to confirm, with broadcasts you mean this menu, don't you?


image

ijm8710:
Yup exactly

PR #2649, opened 2026-02-19
upgrade chessground to ^8.0.1

tom-anders:
Fixing API breaks was trivial, mostly changing `isDrop` to `isDragAndDrop` and `NormalMove?` in the move callbacks to `Move?`.

### DIFF
--- a/lib/src/model/common/eval.dart
+++ b/lib/src/model/common/eval.dart
@@ -242,6 +242,7 @@ ISet<Shape> moveShapes({
       ),
   }.toISet(),
   DropMove(role: final role, to: _) => {
+    Circle(color: color, orig: move.to, scale: scale),
     PieceShape(
       color: color,
       orig: move.to,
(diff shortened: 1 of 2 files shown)


lib/src/model/study/study_controller.dart
```diff
@@ -396,15 +397,27 @@ class StudyController extends AsyncNotifier<StudyState>
...
+      case NormalMove():
+        _recordChange('anaMove', {
+          'orig': move.from.name,
+          'dest': move.to.name,
+          'variant': state.requireValue.variant.name,
+          'fen': state.requireValue.currentPosition!.fen,
+          'path': state.requireValue.currentPath.value,
+        });
+      case DropMove():
+        _recordChange('anaDrop', {
```
tom-anders, line 413: @veloce FYI I created a crazyhouse study in web and used that to get the correct format here

tom-anders:
> LGTM!
> 
> This is cool, thanks. Crazyhouse support should be "easy" from now on.

Yep, I already have a branch where I added a rudimentary piece menu for crazyhouse and it works great. The main work is to make it look nice and not mess up any of the layouts too much

PR #2651, opened 2026-02-19
Puzzle dashboard: strengths and improvement areas

r3econ:
## Puzzle dashboard: strengths and improvement areas

Extends the puzzle dashboard with per-theme performance sections.

### New features

- Added **Strengths** and **Improvement Areas** sections (like in [web](https://lichess.org/training/dashboard/30/dashboard) version) below the radar chart, each showing the 3 highest/lowest performing themes with their performance rating, solve percentage, and play count.
- Tapping a theme row navigates directly into a puzzle session for that theme.

### Refactors

- Extracted `_ChartSection` and `_PerformanceSection` as dedicated `StatelessWidget`s, replacing large builder methods on `PuzzleDashboardWidget`.
- Moved sorting, title, and subtitle logic onto the `Metric` enum — each case is self-contained and no `switch` is needed at the widget layer.
- Used `sortedByCompare` with a descending comparator for the strength sort, avoiding the extra allocation from `.reversed`.
- `_itemsToShow` is a private constant on `Metric` since it only has meaning there.

### Showcase

puzzledashboard

puzzle dashboard ipad

https://github.com/user-attachments/assets/d3d48b7b-a717-418f-a128-5569002cbf5a

### DIFF
--- a/lib/src/constants.dart
+++ b/lib/src/constants.dart
@@ -63,7 +63,6 @@ const kMaxClockTextScaleFactor = 1.94;
 const kEmptyWidget = SizedBox.shrink();
 const kEmptyFen = '8/8/8/8/8/8/8/8 w - - 0 1';
 const kTabletBoardTableSidePadding = 16.0;
-const kAnalysisPortraitBoardSidePadding = 8.0;
 const kBottomBarHeight = 56.0;
 const kMaterialPopupMenuMaxWidth = 500.0;
 
--- a/lib/src/model/analysis/analysis_controller.dart
+++ b/lib/src/model/analysis/analysis_controller.dart
@@ -8,6 +8,7 @@ import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:freezed_annotation/freezed_annotation.dart';
 import 'package:intl/intl.dart';
 import 'package:lichess_mobile/src/model/account/account_service.dart';
+import 'package:lichess_mobile/src/model/analysis/analysis_player.dart';
 import 'package:lichess_mobile/src/model/analysis/analysis_preferences.dart';
 import 'package:lichess_mobile/src/model/analysis/common_analysis_state.dart';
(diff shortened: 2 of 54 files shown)


ijm8710:
This is sweet!

when you click the puzzle perf tile on your profile, there's also a radar chart there as well. Do you think it would make sense to also have the same view of this PR there as well, especially since the puzzle stuff can be a little bit hidden

r3econ:
> This is sweet!
> 
> when you click the puzzle perf tile on your profile, there's also a radar chart there as well. Do you think it would make sense to also have the same view of this PR there as well, especially since the puzzle stuff can be a little bit hidden

@ijm8710, it's already there :) No matter how you access the puzzle dashboard (where the spider diagram is), the new stuff, _strengths_ and _improvement areas_ will be there as well.

Home|Profile
--|--
|

ijm8710:
Oh then I said the question backwards. on the puzzles tab on the top right the left of the two icons if you click from there, is this stuff there as well?

r3econ:
> Oh then I said the question backwards. on the puzzles tab on the top right the left of the two icons if you click from there, is this stuff there as well?

Yes. This is shown in the video in the PR's description.

ijm8710:
Perfect! My apologies for confusing myself!

ijm8710:
Hey @r3econ any insight into why the areas shared on web don't align with what we see in the app? Admittedly my recent puzzle activity is somewhat scarce but if it's enough sample size to generate, shouldn't they match? Is this potentially an issue with the api?

image
image
image
image
image

r3econ:
@ijm8710, I'll have a look what the web does. What I did is sorting puzzle categories by user performance and show first 3 as strengths and last 3 as improvement areas. Probably web does something more sophisticated when choosing what to show

ijm8710:
Fair enough. There may be minimums when less data is available

PR #2655, opened 2026-02-20
Variant material diff

tom-anders:
- Hide material in Horde and Crazyhouse
- Display number of checks in 3-check

| Before | After |
|---------|---------|
|  |  |
|  |  |


veloce:
Is the website also displaying the checks like that? Some white space could be useful to separate the material from the checks.

tom-anders:
> Is the website also displaying the checks like that? Some white space could be useful to separate the material from the checks.

Pretty much looks the same on web, but there's indeed some horizontal padding between the king and the other pieces. So yeah we should do it as well in the app

PR #2660, opened 2026-02-21
feat: variant support in standalone analysis board

tom-anders:
Screenshot_1771667929

[variant_analysis.webm](https://github.com/user-attachments/assets/9dff6fef-8905-4d31-a362-d74c062a7099)

### DIFF
--- a/lib/src/view/analysis/analysis_screen.dart
+++ b/lib/src/view/analysis/analysis_screen.dart
@@ -602,7 +602,15 @@ class _BottomBar extends ConsumerWidget {
                     )
                     .toList(),
                 selectedItem: analysisState.variant,
-                labelBuilder: (Variant variant) => Text(variant.label),
+                labelBuilder: (Variant variant) => Text.rich(
+                  TextSpan(
+                    children: [
+                      WidgetSpan(child: Icon(variant.icon), alignment: PlaceholderAlignment.middle),
+                      const WidgetSpan(child: SizedBox(width: 8)),
+                      TextSpan(text: variant.label),
+                    ],
+                  ),
+                ),
                 onSelectedItemChanged: (Variant variant) =>
                     WidgetsBinding.instance.addPostFrameCallback((_) {
                       ref
--- a/test/view/analysis/analysis_screen_test.dart
+++ b/test/view/analysis/analysis_screen_test.dart
@@ -137,18 +137,18 @@ void main() {
           await tester.tap(find.text('Variant'));
           await tester.pumpAndSettle(); // wait for dialog to open
 
-          expect(find.text('Standard'), findsOneWidget);
-          expect(find.text('Chess960'), findsNothing);
-          expect(find.text('From Position'), findsNothing);
-          expect(find.text('Antichess'), findsOneWidget);
-          expect(find.text('King of the Hill'), findsOneWidget);
-          expect(find.text('Three Check'), findsOneWidget);
-          expect(find.text('Atomic'), findsNothing);
-          expect(find.text('Horde'), findsOneWidget);
-          expect(find.text('Racing Kings'), findsOneWidget);
-          expect(find.text('Crazyhouse'), findsNothing);
+          expect(find.textContaining('Standard'), findsOneWidget);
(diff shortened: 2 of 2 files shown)


lib/src/model/analysis/analysis_controller.dart
```diff
@@ -202,7 +204,7 @@ class AnalysisController extends AsyncNotifier<AnalysisState>
           // We want to keep the standalone analysis session alive even if the user navigates away
           ref.onCancel(() {
             if (_root.mainline.isNotEmpty) {
-              _savedStandalone = (root: _root, path: _currentPath);
+              _savedStandalone = (root: _root, path: _currentPath, variant: _variant);
             }
```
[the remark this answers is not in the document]
tom-anders: Not sure if we actually want this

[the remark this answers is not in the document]
veloce: nah

lib/src/view/analysis/analysis_screen.dart
```diff
@@ -581,13 +582,52 @@ class _BottomBar extends ConsumerWidget {
...
             makeLabel: (context) => Text(context.l10n.clearSavedMoves),
             onPressed: () => ref
                 .read(analysisControllerProvider(options).notifier)
                 .clearSavedStandaloneAnalysis(),
           ),
+          // Only allow changing the variant if this is standalone analysis entered from the home screen,
+          // but not for any other case like puzzle analysis or an active correspondence game.
+          if (pgn.isEmpty)
+            BottomSheetAction(
+              makeLabel: (context) => const Text('Change variant'), // TODO l10n
```
veloce: We could use the already available `context.l10n.variants`; it is obvious that the dialog will change the variant

lib/src/model/analysis/analysis_controller.dart
```diff
@@ -192,7 +192,9 @@ class AnalysisController extends AsyncNotifier<AnalysisState>
         }
       case Standalone(:final variant, pgn: final gamePgn):
         {
-          _variant = variant;
+          _variant = gamePgn.isEmpty && _savedStandalone != null
+              ? _savedStandalone!.variant
```
veloce, line 196: What if we have an already saved analysis without variant in previous app version? it will crash because variant will be `null` it seems.

lib/src/view/analysis/analysis_screen.dart
tom-anders: Makes sense, fixed

lib/src/model/analysis/analysis_controller.dart
tom-anders: The saved analysis does not survive restarts, does it? It's saved in RAM only, so an app update should clear it anyway...?

veloce: oh, right; I thought it was persisted, my bad ;)

tom-anders: I'm sure persisting is on @HaonRekcef's TODO list, but I was quick enough here :D

veloce (approved):

ijm8710:
Hey tom super super cool, any chance to ask 2 quick questions:

1) your photo has logos before variant name but video does not. Any reason why? I think it looks even better with
2) is 960 omitted since it's merged with standard? Prob still makes sense to follow web here if so and have a separate option for 960 or if not should it be standard/960?

tom-anders:
> Hey tom super super cool, any chance to ask 2 quick questions:
> 
> 1. your photo has logos before variant name but video does not. Any reason why? I think it looks even better with

Final version has the icons, video was an older version

> 2. is 960 omitted since it's merged with standard? Prob still makes sense to follow web here if so and have a separate option for 960 or if not should it be standard/960?

Yes, but it's not much value unless it also allows you to choose one of the starting positions (or a random one). So something for the future

lib/src/model/analysis/analysis_controller.dart
HaonRekcef: Ha yes! But I dont think persisting to storage automatically is needed, but rather via a save pgn option.

PR #2661, opened 2026-02-21
fix: do not flip board in Racing Kings if POV is black

tom-anders:
In racing kings, both side's pieces move in the same direction, so always orient the board as "white" regardless of who is playing, unless the board is explicitly turned.


veloce:
Would be nice to add tests for this. At least in the game screen. Thanks!

tom-anders:
Hmm do we have some helpers to assert the board orientation in tests? 

I now added a tests that directly tests `variantBoardOrientation()`

PR #2663, opened 2026-02-21
feat: add explosion sound for captures in Atomic Chess

tom-anders:
:boom: 

sounds copied over from lila and converted to aifc via ffmpeg.


veloce (approved):
Thanks!

PR #2664, opened 2026-02-21
Add PGN Sharing to the Openings Explorer

r3econ:
- This PR adds PGN sharing to the openings explorer
- Same functionality is already available in the analysis screen (and in the web version)
- Requested in https://github.com/lichess-org/mobile/pull/2632#issuecomment-3937798763

### DIFF
--- a/lib/src/view/explorer/opening_explorer_screen.dart
+++ b/lib/src/view/explorer/opening_explorer_screen.dart
@@ -52,7 +52,7 @@ class OpeningExplorerScreen extends ConsumerWidget {
           title: Text(context.l10n.openingExplorer),
           actions: [
             SemanticIconButton(
-              semanticsLabel: 'Share game',
+              semanticsLabel: context.l10n.studyShareAndExport,
               onPressed: () => _showShareMenu(context, ref),
               icon: const PlatformShareIcon(),
             ),


lib/src/view/explorer/opening_explorer_screen.dart
```diff
@@ -51,17 +49,37 @@ class OpeningExplorerScreen extends ConsumerWidget {
...
         title: Text(context.l10n.openingExplorer),
         actions: [
-          if (boardState != null)
-            SemanticIconButton(
-              semanticsLabel: context.l10n.mobileSharePositionAsFEN,
-              onPressed: () => launchShareDialog(context, ShareParams(text: boardState.fen)),
-              icon: const PlatformShareIcon(),
-            ),
+          SemanticIconButton(
+            semanticsLabel: context.l10n.mobileSharePositionAsFEN,
```
veloce: label is wrong now that the button opens the share menu

r3econ: @veloce, thanks for spotting it 🎣  It looks like we do not have anything generic that would fit here 🤔 

SCR-20260222-rcyv

For now, I hardcoded it to _Share game_ and added a new translation string 59a50246c9455ff150d862d6cc97999575e38fc6. 

Is this a chicken - egg problem, isn't it? I can't use the translation key via `context.l10n.mobileShareGame` **yet**; it has to be generated and pulled by the CI action first.

If my understanding is correct I suggest we merge the PR and I'll update the code to use the translation logic once the key gets generated

HaonRekcef: @r3econ take a look here: https://github.com/lichess-org/mobile/blob/main/docs/internationalisation.md you can add it to the mobile.xml and then generate the files

r3econ: Perfect, @HaonRekcef thanks for the hint! I generated the files ☑️ @veloce, the PR is ready 🚀

HaonRekcef:
Mmh actually we maybe could also use share and export like we do in the analysis board so we do not need a new translation

r3econ:
One extra translation key is just a tip of an iceberg. IMHO, what we need is a broader refactor of game sharing functionality - having one easily reusable component that would allow for sharing:
- FEN
- PGN
- GIF
- URL
- .. and allow for extending with further sharing options

Obviously, out of scope of this PR 🤔

veloce:
Yes as said by HaonRekcef let's not add a new translation if we can avoid it (as it takes people time to translate). So please remove the string in mobile.xml and use the available `studyShareAndExport` key.

PR #2665, opened 2026-02-21
Make Clock Background Reflect Player Side when Active

### LINKED ISSUE #822: Feature request: click turning white black depending on side is a nice touch
image

r3econ:
Closes https://github.com/lichess-org/mobile/issues/822 - by @ijm8710

## What

When it's your turn, the active clock background now reflects the side you're playing: white background for the white clock, black background for the black clock. A grey border is added to maintain contrast against both light and dark app themes.

UI/UX is often a matter of taste. I myself am not sure if this is an improvement. I am curious what the others think 🤔 

## Behaviour

- Side colour only applies when the clock is **active** (your turn)
- **Emergency colour always takes priority** over the side colour
- No change to inactive clock appearance

#### Light Mode

https://github.com/user-attachments/assets/0fd98c4c-8345-49e4-a76a-0fd93e5d6c14

#### Dark Mode

https://github.com/user-attachments/assets/9b38ee74-9b3a-41ae-84a5-978aca73a02f


ijm8710:
I love it! May not be for everyone but if you have the option to set clock on right/left I think a lot of people would love the ability to enable this even if it's not the default

r3econ:
Thanks for the feedback. I will proceed with making that a setting similar to _Clock position_ and report back 🤖 

image

veloce:
I'd like not to have too many settings. A settings for that one is way overkill imo. Either the clocks are better like that and we switch, or we keep the old ones.

A before/after video comparison would help here :)

Also the new clock style must fit well with all the background themes the app supports.

ijm8710:
> I'd like not to have too many settings. A settings for that one is way overkill imo. Either the clocks are better like that and we switch, or we keep the old ones.
> 
> 
> 
> A before/after video comparison would help here :)
> 
> 
> 
> Also the new clock style must fit well with all the background themes the app supports.
> 
> 

Not as beautifully formatted as the way recon makes his video demos but the only significant difference is when it's black turn it inverts it. So formerly white-attention was used regardless of if it was black or whites turn but now it gives a more black or white feel.

To me this is especially helpful when spectating a game that's in the endgame where it's not super easy to see who is black and who is white. 

As recon implied, should be no theming issue as white text was already used for player names. And the black color is given a white perimeter.

Top row: black (old vs new)
Bottom row: white (old vs new)

image

r3econ:
@ijm8710 thanks for testing it, it looks very nice 🥇 I resolved the conflicts and reopened the PR 🙂

veloce:
Still not sure about this (not sure this is an improvement). Will need to test that with the custom backgrounds too. I'll review and test after the next release.

PR #2671, opened 2026-02-22
Open PGN files from file explorer, messages,... in the App on Android

### LINKED ISSUE #2425: [Feature] Utilize android's intent/share functionality to import PGN
You can currently do this by copy, open lichess, more, load position, paste, go. Would be nice to save a few steps.

HaonRekcef:
closes #2425
Allows to open `.pgn` files from outside the app on android. Preview:

https://github.com/user-attachments/assets/96b55db6-cc74-468b-aa97-3ba4e44be74b

Successfully tested on Android 8 and Android 15

I used the package `receive_sharing_intent` which seems to only be sporadically maintained, but I couldnt find a better option.
I needed to give a direct link to github for the package because of a fix that is not released yet: https://github.com/KasemJaffer/receive_sharing_intent/pull/333

Needs to be implemented on the iOS side as well, but I don't have the devices to do it.

I changed the launchmode to single task to prevent multiple instances of the app getting created.


r3econ:
Cool stuff! 😎 Once this gets merged I'll work on the iOS support for this 🍏

android/app/src/main/AndroidManifest.xml
```diff
@@ -21,7 +21,7 @@
         <activity
             android:name=".MainActivity"
             android:exported="true"
-            android:launchMode="singleTop"
+            android:launchMode="singleTask"
```
veloce, line 24: I am not sure to understand all the implications of this change. Could you elaborate why you had to change this, and what was the "problem" with "singleTop".

According to the doc here: https://developer.android.com/guide/topics/manifest/activity-element#lmode, "singleTask" uses case are:

> Specialized launches (not recommended for general use)

Then I'm not sure to understand all the differences between the launch modes in that doc. But when I see,

- for "singleTop"
> If an instance of the activity already exists at the top of the target task, the system routes the intent to that instance through a call to its [onNewIntent()](https://developer.android.com/reference/android/app/Activity#onNewIntent(android.content.Intent)) method, rather than creating a new instance of the activity.

- for "singleTask"
> The system creates the activity at the root of a new task or locates the activity on an existing task with the same affinity. If an instance of the activity already exists and is at the root of the task, the system routes the intent to existing instance through a call to its [onNewIntent()](https://developer.android.com/reference/android/app/Activity#onNewIntent(android.content.Intent)) method, rather than creating a new one.

I'm not sure why we need to switch to singleTask, looks like singleTop should be good, no?

veloce (commented):
Except for the launch mode change, nothing to comment.

HaonRekcef:
@veloce the behavior I do not like about `singleTop` is this: 

https://github.com/user-attachments/assets/55899977-ceca-414c-9a20-faf813fec050

It opens a second app in the file explorer itself, which I always find strange.

veloce:
I see. We are kind of used to it with other applications like the browser, or a map app, though. But for lichess maybe it is strange.

veloce:
Since it is the flutter default, I'm afraid changing it could break some other parts of the app.

I tend to think we should conservatively keep "singleTop" as this is not such a big deal that the app opens in the file explorer.

With the change to single task, I'm wondering for instance, what are the differences when: opening app from notifications (should be ok), split screens (not sure here), OAuth flow, etc?

HaonRekcef:
> Since it is the flutter default, I'm afraid changing it could break some other parts of the app.
> 
> I tend to think we should conservatively keep "singleTop" as this is not such a big deal that the app opens in the file explorer.
> 
> With the change to single task, I'm wondering for instance, what are the differences when: opening app from notifications (should be ok), split screens (not sure here), OAuth flow, etc?

I think all of those issues should actually be even less likely, because with singleTask there is only one instance of the app.

Also the package used, suggests this: `<!--Set activity launchMode to singleTask, if you want to prevent creating new activity instance everytime there is a new intent.-->` https://pub.dev/packages/receive_sharing_intent

veloce:
We can test it for a while with "singleTask" and revert if there is an issue.

Another question: the screen recording you posted, was it in release or debug mode? The loading time for the cold start is quite slow I found and strangely we don't see the splash screen but a white screen.

PR #2674, opened 2026-02-23
Animated Native Splash Logo (Android 12+)

HaonRekcef:
Adds an animated native splash logo on Android 12+. Preview: 

https://github.com/user-attachments/assets/ee2d85a5-e6de-439b-bb34-3f5a636ae7f0

The animation runs for 0.75 seconds, which is slightly shorter than the time my Nothing Phone 3a and the emulators need to load the app.

I would appreciate feedback from anyone willing to test this on their device in **release mode** and share how the animation timing feels relative to the app startup time.

Notes

- The animation does not affect Flutters loading time.
- If Flutter finishes loading before the animation completes, the app UI will be shown immediately.
- On Android < 12, the previous static splash screen is used (tested on Android 8, still works).
- On iOS the static splash screen is also used, since animated splash screens are not supported natively.

## Creation Documentation for future reference

Files used to generate `animated_logo.xml`

- [spinner.svg](https://github.com/user-attachments/assets/c80dc915-a2d9-4f41-8283-d66134c9d7cc)  
  Based on:  
  https://github.com/lichess-org/lila/blob/fc960b9a82ab042f1e6538e5a18b07f3ffde6f6d/modules/ui/src/main/helper/HtmlHelper.scala#L10

- [mask.svg](https://github.com/user-attachments/assets/a58766cf-0d74-4490-b5d8-84a7cd8598c0)  
  Based on:  
  https://github.com/lichess-org/lila/blob/fc960b9a82ab042f1e6538e5a18b07f3ffde6f6d/modules/web/src/main/ui/layout.scala#L203

To help generate the XML, I used:

https://shapeshifter.design/

### DIFF
--- a/android/app/src/main/res/drawable/animated_logo.xml
+++ b/android/app/src/main/res/drawable/animated_logo.xml
@@ -40,8 +40,7 @@
         <aapt:attr name="android:animation">
             <objectAnimator
                 android:propertyName="trimPathEnd"
-                android:startOffset="300"
-                android:duration="300"
+                android:duration="250"
                 android:valueFrom="0"
                 android:valueTo="1"
                 android:valueType="floatType"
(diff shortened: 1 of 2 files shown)


veloce:
> I would appreciate feedback from anyone willing to test this on their device in release mode and share how the animation timing feels relative to the app startup time.

Will try on a OnePlus 6T.

HaonRekcef:
> Will try on a OnePlus 6T.

You should not see an animation though, as this devices only has up to Android 11 according to a quick search. Nevertheless good to test if it doesn't break anything.

veloce:
Merging it to let people in beta testing group see it.

PR #2684, opened 2026-02-27
fix: use buildScreenRoute when replacing AnalysisScreen

### LINKED ISSUE #2682: Enabling variants in standalone analysis (ie horde) turn background black and lock out the Backswipe
https://github.com/user-attachments/assets/eb2a7a37-dfd1-476c-93c1-76fa3968dbfc

tom-anders:
closes #2682

### DIFF
--- a/lib/src/view/analysis/analysis_screen.dart
+++ b/lib/src/view/analysis/analysis_screen.dart
@@ -602,8 +602,9 @@ class _BottomBar extends ConsumerWidget {
                           .read(analysisControllerProvider(options).notifier)
                           .clearSavedStandaloneAnalysis();
                       Navigator.of(context, rootNavigator: true).pushReplacement(
-                        PageRouteBuilder<dynamic>(
-                          pageBuilder: (context, animation, secondaryAnimation) => AnalysisScreen(
+                        buildScreenRoute<dynamic>(
+                          context,
+                          screen: AnalysisScreen(
                             options: (options as Standalone).copyWith(variant: variant),
                           ),
                           transitionDuration: Duration.zero,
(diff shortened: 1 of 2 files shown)


lib/src/view/analysis/analysis_screen.dart
```diff
@@ -602,8 +602,9 @@ class _BottomBar extends ConsumerWidget {
                           .read(analysisControllerProvider(options).notifier)
                           .clearSavedStandaloneAnalysis();
                       Navigator.of(context, rootNavigator: true).pushReplacement(
-                        PageRouteBuilder<dynamic>(
-                          pageBuilder: (context, animation, secondaryAnimation) => AnalysisScreen(
+                        buildScreenRoute<dynamic>(
+                          context,
+                          screen: AnalysisScreen(
```
veloce, line 607: I don't get the fix since you are not using the transitionDuration parameter introduced above.

tom-anders: I *do* use it, it's just that the PageRouteBuilder also has a parameter with that exact name, so it doesn't show up in the diff. See line 610

tom-anders: A bit more context: When implementing changing the variant, I initially used `Navigator.of(context, rootNavigator: true).pushReplacement(buildScreenRoute<dynamic>(...)`. However, it looked a bit weird because of the transition animation. That's why I switched to `Navigator.of(context, rootNavigator: true).pushReplacement(PageRouteBuilder<dynamic>(...)` because of the `transitionDuration` parameter. But I didn't take into acount that not using `buildRoute` breaks the app's background. So that's why I switched to `buildRoute` here and added the `transitionDuration` parameter to that function.

PR #2702, opened 2026-03-02
Grey logo  + increase animation duration

HaonRekcef:
- Adds a grey logo below the animation, to improve the way it looks if the app loads too quickly
- Increases the animation duration slightly to 0.9s (longer is not recommeded by android docs)
- slight delay before the animation plays of 0.1s (longer is not recommeded by android docs)

Light mode:

https://github.com/user-attachments/assets/61debda0-0e0f-43f7-8006-b5189edf823d

Dark mode:

https://github.com/user-attachments/assets/ba663051-b11d-4bf6-9914-23274b80bd9f


veloce:
Thanks :)

PR #2715, opened 2026-03-07
feat: add "Next to play" side choice for games from position

### LINKED ISSUE #2714: Add "Next to play" side choice when starting a game from a position
## Feature request

When starting a game against the computer from a custom position (e.g. via **Continue from here** after solving a puzzle), the side picker currently offers White, Random, and Black.

### Use case

A common workflow is:
1. Solve a puzzle
2. Click "Continue from here" → "Play against computer"
3. Play as the winning side against a high-level bot under time pressure

Because the user always wants to play as the side that has the move (the winning side), they currently have to manually switch the side selection each time. Since the last selection is remembered, this means toggling between White and Black depending on the puzzle.

### Proposal

Add a 4th option — **"Next to play"** — that automatically selects whichever side has the move in the current FEN position. This option would:
- Only appear when starting from a custom position (i.e. when `initialFen` is provided)
- Be the default selection in that context
- Not appear when starting a standard new game (where it has no meaning)

This small UX improvement eliminates a repetitive manual step in a common puzzle-to-game workflow.

lacostej:
## Summary

- Adds a **"Next to play"** option to the side picker when starting an offline computer game from a custom position (via "Continue from here")
- Automatically selects the side that has the move in the FEN, removing the need to manually toggle White/Black each time
- Shows **"White to play"** or **"Black to play"** below the mini board preview so the user knows which side moves first
- Only shown when `initialFen` is provided; defaults to "Next to play" in that context
- Adds 4 widget tests covering the new behavior
- "Next to play" label is hardcoded in English for now, following the project convention of adding translations after the feature is stable

Closes #2714

## Screenshot

next-to-play-1

## Video demo

https://github.com/user-attachments/assets/ce5695f2-3895-4ee2-9885-102970b74161

## Test plan

- [x] `flutter analyze` — no issues
- [x] `dart format` — properly formatted
- [x] `flutter test` — all tests pass (4 new tests added)
- [x] Manual testing on Android emulator (Pixel, API 35):
  - Solve a puzzle → Continue from here → Play against computer
  - Side defaults to "Next to play"
  - "White to play" / "Black to play" shown below mini board
  - Correctly resolves to white/black based on position's side to move
  - "Next to play" does not appear when starting a standard game (no initialFen)

### DIFF
--- a/lib/src/model/offline_computer/offline_computer_game_preferences.dart
+++ b/lib/src/model/offline_computer/offline_computer_game_preferences.dart
@@ -65,12 +65,18 @@ class OfflineComputerGamePreferences extends Notifier<OfflineComputerGamePrefs>
 enum SideChoice {
   white,
   random,
-  black;
+  black,
+  nextToPlay;
 
-  Side? toSide() => switch (this) {
+  /// Resolves the side choice to a [Side].
+  ///
+  /// When [fen] is provided and this is [nextToPlay], returns the side to move
+  /// from the FEN. Returns `null` for [random] (caller should pick randomly).
+  Side? toSide({String? fen}) => switch (this) {
     SideChoice.white => Side.white,
     SideChoice.random => null,
     SideChoice.black => Side.black,
+    SideChoice.nextToPlay => fen != null ? Setup.parseFen(fen).turn : null,
   };
 
   static SideChoice fromSide(Side? side) => switch (side) {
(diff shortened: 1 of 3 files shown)


veloce:
Nice! thank you.

I have one suggestion: we don't see which side is to play in the bottom sheet. Perhaps we could add a "White to play" or "Black to play" below the mini board. These strings are already translated.

lacostej:
> Nice! thank you.
> 
> I have one suggestion: we don't see which side is to play in the bottom sheet. Perhaps we could add a "White to play" or "Black to play" below the mini board. These strings are already translated.

OK. let me see what I can do. I also need to update the code.

lacostej:
Screenshot 2026-03-10 at 12 52 55

Added the text below the mini board. Text for "Next to Play" still hardcoded. 

Are we ok with this text as is? Can I move it to translations?

veloce:
Thanks! "Next to play" is fine I think. Let's keep it hardcoded for now.

ijm8710:
Hi @lacostej this was very cool
And it worked everywhere else but once place, is it supposed to
Be able to work from
The editor?

PR #2718, opened 2026-03-07
Broadcasts: support `customScoring` and multi time-control stats

### LINKED ISSUE #1834: Broadcast Update: Scores can now be customized
With update: https://github.com/lichess-org/lila/pull/17599
Image

In app show now:

Image

allanjoseph98:
* Support custom scoring for broadcasts. Closes #1834. Show points scored and color code the result. Affects the: 
    i. Game screen
    ii. Player's games screen
    iii. Boards screen
    iv. Teams tab

https://github.com/user-attachments/assets/60d536bf-b329-4371-9f47-6a53250e6cbe

* Support multi time-control broadcast stats. Affects the
   i. Players screen
   ii. Player's games screen

https://github.com/user-attachments/assets/3522b15a-b062-438a-af6d-486641fc2d0d

### DIFF
--- a/lib/src/view/broadcast/broadcast_player_results_screen.dart
+++ b/lib/src/view/broadcast/broadcast_player_results_screen.dart
@@ -345,7 +345,7 @@ class _OverallStatPlayer extends StatelessWidget {
                                 mainAxisAlignment: .center,
                                 spacing: 4.0,
                                 children: [
-                                  Icon(tc.icon, size: 16),
+                                  if (performances.length > 1) Icon(tc.icon, size: 16),
                                   Text(
                                     '$p${games.count((g) => g.fideTC == tc) < 4 ? '?' : ''}',
                                     style: const TextStyle(fontSize: 16),
@@ -370,7 +370,7 @@ class _OverallStatPlayer extends StatelessWidget {
                                 mainAxisAlignment: .center,
                                 spacing: 4.0,
                                 children: [
-                                  Icon(tc.icon, size: 16),
+                                  if (ratingDiffs.length > 1) Icon(tc.icon, size: 16),
                                   Text(
                                     ratingsMap?.get(tc)?.toString() ?? '',
                                     style: const TextStyle(fontSize: 16),


ijm8710:
Hi @allanjoseph98 

This is very cool, I had a few open tickets for some of the things in your first video so I'm happy to see them.

Curious to grab your thoughts on 2 things

- I love how you colored wins/losses on the player results screen (consistent with web). The scenario where you have a red rating change while the tie is non-colored can cause some imbalance/noise. 
Personally I think it makes sense to move the rating change to the left side after the rating itself rather than have it clashing with the two right next to each other.
What are your thoughts?

  - benefit 1 is it allows the user to see the win/loss color with less noise
  - Benefit 2 is it becomes similar to the "standings" section  where the rating change is right after the rating. Plus that just makes sense for it to be naturally anyway

- Not everyone has agreed with me but I really like how the player results screen on web is numbered down. It makes it much easier to follow an event with a bunch of rounds in a row so you know where you last left off. And this is how it is on web too. Thoughts?

image
image

allanjoseph98:
1. The rating change on the results screen is for the player you're viewing not for the opponent in the list. You can't move one player's rating diff next to their opponent's rating.
2. This is a separate feature request that you should file an issue for.

ijm8710:
> 1. The rating change on the results screen is for the player you're viewing not for the opponent in the list. You can't move one player's rating diff next to their opponent's rating.
> 

I meant move the rating diff next to the rating itself on the players games list. This will mimic how it is for standings. And by doing so, it would allow the win (green) loss (red) tie (no color) to be on right side alone without the distraction of the rating change while allowing the rating change to be next to the rating itself 

image
image

allanjoseph98:
I understood you. You didn't understand me. Those points are what Gergely gained for beating Marcell. Moving Gergely's rating gain for that game next to Marcell's rating reads like Marcell gained 3 rating points for that game even though he lost. Suffer through this hideous diagram for your understanding.

ijm8710:
> 

Ah apologies and thanks for taking the time to break that down. Now I hear what you're saying

I still think it gets a little combative where the color from the rating change and the color from the result are stacked in the exact same column

I'm not sure the best way around that whether perhaps putting it next to the name or even to the right in another column like on web.

I'm assuming most feel it's fine as is

I hear you on what you're saying now but seeing two colors at once in the same place that won't always be the same to me can feel a little misdirection

ijm8710:
.. Or even part of me feels then maybe to remove the color for the rating diff. To me the result color trumps the importance of the rating diff color

PR #2721, opened 2026-03-07
Fix: Add variant ending to the controller and UI in offline game screen

### LINKED ISSUE #2713: Play against computer has no end game dialog for multiple varients
### What is the problem?

When playing against a computer, there is no end game dialog for atomic chess. When the game ends there is no dialog and the pieces are frozen.

On further testing, the endgame condition doesn't work for antichess, atomic,  king of the hill, three check, racking kings and horde.

https://github.com/user-attachments/assets/104318c0-7dbd-4cd3-9431-cda0a5a177fc

1. Start a game against computer with the atomic chess variant
2. Win the game

### App version

0.22.0

### Device Info

Pixel 7a/Android 16

### Additional information

_No response_

nav-28:
Update the controller and the game screen to check for variant ending condition. 

Closes #2713

### DIFF
--- a/lib/src/view/offline_computer/offline_computer_game_screen.dart
+++ b/lib/src/view/offline_computer/offline_computer_game_screen.dart
@@ -940,6 +940,8 @@ class OfflineComputerGameResultDialog extends StatelessWidget {
       variant: game.meta.variant,
       status: game.status,
       lastPosition: game.lastPosition,
+      winner: game.winner,
+      isThreefoldRepetition: game.isThreefoldRepetition,
     );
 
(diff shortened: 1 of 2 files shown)


tom-anders:
There are some specific texts defined for variant endings, e.g. "raceFinished" : https://github.com/lichess-org/mobile/blob/05b713e18b3c2655e5d3bfcd0a819f9d3c3160f2/lib/src/view/game/status_l10n.dart#L63

Could we reuse that gameStatusL10n function here?

veloce (approved):
Thanks!

PR #2741, opened 2026-03-10
fix: disable predictive back gesture for tab navigators on Android

HaonRekcef:
Currently we have the following issues on newer android versions with predictive back: 

https://github.com/user-attachments/assets/21cd8587-84e0-4484-8fa4-bbbbc576cf13

Notice how the broadcast list and even the home tab navigation got closed too. This is related to flutter issue: https://github.com/flutter/flutter/issues/152323

This PR overrides the predictive back animation for the tab navigators which appears to fix the issue.
Preview: 

https://github.com/user-attachments/assets/453b039a-f43a-4d37-8ae3-c59455714ed3

This fixes the tabs closing unintentionally, with the cost that the predictive back animation is disabled for tab navigation. Predictive back still works correctly for root navigator screens.


veloce:
Not sure about this yet. The flutter issue is pretty old too... but have you done extensive research about it? Changing the page transition to fix the issue is not something we want to do if we can avoid it.

HaonRekcef:
> Not sure about this yet. The flutter issue is pretty old too... but have you done extensive research about it? Changing the page transition to fix the issue is not something we want to do if we can avoid it.

I agree that changing the page transition is not something we want to do. But I can still replicate the issue in the latest flutter release and there are some what recent discussions and open pr.

But having the bug isnt better imo. This is a screen recording on my real device that maybe shows the issue more clearly.

https://github.com/user-attachments/assets/3fc1f547-671f-4938-897b-e8638888cbef

See the frames where the broadcast list closes itself. 

If we can fix this just by disabling predictive back for the tab navigators, I think it is worth it.
The other approaches are likely more involved (at least I wouldn't know how to fix it yet) and probanly would mean complicating our navigation logic to patch something that Flutter should fix, not us.

veloce:
Some recent work has been done regarding predictive back gestures so I don't think this is relevant anymore.

HaonRekcef:
this is still relevant on my Nothing Phone with android 16 and predictive back enabled

PR #2744, opened 2026-03-10
Crazyhouse fixes

tom-anders:
- Fix drop moves in offline computer games (use Move.parse instead of NormalMove.fromUci)
- Fix position rule when initial FEN is given, analogous fix to https://github.com/lichess-org/mobile/pull/2729
- Support DropMoves in engine lines, opening explorer, tablebase (also by switching to Move.parse)


veloce (commented):
We should also have basic tests that ensure crazy house drop moves are working for the other implied widgets like engine lines and the explorer screen.

PR #2753, opened 2026-03-12
[WIP] feat: enable Crashlytics NDK native symbol upload

### LINKED ISSUE #2752: Catching native crashes on Android
#2749 made me wonder if native crashes are caught by crashlytics. If not we would need to enable them and upload debug symbols. I checked the code and while iOS seems to be configured to report native crashes, android ones don't seem to be collected.

lacostej:
## Summary

- Upload the `native-debug-symbols.zip` (which includes both native `.so` symbols and Dart `libapp.so` symbols) to Firebase Crashlytics, replacing the previous Dart-only symbols upload
- Enables symbolicated native crash reports for bundled libraries like `libflutter.so` and Stockfish
- The zip is already built by `add-debug-symbols.sh` and uploaded to Google Play Console, but was not being sent to Crashlytics

Closes #2752

## Test plan

- [ ] Verify the deploy workflow succeeds on next Play Store deployment
- [ ] Verify native crash symbols appear in the Firebase Crashlytics console

### DIFF
--- a/.github/workflows/deploy_play_store.yml
+++ b/.github/workflows/deploy_play_store.yml
@@ -69,5 +69,5 @@ jobs:
 
       - name: Install Firebase CLI
         run: curl -sL https://firebase.tools | bash
-      - name: Deploy symbols to Firebase Crashlytics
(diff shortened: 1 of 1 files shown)


lacostej:
~Well I just realized there is some non conventional upload being done under `.github/workflows/deploy_play_store.yml`. I'll dig a bit more~

Updated, I hope I got it right now

veloce:
Thanks! we will test it in the next deploy then.

PR #2758, opened 2026-03-13
feat: support clockTenths pref from the website

tom-anders:
The current behavior is to always show tenths below 10s, this PR uses the existing preference from account settings to also support "never" and "always".

Needs https://github.com/lichess-org/lila/pull/19799


tom-anders:
Ah yes, I forgot about correspondence. Will test and post some Screenshots

veloce:
Not only correspondence, but classical too.

tom-anders:
Looks fine with hours (see screenshot below). And in correspondence, we're actually using a separate clock widget, so the preference does not actually have any effect there.

Screenshot_1773774802

Issue #2742, opened 2026-03-10
The variant icons are a nice touch and visually distinctive from one another. Can we show them in the selector rather than only once we already selected?


ijm8710:
- Create a game
- Play a bots 
- Otb
- Challenge someone
- Play computer 
- Explorer (I think explorer variant support is still to come tho)

I'm not sure there's any valid places that may require it here but I may be wrong
- Tv
- Tourney
- Study

There's a chance I'm missing one or two but I think this covers Most

PR #2760, opened 2026-03-13
Variant icons in choice picker

r3econ:
- Resolves https://github.com/lichess-org/mobile/issues/2742

#### Before & After

SCR-20260313-ufoy-tile

### DIFF
--- a/test/view/analysis/analysis_screen_test.dart
+++ b/test/view/analysis/analysis_screen_test.dart
@@ -243,7 +243,7 @@ void main() {
       await tester.tap(find.text('Play against computer'));
       await tester.pumpAndSettle(); // wait for play menu to open
       // Variant we set previously should be preselected
-      expect(find.text('Atomic'), findsOneWidget);
+      expect(find.textContaining('Atomic'), findsOneWidget);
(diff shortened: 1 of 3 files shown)


HaonRekcef:
Looks nice
Tom, Veloce any reason we dont support Variant Correspondce yet or just an oversight?

tom-anders:
> Looks nice Tom, Veloce any reason we dont support Variant Correspondce yet or just an oversight?

Hmm, conditional premoves in Crazyhouse might need to be tested (but I think it should just work), don’t see any other issues

tom-anders:
> > Looks nice Tom, Veloce any reason we dont support Variant Correspondce yet or just an oversight?
> 
> Hmm, conditional premoves in Crazyhouse might need to be tested (but I think it should just work), don’t see any other issues

No changes needed on the app for conditional premoves in crazyhouse, the layout looks fine and no changes needed for `forecast.dart`.

However, while testing I found that the server currently does accept drop moves in conditional premove lines, but then does not actually play them when the opponent has moved.

I'll open an issue in lila github for that

Issue #2738, opened 2026-03-09
Would it make sense for the chess clock to have a low time warning like real time games do


ijm8710:
Flashes clock red and sounds an alert, perhaps at 20 seconds?

llsant:
Maybe but somewhat less obnoxius and it must be user-configurable (for good players 20 seconds is a lot and they want to focus, not have something flashing strong and making a sound xD)

PR #2771, opened 2026-03-15
Clock Tool -  Low time warning settings

r3econ:
- Resolves https://github.com/lichess-org/mobile/issues/2738
- Adds settings for low time warnings for the clock tool
- @ijm8710 please have a look if that's what you were thinking about

#### Demo

SCR-20260315-tsgi-tile

https://github.com/user-attachments/assets/ecd57f0e-1175-4a5a-b24d-e1f0c61eb0d8

### DIFF
--- a/lib/src/model/clock/clock_tool_controller.dart
+++ b/lib/src/model/clock/clock_tool_controller.dart
@@ -68,8 +68,9 @@ class ClockToolController extends Notifier<ClockState> {
     if (activeSide == null) return;
     if (_hasPlayedLowTimeSound[activeSide]!) return;
     final warning = ref.read(clockToolPreferencesProvider).lowTimeWarning;
-    final initialTime =
-        activeSide == ClockSide.top ? state.options.topTime : state.options.bottomTime;
+    final initialTime = activeSide == ClockSide.top
+        ? state.options.topTime
+        : state.options.bottomTime;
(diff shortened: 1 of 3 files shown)


ijm8710:
Hey @r3econ love your work you've been contributing but I feel for this one a setting isn't really needed. To be honest I'm not even that picky on when the warning kicks in, 20 or 30 seconds to match a regular timed game in the app works for me but I'm down for whatever, but I don't think it really needs a 7th button. The sound that you hear when time is running low in an in-app timed game would be nice. But I don't see a big deal where people
Will complain or care much when the warning kicks in

r3econ:
Yes, I've been thinking that a setting is an overkill.

We could simplify it by only having the animation (see the video) and use the already existing time interval of 10% of the time set for the threshold when the animation would start. And remove the setting whatsoever.

The sound you're talking about is being played at the emergency threshold. This is already there

ijm8710:
> Yes, I've been thinking that a setting is an overkill.
> 
> We could simplify it by only having the animation (see the video) and use the already existing time interval of 10% of the time set for the threshold when the animation would start. And remove the setting whatsoever.
> 
> The sound you're talking about is being played at the emergency threshold. This is already there

Sounds great to me!

r3econ:
Ok, I'll close this PR and implement the change in a new one 👍🏾

PR #2777, opened 2026-03-16
fix: add request timeout to prevent indefinite loading

lacostej:
## Summary
- Adds a 15-second default timeout to all HTTP requests in `LichessClient.send()`
- Prevents screens from getting stuck in loading state when requests hang indefinitely (e.g. friends screen)

The timeout is applied at the `LichessClient` layer, so it covers all API requests without needing per-call configuration. The timeout is a `static const` that can be referenced if specific calls need to document why they override it.

## Test plan
- [x] All existing tests pass (743/743)
- [ ] Verify friends screen loads or shows error within 15 seconds on poor network
- [ ] Verify normal API calls are unaffected by the timeout

EDIT: (maintainer)
I removed the issue marked as fixed.

### DIFF
--- a/lib/src/network/http.dart
+++ b/lib/src/network/http.dart
@@ -332,6 +332,8 @@ class _RegisterCallbackClient extends BaseClient {
 class LichessClient implements Client {
   LichessClient(this._inner, this._ref);
 
+  static const defaultRequestTimeout = Duration(seconds: 15);
+
   final Ref _ref;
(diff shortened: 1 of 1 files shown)


veloce:
I merged it because it is good to have a global timeout. But I don't think it fixes the loading issue in friends screen. As I understand the issue the server returns data, so we may have another problem in the app.

Issue #2747, opened 2026-03-10
Variant descriptions of what they are


ijm8710:
Depending on how challenging it is to code.

it is nice that in web the variants have a one sentence description of what they are. I know it couldn’t be same in app to
Show that outright but thoughts to an info button or maybe a down arrow to expand and show the description?

r3econ:
@ijm8710, I would have a relatively simple solution. It works well for iOS. I would need to spend more time on Android. What are your initial thoughts when you look at the screenshots? I guess android could look good if the description font would be smaller, and variant title left-aligned?

<img width="2468" height="2563" alt="Image" src="https://github.com/user-attachments/assets/b04df06a-7533-4da1-97e0-686f2b20a72e" />

ijm8710:
> @ijm8710, I would have a relatively simple solution. It works well for iOS. I would need to spend more time on Android. What are your initial thoughts when you look at the screenshots? I guess android could look good if the description font would be smaller, and variant title left-aligned?
> 
> <img width="2468" height="2563" alt="Image" src="https://github.com/user-attachments/assets/b04df06a-7533-4da1-97e0-686f2b20a72e" />

I think iOS especially works. @tom-anders since you have good opinions usually and helped land us variants in first place, what are your thoughts before it's shared to Veloce. It's slightly busy but I do think the descriptions are helpful to some degree.
I'd say yes on the iOS design, agree?

tom-anders:
> [@ijm8710](https://github.com/ijm8710), I would have a relatively simple solution. It works well for iOS. I would need to spend more time on Android. What are your initial thoughts when you look at the screenshots? I guess android could look good if the description font would be smaller, and variant title left-aligned?
[...]

PR #2780, opened 2026-03-16
Variant Description in Variant Picker

r3econ:
- Adds variant description to the variant picker
- Resolves https://github.com/lichess-org/mobile/issues/2747

### Demo

Screenshot_20260316_221519-tile

### DIFF
--- a/test/view/analysis/analysis_screen_test.dart
+++ b/test/view/analysis/analysis_screen_test.dart
@@ -142,6 +142,8 @@ void main() {
       expect(find.textContaining('Racing Kings'), findsOneWidget);
       expect(find.textContaining('Crazyhouse'), findsOneWidget);
 
+      await tester.ensureVisible(find.textContaining('Horde')); // scroll if needed
+      await tester.pumpAndSettle();
       await tester.tap(find.textContaining('Horde'));
       await tester.pumpAndSettle(); // wait for dialog to close and new variant to be loaded
 
@@ -155,6 +157,8 @@ void main() {
       await tester.pumpAndSettle(); // wait for menu to open
       await tester.tap(find.text('Variant'));
       await tester.pumpAndSettle(); // wait for dialog to open
+      await tester.ensureVisible(find.textContaining('Crazyhouse')); // scroll if needed
+      await tester.pumpAndSettle();
       await tester.tap(find.textContaining('Crazyhouse'));
       await tester.pumpAndSettle(); // wait for dialog to close and new variant to be loaded
 


lib/src/widgets/variant_app_bar_title.dart
```diff
@@ -12,17 +16,32 @@ class VariantLabel extends StatelessWidget {
...
-            alignment: PlaceholderAlignment.middle,
-            child: Icon(variant.icon, size: style.fontSize, color: style.color),
+    final alignment = Platform.isAndroid ? CrossAxisAlignment.start : CrossAxisAlignment.center;
+    final descriptionStyle = Platform.isAndroid
+        ? TextStyle(
+            height: 1.2,
+            fontSize: 12.0,
+            color: DefaultTextStyle.of(context).style.color?.withValues(alpha: 0.6),
+          )
+        : Styles.subtitle.copyWith(color: textShade(context, Styles.subtitleOpacity));
```
r3econ, line 26: As discussed in https://github.com/lichess-org/mobile/issues/2747, iOS and Android design differs here slightly. The Android style is the same as in the puzzles tab and seems to fit nicely

Image

ijm8710:
Hey @veloce @r3econ the sizing of the menu feels enlarged versus the pr. Do you think it may look better without a scroll. There's a tiny scroll for the first picture when starting a game (half crazyhouse is below the fold).
And when you challenge someone (photo 2) which includes the "from position option" the scroll becomes even more (standard game is completely out of view).

image


image

r3econ:
I guess it can't be fixed without sacrificing something. I think whether the list becomes scrollable depends on the device size and font size (that can also be enlarged in system settings)

We could:

1. Remove the descriptions
2. Make the description font even smaller or allow only single line and truncate it
3. Think of different design 
4. Accept how it is. Maybe tweak it slightly so Standard is always visible when the menu opens to keep it consistent

ijm8710:
> I guess it can't be fixed without sacrificing something. I think whether the list becomes scrollable depends on the device size and font size (that can also be enlarged in system settings)
> 
> We could:
> 
> 1. Remove the descriptions
> 2. Make the description font even smaller or allow only single line and truncate it
> 3. Think of different design 
> 4. Accept how it is. Maybe tweak it slightly so Standard is always visible when the menu opens to keep it consistent 

Hmm my confusion lies when I compare your iOS photo to the fist photo I replied with. Is yours a smaller font? It must be because your mockup has antichess description one line and I see it as two

I have never changed my system text size, it's the default.

The sizing of the mockup seems perfect.

I'm not sure if all font sizes are uniform in the app. If all fonts in the app are not-uniform, I feel it wouldn't be a bad trade off to shrink it slightly so that all 10 options (including from variant) fit without a scroll assuming it still looked big enough.

If that's not plausible then  I guess it's fine as is but I figured it was a possibility because the mockup seemed to fit everything easier

veloce:
Let's accept how it is... it just can't be perfect in all devices.

ijm8710:
> I guess it can't be fixed without sacrificing something. I think whether the list becomes scrollable depends on the device size and font size (that can also be enlarged in system settings)
> 
> We could:
> 
> 1. Remove the descriptions
> 2. Make the description font even smaller or allow only single line and truncate it
> 3. Think of different design 
> 4. Accept how it is. Maybe tweak it slightly so Standard is always visible when the menu opens to keep it consistent 

Rafa last message on this one promise :) on the challenge a friend screen, crazy house is completely below the fold so you may not even know it exists (see photo).
You mentioned sacrificing, but wouldn't just making all text ~70-80%  proportionately of the size it is now make it just read a good bit better 
For some reason the text size in these menus is bigger than it is in like any other place in the app

Veloce is the boss, don't want to drag out what he says is fine but to me that's such a tiny simple solution that I think might improve readability without any true shift or even time spent to fix it. 


image

r3econ:
@ijm8710, I think it could be that the text size is context-based and it indeed is bigger in some screens. In such case it would be a bug. I'll have a look into it and report back with my findings 🕵️

PR #2783, opened 2026-03-17
crazyhouse: fix drop moves in online games

tom-anders:
There's a special `drop` topic that the server sends/expects, instead of the normal `move` topic.

When receiving, luckily we can just parse the `drop` event in the same way as the `move` event. `drop` contains the dropped piece as an additional field, but we don't need that since we're parsing it from the UCI anyway.

### DIFF
--- a/lib/src/view/game/game_loading_board.dart
+++ b/lib/src/view/game/game_loading_board.dart
@@ -220,6 +220,7 @@ class StandaloneGameLoadingContent extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
+    final lastMove = position?.lastMove;
     return Shimmer(
       child: SafeArea(
         child: GameLayout(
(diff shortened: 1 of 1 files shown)


lib/src/view/game/game_loading_board.dart
```diff
@@ -229,7 +230,14 @@ class StandaloneGameLoadingContent extends StatelessWidget {
             variant: Variant.standard,
             pockets: null,
           ),
-          lastMove: position?.lastMove,
+          lastMove: switch (lastMove) {
+            // In crazyhouse games, the "ongoing games" endpoint does not return the correct UCI for crazyhouse games,
+            // e.g. instead of P@c4 the UCI will be c4c4.
```
tom-anders, line 235: @ornicar FYI

veloce: Can be wrong here but `P@c4` does look like a SAN notation, not UCI.

tom-anders: I think for crazyhouse drops, SAN and UCI is actually the same: https://github.com/fairy-stockfish/chess-variant-standards?tab=readme-ov-file#coordinate-notation

tom-anders: ah, looks like it's intentional:

https://github.com/lichess-org/lila/blob/060b7fcc2afbb4eb53e376fa5a02ecc4ca5c0b0b/modules/core/src/main/game/Game.scala#L109

Issue #2253, opened 2025-10-07
There is a pull refresh for own profile should there be one for other players profiles


ijm8710:
If you open the profile of a player from watch TV  their recent game log can be stale and miss the recently completed game.

Pull refresh should probably be here but honestly this should probably update dynamically automatically

PR #2803, opened 2026-03-19
Pull to Refresh in User Screen

r3econ:
- Adds pull to refresh loading indicator in the user screen
- Resolves https://github.com/lichess-org/mobile/issues/2253

#### Demo 

https://github.com/user-attachments/assets/76d59335-789a-4dd6-8efa-3ab108a5f462

### DIFF
--- a/lib/src/view/user/user_screen.dart
+++ b/lib/src/view/user/user_screen.dart
@@ -139,7 +139,12 @@ class _UserScreenState extends ConsumerState<UserScreen> {
 }
 
 class _UserProfileListView extends ConsumerWidget {
-  const _UserProfileListView(this.data, this.isLoading, this.setIsLoading, {required this.onRefresh});
+  const _UserProfileListView(
+    this.data,
+    this.isLoading,
+    this.setIsLoading, {
+    required this.onRefresh,
+  });
 
   final UserScreenData data;
   final bool isLoading;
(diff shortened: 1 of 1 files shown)


ijm8710:
Hey @r3econ thanks for adding this. You're killing it! 

Had 2 quick questions for ya if you didn't mind

1) let's use safari on iOS if you're familiar with it. When you pull to refresh you feel the haptic as you're pulling down. For some reason, maybe it's a result of the app being built in flutter, the haptic in lichess beta, you don't feel until after it's released which feels a little less natural than how it's handled throughout the rest of the iOS ecosystem. And it feels a little weaker/less full. Any chance you notice what I'm saying and is that something the feasibly can be refactored to work more like normal-iOS or because of flutter that's just how the refresh works here?

2) pull to refresh is great. But Veloce also added a lot of natural refresh dynamically without needing to pull. For example on the home screen once you finish a game, and return back to home, you should see the game listed in 10 recent games without needing to pull. Do you think player profiles should dynamically refresh similarly? I still think it makes sense to have the pull to refresh regardless, but curious your thoughts on the dynamic refresh automatically or you think pull to refresh is good enough. The causality is the data is stale unless a user actions it rather than it always being up to date

r3econ:
@ijm8710, good input. I think creating a GitHub Issue for the first point you mentioned would be a good idea. I could definitely look into it and make haptic feedback feel more native

The second is out of scope (I think) since as you say the dynamic refresh is implemented in critical places and for non-critical places a simple pull to refresh will suffice

PR #2817, opened 2026-03-21
fix(play): prevent duplicate correspondence challenge submissions

### LINKED ISSUE #2815: The challenge panel doesn't dismiss after the button was pressed
### What is the problem?

If one creates a challenge against another player, it is very confusing how the creation panel doesn't disappear after the creation button was pressed. 

The panel should disappear and it would be cool if some kind of notification/toast/whatever appropriate would be displayed to confirm to the user that the game was successfully created. 

Currently it happens often that s user ends up creating multiple games of the same kind because he doesn't get any kind of feedback from the UI.

https://github.com/user-attachments/assets/c4cfecef-a2a5-4554-8812-c87b4963799a

1. Go to a player's profile
2. Create a challenge
3. Notice how the panel doesn't disappear

### App version

0.22.6

### Device Info

Pixel 10

### Additional information

_No response_

zerox13:
Fixes #2815 

Improves correspondence challenge submission flow in create_challenge_bottom_sheet to make it safer and clearer for users.
- Adds explicit in-flight state to prevent duplicate correspondence/unlimited challenge submissions.
- Shows a loading indicator on the submit button while the challenge request is pending.
- Hardens async UI handling by capturing l10n/root context before await and using consistent success/error snackbars after the sheet closes.

### DIFF
--- a/lib/src/view/play/create_challenge_bottom_sheet.dart
+++ b/lib/src/view/play/create_challenge_bottom_sheet.dart
@@ -424,7 +424,8 @@ class _CreateChallengeBottomSheetState extends ConsumerState<CreateChallengeBott
                               }
                             : null,
                     },
-                    child: isWaitingForCorrespondenceChallenge &&
+                    child:
+                        isWaitingForCorrespondenceChallenge &&
                             (timeControl == ChallengeTimeControlType.correspondence ||
                                 timeControl == ChallengeTimeControlType.unlimited)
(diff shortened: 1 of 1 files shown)


veloce:
Closing as there was a much simpler fix: https://github.com/lichess-org/mobile/pull/2825/changes/4bb7ef75fb6ae30efc155e5bf90ccbb8b16de076

PR #2827, opened 2026-03-22
feat: support opening challenge and game deep links

### LINKED ISSUE #2734: Catch "open link in app" intents to join games in the app. Also add android:autoVerify.
Right now, as of 2026-08-03, I can open exercises in the app. For instance, if I have a Discord or Whatsapp message with the following link:

https://lichess.org/training/LqCuD

I can open it in the app and it works, **however I need to manually enable it in applications -> lichess -> defaults -> open links in app -> lichess.org -> enable toggle**.

However, that doesn't work for games. If I try to open

https://lichess.org/zkW7gb3z

I just go into my web browser. The new lichess app is not catching that link even with the "lichess.org" domain manually enabled for the app in the system settings. This is pretty inconvenient because if someone challenges you to a game, and you are on mobile, you can't play on the app (at least I haven't found a way to manually enter the game ID on the app itself). So you end up playing in your mobile browser.

Also, as I mentioned previously, even training exercises only open in the app if you manually enable it, which didn't happen with the old app. I took the time to investigate the issue and both AndroidManifest.xml's. It turns android:autoVerify is not enabled in the new app manifest. However, the domain is correctly set up for both apps, most of the work is already done:

`
 curl https://lichess.org/.well-known/assetlinks.json
`

`
[
  {
    "relation": [
      "delegate_permission/common.handle_all_urls"
    ],
    "target": {
      "namespace": "android_app",
      "package_name": "org.lichess.mobileapp",
      "sha256_cert_fingerprints": [
        "F3:7D:99:F0:1B:15:9F:55:05:F8:1A:93:E5:6E:72:5D:46:08:50:A7:ED:09:A8:D2:63:3C:98:01:FA:BD:0E:B9"
      ]
    }
  },
  {
    "relation": [
      "delegate_permission/common.handle_all_urls"
    ],
    "target": {
      "namespace": "android_app",
      "package_name": "org.lichess.mobileV2",
      "sha256_cert_fingerprints": [
        "5B:62:4B:BF:9D:72:D3:18:EA:4F:9E:E3:E2:DA:EB:A7:72:19:E5:97:11:6C:B5:93:12:30:7A:AD:18:44:94:5E"
      ]
    }
  },
  {
    "relation": [
      "lookalikes/allowlist"
    ],
    "target": {
      "namespace": "web",
      "site": "https://lichess.org"
    }
  },
  {
    "relation": [
      "lookalikes/allowlist"
    ],
    "target": {
      "namespace": "web",
      "site": "https://lichess-org.github.io"
    }
  }
]
`

To summarize, add autoVerify to the manifest and also handle game ID links in the app.

Have a nice day!

tom-anders:
- Challenge links now open the "challenge requests" screen and trigger the dialog to accept/decline the challenge
- Open game links in TvScreen if the game is still ongoing
- Open game links in AnalysisScreen if the game is finished

Closes #2734

### DIFF
--- a/lib/src/app_links_service.dart
+++ b/lib/src/app_links_service.dart
@@ -125,8 +125,8 @@ class AppLinksService {
       ref.read(challengeServiceProvider).showConfirmDialog(context, challenge);
 
       return true;
-    } catch (e) {
-      _logger.info('Not a challenge link: $e');
+    } catch (e, st) {
+      _logger.info('Not a challenge link: $e', e, st);
     }
     return false;
   }
@@ -159,8 +159,8 @@ class AppLinksService {
       if (user != null) {
         return [TvScreen.buildRoute(context, gameId: gameId, user: user, orientation: orientation)];
       }
-    } catch (e) {
-      _logger.info('Not a game link: $e');
+    } catch (e, st) {
+      _logger.info('Not a game link: $e', e, st);
     }
 
     return null;
--- a/lib/src/model/challenge/challenge_service.dart
+++ b/lib/src/model/challenge/challenge_service.dart
@@ -206,7 +206,7 @@ class ChallengeService {
   void showConfirmDialog(BuildContext context, Challenge challenge, {String? title}) {
     showAdaptiveActionSheet<void>(
       context: context,
-      title: challenge.variant.isPlaySupported
+      title: challenge.challenger != null && challenge.variant.isPlaySupported
           ? Text(
               title ??
                   '${challenge.challenger!.user.name} challenges you: ${challenge.description(context.l10n)}',
--- a/test/model/challenge/challenge_service_test.dart
+++ b/test/model/challenge/challenge_service_test.dart
@@ -1,8 +1,12 @@
 import 'package:fake_async/fake_async.dart';
 import 'package:fast_immutable_collections/fast_immutable_collections.dart';
+import 'package:flutter/material.dart';
 import 'package:flutter_local_notifications/flutter_local_notifications.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:flutter_test/flutter_test.dart';
+import 'package:lichess_mobile/l10n/l10n.dart';
 import 'package:lichess_mobile/src/model/challenge/challenge.dart';
+import 'package:lichess_mobile/src/model/challenge/challenge_repository.dart';
 import 'package:lichess_mobile/src/model/challenge/challenge_service.dart';
 import 'package:lichess_mobile/src/model/common/chess.dart';
 import 'package:lichess_mobile/src/model/common/game.dart';
@@ -11,17 +15,56 @@ import 'package:lichess_mobile/src/model/common/speed.dart';
 import 'package:lichess_mobile/src/model/notifications/notification_service.dart';
 import 'package:lichess_mobile/src/model/user/user.dart';
 import 'package:lichess_mobile/src/network/socket.dart';
+import 'package:lichess_mobile/src/tab_scaffold.dart' show currentNavigatorKeyProvider;
+import 'package:lichess_mobile/src/view/game/game_screen.dart';
 import 'package:mocktail/mocktail.dart';
 
 import '../../network/fake_websocket_channel.dart';
 import '../../network/socket_test.dart';
 import '../../test_container.dart';
+import '../../test_helpers.dart';
+import '../../test_provider_scope.dart';
 import '../auth/fake_auth_storage.dart';
 
 class NotificationDisplayMock extends Mock implements FlutterLocalNotificationsPlugin {}
 
+class MockChallengeRepository extends Mock implements ChallengeRepository {}
+
+class _ShowConfirmDialogWidget extends ConsumerWidget {
+  const _ShowConfirmDialogWidget({required this.challenge});
+
+  final Challenge challenge;
+
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {
+    return ElevatedButton(
+      onPressed: () => ref.read(challengeServiceProvider).showConfirmDialog(context, challenge),
+      child: const Text('Open Dialog'),
+    );
+  }
+}
+
+class _ShowDeclineDialogWidget extends ConsumerWidget {
+  const _ShowDeclineDialogWidget({required this.challengeId});
+
+  final ChallengeId challengeId;
+
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {
+    return ElevatedButton(
+      onPressed: () => ref.read(challengeServiceProvider).showDeclineDialog(context, challengeId),
+      child: const Text('Open Dialog'),
+    );
+  }
+}
+
 void main() {
   TestWidgetsFlutterBinding.ensureInitialized();
+
+  setUpAll(() {
+    registerFallbackValue(const ChallengeId(''));
+    registerFallbackValue(ChallengeDeclineReason.generic);
+  });
 
   final notificationDisplayMock = NotificationDisplayMock();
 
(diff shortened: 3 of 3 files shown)


apple-app-site-association
```diff
@@ -9,6 +9,7 @@
           { "/": "/tournament/????????", "comment": "Tournaments" },
           { "/": "/broadcast/*/*/????????", "comment": "Broadcast rounds" },
           { "/": "/broadcast/*/*/????????/????????", "comment": "Broadcast games" },
+          { "/": "/????????", "comment": "Games and challenges" },
```
tom-anders, line 12: @fitztrev can you add this to the server when the PR is merged? Thanks!

lib/src/app_links.dart
```diff
@@ -52,33 +61,59 @@ List<Route<dynamic>>? resolveAppLinkUri(BuildContext context, Uri appLinkUri) {
...
-      if (gameId.isValid) {
+      final id = appLinkUri.pathSegments[0];
+
+      try {
+        final gameId = GameId(id);
+        final game = await ref.read(gameRepositoryProvider).getGame(gameId);
+        final orientation = appLinkUri.pathSegments.getOrNull(1) == 'black'
+            ? Side.black
+            : Side.white;
+        final int ply = int.tryParse(appLinkUri.fragment) ?? 0;
```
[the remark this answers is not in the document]
tom-anders: GameId constructor already throws if it's not a valid id

lib/src/view/user/challenge_requests_screen.dart
```diff
@@ -19,24 +19,124 @@ import 'package:lichess_mobile/src/widgets/feedback.dart';
 import 'package:lichess_mobile/src/widgets/list.dart';
 
 class ChallengeRequestsScreen extends StatelessWidget {
-  const ChallengeRequestsScreen({super.key});
+  const ChallengeRequestsScreen({required this.incomingChallenge, super.key});
```
veloce: I don't think this screen should take an incoming challenge param. See other comments for the explanation.

lib/src/view/user/challenge_requests_screen.dart
```diff
@@ -19,24 +19,124 @@ import 'package:lichess_mobile/src/widgets/feedback.dart';
...
     return Scaffold(
       appBar: AppBar(title: Text(context.l10n.preferencesNotifyChallenge)),
-      body: _Body(),
+      body: _Body(incomingChallenge: incomingChallenge),
     );
   }
 }
 
-class _Body extends ConsumerWidget {
+void _showConfirmDialog(WidgetRef ref, BuildContext context, Challenge challenge) {
```
veloce: There is already the logic in `ChallengeService` to show a confirm dialog. 

I know the code was already there, but this is a good opportunity to refactor and simplify. We should reuse the challenge service instead of duplicating the code here.

lib/src/view/user/challenge_requests_screen.dart
```diff
@@ -19,24 +19,124 @@ import 'package:lichess_mobile/src/widgets/feedback.dart';
...
+        makeLabel: (context) => Text(context.l10n.decline),
+        leading: Icon(Icons.clear, color: context.lichessColors.error),
+        isDestructiveAction: true,
+        onPressed: () => _showDeclineDialog(ref, context, challenge),
+      ),
+    ],
+  );
+}
+
+Future<void> _declineChallenge(
```
veloce: Same for decline dialog, we should reuse the challenge service.

lib/src/view/user/challenge_requests_screen.dart
```diff
@@ -19,24 +19,124 @@ import 'package:lichess_mobile/src/widgets/feedback.dart';
...
+            onPressed: () async {
+              await _declineChallenge(ref, challenge, reason);
+            },
+          ),
+        )
+        .toList(),
+  );
+}
+
+Future<void> _acceptChallenge(WidgetRef ref, BuildContext context, Challenge challenge) async {
```
veloce: Same here, we should use the challenge service.

lib/src/view/user/user_context_menu.dart
```diff
@@ -44,7 +44,7 @@ class UserContextMenu extends ConsumerWidget {
                   const SizedBox(height: 8.0),
                   if (value.profile?.bio != null)
                     Linkify(
-                      onOpen: (link) => onLinkifyOpen(context, link),
+                      onOpen: (link) async => await onLinkifyOpen(context, link, ref),
```
veloce: I'm not a fan of passing a WidgetRef around to global functions. See other comments for refactoring suggestion.

lib/src/app_links.dart
veloce, line 1: Since now app links need a `ref` we should refactor the whole thing as a service.

We should have an `AppLinksService` that is started in `app.dart` like the other ones for consistency.

This service can then delegate to the ChallengeService the handling of challenge links.

tom-anders: Agreed, will refactor 👍

tom-anders: Done!

lib/src/app.dart
```diff
@@ -158,8 +158,8 @@ class _AppState extends ConsumerState<Application> {
     );
   }
 
-  Future<void> _initAppLinks() async {
-    _linkSubscription = _appLinks.uriLinkStream.listen((uri) {
+  Future<void> _initAppLinks(WidgetRef ref) async {
```
veloce: Should be done in the AppLinksService.start() method.

tom-anders: ah forgot about that, fixed

test/app_links_service_test.dart
veloce, line 1: Should be renamed to app_links_service_test.dart

lib/src/app_links_service.dart
```diff
@@ -0,0 +1,200 @@
...
+        return [
+          PuzzleScreen.buildRoute(
+            context,
+            angle: PuzzleAngle.fromKey('mix'),
+            puzzleId: PuzzleId(id),
+          ),
+        ];
+      case _:
+        final gameRoutes = await _tryResolveGameLink(context, appLinkUri);
+        if (gameRoutes != null) return gameRoutes;
```
veloce, line 112: For future readers, I'd add a comment here to explain that `_` handles both challenges and games links.

We try first to resolve it as a game (if it is a game it cannot be a challenge) if that fails this method returns `null` and the challenge link is resolved later.

I realise now that for each received challenge via deep link we make an extra game request. I don't think that is a big deal, so let's not bother with that in this PR. But for the sake of questioning I wonder if there would be a better to deal with it?

tom-anders: Yeah, I couldn't find any better API, maybe it can be added at some point. I'll add a comment

lib/src/app_links_service.dart
```diff
@@ -0,0 +1,200 @@
...
+
+  static const kLichessLinkifiers = [UrlLinkifier(), EmailLinkifier(), UserTagLinkifier()];
+
+  /// Handles link clicks in Linkify widgets throughout the app.
+  Future<void> onLinkifyOpen(BuildContext context, LinkableElement link) async {
+    if (link is UrlElement && link.url.startsWith(RegExp('https?:\\/\\/$kLichessHost'))) {
+      // Handle Lichess links specifically
+      final appLinkUri = Uri.parse(link.url);
+      await handleAppLink(context, appLinkUri);
+    } else if (link.originText.startsWith('@')) {
```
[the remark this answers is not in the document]
tom-anders: Not needed I think, AndroidManifest.xml and apple-app-side-association already filter the host for us

lib/src/app_links_service.dart
```diff
@@ -0,0 +1,200 @@
...
+
+      final user = game.playerOf(orientation).user;
+      if (user != null) {
+        return [TvScreen.buildRoute(context, gameId: gameId, user: user, orientation: orientation)];
+      }
+    } catch (e) {
+      _logger.info('Not a game link: $e');
+    }
+
+    return null;
```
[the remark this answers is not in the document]
tom-anders: don't really see how this makes things better

lib/src/app_links_service.dart
```diff
@@ -0,0 +1,200 @@
...
+  Future<bool> _tryResolveChallengeLink(BuildContext context, Uri appLinkUri) async {
+    try {
+      final challengeId = ChallengeId(appLinkUri.pathSegments[0]);
+      if (!challengeId.isValid) return false;
+      final challenge = await ref.read(challengeRepositoryProvider).show(challengeId);
+      if (!context.mounted) return false;
+
+      ref.read(challengeServiceProvider).showConfirmDialog(context, challenge);
+
+      return true;
```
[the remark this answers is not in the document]
tom-anders: PR description is not up to date anymore here

lib/src/model/challenge/challenge_service.dart
```diff
@@ -169,14 +162,17 @@ class ChallengeService {
...
     // Cancel any pending lobby seek before accepting, to prevent being matched into a new game
     // while accepting a challenge.
     try {
       await ref.read(createGameServiceProvider).cancelSeek();
     } catch (_) {}
 
-    final fullId = await acceptChallenge(id);
+    final challengeRepo = ref.read(challengeRepositoryProvider);
+    await challengeRepo.accept(id);
+    final fullId = await challengeRepo.show(id).then((challenge) => challenge.gameFullId);
```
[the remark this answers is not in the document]
tom-anders: I added try/catch around `handleAppLink`, so I think we should be good now

lib/src/model/common/id.dart
```diff
@@ -51,6 +51,8 @@ extension type const UserId(String value) implements StringId {
 
 extension type const ChallengeId(String value) implements StringId {
   ChallengeId.fromJson(dynamic json) : this(json as String);
+
+  bool get isValid => RegExp(r'''[\w-]{8}''').hasMatch(value);
 }
```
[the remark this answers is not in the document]
tom-anders: @veloce hmm, why is it matching literal `-` in the regex here anyway?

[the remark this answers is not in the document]
veloce: This is the game id regex coming from Lila: https://github.com/veloce/lila/blob/b65600921a925205a20e4a713c6a5f20ca5f848c/modules/core/src/main/id.scala#L15

As you can see in that file, there is no validation of challenge id in Lila.

lib/src/view/user/challenge_requests_screen.dart
```diff
@@ -88,93 +82,28 @@ class _ChallengeListItem extends ConsumerWidget {
```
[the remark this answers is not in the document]
veloce: @tom-anders I am not sure we can remove the auth user check. Can we?

If we want to support the open challenges for anons, the logic needs to be updated. But if the challenge was created by an user and targeted to another user, then we probably need to check the user match, don't we?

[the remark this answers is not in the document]
tom-anders: Hmm I was under the impression that you can never actually open the challenge requests screen from the home screen if you're not logged in.

And if you get a challenge via deep link, we also don't open this screen, but rather show the accept/decline dialog directly.

Is there something I'm missing?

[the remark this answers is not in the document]
veloce: Maybe not. I was just wondering why we were using the `authUser` in the challenge list item before.

lib/src/model/common/id.dart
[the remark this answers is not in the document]
veloce: @tom-anders I don't know why there is no validation server side. I think we can (should) do validation in the app. But then let's use the same regex as in game ID (should be declared as a private).

lib/src/view/user/challenge_requests_screen.dart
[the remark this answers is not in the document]
tom-anders: I think it was already redundant before my PR - Maybe add some point in the past it was possible to open this screen w/o being logged in, but that's not the case anymore

lib/src/model/common/id.dart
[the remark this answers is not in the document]
tom-anders: Factored out to a private function now :+1:

PR #2833, opened 2026-03-22
Add search to app logs

r3econ:
- Adds a search bar to the App Logs screen that filters log entries by message, logger name, or error text. 
- Filtering is done in SQLite via a parameterized `LIKE` query across all three fields. 

### Demo

https://github.com/user-attachments/assets/c9104967-ce47-45f5-a4be-b1c5127ce416


lib/src/view/settings/app_log_settings_screen.dart
```diff
@@ -126,6 +130,20 @@ class _AppLogSettingsScreenState extends ConsumerState<AppLogSettingsScreen> {
               ),
             ],
           ),
+          Padding(
```
veloce: Not sure about a static search field. Since we already have the log level selector, it takes too much vertical space imo.

Also ideally this screen should use the PlatformScaffold to handle iOS app bar transparency, and instead of a Column the body should be the list view directly and the static part should be put in the `bottom` section of the `AppBar`.

r3econ: Thanks  for the feedback 🙏🏾  I've put the log level on the same line as the search. It looks better, before & after pics:

SCR-20260323-qkmj-tile

Issue #1790, opened 2025-05-23
Platform home widget


veloce:
It would be nice to have more native integration with widgets on the OS side. This would be useful to display things that does not depend on the live connection, like daily puzzle, correspondence games, etc.

![Image](https://github.com/user-attachments/assets/0b2ef468-be27-43e0-bdde-27ee489fb886)

Using this package:

https://pub.dev/packages/home_widget

### Ideas
- correspondence games where it is my turn
- daily puzzle
- featured blog post

cappylot:
Hello, I was wondering if someone is already working on this? If not, I would take some time to implement some widgets in the upcoming days, as this is a feature I also desperately want 😇

tom-anders:
There's a PR here that adds a widget for the daily puzzle, however it looks like the author has abondoned it: https://github.com/lichess-org/mobile/pull/2677 I don't think anyone else is working on this issue currently. 

You can probably use that PR as a starting point.

PR #2834, opened 2026-03-22
iOS Blog Feed Widget

r3econ:
- Adds native iOS home screen widget that displays recent posts from Lichess blog feeds. The PR also lays the groundwork for adding more iOS widgets in the future.

> Disclaimer: iOS home screen widgets can't be written using flutter. iOS requires writing them using native code.

- Feed content is configurable via the widget settings sheet and supports:
  - Official Blog (default)
  - Community Blog
  - User Blog — shows a username field
- Feeds are fetched as Atom (all Lichess feeds use Atom format)
- Tapping a post deep-links into the app and opens it in the in-app browser.
- See `EXTENSIONS.md` for how to sign and test the extension as well as what is required to make it work when releasing in the App Store.
- Partially resolves https://github.com/lichess-org/mobile/issues/1790 
- Feedback is highly welcomed 🙂 🆗 

#### iPhone

iphone

#### iPad

ipad

#### Widget Catalog

widget_catalog

#### Settings

settings

#### Video

https://github.com/user-attachments/assets/dc623485-ae0f-4402-bf7d-dafbe4ed39ad


veloce:
Whoa, man, that is truly awesome!

Just a small piece of feedback for now (will review the code): for the community blog, the author name (lichess user) should be displayed somewhere.

r3econ:
Good idea. Adding post author will be easy, I'll add it 👍🏾 

image

The other thoughts I had while developing:
- Having a way to specify the language for the posts that should appear in the feed
- Technical: Maybe we could set the minimal deployment target for the widgets to iOS 26. It would lower the testing effort and maintenance

veloce:
> Having a way to specify the language for the posts that should appear in the feed

lichess allows it but if it's too complicated in the widget I'd say let's not bother

> Technical: Maybe we could set the minimal deployment target for the widgets to iOS 26. It would lower the testing effort and maintenance

sure, no problem

veloce:
Another feedback: instead of showing a relative date for the blog post, always show the day and month, and the year if it is a different year (we don't need the hour), like it is done on lichess.

We already have the hour for when the widget has been updated, which is great.

r3econ:
Nice, I've implemented the feedback:

- Show author in the bommunity blog ☑️ 
- Format date ☑️ 

image

ios/LichessWidgets/Resources/Assets.xcassets/OfficialBlog/OfficialBlogPlaceholderImage2.imageset/OfficialBlogPlaceholderImage2.png
veloce, line 1: What is the purpose of this image? If it is a placeholder the image should be generic.

ios/LichessWidgets/Resources/Assets.xcassets/OfficialBlog/OfficialBlogPlaceholderImage3.imageset/OfficialBlogPlaceholderImage3.png
veloce, line 1: Same here, it doesn't look as a placeholder image.

ios/EXTENSIONS.md
```diff
@@ -0,0 +1,26 @@
...
+
+These changes are local only and should not be committed.
+
+### Merging new extensions
+
+When a PR that adds a new extension is ready to merge, a Lichess org member with Apple Developer access needs to:
+
+1. Register the new App ID (for the extension's bundle ID) in the Apple Developer portal.
+2. Create a provisioning profile for it.
+3. Ensure any required entitlements are added to the main app profile.
```
veloce: Could you precise here what entitlements are required?

ios/EXTENSIONS.md
```diff
@@ -0,0 +1,26 @@
...
+3. Do the same for the extension target (e.g. **LichessWidgetsExtension**), using a matching sub-ID (e.g. `com.yourname.lichess.widget`).
+4. Build and run from Xcode — the extension will be signed with your profile and work on your device.
+
+These changes are local only and should not be committed.
+
+### Merging new extensions
+
+When a PR that adds a new extension is ready to merge, a Lichess org member with Apple Developer access needs to:
+
+1. Register the new App ID (for the extension's bundle ID) in the Apple Developer portal.
```
veloce, line 24: We're using [fastlane match](https://docs.fastlane.tools/actions/match/#fastlane) for code signing in release mode in this project. It would be nice to update these instructions to take that into account.

lib/src/app.dart
```diff
@@ -168,6 +168,10 @@ class _AppState extends ConsumerState<Application> {
         ref.read(oauthCallbackProvider).add(uri);
         return;
       }
+      if (uri.scheme == kOAuthRedirectUriScheme && uri.host == 'open-web') {
```
veloce: Since now we use the scheme for something else than OAuth we should rename that constant to sth like `kLichessUriScheme`.

ios/LichessWidgets/Resources/Assets.xcassets/OfficialBlog/OfficialBlogPlaceholderImage2.imageset/OfficialBlogPlaceholderImage2.png
r3econ: All `PlaceholderThumbnail..` images are preview images used in the iOS widget gallery where user can see the widgets before adding them to the screen. [Apple guidelines](https://github.com/lichess-org/mobile/issues/2839) suggest providing a realistic preview. I used 4 first entries of the Lichess feed and saved their content so that a realistic feed preview can be shown in the gallery - [exact](https://github.com/lichess-org/mobile/pull/2834/changes#diff-bb9f205864d21647c70f11ef13868f529bf2113956e6ef1935b760d9aa682209R73) place in code.

SCR-20260324-rqrm-tile

I would not recommend using generic images like grey boxes etc. The other apps show placeholder content that looks real - below feeds from Reuters, Reddit and Wikipedia:

IMG_5406-tile

veloce: I get it, thanks for the explanation!

veloce:
@r3econ in case you missed it:

> What I found is that community and user blogs are way less discoverable since you have to modify the widget to see them. Would it be possible to include them in the widgets list?

do you think this is feasible?

r3econ:
@veloce, yes, very good point, the widgets must be easily discoverable. I've updated the code so that all 3 kinds of widgets appear in the widget gallery (_Community_, _Official_ and _User_ blog feeds). Below a preview:

widget_gallery

https://github.com/user-attachments/assets/afe4f537-a723-469c-84ad-352531c18dac

Additionally I've improved the design when widget content can't be loaded (e.g. when there is no internet connection).

errors

ios/EXTENSIONS.md
r3econ: I've removed that point. If I remember it correctly some extensions require creating an entitlement file but I think the person adding extensions will know it (or will find out it's needed) - so let's keep the readme as simple as possible

lib/src/app.dart
r3econ: Good point 👍🏾 I fixed it 🧑‍🔧

ios/EXTENSIONS.md
r3econ: Yes, definitely. I added fastlane instructions in daba1a07283472c9eb4087537921773b3f115fe4. They might require some fine-tuning once you create a build and see how fastlane handles the new widget extension. Hopefully it works with no changes 🚀

ios/LichessWidgets/Blog Feed Widget/Model/BlogFeedChoice.swift
```diff
@@ -0,0 +1,25 @@
+import Foundation
+
+enum BlogFeedChoice: String {
+    case officialBlog
+    case communityBlog
+    case userBlog
+
+    var displayName: String {
+        switch self {
+        case .officialBlog: return "Official Blog"
```
veloce, line 10: Not a requested change in this PR but a question for the future: how complicated would it be to include the crowdin translations in the swift code?

r3econ: Yes, I've been thinking about it and decided not to include it in the first version to keep the complexity low. If I am not wrong, we would have to include the translations in the widget's target - how we do it is something to find out (maybe they could be generated and added from Flutter (probably preferred), or maybe fetched via CrowdIn SDK in iOS Project). I will research this 🧑‍🔬

PR #2845, opened 2026-03-25
fix incorrect circle color in  Tournament Player Details

HaonRekcef:
The color of the circle indicating the color the player had in the game was swapped. Reported here: https://lichess.org/forum/community-blog-discussions/ublog-VGPcQtN3?page=5#DPksKbS9
Created a new side indicator widget and also used it in the `broadcast_player_results_screen.dart‎ `.

dark mode: 

image

light mode:

image

### DIFF
--- a/lib/src/view/broadcast/broadcast_player_results_screen.dart
+++ b/lib/src/view/broadcast/broadcast_player_results_screen.dart
@@ -17,6 +17,7 @@ import 'package:lichess_mobile/src/view/broadcast/broadcast_player_widget.dart';
 import 'package:lichess_mobile/src/widgets/network_image.dart';
 import 'package:lichess_mobile/src/widgets/platform.dart';
 import 'package:lichess_mobile/src/widgets/progression_widget.dart';
+import 'package:lichess_mobile/src/widgets/side_indicator.dart';
 import 'package:lichess_mobile/src/widgets/stat_card.dart';
 
 final broadcastTournamentIdProvider = FutureProvider.autoDispose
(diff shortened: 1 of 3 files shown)


lib/src/view/tournament/tournament_screen.dart
```diff
@@ -1492,7 +1492,7 @@ class _PairingTile extends ConsumerWidget {
               padding: EdgeInsets.only(right: 8.0),
               child: Icon(LichessIcons.body_cut, size: 20),
             ),
-          Icon(pairing.color == Side.white ? Icons.circle_outlined : Icons.circle, size: 20),
+          Icon(pairing.color == Side.white ? Icons.circle : Icons.circle_outlined, size: 20),
```
veloce: I believe it depends on the background theme as well (dark vs light). Does it? In that case we should use an helper.

HaonRekcef: Ah yes it does, good catch

HaonRekcef: fixed now

lib/src/view/broadcast/broadcast_player_results_screen.dart
```diff
@@ -495,33 +496,16 @@ class _GameResultListTile extends StatelessWidget {
...
-                      .white => Colors.white.withValues(alpha: 0.9),
-                      .black => Colors.black.withValues(alpha: 0.9),
-                    },
-                  ),
-                ),
-              ),
+              child: Center(child: SideIndicator(side: color, size: 15)),
             ),
-            if (showTCIcon) SizedBox(width: 12, child: Icon(fideTC.icon, size: 15)),
+            if (showTCIcon) Icon(fideTC.icon, size: 15),
```
HaonRekcef, line 508: unrelated change, but I don't think we should wrap the Icon with Size 15 in a SizedBox with lesser size, so I simply removed it as I don't see a reason for it.

veloce (approved):
Thanks!

Issue #2750, opened 2026-03-12
"Learn from your mistakes" sometimes remains stuck loading 


giordy:
### What is the problem?

I'm using a Pixel 10.

I noticed a couple of times already that when I try to use "Learn from your mistakes" right after playing a game it doesn't load promptly even after 2-3 attempts.

https://github.com/user-attachments/assets/4d1150be-6b3c-493a-9115-01cd92eec8e5

I had already requested the analysis, I'm not sure if that's a factor.

If I go back to my profile and try again then it often works

### Steps to reproduce the bug

1. Play a game
2. Request analysis right away
3. Proceed with "Learn from your mistakes"

### App version

0.21.9

### Device Info

Pixel 10

giordy:
Still happens in 0.22.7.

https://github.com/user-attachments/assets/8f6d0f5e-9db2-4056-9128-d1d6ff73b06d

sviluppatoredisoftware:
Fixed in pre-release 0.23.2
#2859

Issue #2564, opened 2026-01-22
Learn from your misstake "something went wrong"


TBestLittleHelper:
>Clicking "learn from your mistakes" after requesting a computer analysis seams to start a new computer analysis which takes some time until I get a screen telling me "Something went wrong". I was able to reproduce this bug twice. I do know that I can also start by clicking learn from your mistakes, as this performs a computer analysis as well, but this is the more intuitive way for me as it is the same as in the browser.

From : https://discord.com/channels/280713822073913354/1463552462211518709/1463552462211518709

tom-anders:
Reproduction steps on 0.19.4:

1) Open game without server analysis in analysis board
2) swipe left and request server analysis via the button
3) open hamburger menu and press the "Learn from your mistakes" button

When omitting step 2), the bug does not appear. Closing the analysis board between 2) and 3) fixes it as well

ro011110ot:
Same here

App-Log

[Lichess shared text.txt](https://github.com/user-attachments/files/24815604/Lichess.shared.text.txt)

sviluppatoredisoftware:
Fixed in pre-release 0.23.2
#2859

PR #2859, opened 2026-03-28
Fix retro reuse existing analysis

sviluppatoredisoftware:
Related to #2564 and #2750

## What this changes

This PR tries to fix the case where **Learn from your mistakes** can get stuck / keep loading after a computer analysis was already requested first.

The issue seems to come from the retro flow attaching too late to analysis progress and potentially trying to request / restart analysis again for the same game.

## Changes

- avoid restarting server analysis if the same game is already being analyzed
- attach the retro listener earlier
- reuse an already available analysis progress event for the same game
- keep tracking the current game even when the server responds with `400 already requested`

## Why

This should help with the cases described in:

- #2564 — `Learn from your misstake "something went wrong"`
- #2750 — `"Learn from your mistakes" sometimes remains stuck loading`

### DIFF
--- a/lib/src/model/analysis/server_analysis_service.dart
+++ b/lib/src/model/analysis/server_analysis_service.dart
@@ -51,9 +51,7 @@ class ServerAnalysisService {
   Future<void> requestAnalysis(GameId id, [Side? side]) async {
     // If we are already listening for analysis updates of this exact game,
     // don't tear everything down and reconnect.
-    if (_currentAnalysis.value == id &&
-        _socketSubscription != null &&
-        _analysisCompleter != null) {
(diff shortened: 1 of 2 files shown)


lib/src/model/analysis/server_analysis_service.dart
```diff
@@ -95,6 +101,7 @@ class ServerAnalysisService {
       // of analyses is reached.
       if (e.statusCode == 400) {
         debugPrint('Analysis already requested for game $id');
+        _currentAnalysis.value = id.gameId;
```
veloce, line 104: Does not look useful: we already assign it in line 96.

PR #2864, opened 2026-03-29
Fix: save all PGN header fields before exporting

### LINKED ISSUE #2850: Black elo missing from Sharing & export PGN from analysis board
### What is the problem?

Black elo missing from Sharing & export PGN from analysis board. Even is elo number is entered in export form a ? is shown

1 tools
2 analysis board
3 make moves
4 burger meny share & export
5 enter values for elo fields
6 inspect the PGN shared text

### App version

0.22.7

### Device Info

Samsung Galaxy s22 Ultra / android 16

### Additional information

_No response_

MaartenD:
When pressing the share button while still in the last PGN header field (BlackElo), 
the value was not saved. Fixed by reading all controllers before exporting.

Fixes #2850

### DIFF
--- a/lib/src/view/analysis/analysis_share_screen.dart
+++ b/lib/src/view/analysis/analysis_share_screen.dart
@@ -166,9 +166,13 @@ class _EditPgnTagsFormState extends ConsumerState<_EditPgnTagsForm> {
                     return FilledButton(
                       onPressed: () {
                         for (final entry in pgnHeaders.entries) {
+                          final controller = _controllers[entry.key];
+                          if (controller == null) {
+                            throw StateError('Controller missing for key: ${entry.key}');
+                          }
                           ref
                               .read(ctrlProvider.notifier)
-                              .updatePgnHeader(entry.key, _controllers[entry.key]!.text);
+                              .updatePgnHeader(entry.key, controller.text);
                         }
                         launchShareDialog(
                           context,


devtools_options.yaml
veloce, line 1: This file should not be part of this PR. Especially since it does nothing here.

MaartenD: Sorry, i will take a look and try to adjust the commit. I’m a newbie.

MaartenD: I hope everything is alright now, if not let me know.

lib/src/view/analysis/analysis_share_screen.dart
```diff
@@ -162,6 +162,11 @@ class _EditPgnTagsFormState extends ConsumerState<_EditPgnTagsForm> {
                   builder: (context) {
                     return FilledButton(
                       onPressed: () {
+                        for (final entry in pgnHeaders.entries) {
+                          ref
+                              .read(ctrlProvider.notifier)
+                              .updatePgnHeader(entry.key, _controllers[entry.key]!.text);
```
veloce: I suppose the `_controllers[entry.key]` cannot be `null` with the current code, but we should still guard it for the future. So let's add a check to avoid the null error.

.gitignore
```diff
@@ -41,3 +41,6 @@ doc/api/
 
 # VS Code
 .vscode/
+
```
veloce: At some point we may want to include some dev tools prefs for the project so there's no reason to add this for now.

If you are using dev tools options on your own you should use `.git/info/exclude` to ignore it.

lib/src/view/analysis/analysis_share_screen.dart
```diff
@@ -59,7 +59,11 @@ class _EditPgnTagsFormState extends ConsumerState<_EditPgnTagsForm> {
       _focusNodes[entry.key] = FocusNode();
       _focusNodes[entry.key]!.addListener(() {
         if (!_focusNodes[entry.key]!.hasFocus) {
-          ref.read(ctrlProvider.notifier).updatePgnHeader(entry.key, _controllers[entry.key]!.text);
+          final controller = _controllers[entry.key];
+          if (controller != null) {
+            ref.read(ctrlProvider.notifier).updatePgnHeader(entry.key, controller.text);
+          }
+          //ref.read(ctrlProvider.notifier).updatePgnHeader(entry.key, _controllers[entry.key]!.text);
```
HaonRekcef: You should remove the comment

lib/src/view/analysis/analysis_share_screen.dart
veloce: @MaartenD we're still missing the null check here

MaartenD: @veloce when you scroll down in this list there you would see the null check.  I did 2 new comits yesterday evening. I’m sorry if this is not the correct way, it’s a whole new world for me (just an old ABAP programmer ;-) ).

veloce: You added a check line 63.

There's nothing to scroll, I'm just seeing the code. Here, line 171 I see `_controller[entry.key]!.text`, there is still the non-null assertion operator without null check.

PR #2879, opened 2026-03-31
feat(engine): show 0.0 on evaluation gauge in neutral positions

flcsezz:
## Description
    This PR improves the user feedback on the evaluation gauge (eval bar) in the analysis screen.
      Previously, when a position was perfectly equal (0.0) or the game ended in a draw, the
      evaluation text was hidden, which could lead users to believe the engine was not working or
      still loading.
    
    The gauge now explicitly displays **0.0** in these scenarios to provide immediate
      confirmation of the engine's state.
    
## Changes
**`lib/src/view/engine/engine_gauge.dart`**: 
 - Updated `evalDisplay` logic to return `'0.0'` for draw outcomes (no winner).
   Removed the visibility constraint that hid the text when the evaluation was exactly
   `0.5` (neutral).
  - **`test/view/engine/engine_gauge_test.dart`**:
  - Added a new test file with widget tests to verify that `0.0` is rendered for both equal
     positions and draw outcomes.

   ## Verification
    - Created and passed `test/view/engine/engine_gauge_test.dart`.
    - Ran `flutter analyze` and `flutter test` - **No issues found.**


veloce:
Can I see screenshots or better screen recording before I review this? Thanks.

flcsezz:
Screenshot_20260331-225636_Lichess - DEBUG
Screenshot_20260331-225740_Lichess

I've attached screenshots comparing the current Play Store version with the latest build.

In the new build, the gauge now explicitly displays 0.0 during a "Theoretical Draw" scenario. Previously, this state could be misinterpreted as the engine failing to initialize or hanging. By clearly showing the numerical value, we provide immediate feedback to the player that the engine is active and the UI is functioning as intended.

Overall, this change removes any ambiguity regarding the "ready" state of the bar and results in a much cleaner user experience. ^^

ijm8710:
One quick comment. If I look at say Hikarus game today in candidates which was a draw.  Most times the engine reads 0.0 on web it doesn't get the plus sign. There are a couple times in the app where it will actually edgecase and show +0.0 or -0.0 (this happens where it starts to drift towards 0.1 when it dives deeper) but when it's as theoretically 0.0 as possible it has no sign.


I feel in the situations where it's fully 0.0 no sign should be used to be consistent with web

Not sure if it saying +0.0 is one of those rare edge cases where the sign is affixed mentioned above or if you made it always affix the + as +0.0 which I feel should stay as it was before as 0.0

flcsezz:
right that the sign is unnecessary when the evaluation rounds to 0.0. I've updated the evaluation logic to ensure that any score rounding to 0.0 (like +0.02 or -0.02) is displayed as a plain '0.0' without a sign.

flcsezz:
Screenshot_20260401-001322_Lichess - DEBUG

veloce (approved):
LGTM.

Thanks! And sorry for the delay.

Issue #2724, opened 2026-03-08
Friends screen doesn't load


ornicar:
### What is the problem?

The friends screen shows a loading animation that never completes.

On the server side, the request has successfully completed with data:

```
GET /api/rel/following HTTP/2.0 200 152878 "-" "Lichess Mobile/0.21.9 as:thibault sri:xxxxxxxxxxxxxx os:Android/15 dev:Fairphone 6" "-"
```

152878 is the bytes sent

### Steps to reproduce the bug

1. Log in
2. More
3. Friends

I'm logged in as `thibault`.

### App version

0.21.9

### Device Info

`os:Android/15 dev:Fairphone 6`

ijm8710:
Have also experienced this on iOS when accessing friends from the top right icon on home tab. Videos of it happening in #2043 and #1639 (prematurely closed)

Have also mentioned it to dev haonrekcef who has reproduced it a few sporadic times as well

lacostej:
Having a default timeout could be a good thing. I am not 100% sure it is enough to fully fix the issue though. There might be more at play.

lacostej:
I played around retrying on more type of issues, but I hit some issues with the way the tests are architected. So I left that for later.

lacostej:
Issue #2043, with its "wake from standby" pattern made me wonder if there were some generic issues in the network layer.
Hence the default timeout in #2777. Then I added #2779 to let the RetryClient handle other type of possible failures. I hope this hardens a bit more the network layer and will reduce potential errors.

CloudyDino:
I think the issue might be that the call `Response.fromStream` in `_sendUnstreamed` fails and then causes an `AsyncError` which is currently being handled by a loading spinner. I think if we make sure that isn't the issue and that we are indeed always loading, then I wonder if  `Response.fromStream` needs a timeout as well.

I created a PR here that might be able to address the infinite loading on error and another potential issue I found with `getUsersStatuses` : https://github.com/lichess-org/mobile/pull/2884

veloce:
I believe this is fixed by 87e234f6ddd99e021dbbf2c6dc8424b74d1dd134

PR #2884, opened 2026-04-02
Show full-screen retry on friends load error

CloudyDino:
Maybe a partial fix for https://github.com/lichess-org/mobile/issues/2724 and related issues?

**Changes**
 1. ~~Add a timeout on `Response.fromStream()` in `LichessClient._sendUnstreamed`. That way in case `Response.fromStream()` reads the body and hangs, we can just timeout instead. I think this can fix the infinite loading potentially? But then what if the user is following so many people that they always trigger the `Response.fromStream()` timeout?~~
 2. Call `getUsersStatuses` with only the first 100 user IDs, since the Lichess API silently truncates to 100 IDs (`.take(100)` in [Api.scala](https://github.com/lichess-org/lila/blob/21aec830ca771c12ee9a31416214c7a09472a043/app/controllers/Api.scala#L82)).
 3. Replace `firstWhere` with `firstWhereOrNull` so a missing status defaults to offline instead of throwing.
 4. Don't show the loading spinner forever for the `AsyncError` case at least. 

**Investigation**

I see that we call `/api/users/status` right after the call to `/api/rel/following`. Based on the [API documenation](https://lichess.org/api#tag/users/GET/api/users/status), looks like we limit the URL to have 100 user IDs in the query. Based on what I'm seeing in [Api.scala](https://github.com/lichess-org/lila/blob/21aec830ca771c12ee9a31416214c7a09472a043/app/controllers/Api.scala#L82), this limit is done by taking the first 100 user IDs. 

That means when `_isOnline` is called, the `statuses.firstWhere` call will fail for everyone besides the first 100 user IDs and throw a `StateError`. I think that might be causing an issue as well so I've added chunking for the `/api/users/status` call and replaced `firstWhere` with `firstWhereOrNull` so that a missing status defaults to offline instead of throwing an error.

Let me know if I should continue using `firstWhere` so that we fail instead of showing potentially wrong information and hiding a bug!

### DIFF
--- a/lib/src/view/relation/friend_screen.dart
+++ b/lib/src/view/relation/friend_screen.dart
@@ -30,15 +30,11 @@ final followingStatusesProvider = FutureProvider.autoDispose<(IList<User>, IList
     return (IList<User>(), IList<UserStatus>());
   }
 
-  final ids = following.map((user) => user.id).toIList();
-  final statuses = <UserStatus>[];
   // The /api/users/status endpoint accepts at most 100 IDs per request.
-  for (var i = 0; i < ids.length; i += 100) {
-    final chunk = ids.sublist(i, i + 100 < ids.length ? i + 100 : ids.length);
-    final result = await ref.read(userRepositoryProvider).getUsersStatuses(chunk.toISet());
-    statuses.addAll(result);
-  }
-  return (following, IList(statuses));
+  final statuses = await ref
+      .read(userRepositoryProvider)
+      .getUsersStatuses(following.take(100).map((user) => user.id).toISet());
+  return (following, statuses);
 });
 
 class FriendScreen extends ConsumerStatefulWidget {


CloudyDino:
I'm not sure exactly how to test if this is fixing the issue, @ijm8710 did you have a reliable way to trigger the problem or was it happening to you only occasionally?

ijm8710:
> I'm not sure exactly how to test if this is fixing the issue, @ijm8710 did you have a reliable way to trigger the problem or was it happening to you only occasionally?

No direct trigger but I would see it all the time.

I don't want to misspeak as Veloce was skeptical of it having a huge effect but I feel #2777 may have had a greater impact than anticipated. Since that merge, I haven't seen the issue at all so I think those PRs helped even more than anticipated

lib/src/view/relation/friend_screen.dart
```diff
@@ -29,10 +30,15 @@ final followingStatusesProvider = FutureProvider.autoDispose<(IList<User>, IList
...
-  final statuses = await ref
-      .read(userRepositoryProvider)
-      .getUsersStatuses(following.map((user) => user.id).toISet());
-  return (following, statuses);
+  final ids = following.map((user) => user.id).toIList();
+  final statuses = <UserStatus>[];
+  // The /api/users/status endpoint accepts at most 100 IDs per request.
+  for (var i = 0; i < ids.length; i += 100) {
+    final chunk = ids.sublist(i, i + 100 < ids.length ? i + 100 : ids.length);
+    final result = await ref.read(userRepositoryProvider).getUsersStatuses(chunk.toISet());
```
veloce: The request should not be done in the loop.

lib/src/view/relation/friend_screen.dart
```diff
@@ -29,10 +30,15 @@ final followingStatusesProvider = FutureProvider.autoDispose<(IList<User>, IList
...
   }
 
-  final statuses = await ref
-      .read(userRepositoryProvider)
-      .getUsersStatuses(following.map((user) => user.id).toISet());
-  return (following, statuses);
+  final ids = following.map((user) => user.id).toIList();
+  final statuses = <UserStatus>[];
+  // The /api/users/status endpoint accepts at most 100 IDs per request.
+  for (var i = 0; i < ids.length; i += 100) {
```
veloce: Instead of this for loop you can simply do:

```dart
final statuses = await ref
      .read(userRepositoryProvider)
      .getUsersStatuses(following.take(100).map((user) => user.id).toISet());
```

CloudyDino: Thanks, updated!

lib/src/view/relation/friend_screen.dart
```diff
@@ -268,6 +275,6 @@ class _Following extends ConsumerWidget {
   }
 
   bool _isOnline(User user, IList<UserStatus> statuses) {
-    return statuses.firstWhere((status) => status.id == user.id).online ?? false;
+    return statuses.firstWhereOrNull((status) => status.id == user.id)?.online ?? false;
```
CloudyDino, line 278: @veloce Should we just return a nullable boolean here? `UserListTile.isOnline` is already of type `bool?`. Then we have a few options:
1. Nothing. In `widgets/user.dart`, we call `ConnectedIcon(isConnected: user?.isOnline == true)`. That means it'll have the same behavior, but this `_isOnline` function isn't hiding the fact we don't know about online status.
2. Hide the `ConnectedIcon` when `isOnline == null` so we don't display false info to the user.
3. Update `ConnectedIcon` to handle a null `isOnline` argument and display the `Symbols.android_wifi_4_bar_question` icon when it's null.

veloce:
I just kept the retry screen from your PR. All the rest was unnecessary because we don't show anymore the connection status, so I simplified the whole thing.

PR #2887, opened 2026-04-02
fix now it saves skip state immediately in Puzzle Streak

### LINKED ISSUE #2886: Puzzle Streak: Infinite "Skip This Move" Exploit
### What is the problem?

The method I'm describing makes it possible to reach arbitrary high scores on Puzzle Streak:

When playing Puzzle Streak, it's possible to click on the "Skip This Move" (STM) button at any point, remember the sequence of moves, and return to the main menu using the back button at the top left. After confirming that the state will be saved locally, you get to the main menu. After reopening of Puzzle Streak again, the puzzle begins from the beginning of the sequence, and the STM button is reset and can be used again.

At least, I would expect that the local state also saves that the STM button has already been clicked.
Even better, that the sequence was revealed until some point (the advantage that the STM button gave you) and move forward to this position.

1. Open the Lichess mobile app.
2. Start a new Puzzle Streak.
4. Tap the STM button to skip a move (and memorize the sequence).
5. Use the top-left back button to exit to the main menu.
6. Confirm any prompt that says the state will be saved locally.
7. From the main menu, reopen Puzzle Streak (resume the same puzzle/session).

### App version

0.22.7

### Device Info

iPhone, iOS 26.3.1 (a)

### Additional information

_No response_

glitchfl:
The "skip this move" button state was only stored in memory and not persisted to local storage. This allowed users to use the skip than exit to the main menu and resume the streak with the skip button available again and by that enabling unlimited skips. The fix persists the hasSkipped flag to local storage immediately when the skip is used, matching the same save pattern already used when going to the next puzzle.

Fixes #2886 - "Puzzle Streak: Infinite "Skip This Move" Exploit"

### DIFF
--- a/test/view/puzzle/streak_screen_test.dart
+++ b/test/view/puzzle/streak_screen_test.dart
@@ -5,6 +5,7 @@ import 'package:http/testing.dart';
 import 'package:lichess_mobile/src/network/http.dart';
 import 'package:lichess_mobile/src/utils/navigation.dart';
 import 'package:lichess_mobile/src/view/puzzle/streak_screen.dart';
+import 'package:lichess_mobile/src/widgets/bottom_bar.dart';
 
 import '../../test_helpers.dart';
 import '../../test_provider_scope.dart';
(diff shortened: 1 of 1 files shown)


tom-anders:
Thanks for the quick fix! I think it would be nice to have a regression test for this. We already have a widget test that tests saving the streak to storage, so you should be able to extend that one.

glitchfl:
Done! I added a regression test that works like the existing "Score is saved when exiting screen" test. it solves a puzzle, taps "Skip This Move", exits via the back button, re-enters the streak and verifies the skip button is still disabled. needless to say that all existing tests continue to pass.

PR #2889, opened 2026-04-02
Fix antichess stalemate variant outcome

### LINKED ISSUE #2814: Antichess Mobile OTB Stalemate shows as draw
### What is the problem?

When playing Antichess OTB or against computer stalemate is shown as draw 1/2-1/2.

1. Start Antichess OTB on mobile
2. Play until a player has no legal moves
3. Popup says 1/2-1/2 Stalemate

### App version

0.22.6

### Device Info

Samsung Galaxy S21 / Android 15

### Additional information

Image

Image

Image

Image

CloudyDino:
# Fix antichess stalemate incorrectly shown as draw

Fixes #2814

## Summary

- In antichess, stalemate means the stalemated player **wins** — but the app was showing it as a draw (½-½)
- Root cause: the game-over logic in both OTB and offline computer controllers checked `isStalemate` before `variantOutcome`, so the variant-specific winner was never set
- Fix: reorder the conditionals to check `variantOutcome` first, allowing the dartchess library to correctly report the winner

## Changed files

- `lib/src/model/over_the_board/over_the_board_game_controller.dart` — reorder stalemate/variant checks
- `lib/src/model/offline_computer/offline_computer_game_controller.dart` — same reorder
- `test/view/over_the_board/over_the_board_screen_test.dart` — add regression test

## Test plan

- [x] New test: "Antichess stalemate is a win, not a draw" — starts an antichess OTB game from a position one move before stalemate, plays the move, and verifies the result is `0-1` / `GameStatus.variantEnd` / `winner: Side.black` (not `½-½`)
- [x] Verified the test **fails** with the old code and **passes** with the fix
- [x] All existing OTB tests pass (15/15)
- [x] Manual test: play an antichess game OTB to stalemate and confirm the result dialog shows the correct winner

| Before | After |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/375eeeff-212c-4964-9e5b-041f11fc15b6"/> | <video src="https://github.com/user-attachments/assets/c93a1d2e-473b-49ef-a87a-860a6eb18a02"/> |

### DIFF
--- a/lib/src/model/offline_computer/offline_computer_game_controller.dart
+++ b/lib/src/model/offline_computer/offline_computer_game_controller.dart
@@ -209,10 +209,6 @@ class OfflineComputerGameController extends Notifier<OfflineComputerGameState> {
       state = state.copyWith(
         game: state.game.copyWith(status: GameStatus.mate, winner: state.turn.opposite),
       );
-    } else if (state.currentPosition.isStalemate) {
-      state = state.copyWith(game: state.game.copyWith(status: GameStatus.stalemate));
-    } else if (state.currentPosition.isInsufficientMaterial) {
-      state = state.copyWith(game: state.game.copyWith(status: GameStatus.draw));
     } else if (state.currentPosition.isVariantEnd) {
       state = state.copyWith(
         game: state.game.copyWith(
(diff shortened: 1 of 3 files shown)


veloce (approved):
Thanks for the fix and test!

PR #2893, opened 2026-04-03
Fix: Url's clickable in Studies

### LINKED ISSUE #2831: Links in study are not clickable
There could be various links to external or lichess resources, but you can't click on them

MaartenD:
http(s) text are now clickable in Studies

https://github.com/user-attachments/assets/275f3cba-372d-41fe-8393-6e7f24d59850

Fixes #2831

### DIFF
--- a/lib/src/widgets/pgn.dart
+++ b/lib/src/widgets/pgn.dart
@@ -8,7 +8,6 @@ import 'package:flutter/material.dart';
 import 'package:flutter_linkify/flutter_linkify.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:lichess_mobile/src/app_links_service.dart';
-import 'package:linkify/linkify.dart';
 import 'package:lichess_mobile/src/model/account/account_preferences.dart';
 import 'package:lichess_mobile/src/model/common/node.dart';
 import 'package:lichess_mobile/src/model/common/uci.dart';
@@ -19,6 +18,7 @@ import 'package:lichess_mobile/src/utils/l10n_context.dart';
 import 'package:lichess_mobile/src/utils/rate_limit.dart';
 import 'package:lichess_mobile/src/widgets/adaptive_bottom_sheet.dart';
 import 'package:lichess_mobile/src/widgets/list.dart';
+import 'package:linkify/linkify.dart';
 import 'package:url_launcher/url_launcher.dart';
 
 const innacuracyColor = LichessColors.cyan;
--- a/lib/src/widgets/pgn.dart
+++ b/lib/src/widgets/pgn.dart
@@ -3,6 +3,7 @@ import 'package:collection/collection.dart';
 import 'package:dartchess/dartchess.dart';
 import 'package:dynamic_system_colors/dynamic_system_colors.dart';
 import 'package:fast_immutable_collections/fast_immutable_collections.dart';
+import 'package:flutter/gestures.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:lichess_mobile/src/model/account/account_preferences.dart';
@@ -15,6 +16,7 @@ import 'package:lichess_mobile/src/utils/l10n_context.dart';
 import 'package:lichess_mobile/src/utils/rate_limit.dart';
 import 'package:lichess_mobile/src/widgets/adaptive_bottom_sheet.dart';
 import 'package:lichess_mobile/src/widgets/list.dart';
+import 'package:url_launcher/url_launcher.dart';
 
 const innacuracyColor = LichessColors.cyan;
 const mistakeColor = Color(0xFFe69f00);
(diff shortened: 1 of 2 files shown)


lib/src/widgets/pgn.dart
```diff
@@ -1456,11 +1458,42 @@ class _MoveContextMenu extends ConsumerWidget {
...
-List<TextSpan> _comments(Iterable<String> comments, {required TextStyle textStyle}) => comments
-    .map(
-      (comment) => TextSpan(
-        text: comment,
-        style: textStyle.copyWith(fontSize: textStyle.fontSize! - 2.0),
-      ),
-    )
-    .toList(growable: false);
+List<TextSpan> _comments(Iterable<String> comments, {required TextStyle textStyle}) {
+  final urlRegex = RegExp(r'(https?:\/\/[^\s]+)');
```
veloce: Instead of a custom solution you should use `Linkify` as in the rest of the app.

Example: https://github.com/lichess-org/mobile/blob/1fda24c178b4a9ef71f70ef2d742c6a2dca1c24c/lib/src/view/chat/chat_screen.dart#L212

lib/src/widgets/pgn.dart
```diff
@@ -1456,11 +1473,23 @@ class _MoveContextMenu extends ConsumerWidget {
...
-      (comment) => TextSpan(
-        text: comment,
-        style: textStyle.copyWith(fontSize: textStyle.fontSize! - 2.0),
-      ),
-    )
-    .toList(growable: false);
+List<InlineSpan> _comments(Iterable<String> comments, {required TextStyle textStyle}) {
+  return comments
+      .expand((comment) {
+        final elements = linkify(comment, options: const LinkifyOptions(humanize: false));
```
veloce: Cannot find where that `linkify` function is coming from?

lib/src/widgets/pgn.dart
```diff
@@ -1456,11 +1473,23 @@ class _MoveContextMenu extends ConsumerWidget {
...
 
-List<TextSpan> _comments(Iterable<String> comments, {required TextStyle textStyle}) => comments
-    .map(
-      (comment) => TextSpan(
-        text: comment,
-        style: textStyle.copyWith(fontSize: textStyle.fontSize! - 2.0),
-      ),
-    )
-    .toList(growable: false);
+List<InlineSpan> _comments(Iterable<String> comments, {required TextStyle textStyle}) {
```
veloce, line 1485: Don't have the time to good deeply in this for now.

What I can tell so far is that:

- we don't want to mix Linkfiy and TapGestureRecognizer


The whole thing probably needs another refactoring to support comments.

MaartenD:
**Comment 1 — TapGestureRecognizer leak & inconsistent link handling**

> _comments() creates a new TapGestureRecognizer() on every build and never disposes it. Also, the tap handler calls launchUrl directly, bypassing AppLinksService.onLinkifyOpen.

**Resolved.** Replaced the manual TextSpan + TapGestureRecognizer approach in _comments() with a WidgetSpan containing a Consumer + Linkify widget, consistent with all other comment rendering in the file. This means:

Recognizer lifecycle is managed internally by Linkify — no more leaks.
All link taps are routed through AppLinksService.onLinkifyOpen, so internal Lichess links and @user tags are handled consistently with the rest of the app.
AppLinksService.kLichessLinkifiers is now used, adding @user tag support that was previously missing.
Three imports that became unused after the refactor were removed: gestures.dart, linkify/linkify.dart, and url_launcher.

**Comment** 2 — Hard-coded line numbers in test doc comments

>Doc comments reference specific line numbers (e.g. "line 843", "line 524") which will drift as the file changes.

**Resolved.** All line number references in pgn_test.dart have been replaced with the names of the relevant widgets:

line 524 → _PgnTreeView root comment rendering
line 843 → _TwoColumnMainlinePart comment rendering

veloce:
I have updated your branch @MaartenD 

I have some concerns about the utility of this PR though. It adds some overhead (parsing all the comments several times, for each linkifier). I wonder if it can slow down the rendering of some big PGN files?

And the utility. It seems we can have some links in PGN but it is not the most common.

There are other parts in studies where we have text that might contain links. It might be more useful to linkify these, than the PGN.

MaartenD:
> I have updated your branch @MaartenD
> 
> I have some concerns about the utility of this PR though. It adds some overhead (parsing all the comments several times, for each linkifier). I wonder if it can slow down the rendering of some big PGN files?
> 
When it's a big PGN file, i can imagine that this will slow down the process.

> And the utility. It seems we can have some links in PGN but it is not the most common.
> 
Also agree that's not a common use case.

> There are other parts in studies where we have text that might contain links. It might be more useful to linkify these, than the PGN.

Can you tell me which parts could contain text (as a link) in studies where you want to use linkify? I will remove te linkify part from the comments and add the parts you where you want the links to be available. Is that an option? Let me know.

veloce:
See that study for instance: https://lichess.org/study/UDvstVM0 (using web not the app).

There is a description at the bottom, containing links. We are not showing that in the app at the moment but we should somehow. Then this part should be linkified.

I think this PR can be closed for now. Sorry for not having found earlier the potential performance issue and the fact it is not really useful earlier.

MaartenD:
No problem!

PR #2896, opened 2026-04-04
Handle broadcast deep links with #players/{playerId} fragment

### LINKED ISSUE #2895: Broadcast deep links with #players/{id} fragment don't navigate to specific player
### What is the problem?

Opening a broadcast URL with a #players/{fideId} fragment (e.g. https://lichess.org/broadcast/grenke-chess-festival-2026--freestyle-open-a/round-3/ioIYmuar#players/250511) navigates to the broadcast round screen but does not open the specific player's results page. The same URL works correctly on the website, where it opens the players tab and shows the selected player's results.

  1. Open a broadcast URL with a player fragment, e.g.:
  https://lichess.org/broadcast/grenke-chess-festival-2026--freestyle-open-a/round-3/ioIYmuar#players/250511
  2. The app opens the broadcast round screen on the default tab
  3. Expected: the app should open the players tab and navigate to that specific player's results screen (same as the
  website)

### App version

0.22.7

### Device Info

Android 16

### Additional information

In app_links_service.dart, the broadcast deep link handler passes the full fragment (e.g. players/250511) to BroadcastRoundTab.tabOrNullFromString(), which does exact string matching. Since players/250511 doesn't match players, it returns null and the default tab is shown.
I'll try make a PR to fix this

wesselvankessel:
## Summary

- Handle `#players/{playerId}` URL fragments in broadcast deep links, navigating to the player results screen with the players tab underneath (matching lichess.org website behavior)
- Made `player` parameter optional in `BroadcastPlayerResultsScreenLoading` so the screen can be opened from a deep link without prior player data (resolves player info from the API)
- Added tests for both `#players` (tab only) and `#players/{fideId}` (player screen + back navigation)

Fixes #2895

## Test plan

- [x] All 793 tests pass
- [x] `flutter analyze` clean
- [ ] Open `https://lichess.org/broadcast/.../roundId#players/250511` — should navigate to player results screen
- [ ] Press back — should land on the round screen with players tab selected
- [ ] Open `https://lichess.org/broadcast/.../roundId#players` — should open round screen on players tab
- [ ] Existing broadcast deep links (round, game) still work

### DIFF
--- a/lib/src/view/broadcast/broadcast_player_results_screen.dart
+++ b/lib/src/view/broadcast/broadcast_player_results_screen.dart
@@ -105,8 +105,8 @@ class BroadcastPlayerResultsScreen extends ConsumerWidget {
   Widget build(BuildContext context, WidgetRef ref) {
     final displayPlayer =
         player ??
-        switch (ref.watch(_playerAndTournamentProvider((tournamentId, playerId)))) {
-          AsyncData(value: final data) => data.$1.playerWithOverallResult.player,
+        switch (ref.watch(broadcastPlayerProvider((tournamentId, playerId)))) {
+          AsyncData(value: final data) => data.playerWithOverallResult.player,
           _ => null,
         };
 


SergioGlorias:
@wesselvankessel please note: have players without fide id
for this cases is player name like: https://lichess.org/broadcast/tcec-s29-leagues--superfinal/match/RSIGxDYD#players/Stockfish%20dev-20260318-d173a065 (is a bot but is same thing)

wesselvankessel:
Thanks for flagging this. In mobile, the fragment suffix is treated as a generic `playerId`, so it works both for FIDE IDs like `250511` and for name-based player IDs when no FIDE ID is available.

I’ve updated the deep-link handling to decode the fragment suffix before navigating, and I added a regression test for the case you mentioned: `#players/Stockfish%20dev-20260318-d173a065`

The new test verifies that this resolves to playerId `Stockfish dev-20260318-d173a065`, opens the player results screen, and keeps the players tab underneath for back navigation.

lib/src/view/broadcast/broadcast_player_results_screen.dart
```diff
@@ -102,10 +102,19 @@ class BroadcastPlayerResultsScreen extends StatelessWidget {
   }
 
   @override
-  Widget build(BuildContext context) {
+  Widget build(BuildContext context, WidgetRef ref) {
+    final displayPlayer =
+        player ??
+        switch (ref.watch(_playerAndTournamentProvider((tournamentId, playerId)))) {
```
veloce: Technically you don't need the tournament so you should call directly the `broadcastPlayerProvider`

lib/src/view/broadcast/broadcast_player_results_screen.dart
```diff
@@ -102,10 +102,19 @@ class BroadcastPlayerResultsScreen extends StatelessWidget {
...
+          AsyncData(value: final data) => data.$1.playerWithOverallResult.player,
+          _ => null,
+        };
+
     return PlatformScaffold(
       appBar: PlatformAppBar(
-        title: BroadcastPlayerWidget(player: player, showFederation: false, showRating: false),
+        title: displayPlayer != null
```
veloce, line 115: What are the cases we can have a `playerId` but no player?

lib/src/view/broadcast/broadcast_player_results_screen.dart
wesselvankessel: Good point, I'll change it

lib/src/view/broadcast/broadcast_player_results_screen.dart
wesselvankessel: When opening from a deep link (#players/{id}), we only have the ID from the URL fragment, not the full player object. The player object is only available when navigating from within the app.

PR #2897, opened 2026-04-04
Prevent http requests we know will fail

freebeartogoodhome:
If the user is not logged in the Lichess server will deny requests to access the master game database.
I would like to prevent requests to the master database when we are not logged in.
I noticed this when I tried Practice Mode on my phone.
logscreenshot


veloce:
Well, thanks for the report. ~~Let me check that with the server dev team first.~~

veloce (approved):

PR #2904, opened 2026-04-05
Only enable "Challenge a friend" when online

freebeartogoodhome:
This will change it to match the other items in the menu that don't work while offline.

### DIFF
--- a/lib/src/view/play/play_menu.dart
+++ b/lib/src/view/play/play_menu.dart
@@ -28,6 +28,7 @@ class PlayMenu extends ConsumerWidget {
         _Section(
           children: [
             ListTile(
+              enabled: isOnline,
               onTap: () {
                 // Pops the play bottom sheet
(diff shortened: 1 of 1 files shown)


veloce (approved):
Good catch!

J0J0XMR:
This means I can only send a challenge only when I am online or only when my friend is?

freebeartogoodhome:
The app user must be online.

If you are offline and click on the Play button some things will be greyed out. Arena tournaments, for example. This is just like that. We prevent the app user from clicking on it because we know it won't work.

J0J0XMR:
> The app user must be online.
> 
> If you are offline and click on the Play button some things will be greyed out. Arena tournaments, for example. This is just like that. We prevent the app user from clicking on it because we know it won't work.

OK great, because sending a request to an offline friend is very useful (they'll accept when they're ready).

PR #2911, opened 2026-04-07
Fix last edited PGN header not saved on share (#2850)

### LINKED ISSUE #2850: Black elo missing from Sharing & export PGN from analysis board

CloudyDino:
Fixes #2850

## Summary

- When exporting PGN from the analysis board's Share & Export screen, whichever text field was still focused when tapping "Share PGN" would retain its default `?` value in the exported PGN. This most commonly affected BlackElo since it's the last rating field.
- We now explicitly flush all text controller values into state before generating the PGN, making it independent of focus timing.

## Changes

- `analysis_share_screen.dart`: Iterate through all text controllers and call `updatePgnHeader` for each before calling `makeExportPgn()`
- New test file `analysis_share_screen_test.dart`: Verifies that the last edited field (BlackElo, WhiteElo) is correctly persisted when sharing without first unfocusing

## Test plan

- [x] Run `flutter test test/view/analysis/analysis_share_screen_test.dart`
- [x] Tested manually with different fields to confirm that the fix works.

### DIFF
--- a/lib/src/view/analysis/analysis_share_screen.dart
+++ b/lib/src/view/analysis/analysis_share_screen.dart
@@ -162,6 +162,18 @@ class _EditPgnTagsFormState extends ConsumerState<_EditPgnTagsForm> {
                   builder: (context) {
                     return FilledButton(
                       onPressed: () {
+                        // Flush all text field values into state before
+                        // generating the PGN. We can't rely on unfocus
+                        // alone because the focus listener may not fire
+                        // synchronously or may not fire at all if focus
+                        // was already lost.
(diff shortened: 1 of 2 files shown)


veloce (approved):
LGTM

PR #2929, opened 2026-04-08
Fix bottom nav staying visible when opening deep links

### LINKED ISSUE #2910: Bottom menu doesn't disappear when opening deep links
### What is the problem?

When opening a game deep link, the bottom menu doesn't disappear, making the analysis screen shrink in size. Please refer to the following screenshot.

Image

1. Open a game Deep Link
2. The bottom menu stays on the screen instead of getting replaced by the Analysis screen menu
3. The whole Analysis screen is shrinking because of that

### App version

0.23.2

### Device Info

Xiaomi Redmi Note 10 5G (MIUI 14 / Android 13)

### Additional information

_No response_

CloudyDino:
# Fix bottom nav staying visible when opening deep links

## Summary

- Use `rootNavigator: true` when pushing deep link routes so they display above the tab scaffold instead of inside it, hiding the bottom navigation bar
- This is a one-line change in `handleAppLink()` in `app_links_service.dart`
- Fixes #2910

## Root Cause

`handleAppLink()` was calling `Navigator.of(context).push(route)` which pushed onto the **tab navigator** (since the context comes from `currentNavigatorKeyProvider`, a tab-specific key). This meant deep link screens were rendered *within* the tab, leaving the bottom navigation bar visible and shrinking the screen.

The fix changes this to `Navigator.of(context, rootNavigator: true).push(route)`, which pushes onto the **root navigator** above `MainTabScaffold` -- consistent with how every other full-screen navigation in the app works.

## Deep links tested

All deep link types go through `handleAppLink()` and are affected by this fix. Tested on Android emulator (API 36) against lichess.dev:

### Game (`/3K42eD3l`)

| Before | After |
|--------|-------|
| 1_game | 1_game |

### Study (`/study/9hdiWME4`)

| Before | After |
|--------|-------|
| 2_study | 2_study |

### Tournament (`/tournament/HWT5E5Ja`)

| Before | After |
|--------|-------|
| 3_tournament | 3_tournament |

### Broadcast (`/broadcast/.../BoiVuw5O`)

| Before | After |
|--------|-------|
| 4_broadcast | 4_broadcast |

### DIFF
--- a/lib/src/app_links_service.dart
+++ b/lib/src/app_links_service.dart
@@ -208,7 +208,7 @@ class AppLinksService {
 
     if (routes != null) {
       for (final route in routes) {
-        Navigator.of(context).push(route);
+        Navigator.of(context, rootNavigator: true).push(route);
       }
     } else {
       final isChallengeLink = await _tryResolveChallengeLink(context, uri);


veloce (approved):
Thanks!

PR #2932, opened 2026-04-09
fix: show puzzle storm-only days in user activity

sviluppatoredisoftware:
Fixes the mobile activity feed so days with only Puzzle Storm are no longer treated as empty.

The issue was that `storm` was rendered in the UI, but not included in `UserActivity.isEmpty`, so those days were filtered out before rendering.


veloce (approved):
Thanks!

Issue #2686, opened 2026-02-28
Few other minor requests for "play vs computer"


ijm8710:
- Can you handle immersive as you do in other parts of app (like puzzle streak and storm) whereby you can Backswipe until once game starts. Then once game starts it's locked but once it ends again you can Backswipe out?
- from a muscle memory perspective, resign is towards left side for regular games. So perhaps resign before undo move for consistency?
- when a normal game ends, the resign button turns into a popup showing the game result and how it ended (along with analysis board access). Thoughts of handling it similarly for here?
- lastly, back and fwd arrows are nice to be able to rewind moves easier. Are you fully opposed to having them here like we have on bottom right in a normal game?

Other things I'd like to see down road:
- archive of last 5-10
- Computer analysis/learn from mistakes
- Timed games 
- Premoves

And things not working great:

- Feedback can be inconsistently over-aggressive
- Feedback freezing up
- Computer will play super normal then make series of horrific moves (not balanced play)

PR #2946, opened 2026-04-10
Change the order of resign and undo move buttons

r3econ:
- Changes the order of buttons when playing against the computer - _resign_ before _undo move_
- Requested in https://github.com/lichess-org/mobile/issues/2686

#### Before and after

SCR-20260410-pmsa-tile

### DIFF
--- a/lib/src/view/offline_computer/offline_computer_game_screen.dart
+++ b/lib/src/view/offline_computer/offline_computer_game_screen.dart
@@ -360,16 +360,16 @@ class _BottomBar extends ConsumerWidget {
           icon: Icons.menu,
         ),
         BottomBarButton(
+          label: context.l10n.resign,
+          onTap: gameState.game.resignable ? () => _showResignDialog(context, ref) : null,
+          icon: CupertinoIcons.flag,
+        ),
+        BottomBarButton(
           label: context.l10n.takeback,
           onTap: gameState.canTakeback && (gameState.game.casual || gameState.game.practiceMode)
               ? () => ref.read(offlineComputerGameControllerProvider.notifier).takeback()
               : null,
           icon: CupertinoIcons.arrow_uturn_left,
-        ),
-        BottomBarButton(
-          label: context.l10n.resign,
-          onTap: gameState.game.resignable ? () => _showResignDialog(context, ref) : null,
-          icon: CupertinoIcons.flag,
         ),
         BottomBarButton(
           label: context.l10n.getAHint,


whole101:
Please add an option to change it back, this is super annoying and makes zero sense.

ijm8710:
>  Please add an option to change it back, this is super annoying and makes zero sense.

I can assure you something like this is not going to have a setting option. It's gonna be one way or another either way.

Why do you prefer the prior way?

The reason the change was suggested was to be consistent with how it is for normal games. 
Plus takeback and hint kind of go together and it makes sense for them to be near each other

Besides the fact that you may have gotten used to the prior way (computer has only been around for a month or two) and therefore you'll get used to it very quickly, may I ask any logical reasons you think the prior way made more sense?

whole101:
@ijm8710 it isnt a "normal game", so I dont understand why there even has to be is a resign button, and not just a "new game" button. But ok, for some reason there is a resign button. 

It was better placed on the left side cuz most people are right handed, and use the right hand to move the pieces while the left does kinda nothing, but could access the undo button very easy and fast. 

Its less about muscle memory its more that _undo_  is just less comfortable to reach now. The UI got a lot worse in favor for a completely useless resign button. 

The _undo_ function is literally the only reaon to even use "play against Computer". Try stuff, see how it might turn out, undo, undo, undo, undo, try something new. And so on, all day long maybe. And then try it in a normal game, and resign there if you really want to.

ijm8710:
> @ijm8710 it isnt a "normal game", so I dont understand why there even has to be is a resign button, and not just a "new game" button. But ok, for some reason there is a resign button. 
> 
> 
> 
> It was better placed on the left side cuz most people are right handed, and use the right hand to move the pieces while the left does kinda nothing, but could access the undo button very easy and fast. 
> 
> 
> 
> Its less about muscle memory its more that _undo_  is just less comfortable to reach now. The UI got a lot worse in favor for a completely useless resign button. 
> 
> 
> 
> The _undo_ function is literally the only reaon to even use "play against Computer". Try stuff, see how it might turn out, undo, undo, undo, undo, try something new. And so on, all day long maybe. And then try it in a normal game, and resign there if you really want to.

Two things

1) I use resign much more than undo. I will almost never undo and prob will resign a completely lost game. I use computer more to play offline rather than to teach. But obv that makes neither of us wrong. So users will be split there

2) look at my ChatGPT prompt. That was first attempt unbiased. The way you hold it is totally fine but based on what you said and your use case. A right handed person will have preference to bottom right. That's unfortunate for you but more beneficial for your use case to majority of users.

It stinks but something like this will almost never have a settings toggle though.

image

whole101:
Lol so generative AI still didn't figure out how hands and fingers work? But maybe ask chatgpt to explain my reply to you, I think you kinda missed what I tried to say.

ijm8710:
> Lol so generative AI still didn't figure out how hands and fingers work? But maybe ask chatgpt to explain my reply to you, I think you kinda missed what I tried to say.

Did you not say the button you use the most you prefer to be most accessible to you which is bottom left.

But per the included chart, seems you hold it a way that's in minority. Generative ai wouldn't just make something up, it's fetching from relevant articles to feed it. This is not opinion. 

The other half of whether  or not you resign button should be shown outright has less relevance because that's more a you thing. I know I and many users want it and so did the developer. 

image

r3econ:
>  Please add an option to change it back, this is super annoying and makes zero sense.

Hi @whole101 👋 author of the PR here. I do think this PR improves the UX and it is most likely why it was accepted. The change here is optimized for the majority of the users (right-handed).

Having said that, there will always be users that do not like or find the interface suboptimal. We sadly can't make everybody happy.

If you really want to get the visibility for this particular UX matter I suggest opening a new GitHub Issue where it would get more visibility. Here, because the PR is closed, only 4 people see the discussion

PR #2952, opened 2026-04-11
fix challenge title if unliminited time

tom-anders:
Also, factor out the perf+time `Row` that we had in three places into `_GameParamsDisplay` (feel free to suggest a better name :D )

## Before:

Screenshot_1775900955

## After:

Screenshot_1775900965


ijm8710:
Should the word unlimited be added after the icon as well in the top bar in the backdrop?

veloce (approved):
Thanks!

PR #2955, opened 2026-04-11
Add Quick Pairing iOS widget

cappylot:
@veloce @r3econ I redid the IOS widget and complied with the existing structure you mentioned. I hope this is correct now!


r3econ:
@cappylot, this looks awesome! 🤩 

SCR-20260411-rvib

---

I've tested it. It works well, I am pretty sure the users will love it 😍 I gathered my feedback, ideas and thoughts below.

1. Maybe we don't need the medium-sized widget since half of it if empty anyways?
2. The current design does not fit the other widget, it looks foreign. Some ideas how to improve it:
	- Adding Lichess logo like in the blog widgets
	- Maybe a divider below title?
	- If we drop the medium widget and only use the small one the title could simply be _Play_

SCR-20260411-shhs

3. Kid mode should not restrict quick pairing. We have a kid mode check in blog widgets, because blogs are not shown when the kid mode is active. This does not apply to quick pairing
4. _Play once to set Quick Pairing_ placeholder is a little bit misleading. I had played a game with a friend and the widget still showed me that message which made me think it is broken. It took me a while to figure out that the widget takes the last settings when the game was started using _quick pairing_ and not directly by challenging someone. I see a couple of solutions:
    - Instead of showing _Play once to set Quick Pairing_ just show widget with default settings (e.g. Standard, 5+3)
    - Open the app and show the game settings selection screen (screenshot below). However, this will probably require a separate deeplink and navigation flow, so maybe the first solution is better
    - Additionally, also the game settings when starting the game directly by challenging a user/friend could be saved and picked up by the widget 

image

SCR-20260411-quek

Let me know what you think 🙂

cappylot:
@r3econ Thanks for the reply :)

1. I agree with you on the medium-sized widget, I will get rid of it for now. Maybe later on I'll introduce a medium-sized one that lets the user pick between three different time controls or something similar. 
2. I'll working on making the widget design fit the blog widget
3. Got it! 
4. I will make it that the placeholder shows the default configuration. Since the widget's purpose is to make starting a game as fast as possible I would refrain from the second option you mentioned and start game search immediately. 

I will get to work now 👍

r3econ:
Cool! 😀 🦾 

I am currently also working on a widget, with puzzles. It's not ready yet but I am slowly getting there with the design. I thought sharing current state might help you with the design (which I always find the hardest to get right or at least good enough)

SCR-20260412-mtgp-tile

cappylot:
@r3econ Looks great! 

I just finished the changes we discussed and will push them now 🔥

- fits design of blog widget
- shows time control icon
- no limitation for kid mode 
- no medium sized version
- shows default 5+3 time control for placeholder

Simulator Screenshot - iPhone 17 Pro - 2026-04-12 at 16 36 00

ios/LichessWidgets/Quick Pairing Widget/QuickPairingWidgetLayout.swift
```diff
@@ -0,0 +1,41 @@
+//
+//  QuickPairingWidgetLayout.swift
+//  Runner
+//
+//  Created by Orfeo Kaplow on 12.04.26.
+//
+
```
r3econ: Remove the header comment

ios/LichessWidgets/Quick Pairing Widget/QuickPairingWidgetLayout.swift
```diff
@@ -0,0 +1,41 @@
...
+import CoreGraphics
+
+/// Layout constants shared across blog feed widget views.
+enum QuickPairingWidgetLayout {
```
r3econ: Here the comment is incorrect. It could also be dropped, it's clear what this file is for

ios/LichessWidgets/Quick Pairing Widget/QuickPairingWidgetLayout.swift
```diff
@@ -0,0 +1,41 @@
...
+    // Thumbnail
+    static let minThumbHeight: CGFloat = 20
+    static let thumbnailCornerRadius: CGFloat = 6
+
+    // Small widget footer
+    static let smallFooterTopPadding: CGFloat = 6
+
+    // Error state
+    static let errorIconSize: CGFloat = 20
+    static let errorStackSpacing: CGFloat = 6
```
r3econ: Is this whole file actually needed? I can see that the constants defined here in `QuickPairingWidgetLayout` are not used in `QuickPairingWidget` - the layout values are still hardcoded there. It looks like this file was copied from blog widgets, renamed and is not used

r3econ (commented):
The new design looks way better! 🤩 🚀 

I've found some minor problems and commented them in the diff. 

I've also found that the whitespaces used in the new or modified files are incorrect - we use spaces and not tabs (defined in `.editorconfig`, also Xcode's default). Could you reformat the files so it fits the convention? At the moment it's mixed, sometimes tabs, sometimes spaces. Should be only spaces 

Image

veloce:
Thanks @cappylot and @r3econ for your work, this looks really promising!

For now I just wanted to add my thoughts about the behaviour of the widget, reacting to that comment:

> Instead of showing Play once to set Quick Pairing just show widget with default settings (e.g. Standard, 5+3)
Open the app and show the game settings selection screen (screenshot below). However, this will probably require a separate deeplink and navigation flow, so maybe the first solution is better
Additionally, also the game settings when starting the game directly by challenging a user/friend could be saved and picked up by the widget

I think showing the default settings instead of "play once ... " is a good idea.

Then what was the plan to select the ideal time control? I don't remember if we've discussed this already. I see a couple of options:

- use the last played one (whether it is from quick pairing or challenge, any online game actually)
- store locally all the time controls played (with a cap) and show the most played one
- let the user choose the time control for the widget

Option 3 is prob overkill; my pref goes to 2 I guess, but no strong feelings here.

Also, should this behave similarly to the existing quick actions?

Simulator Screenshot - iPhone 17 - 2026-04-13 at 14 26 05

cappylot:
Hey @veloce, regarding the 2nd option: for a lot of people I think, they play in sort of sessions or based on their current mood. If for example I am playing a lot of blitz currently or recently I would want my widget to let me launch a blitz game. If my total amount of games played were dominated by classical though, my widget would suggest classical even though I want to play blitz. 

That's why I would either keep it as is, meaning the widget suggest the time control that was last played, or we do the 2nd option but only count games of the last 2-3 days so it fits the users current mood?

veloce:
Hey @cappylot you're right, I hadn't thought of that. Let's keep the option 2 (and this is the way it works for the quick actions too).

As an improvement (that can be done later) we could allow the user to modify the widget to choose their own time control. Ie: choose between last played or custom with a time control selector.

cappylot:
@veloce just to make sure there is no misunderstanding here: do you want me to patch the remaining things that @r3econ commented and leave as is, or implement the 2nd option you mentioned with the widget showing users most popular time control but based on the last 2-3 days?

veloce:
@cappylot sorry I realise that wasn't clear.

Yeah just patch the remaining things that were commented. And for the rest leave as is. It is already using last time control played which is fine. Just make sure it takes it from both quick pairing and challenges (quick action logic should be updated like that too but that is for another PR).

cappylot:
@veloce got it, thanks! I'll try to get this down later today

ios/LichessWidgets/Quick Pairing Widget/QuickPairingWidgetLayout.swift
cappylot: Yeah, I was lazy and had just copied it from your widget. I remove the unused constants now.

cappylot:
@veloce @r3econ should be all correct now, lmk 🤝

veloce:
@cappylot you have conflicts in your branch; please solve them in order to proceed.

veloce:
@cappylot sorry for the delay on this and this is taking longer. Now the analysis is failing, I think due to some changes in app_links_service. Can you see what it is? Thanks!

PR #2965, opened 2026-04-12
 Add iOS Daily Puzzle widget

r3econ:
Adds a new iOS home screen widget that displays today's chess puzzle from lichess.org. Scroll down for demos.

## Features

- Large widget — board + header with the Lichess logo, title, side-to-move indicator, update time, and an optional rating (toggled via a widget configuration intent)

### Board rendering:

- Parses the puzzle FEN and renders the full 8×8 position with piece images
- Highlights the last move square
- Flips the board to always show the side to move at the bottom
- Reads the board theme and piece set from the shared _App Group_, so the widget automatically matches whatever the user has configured in the app
- Supports all board themes — both solid-colour (brown, blue, green, etc) and image-backed (wood, marble, etc.)

### Data & refresh:

- Fetches from `/api/puzzle/daily`; uses the host configured in the main app (supports local dev server)
- Refreshes every 6 hours
- Tapping the widget opens the puzzle natively via a universal link

### Notes for reviewers

- Board texture and piece images are sourced from the `ChessgroundAssets` swift package from `flutter-chessground`

- `EXTENSIONS.md` documents how external contributors can build and test the widget locally using their own Apple Developer account, and what a Lichess org member needs to do when merging a new extension.

#### Theme support

themes1

themes_2

#### Error state

errors

#### Configuration

config

#### Videos

- Opening puzzle screen from the widget + widget config

https://github.com/user-attachments/assets/5ca9ccce-2a77-4178-8c8b-5aafd4618533

- Changing app and widget chessboard theme and piece set

https://github.com/user-attachments/assets/b9a330a9-74f0-4f16-8508-f695c60f76e9


veloce:
What an impressive work! I will take the time to test and review it. Just responding to your questions for now:

> Is small puzzle widget of any value? It only shows the board, it is very small and if we ever add another chessboard-based widget it will bring confusion since it has no title. Maybe we should simply drop it and only have a big one, or a medium one with title (which I implemented and removed but could bring it back).

Yes indeed the small puzzle widget has low value. Moreover if we add later other board widgets (like correspondence games for instance) we'd not be able to distinguish them. So I think you should drop it.
Whether we do the medium size is up to you. To me the added value is that it takes less vertical space than the full size one so it is interesting for people wanting to save space. That being said it could be added in another PR as the full size one is already good for a first version.

> Image assets are currently copied from flutter-chessground using a script. This is convinient and rather low maintenance. However, I've been thinking of having a separate repo (e.g. native-chessground) that would provide a simple Swift and Android libraries containing chessboard and piece sets in native formats. This would mean that we do not have to have these files in the mobile repo. @veloce, if you find this idea legit, please decide - I could either prep this now and we wait with merging the PR or we do merge it and extract the resources later.

I think this is a good idea yes. If you feel like it is not too much work we could do it already in this PR to avoid adding too much assets in this repo.

r3econ:
I think we're slowly getting there 😎 
- small widget dropped ✅ 
- updated refresh logic (every 6 hours, random start) ✅ 
- chessboard resources fetched from external package (added in https://github.com/lichess-org/flutter-chessground/pull/92) ✅ 
- improved design with chessboard corners matching the container shape ✅ 

SCR-20260413-udsd-tile

HaonRekcef:
Looks amazing :100: 

Minor nitpick:
Can we have a nice puzzle in the preview instead of a standard opening position?

r3econ:
> Looks amazing 💯
> 
> Minor nitpick: Can we have a nice puzzle in the preview instead of a standard opening position?

@HaonRekcef of course, I'd just need the FEN, do you have something specific in mind? 🙂

HaonRekcef:
What about a smothered mate like this: 
image

FEN:
`1n3rk1/4ppbp/rq1p2p1/3P4/2p1P3/2N2P1n/PPN3PP/R1BQ1R1K b - - 11 18`

from this puzzle: https://lichess.org/training/cGsUy

r3econ:
Indeed, it looks classy 🥇 

image

ijm8710:
Noah's puzzle pointed out another minor graphical bug

Puzzles over 999 should have no comma 😉

r3econ:
I'll check tomorrow but I suspect it's not a bug but a feature. Number formatting depends on the device's region setting. So it could well be it is correct

veloce (approved):
LGTM! Great work as usual.

PR #2972, opened 2026-04-13
Fix show solution giving wrong answer in puzzle

### LINKED ISSUE #2876: Show Solution giving wrong answer In Todays Daily Puzzle
https://github.com/user-attachments/assets/29affd68-a4ac-4d94-9b45-0aae4c5c6060

Show solution doesn't play the correct move, or if the user tried to make a move before it will simply play that wrong move again.

An example puzzle where this happens is todays puzzle of the day:
https://lichess.org/training/9gsMd

Version 0.22.7

HaonRekcef:
The problem was that castling moves from the solution string were converted into Move objects without being normalized.
This would lead to a different Move object being created compared to when a user plays the move manually and then the solution would not be played.

Includes a test that fails before and passes now.
Puzzle to test it on:
https://lichess.org/training/9gsMd

Fixes #2876


veloce (approved):
Thanks!

Issue #2914, opened 2026-04-07
Otb time control menus can get out of sync


ijm8710:
![image](https://github.com/user-attachments/assets/63c2552c-74bc-4f61-82fb-876d9ed3e424)

The way to reproduce is to start a game and resign or draw, then switch options but don't start the new game. Upon reentering the menu, several times I've seen the increment with unlimited time (as seen in attached photo) or even the reverse (real time chosen but no increment slider). 

*May take repeating reproduction steps a few times but I can usually force it to happen after a few sequences

ghost:
I can take a look at that. I was able to reproduce it easily in the app.

ijm8710:
> I can take a look at that. I was able to reproduce it easily in the app.

Thanks man, what would also be nice is if it was able to leverage the previous time control. For some reason it seems to always want to go back to 5+12 regardless of what I played last.

Lmk if you need it as a separate issue

Issue #2050, opened 2025-08-06
Have LeelaOdds retain previous time controls like all the other bots


ijm8710:
Currently, I believe all the other bots will retain your prior selected time controls which is nice.

The only exception seems to be the LeelaOdds bots, I think becuase their game creation setup is slightly different.

It will still use the most recent time control you have for the other bots but if this is change, that change is not retained.

This would be particularly helpful since we currently cannot rematch the bot, so being able to have it retain the time control you just set would be a temporary assist there.

PR #2976, opened 2026-04-13
Fix: Inconsistant UI for OTB time controls

### LINKED ISSUE #2956: Inconsistent UI for OTB time controls
### What is the problem?

Image

When both sliders (time per side and increment) are set to 0, they disappear and the game starts with unlimited time—which makes sense.

Image

However, the time control menu still shows “Real Time” instead of “Unlimited,” which is confusing.

Not a major issue, but I figured I share it anyway.

Keep up the good work and thanks a lot.

See description.

### App version

0.22.7

### Device Info

Android 16

### Additional information

_No response_

MaartenD:
See attached video.

https://github.com/user-attachments/assets/fa5bd513-bdd8-4f52-8b3c-e54ab43679ac

Fixes #2956

### DIFF
--- a/lib/src/model/settings/over_the_board_preferences.dart
+++ b/lib/src/model/settings/over_the_board_preferences.dart
@@ -70,14 +70,14 @@ sealed class OverTheBoardPrefs with _$OverTheBoardPrefs implements Serializable 
     required bool flipPiecesAfterMove,
     required bool symmetricPieces,
     @Default(TimeControlType.realTime) TimeControlType timeControlType,
-    @Default(TimeIncrement(300, 3)) TimeIncrement timeIncrement,
+    @Default(TimeIncrement.blitzDefault()) TimeIncrement timeIncrement,
   }) = _OverTheBoardPrefs;
 
   static const defaults = OverTheBoardPrefs(
     flipPiecesAfterMove: false,
     symmetricPieces: false,
     timeControlType: TimeControlType.realTime,
-    timeIncrement: TimeIncrement(500, 3),
+    timeIncrement: TimeIncrement.blitzDefault(),
   );
 
   factory OverTheBoardPrefs.fromJson(Map<String, dynamic> json) {
--- a/lib/src/model/settings/over_the_board_preferences.dart
+++ b/lib/src/model/settings/over_the_board_preferences.dart
@@ -1,6 +1,7 @@
 import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:freezed_annotation/freezed_annotation.dart';
 import 'package:lichess_mobile/l10n/l10n.dart';
+import 'package:lichess_mobile/src/model/common/time_increment.dart';
 import 'package:lichess_mobile/src/model/settings/preferences_storage.dart';
 
 part 'over_the_board_preferences.freezed.dart';
@@ -41,6 +42,10 @@ class OverTheBoardPreferencesNotifier extends Notifier<OverTheBoardPrefs>
   Future<void> setTimeControlType(TimeControlType type) {
     return save(state.copyWith(timeControlType: type));
   }
+
+  Future<void> setTimeIncrement(TimeIncrement timeIncrement) {
+    return save(state.copyWith(timeIncrement: timeIncrement));
+  }
 }
 
 enum TimeControlType {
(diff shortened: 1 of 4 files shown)


ijm8710:
1. May be nitpicky but is the way the word minutes remains vertically for a few seconds hard to clean out. If it's a lot of work no worries but does look odd.
2. I had also created a ticket for #2914 which @freebeartogoodhome had planned to take a look at but may make sense to approach those in tandem (that ticket and this as they're pretty similar). Part of reason I'm asking is perhaps this would fix that simultaneously already
3. In that same ticket I also noted that a lot of times the otb time controls would randomly reset back to 5+12 or 5+3. Would be great if they could always consistently retain the prior otb time control played. But still keep it separate from the time controls for regular games off the play button
  - this would echo the behavior for bots where they retain prior time controls. I had even an older ticket on this where all the bots do it well besides the leela ones because they have a unique game creation design #2050

freebeartogoodhome:
I'll still get to #2914 if no one else does first. It grew into something bigger as I started looking at the ways offline_computer_game and over_the_board handle different things.

MaartenD:
Hi @ijm8710 
>May be nitpicky but is the way the word minutes remains vertically for a few seconds hard to clean out. If it's a lot of work no worries but does look odd.

I don't know if that's hard to do, if @veloce also want i "removed" i will take a look.

>In that same ticket I also noted that a lot of times the otb time controls would randomly reset back to 5+12 or 5+3. Would be great if they could always consistently retain the prior otb time control played. But still keep it separate from the time controls for regular games off the play button

At first i tought this was only a UI issue, but that wasn't completely the case. I find out that the Increment and Timecontrol weren't synced correctly (in my opinion). I think i fixed it, we shall see if @veloce agrees with my solotion.

freebeartogoodhome:
Yeah, it looks like your `if (timeIncrement.isInfinite)` will close #2914 as well.

ijm8710:
> Yeah, it looks like your `if (timeIncrement.isInfinite)` will close #2914 as well.

Hey, curious, you had mentioned earlier:

> It grew into something bigger as I started looking at the ways offline_computer_game and over_the_board handle different things.

Was there anything still outstanding from that that this pr may not have hit on?

freebeartogoodhome:
> Was there anything still outstanding from that that this pr may not have hit on?

Most of that is stuff involving a game's end, loading an unfinished game, and starting a new game. There can be some odd behavior there.

The one thing that might be worth mentioning here is that I was really glad to see @MaartenD added OTB time prefs. Now we can do a better job starting new OTB games with the user's time control prefs. On first glance I think he could do that a bit more if he wants to look into that. I see `TimeIncrement.blitzDefault()` (the 5+3 time control) in the same places, but I could be missing something. I haven't tested it out yet and need to get to bed.

If the user changes to unlimited by changing the time and inc both to 0 we probably want to just default back to 5+3 the next time they switch to Real Time, but I think we are always switching to 5+3 when they swap that around. Again, I could be wrong.

Regardless, it looks good to me and it looks like he would be closing both of those at once. Well done.

MaartenD:
> I see TimeIncrement.blitzDefault() (the 5+3 time control) in the same places, but I could be missing something.

When the dialog appears and you choose unlimited and than switch back to realtime, the default is 5+3. The naming was already in the code.

ijm8710:
Hmm @MaartenD @freebeartogoodhome the only alternative I'd raise you two is check out challenge the first bot maia1. When you switch back from correspondence back to clock, it doesn't default back to a system wide time pref. It appears to default back to however you last had it. Meaning I think it retains your last time pref for each. (Most bots work this way except the Leela odds bots don't exhibit this behavior but that's another ticket I have)

Am I right about how it works there and do you think that makes more sense?

freebeartogoodhome:
It's worth considering @ijm8710. Maybe it's best to discuss it in another spot though since this PR can be accepted without that change. It fixes the UI issues that it meant to fix.

@MaartenD sorry if any of that sounded critical. It's just something optional we can consider. Good work on the PR.

MaartenD:
@freebeartogoodhome no problem for me. I really like to work on Lichess! If I can help to make it beter, I will help. 

I agree, it's a good idea to create another feature/fix request and discuss it there.

lib/src/model/settings/over_the_board_preferences.dart
```diff
@@ -65,12 +70,14 @@ sealed class OverTheBoardPrefs with _$OverTheBoardPrefs implements Serializable
...
     required bool symmetricPieces,
     @Default(TimeControlType.realTime) TimeControlType timeControlType,
+    @Default(TimeIncrement(300, 3)) TimeIncrement timeIncrement,
   }) = _OverTheBoardPrefs;
 
   static const defaults = OverTheBoardPrefs(
     flipPiecesAfterMove: false,
     symmetricPieces: false,
     timeControlType: TimeControlType.realTime,
+    timeIncrement: TimeIncrement(500, 3),
```
veloce: Here it is 500, and above it is 300. Which one is true? Better to extract that in a constant.

veloce (commented):
It's good except that inconsistent value.

lib/src/model/settings/over_the_board_preferences.dart
MaartenD: @veloce i changed it. The class TimeIncrement already had i constant blitzDefault. That's the one i used now.

veloce:
@copilot resolve the merge conflicts in this pull request

veloce:
Merged in 3b760efcd0fbc484e3c053a004b077093865a7a1

PR #2983, opened 2026-04-14
Fix puzzle widget theme sync

r3econ:
Fixes two bugs related to the puzzle widget (one of them probably also affecting kid mode in blog feed widgets).

#### Bug 1

We have recently changed the entitlements files from `group.org.lichess.mobileV2` to `group.org.lichess.mobileV2.LichessWidgets`, but both the Dart (`_kIosAppGroupId` in `lib/src/app.dart`) and Swift (`LichessAppGroup.id` in `ios/LichessWidgets/LichessAppGroup.swift`) code still referenced the old ID. So the app wrote into one suite and the widget read from another (which fell back to defaults)

Fix: Simple app group name fix

#### Bug 2

An updated would not have `boardTheme`/`pieceSet` written, and the widget would use its Swift-side fallback defaults (`defaultThemeName`/`defaultPieceSet`) instead of the app's actual preferences

Fix: Sync at app start

### Demo

https://github.com/user-attachments/assets/7428f70a-f653-4abe-a0c8-408be220f584

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -361,7 +361,7 @@ The app includes a native iOS WidgetKit extension (`ios/LichessWidgets/`) provid
 ### Architecture
 
 - **`LichessWidgetsBundle.swift`** — `@main` entry point registering all 4 widgets.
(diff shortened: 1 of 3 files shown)


veloce (approved):
Thanks!

PR #2984, opened 2026-04-14
Link to daily puzzle screen from puzzle widget 

r3econ:
- Tapping the daily puzzle widget used to open a generic _Chess tactics trainer_ screen via a universal HTTPS link. 
- This PR fixes it to open the correct _Daily Puzzle_ screen by switching to a custom deeplink `org.lichess.mobile://training/daily[/{id}])` and handling it in `AppLinksService`.

### Reasoning 

The widget includes the specific puzzle id in the link when available — important because the widget caches the daily puzzle for up to 6 hours, so the tapped puzzle can differ from what `/api/puzzle/daily` returns at that moment. 

The handler loads the puzzle by id and marks it as the daily puzzle so the title renders correctly. If the id fetch fails (stale/offline), it falls back to fetching the current daily puzzle rather than silently doing nothing.

### Demo

https://github.com/user-attachments/assets/d8349de8-f214-44c5-bf41-7e27dd59d415

P. S. Also adjusts one line in layout that got lost in https://github.com/lichess-org/mobile/pull/2982

### DIFF
--- a/ios/LichessWidgets/Daily Puzzle Widget/Views/DailyPuzzleWidgetView.swift
+++ b/ios/LichessWidgets/Daily Puzzle Widget/Views/DailyPuzzleWidgetView.swift
@@ -47,8 +47,7 @@ struct DailyPuzzleWidgetView: View {
                 )
         }
         .padding(.horizontal, DailyPuzzleWidgetLayout.horizontalPadding)
-        .padding(.top, DailyPuzzleWidgetLayout.topPadding)
-        .padding(.bottom, DailyPuzzleWidgetLayout.bottomPadding)
+        .padding(.vertical, DailyPuzzleWidgetLayout.verticalPadding)
     }
 
     @ViewBuilder
--- a/lib/src/app_links_service.dart
+++ b/lib/src/app_links_service.dart
@@ -68,13 +68,15 @@ class AppLinksService {
           _handleOpenWebLink(uri);
           return;
         }
+        final context = ref.read(currentNavigatorKeyProvider).currentContext;
         if (uri.scheme == kLichessUriScheme &&
             uri.host == _kDailyPuzzleDeeplinkHost &&
             uri.pathSegments.firstOrNull == _kDailyPuzzleDeeplinkPath) {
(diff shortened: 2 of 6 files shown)


ios/LichessWidgets/Daily Puzzle Widget/Views/DailyPuzzleWidgetView.swift
```diff
@@ -47,8 +47,7 @@ struct DailyPuzzleWidgetView: View {
                 )
         }
         .padding(.horizontal, DailyPuzzleWidgetLayout.horizontalPadding)
-        .padding(.top, DailyPuzzleWidgetLayout.topPadding)
-        .padding(.bottom, DailyPuzzleWidgetLayout.bottomPadding)
+        .padding(.vertical, DailyPuzzleWidgetLayout.verticalPadding)
```
r3econ, line 50: This change got lost in https://github.com/lichess-org/mobile/pull/2984

lib/src/app_links_service.dart
```diff
@@ -150,6 +164,43 @@ class AppLinksService {
...
+  /// home-screen widget.
+  ///
+  /// When [puzzleId] is provided, that exact puzzle is loaded and marked as the
+  /// daily puzzle — the widget caches for up to 6 hours so the tapped puzzle
+  /// may differ from the current `/api/puzzle/daily` result. Without an id,
+  /// falls back to the current daily puzzle.
+  Future<void> _handleDailyPuzzleLink(String? puzzleId) async {
+    try {
+      Puzzle? puzzle;
+      if (puzzleId != null) {
```
veloce: I think we can avoid making that extra request if we change the order of the requests:

1. fetch daily puzzle provider (we have to and it is cached so, when the screen is accessed no more request is made
2. compare the current daily with the puzzle id
3. if it differs, then fetch the puzzle ID
    Also if it differs, don't flag the puzzle as daily so the user don't get confused if they swipe back to puzzle tab and see another daily puzzle

I think the vast majority of the time 3 won't be executed.

PR #2985, opened 2026-04-14
Feature/quick-rematch-with-friends

### LINKED ISSUE #2851: Quick rematch with friends
Current UX requires (at least) 4 clicks through multiple screens to challenge a friend to a match.

For users who frequently play together, this is really inefficient UX.

Suggestion: make each entry in "Recent games" on home page have a swipeable "rematch" button 

Example in another app where swiping reveals the Delete button:

Image

MaartenD:
I created a quick rematch option on the recent games. 
See attached video

https://github.com/user-attachments/assets/163a55ef-f179-44ea-853c-f5f599a1fa63

Fixes: #2851

### DIFF
--- a/lib/src/view/user/game_history_screen.dart
+++ b/lib/src/view/user/game_history_screen.dart
@@ -5,7 +5,11 @@ import 'package:flutter_slidable/flutter_slidable.dart';
 import 'package:lichess_mobile/src/model/account/account_service.dart';
 import 'package:lichess_mobile/src/model/analysis/analysis_controller.dart';
 import 'package:lichess_mobile/src/model/auth/auth_controller.dart';
+import 'package:lichess_mobile/src/model/challenge/challenge.dart';
+import 'package:lichess_mobile/src/model/common/game.dart';
 import 'package:lichess_mobile/src/model/common/perf.dart';
+import 'package:lichess_mobile/src/model/game/game.dart';
+
 import 'package:lichess_mobile/src/model/game/game_filter.dart';
 import 'package:lichess_mobile/src/model/game/game_history.dart';
 import 'package:lichess_mobile/src/model/user/game_history_preferences.dart';
@@ -18,6 +22,8 @@ import 'package:lichess_mobile/src/utils/string.dart';
 import 'package:lichess_mobile/src/view/analysis/analysis_screen.dart';
 import 'package:lichess_mobile/src/view/game/game_list_detail_tile.dart';
 import 'package:lichess_mobile/src/view/game/game_list_tile.dart';
+import 'package:lichess_mobile/src/view/game/game_screen.dart';
+import 'package:lichess_mobile/src/view/game/game_screen_providers.dart';
 import 'package:lichess_mobile/src/widgets/adaptive_bottom_sheet.dart';
 import 'package:lichess_mobile/src/widgets/buttons.dart';
 import 'package:lichess_mobile/src/widgets/feedback.dart';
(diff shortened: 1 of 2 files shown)


MaartenD:
Maybe i was a little to eager yesterday with showing you my result. I watched the video back and there were two questions that come to mind:

1. what color should the dismissable (the swiping from right to left) have?  Now it's ``` color: context.lichessColors.purple ```
2. i think we should show a picture and text. "
   -    as text i suggest Rematch
   -    as a picture a suggest the Sync icon (see attachement)

@veloce what's your opinion? Are there guidelines which i can read and follow accourding to this kind of questions? Sorry for all those questions, still a newbie ;-)


Schermafbeelding 2026-04-15 070919

MaartenD:
Hi,

Here a newer version with correct colors, text and icon.

https://github.com/user-attachments/assets/2af98bc3-37e7-461f-8f26-043dfbf0b693

Enjoy!

veloce:
We can't do it as is @MaartenD 

From what I understand, you added a "Rematch" button to every game tile. But what does the "Rematch" do is to send a challenge to that user.

I see a couple of issues here:
- the game was not necessarily played with a friend
- if you challenge someone that is not your friend and does not accept challenge, it won't work
- you used the `Dismissible` widget but this would conflict with the existing `Slidable` package in the game history screen: https://github.com/lichess-org/mobile/blob/6d73bc511772bce7138e6fdd748510abab2df6db/lib/src/view/user/game_history_screen.dart#L265
- you used random side choice, but a rematch is not supposed to be with the opposite side?

For this feature to work, it needs more logic:
- if the game source was from lobby, it not show rematch; it probably makes sense to show the "New opponent" button then
- if the game source was user challenge, then show a rematch button that creates a challenge with opposite side

Some caveats:
- this will add a new button to existing ones; we probably need shorter labels, it means we'll need to introduce mobile specific labels

I'm curious to hear from other lichess team members @tom-anders and @HaonRekcef about this idea of having a "rematch" / "new opponent" slidable action. Do you think it is a good idea?

tom-anders:
I like the idea in general, but yeah I think the changes you proposed are definitely needed.

HaonRekcef:
Yes the changes suggested are needed, 
I think this feature is cool for powerusers even if the big majority of the players will never notice or it.

J0J0XMR:
The first video UX is great! Very fast and simply - full swipe instantly requests a game. 

@veloce Maybe the solution is to mark games with friends with an icon next the name? And have only those games be swipeable? So that: if game is with contact, enable rematch swipe.

Example UI:

image

ijm8710:
I still stand by my idea from earlier that I championed in the issue ticket:

Rather than have a slidable that will only be able to be inconsistently supported...

Instead, Add it to the long touch menu therefore it would be more natural to Sometimes appear. Also then it would cause no conflicts with the archive game screen.

J0J0XMR:
> I still stand by my idea from earlier that I championed in the issue ticket:
> 
> Rather than have a slidable that will only be able to be inconsistently supported...
> 
> Instead, Add it to the long touch menu therefore it would be more natural to Sometimes appear. Also then it would cause no conflicts with the archive game screen.

If its gonna be conditional, might as well make it easy. The entire point is a quick rematch, not a tap-and-hold-and-press rematch.

Adding the friend icon in list also helps identify friendly games.

MaartenD:
> We can't do it as is @MaartenD
> 
> From what I understand, you added a "Rematch" button to every game tile. But what does the "Rematch" do is to send a challenge to that user.
> 
> I see a couple of issues here:
> 
> * the game was not necessarily played with a friend
> * if you challenge someone that is not your friend and does not accept challenge, it won't work
> * you used the `Dismissible` widget but this would conflict with the existing `Slidable` package in the game history screen: https://github.com/lichess-org/mobile/blob/6d73bc511772bce7138e6fdd748510abab2df6db/lib/src/view/user/game_history_screen.dart#L265
> * you used random side choice, but a rematch is not supposed to be with the opposite side?
> 
> For this feature to work, it needs more logic:
> 
> * if the game source was from lobby, it not show rematch; it probably makes sense to show the "New opponent" button then
> * if the game source was user challenge, then show a rematch button that creates a challenge with opposite side
> 
> Some caveats:
> 
> * this will add a new button to existing ones; we probably need shorter labels, it means we'll need to introduce mobile specific labels
> 
> I'm curious to hear from other lichess team members @tom-anders and @HaonRekcef about this idea of having a "rematch" / "new opponent" slidable action. Do you think it is a good idea?

@veloce,

See attached video's
First is slidable options on friends game, second on game against non-friend.

https://github.com/user-attachments/assets/f26850ae-f7a8-4fed-80f2-19259a92d65e


https://github.com/user-attachments/assets/1be0fb76-a3b6-4633-9455-f6f8c7ac6bfc

lib/src/view/user/game_history_screen.dart
```diff
@@ -297,6 +303,49 @@ class _BodyState extends ConsumerState<_Body> {
                         ? ActionPane(
                             motion: const StretchMotion(),
                             children: [
+                              if (game.source == GameSource.friend)
+                                Builder(
+                                  builder: (context) {
+                                    final opponent = pov == Side.white ? game.black : game.white;
+                                    final sideChoice = pov == Side.white
```
veloce, line 310: You can use `pov.opposite` too.

veloce (approved):

veloce:
We'll do it like that. As said it will be a power user feature.

Issue #2447, opened 2025-12-18
[android] Feature Request: Clock/board placement


shorome2:
Please allow the board to be placed on the right side of the screen in landscape mode (as in the the previous app) and the clock near side on the left so it can be visible. As a right handed tablet user, with a pen in your hand it is impossible to see your own clock because your hand naturally covers both places where your own clock is able to be displayed while operating the pen on the chessboard to move. This is a huge problem.

If the board is unable to switch sides of the screen, is it possible to place the clocks in the upper right quadrant of the screen where your hand cannot obstruct it while you're using the pen? Or the top middle of the screen? Thanks.

Issue #2328, opened 2025-10-31
Ability to hide move list and maximize the board on foldable devices


sungsk:
Device: Samsung Galaxy Z Fold 7

Any page with a board displays a large move list taking up ⅓ of the right side of the screen.

![Image](https://github.com/user-attachments/assets/540ca714-eb25-40a9-81b3-180b48b197e0)

Turning it to landscape makes the board slightly bigger. It would be great if there was a setting to hide or move to a "ticker"-like banner at the top instead so that the full board is maximized.

veloce:
You can already hide the move list in the preferences.

Issue #1604, opened 2025-04-04
Android tablet configuration


NST-d:
On android tablets, there are multiple issues that seem to me like they can be fixed by configuration changes.

- Split screen support (it seems like all multi window options are disabled)
- Empty space for landscape orientation (currently there up is empty space left and right, see image)

![Image](https://github.com/user-attachments/assets/542d7a35-e72e-449a-aac7-5a4b44358fee)

HaonRekcef:
Split screen support for Android tablets was added in a recent update. @NST-d can you check if you still have the same issue?

NST-d:
While wide screen works now without a problem, split screen seems to not load correctly. Currently only the menu bar is visible. The options are clickable and stay highlighted, but there is no content shown, see image. 

![Image](https://github.com/user-attachments/assets/f1b31b7e-6313-4acf-9522-bb965d808ece)

HaonRekcef:
Strange,what device and android version do you have?

NST-d:
It is a Xiaomi Redmi Pad pro with android 15.

NST-d:
The issue seems to be fixed, but I can't tell if this was caused by an app update, os update or _something else_. Do you want me to close this issue or do you first want to find the original cause of this behavior?

Issue #1202, opened 2024-12-01
ability to align board on the right on tablets


leocb:
When playing on a tablet in landscape orientation, the board is aligned on the left of the screen, making it awkward to use if you're right handed, I believe an option to select where the board aligns (left or right, maybe even middle?) would be beneficial.

PR #2993, opened 2026-04-15
fix: use portrait layout on near-square foldable screens

### LINKED ISSUE #2992: Board too small on foldable/near-square screens - tablet layout shrinks board instead of maximizing it
## Device
Samsung Galaxy Z Fold 7, inner display (~756x904dp, aspect ratio ~1.19:1)

## Problem
On near-square foldable screens, the app classifies the display as a "tablet" (`shortestSide > 600dp`) and uses a side-by-side layout with the move list next to the board. This **dramatically shrinks the board**, wasting the screen's largest advantage: its width.

By comparison, Chess.com on the same device fills the full screen width with the board and stacks the move list as a compact horizontal strip above it. The result is a significantly larger, more usable board.

## Root Cause
In `lib/src/widgets/game_layout.dart`, the layout decision is based on `shortestSide > 600dp` for tablet detection. The Z Fold 7 inner screen has `shortestSide = 756dp`, so it triggers tablet layout. But unlike a traditional tablet (aspect ratio ~1.33), this screen is nearly square (1.19:1). The side-by-side tablet layout is wrong for this shape -- it splits already-limited horizontal space between board and move list, making both cramped.

## Expected Behavior
On near-square portrait screens, the app should use a phone-style layout:
- Board fills the full screen width
- Clocks, player info, and move list stack vertically above/below the board
- This maximizes board size while keeping all UI elements accessible

## Previous Issue
Issue #2328 reported this same problem and was closed with "you can already hide the move list in preferences." Hiding the move list is a workaround, not a solution. The layout strategy itself needs to account for near-square aspect ratios. The board should be large *and* the move list should be visible.

## Attempted Fix
I tried a fix in #2993 (now closed) that forced portrait layout on near-square constraints (aspect ratio < 1.1). It correctly enlarges the board on the Z Fold inner display, but introduces a ~24px bottom overflow on the puzzle and game screens because the portrait layout sizes the board to the screen's shortest side -- on near-square constraints that leaves no vertical room for the chrome below the board (feedback/status rows, bottom bar).

A full fix likely needs two parts:
1. Use portrait layout on near-square constraints (not just landscape when `maxWidth > maxHeight`).
2. On near-square portrait, size the board by remaining vertical space, not `shortestSide`.

Key files:
- `lib/src/utils/screen.dart`
- `lib/src/widgets/game_layout.dart`
- `lib/src/view/puzzle/puzzle_screen.dart`
- `lib/src/view/analysis/analysis_layout.dart`

## Related Issues
- #1604 (Android tablet configuration)
- #1202 (board alignment on tablets)
- #2447 (clock/board placement)

CZapin2015:
## Summary

On foldable devices like the Samsung Galaxy Z Fold 7, the inner display has a near-square aspect ratio (~1.19:1). After system UI chrome (status bar, toolbar, navigation) is subtracted, the available layout area can become slightly wider than tall, triggering landscape mode. This places the board and move list side-by-side in a Row, dramatically **shrinking the board** -- the opposite of what these large screens should deliver.

This PR adds a `isNearSquareConstraints()` helper that checks whether the layout constraints have an aspect ratio below 1.1:1. When true, portrait layout is used regardless of which dimension is larger, giving the board full screen width with UI stacked vertically.

## Why 1.1 threshold?

The Z Fold inner display after system chrome subtraction has a constraints ratio of ~1.03. The threshold needs to catch this while leaving legitimate landscape devices alone.

| Device | Constraints Ratio | Affected? |
|--------|------------------|-----------|
| Z Fold inner (after chrome) | ~1.03 | Yes -- stays portrait now |
| Folded Motorola (564.7x482.6) | 1.17 | No -- stays landscape |
| Pixel Fold landscape (841.1x701.0) | 1.20 | No -- stays landscape |

## Related Issues

Fixes #2992. See also #2328, #1604, #1202, #2447


veloce:
Hi! Thanks for contributing.

I'd like to see screenshots please. Of the folded devices, and of regular devices to show they are not impacted by the change.

CZapin2015:
Hi @veloce. after testing on my Z Fold 7, I hit a snag. The orientation switch does fix the small board on the inner display, but it introduced a ~24px bottom overflow on the puzzle and game screens because the portrait board ends up too tall for the chrome underneath. I tried a few follow-up commits to cap the board size, but they broke widget tests and I couldn't get them passing cleanly. Rather than leave this in a half-working state, I'd rather close it out. I'll leave #2992 open in case someone wants to take another pass. the fix probably needs to size the portrait board by remaining vertical space instead of shortestSide when constraints are near-square. 

image

PR #2996, opened 2026-04-16
only display * for ongoing games

### LINKED ISSUE #2939: 'distinguish ongoing vs future games in broadcast players' list' bug fix: Applying the same patch to the app that thibault changed on server
https://github.com/lichess-org/lila/issues/20138

HaonRekcef:
fix #2939

Couldn't test it because there is no tournament currently that provides pairings in advance. But * is still displayed for ongoing games, so I am close to certain it will work.


veloce:
Merged since it does not look dangerous even if not tested ;)

PR #2998, opened 2026-04-16
Share broadcast player

### LINKED ISSUE #2926: Ability to grab link from broadcast player results page
#2895 supports deeplinking for screen in the attached photo.

Since we can Deeplink to it, we should also be able to support grabbing the link in the app like we usually can elsewhere

Prob simply just add share icon in top right 

image

HaonRekcef:
fix #2926 
Adds a share button to the broadcast players. Logic is the same as on the website, if FIDE ID exists use it for the fragment otherwise use the name. Tested and works for both: 

[Screen_recording_20260416_202857.webm](https://github.com/user-attachments/assets/fdf311ba-19c0-4c58-a5ae-2c70f3b004dc)


veloce (approved):
lgtm

PR #3005, opened 2026-04-17
fix/correspondence-clock-reset-issue

### LINKED ISSUE #1955: Clocks don't correctly reset during correspondence games
### What is the problem?

I am playing a correspondence game with 1 day per turn. I have the game open and do not leave the game for the following turns.

Say I am white, I have just played, and I am waiting for black to play. Black's clock reaches 23:55 and they play their turn. The expected behavior is that black's clock should reset to `One day`, however it will sometimes keep decrementing. I then play my turn immediately. Black's clock now reads 23:54, when it should read 23:59. If I leave the game and come back, the clocks are correct.

This appears to occur intermittently, and affects both black's and white's clocks and any given time.

I only really play correspondence, so I'm not sure how this bug manifests in other game types.

1. Create a correspondence game with 1 day per turn
2. Start as white and play your turn.
3. Black's clock should now read 23:59. White's clock should read `One day`
4. Keep the game open
5. Allow some time (a couple minutes) to elapse before black plays their turn
6. Repeat step 5 for each player until clocks are incorrect.

### App version

0.16.4

### Device Info

iPhone 15 Pro / iOS 18.5

### Additional information

_No response_

MaartenD:
The root course of this problem was the method **didUpdateWidget**
```dart 
void didUpdateWidget(CorrespondenceClock oldClock) {
    if (widget.duration != oldClock.duration || (widget.active != oldClock.active)) {
      timeLeft = widget.duration;
    }
    if (widget.active) {
      startClock();
    } else {
      stopClock();
    }
  }
```
The if statement was

```dart
if (widget.duration != oldClock.duration)
```
I played a 1 day correspondece game. The variable timeLeft is only reset when widget.duration changes. But after every move, the server always sends 24:00:00 — so widget.duration stays the same. Meanwhile, the internal timeLeft has already counted down during your turn.

See video how it works now.

https://github.com/user-attachments/assets/ed45eabf-1818-44e3-a6be-986b01fa5fef

Fixes #1955

### DIFF
--- a/lib/src/view/game/correspondence_clock_widget.dart
+++ b/lib/src/view/game/correspondence_clock_widget.dart
@@ -65,7 +65,7 @@ class _CorrespondenceClockState extends State<CorrespondenceClock> {
   @override
   void didUpdateWidget(CorrespondenceClock oldClock) {
     super.didUpdateWidget(oldClock);
-    if (widget.duration != oldClock.duration) {
+    if (widget.duration != oldClock.duration || (widget.active != oldClock.active)) {
       timeLeft = widget.duration;
     }
     if (widget.active) {
--- a/test/view/game/correspondence_clock_widget_test.dart
+++ b/test/view/game/correspondence_clock_widget_test.dart
@@ -1,0 +1,73 @@
+import 'package:flutter/material.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:lichess_mobile/l10n/l10n.dart';
+import 'package:lichess_mobile/src/view/game/correspondence_clock_widget.dart';
+
+Widget _buildClock({required Duration duration, required bool active, VoidCallback? onFlag}) {
+  return MaterialApp(
+    localizationsDelegates: AppLocalizations.localizationsDelegates,
+    home: Scaffold(
+      body: CorrespondenceClock(duration: duration, active: active, onFlag: onFlag),
+    ),
+  );
+}
+
+void main() {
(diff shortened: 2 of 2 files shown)


veloce:
Thanks for the fix and the explanation @MaartenD .

I am not really sure to understand everything here, because when I did this I am pretty it was working. Anyway, I think your fix makes sense indeed.

This code should be tested. This PR is a good opportunity to add tests to make sure we avoid regressions. The widget logic should be tested in a `correspondence_clock_widget_test.dart`.

MaartenD:
**fix/correspondece-clock-reset-issue**

Add widget tests for CorrespondenceClock

Added test/view/game/correspondence_clock_widget_test.dart with 5 tests covering the key behaviors of the CorrespondenceClock widget:

- displays the initial time correctly: verifies the clock renders the correct initial duration.
- does not tick when inactive: verifies the clock does not count down when active is false.
- resets timeLeft when active changes from true to false: verifies that when the clock becomes inactive and a new duration is - provided by the server, timeLeft is reset to the server value instead of keeping the locally ticked value. This is the core behavior tested by the fix.
- resets timeLeft when active changes from false to true: same as above but for the opposite direction.
- resets timeLeft when duration changes without active changing: verifies that a changed duration alone also correctly resets timeLeft.

@veloce what about this set?

veloce:
Thanks for this @MaartenD ; I've added more tests and an even more robust clock update detection (even if yours was fine).

PR #3006, opened 2026-04-17
Fix/game not found opening explorer

### LINKED ISSUE #1008: [Opening explorer] Game not found
### What is the problem?

After the last update, I tried to open some games in rare variants. And sometimes, I manage to catch the error "game not found". I tried to reproduce the error in version 0.10.1 and noticed that the game that was not found in version 0.10.3 simply does not open in version 0.10.1 on iPhone, as in the error https://github.com/lichess-org/mobile/issues/984.

The game in which the error occurs (1. e4 c6 2. a3): https://lichess.org/cLAh6fkU

Another example in which the error occurs (1. d4 d5 2. Bf4 Bf5 3. Nc3 a6): https://lichess.org/8UyzPgSE

If you close the app and open the game again, entering the moves from the example, the game may open. For me, this happened after the 3rd restart.

https://github.com/user-attachments/assets/3a50393c-347e-4a12-bbd8-5589347eac47

https://github.com/user-attachments/assets/6cb591d9-bd9b-455d-b7ff-1b6c94a65c7b

1. Open new Lichess Beta app
2. Open Opening Explorer
3. Enter the moves as in the example or try other rare moves. Try opening games until you encounter the error.

### App version

0.10.3

### Device Info

Pixel 8 / Android 14

### Additional information

_No response_

MaartenD:
I am not able to test this myself.

In the onTap event i added this:
```dart
          final client = ref.read(defaultClientProvider);
          await client.get(lichessUri('/import/master/${widget.game.id}/${widget.pov.name}'));
          if (!context.mounted) return;
```
Fixes #1008

### DIFF
--- a/lib/src/view/game/correspondence_clock_widget.dart
+++ b/lib/src/view/game/correspondence_clock_widget.dart
@@ -65,7 +65,7 @@ class _CorrespondenceClockState extends State<CorrespondenceClock> {
   @override
   void didUpdateWidget(CorrespondenceClock oldClock) {
     super.didUpdateWidget(oldClock);
-    if (widget.duration != oldClock.duration) {
+    if (widget.duration != oldClock.duration || (widget.active != oldClock.active)) {
       timeLeft = widget.duration;
     }
     if (widget.active) {
--- a/lib/src/view/game/correspondence_clock_widget.dart
+++ b/lib/src/view/game/correspondence_clock_widget.dart
@@ -65,7 +65,7 @@ class _CorrespondenceClockState extends State<CorrespondenceClock> {
   @override
   void didUpdateWidget(CorrespondenceClock oldClock) {
     super.didUpdateWidget(oldClock);
-    if (widget.duration != oldClock.duration || (widget.active != oldClock.active)) {
+    if (widget.duration != oldClock.duration) {
       timeLeft = widget.duration;
     }
     if (widget.active) {
--- a/lib/src/view/explorer/opening_explorer_widgets.dart
+++ b/lib/src/view/explorer/opening_explorer_widgets.dart
@@ -6,6 +6,7 @@ import 'package:intl/intl.dart';
 import 'package:lichess_mobile/src/model/analysis/analysis_controller.dart';
 import 'package:lichess_mobile/src/model/common/chess.dart';
 import 'package:lichess_mobile/src/model/explorer/opening_explorer.dart';
+import 'package:lichess_mobile/src/network/http.dart';
 import 'package:lichess_mobile/src/theme.dart';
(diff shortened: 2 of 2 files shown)


lib/src/view/game/correspondence_clock_widget.dart
```diff
@@ -65,7 +65,7 @@ class _CorrespondenceClockState extends State<CorrespondenceClock> {
   @override
   void didUpdateWidget(CorrespondenceClock oldClock) {
     super.didUpdateWidget(oldClock);
-    if (widget.duration != oldClock.duration) {
+    if (widget.duration != oldClock.duration || (widget.active != oldClock.active)) {
```
HaonRekcef: this seems to be an unrelated code change, should be removed

HaonRekcef (changes requested):

lib/src/view/game/correspondence_clock_widget.dart
MaartenD: Hi @HaonRekcef,

I don't agree with your suggestion because when a correspondence game is open and the opponent's clock becomes inactive (i.e. it becomes your turn), the clock was not resetting timeLeft to the value provided by the server, causing the opponent's clock to display an incorrect time (the bug). The fix ensures that timeLeft is also updated when active changes, not just when duration changes.

HaonRekcef: The change makes sense, but you have a seperate pull reqeust for it.

MaartenD: I'm sorry, i see now that this is about the game not found in the opening explorer. I will correct this and do another pull request for this issue.

Again, sorry.

HaonRekcef: No worries :)

You can also keep this one and simply remove changes in the file

MaartenD: I didn't know that, still a newbie in git-land. I hope i did it correctly, i deleted the file.

HaonRekcef: I should have been more clear. I meant you can undo the changes of the file and then push it again to github.

MaartenD: I will push it again.

MaartenD: I pushed it, but no de test fails, i don't think it's an issue on the code side but the GitHub side. Will this process restart automatically?

HaonRekcef: Yes, seems unrelated. No the pipeline needs to be manually re-triggered, but I do not have the permissions.

veloce (approved):

Issue #2963, opened 2026-04-12
Upcoming/live broadcasts widget


ijm8710:
Loving some of the work going on by @cappylot and @r3econ with widgets for daily puzzle, quick play and community blogs.

I think one other widget idea that would be cool is a quick snapshot of the current broadcasts airing and to be aired

May Even be helpful to have a filter setting to toggle between all broadcasts and "top broadcasts" only

r3econ:
@ijm8710 could you give me your feedback on the design please? 😎  It's far from ready atm, but crucial data is there. Of course an image thumbnail would be nice there. Anything else?

<img width="1000" height="1870" alt="Image" src="https://github.com/user-attachments/assets/93429c0d-75b7-4995-b2fc-8b5a665da0f1" />

ijm8710:
Agree with thumbnail being the best to add but otherwise i think it looks great.

Few other ideas but minor:

- Could show relative time if under a day/include the start time if over a day such as the photos below
- maybe show top players playing but I'm not sure that's necessary for this
- maybe show the red person count (it's available on web) when live as a frame of reference to the scale of the event but again not mandatory 

![image](https://github.com/user-attachments/assets/fd954471-d3f5-49c2-90e8-bcddfa848ed1)
![image](https://github.com/user-attachments/assets/d05cc4d6-a7c9-4f81-a3dd-cd896118f178)

veloce:
Broadcast widgets are going to be so cool ❤️

PR #3007, opened 2026-04-17
Add iOS broadcasts widget

r3econ:
Adds a new iOS home-screen widget that displays ongoing and upcoming chess broadcasts from lichess.org, available in Small, Medium, and Large sizes with a configurable filter (All / Top broadcasts).

- Data sourced from `GET /api/broadcast/top`
- Configurable filter via AppIntents (All / Top), where "Top" mirrors lila's tier >= HIGH threshold
- Deep linking — rows tap into the existing `applinks:lichess.org` universal link flow routing to `BroadcastRoundScreen`; no Dart changes needed
- Architecture mirrors the Blog Feed widget one-for-one

#### iPhone

iphone

#### iPad

ipad

#### Catalog

catalog

#### Errors

errors


ijm8710:
Super cool! Did you agree on any of the ideas from before https://github.com/lichess-org/mobile/issues/2963#issuecomment-4232794730

The first bullet is one id support the most and I think there's room

I think the app handles relative times better 

I also would get rid of the tiniest widget altogether. There's no point with the truncation.

But overall looks great!

r3econ:
@ijm8710, the first bullet point from the [ticket](https://github.com/lichess-org/mobile/issues/2963#issuecomment-4232794730) was:

> Could show relative time if under a day/include the start time if over a day such as the photos below

I agree it would be cool but there are some technical reasons why the relative time can't be accurate. Widgets on iOS might not be updated frequently - the OS gives the widget a "refresh budget":

> Reloading widgets consumes system resources and causes battery drain due to additional networking and processing. To reduce this performance impact and maintain all-day battery life, limit the frequency and number of updates you request to what’s necessary.

More about it [here](https://developer.apple.com/documentation/widgetkit/keeping-a-widget-up-to-date). 

In my implementation I've set the refresh interval to 1h - it does not mean the OS will refresh the widget every hour, the request might be ignored based on e.g. battery state, mode (e.g. work, do not disturb, etc.). So showing a relative would almost certainly mean showing inaccurate, wrong info

---

Your other ideas are also cool. Showing player names would be very nice. However I can't think of a nice design for that - the only thing that comes to my mind is having 3 instead of 4 items in the large widget and making the first one occupy two rows (with enough space for bigger thumbnail, and more information). This means we would show one less item in the broadcasts list

ijm8710:
Fair enough on all! I'd maybe consider the time in addition to date but defer to you if you prefer just the date

veloce:
Great work as usual! Really excited to have these widgets, and sorry for the delay :)

r3econ:
It's really cool this is getting merged! 🤩

PR #3027, opened 2026-04-20
Place deleteOldEntries in a transaction

freebeartogoodhome:
I got a database lockout while testing and think it may be due to _deleteOldEntries. I wanted to try placing those in a transaction.


veloce (approved):
👍

PR #3030, opened 2026-04-21
Feature: delete block private messages

### LINKED ISSUE #2023: Add "delete" "report" and "block" for private messages
Thanks for incorporating some of feedback for the private messages, ticketing a few of the others to see if they're possible to add down the road with lesser priority:

This is just to be consistent with actions from web.
Delete especially is helpful to have

MaartenD:
Add the feature to delete message en block a user.

See video for the result.

Enjoy.

https://github.com/user-attachments/assets/efd2c089-1489-4ddf-9917-1031a8235094

Fixes: #2023


ijm8710:
Thanks! Would also be nice to have the block button in top right of within the convo (like you can see with the left device on web) as block, report and delete are all there

Id also love to be able to delete individual messages but not sure if server allows that

MaartenD:
> Thanks! Would also be nice to have the block button in top right of within the convo (like you can see with the left device on web) as block, report and delete are all there

I would love to do that, but i didn't do it directly, because the block button was already created in this user menu. @veloce what's your opinion about this?

> Id also love to be able to delete individual messages but not sure if server allows that.
This isn't possible at web, didn't take a look if it's possible. If it's possible then here the same. Is this the way veloce wants to go.

Happy to here you like it.

ijm8710:
One last request :) 


Understood it's not on web but I've always found these helpful

When it gives you the prompt before it deletes, I always find it super helpful to have the prompt where it clarifies: something like "“This message will be deleted for you only.”

lib/l10n/app_en.arb
```diff
@@ -30,6 +30,9 @@
   "mobileHideVariation": "Hide variation",
   "mobileHomeTab": "Home",
   "mobileLiveStreamers": "Live streamers",
+  "mobileMessageDeleteConversation": "Delete conversation",
+  "mobileMessageDeleteConversationTitle": "Delete conversation",
```
HaonRekcef: We should not add the same string twice, you can reuse it.

HaonRekcef (commented):
https://github.com/lichess-org/mobile/blob/main/docs/internationalisation.md#mobile-translations-tips

Normally we add the features first without the translation strings and then after some time we translate them, to avoid translation contributors needing to translate unneeded/dead translations.

lib/l10n/app_en.arb
MaartenD: Hi @HaonRekcef,

Thanks for the link. I changed it. Deleted the translations and made use of hardcoded text. Re-used the ```code mobileAreYouSure ```.

lib/src/view/message/conversation_screen.dart
```diff
@@ -85,9 +90,37 @@ class _ConversationScreenState extends ConsumerState<ConversationScreen> with Ro
           leading: UserAvatar(widget.user, radius: 16),
           title: UserFullNameWidget(user: widget.user, showFlair: false),
           subtitle: Text(widget.user.isOnline == true ? context.l10n.online : context.l10n.offline),
-          onTap: () =>
-              Navigator.push(context, UserOrProfileScreen.buildRoute(context, widget.user)),
+          onTap: () async {
+            await Navigator.push(context, UserOrProfileScreen.buildRoute(context, widget.user));
+            ref.invalidate(conversationControllerProvider(widget.user.id));
```
veloce: Can you tell me why you invalidate here?

And why after pushing the screen? Usually what we do is invalidate, then push.

lib/src/view/message/conversation_screen.dart
```diff
@@ -399,8 +432,7 @@ class _MessageInput extends ConsumerStatefulWidget {
 class _MessageInputState extends ConsumerState<_MessageInput> {
   final controller = TextEditingController();
 
-  bool get isBlocked =>
-      widget.state.convo.relations.inward == false || widget.state.convo.relations.outward == false;
```
veloce, line 403: You removed that logic. I am not sure why?

We could use tests here. At least to check which convo is blocked or not.

veloce (commented):
This PR not only adds the delete thread feature but changes the logic about blocking. Usually it is better to keep one change per PR to simplify the review process.

The new logic may be better but I want to avoid regressions. That is why I'd like to see tests showcasing the new blocking logic, covering all cases (lichess user, user that blocked, user that did not block, etc.)

lib/src/view/message/conversation_screen.dart
MaartenD: We invalidate the provider after returning from the profile screen because the user may have blocked the contact while viewing their profile. The server's /inbox endpoint does not return blocking status, so the controller calls getUser() to fetch it separately. By invalidating after the push returns, we force the controller to rebuild and call getUser() again picking up any blocking change that happened while on the profile screen.

Invalidating before the push would not help here, since the block action hasn't occurred yet at that point and the controller would just reload with the same unblocked state.

lib/src/view/message/conversation_screen.dart
MaartenD: I removed that logic because the server's /inbox/$userId endpoint returns relations: (inward: null, outward: null) even after blocking so checking inward == false or outward == false would never be true. The original check was effectively dead code. To get the actual blocking status I call getUser() instead, which correctly returns blocking: true when a user is blocked.

I agree that tests would be valuable here to document which conversations are considered blocked and why. I will add those.

MaartenD:
Hi,

Tests in test/model/message/conversation_controller_test.dart

- isBlocked is false for a normal user that is not blocked
- isBlocked is true when you have blocked the user
- isBlocked is false when blocking field is absent from user response
- isBot is true for a bot account
- postable is false when contact does not accept new messages
- postable is true when existing conversation allows further messages

If i missed something, let me know.

lib/src/view/message/conversation_screen.dart
veloce: I see. I actually got confused because this is the first time I see an `async` onTap handler doing it.

In other parts of the code we're doing:

```dart
onTap: () {
   Navigator.push(context, UserOrProfileScreen.buildRoute(context, widget.user))
     // invalidate to refresh potential blocking status change
     .then((_) => ref.invalidate(conversationControllerProvider(widget.user.id)))
}
```

I know `async/await` is most of the time preferred to `.then` but in that case I find it more intuitive (and let's be consistent with the rest).

Also here I think we miss a `if (context.mounted)` after the async gap.

MaartenD: Will make it consist with the rest of the code and also add the ``` if (context.mounted) ``` and push later today.

veloce (approved):
thanks

PR #3031, opened 2026-04-21
Feature: after game added moves in Italic

### LINKED ISSUE #3021: Show after games finish moves in italics
https://discord.com/channels/280713822073913354/1494701216830197961/1494701216830197961

On the website, moves added after the game's actual ending state is styled in italics, as show here 

Image

While in the app, you can't tell if the moves are part of the game or part of the analysis.

Image

Just a small style change for the moves would be nice.

MaartenD:
- **node.dart:** Added isUserAdded field to both Branch (mutable) and ViewBranch (Freezed). Also added isUserAdded parameter to addMoveAt.
- **analysis_controller.dart:** onUserMove now passes isUserAdded: true when adding moves.
- **pgn.dart:** InlineMove.build() applies italic when branch.isUserAdded == true (in addition to sidelines), so user-added analysis moves on the mainline appear in italic.

Fixes: #3021


MaartenD:
Hi, here a small video.

Enjoy.

https://github.com/user-attachments/assets/f5d18951-d3ec-46e0-80d2-5cb1b10524b1

HaonRekcef:
Nice, I think we should add this to broadcasts too.

MaartenD:
> Nice, I think we should add this to broadcasts too.

@HaonRekcef is there a way i can reproduce this? I don't really understand where you would like to have this in broadcasts. On the dev server i only have past events and i don't see a way to add moves myself.  Could you please give me some more context?

Thx in advance.

HaonRekcef:
When analysing a game from a broadcast, the user can make moves, these should be italic similar to what you did in the analysis board now.

MaartenD:
@HaonRekcef thx, found it, will look at it.

MaartenD:
Also added the italic moves in broadcasts section. See video.

Enjoy.

https://github.com/user-attachments/assets/98f48d59-6d1e-4fab-88ff-e0212c63745f

tom-anders:
I don't think it's necessary to make moves in sidelines italic as well, as by definition they are not part of the moves played in the game. I think only mainline moves that are appended to the game's moves should be italic, to differentiate them from the moves played in the game. At least that's also the intention of the original issue I think

MaartenD:
Clear, I will look at it later this week. Only in mainline and not in sideline moves.

MaartenD:
Made the changes after @tom-anders comment. Italic moves now only visible when entered in the mainline. This works for games in the analysis and broadcasts.

Enjoy!

https://github.com/user-attachments/assets/3d8c295f-ba3e-498f-ae9c-c24aa438c75a

veloce (commented):
I'd like to have tests for this please.

Also, I'd like to see that this applies only to analysis with a game associated. A standalone analysis should not have all its move in italic, even though they are user generated.

What about games imported from a PGN? Should we also have the user added moves in italic here? cc @HaonRekcef

HaonRekcef:
> I'd like to have tests for this please.
> 
> Also, I'd like to see that this applies only to analysis with a game associated. A standalone analysis should not have all its move in italic, even though they are user generated.
> 
> What about games imported from a PGN? Should we also have the user added moves in italic here? cc @HaonRekcef

Good point, could be useful but not a necessity.

veloce:
So let's do this only for games played on lichess @MaartenD (not from PGN import).

MaartenD:
**Feature: Show user-added analysis moves in italic for archived Lichess games**

After a Lichess game is finished, the user can continue to make moves on the analysis board. These user-added moves are now displayed in italic in the move list, matching the behavior of the Lichess web version.

**How it works**
A new isUserAdded field was added to Branch and ViewBranch in node.dart. When addMoveAt is called with isUserAdded: true, the field is only set to true if the move extends the mainline (i.e. the parent node has no existing children and the current path is on the mainline). Moves that create a sideline, or moves added within a sideline, always get isUserAdded: false.

In pgn.dart, the InlineMove widget applies italic styling when branch.isUserAdded is true.

**Scope**
The italic behavior is intentionally limited to archived Lichess games only:

- Archived Lichess games (ArchivedGame): isUserAdded: true is passed to addMoveAt, so mainline extensions **are shown in italic**.
- PGN imports and standalone analysis (Pgn, Standalone): isUserAdded defaults to false, **so moves are never shown in italic**.
- Broadcasts: isUserAdded is never set to true, **so moves are never shown in italic**.

**Tests**
Four tests were added to node_test.dart in the addMoveAt isUserAdded group, covering:

- Extending the mainline -> isUserAdded: true
- Creating a sideline mid-game -> isUserAdded: false
- No isUserAdded parameter (PGN import / broadcast) -> isUserAdded: false
- Adding a move within an existing sideline -> isUserAdded: false

veloce:
Thanks for the modifications.

I am not sure it is enough though. Branches can be moved around during an analysis: it can be promoted, promoted to mainline, deleted.

What happens in that case? I assume that moves that were tagged `isUserAdded` on mainline that are moved to a sideline will still have the tag (wrongly).

The view that renders PGN already knows if we're on the mainline or a sideline, so you don't need to check that while adding the node. The `isUserAdded` should only check if it a move added by user (and is a lichess game).

MaartenD:
I will take a look later today. I will keep you informed.

MaartenD:
**Fix: Italic styling for user-added moves after game end**
After a Lichess game is finished, moves added by the user on the analysis board are now displayed in italic in the move list, matching the behavior of the Lichess web version.

**How it works**
A new isUserAdded field was added to Branch and ViewBranch in node.dart. When addMoveAt is called with isUserAdded: true, the field is only set to true if the move extends the current mainline at a position with no existing children meaning it genuinely continues the game past its end. Moves that create a sideline, or moves added within a sideline, always get isUserAdded: false.

In pgn.dart, the InlineMove widget applies italic styling whenever branch.isUserAdded is true, regardless of whether the move is currently on the mainline or a sideline. This matches the web behavior: a user-added move remains italic even if it is later moved to a sideline via promotion of another variation, making it always clear which moves were not part of the original game.

**Scope**
The italic behavior is intentionally limited to archived Lichess games only (ArchivedGame). PGN imports, standalone analysis, and broadcasts never pass isUserAdded: true.

**Tests**
Four tests were added to node_test.dart in the addMoveAt isUserAdded group, covering:
 
- Extending the mainline → isUserAdded: true
- Creating a sideline mid-game → isUserAdded: false
- No isUserAdded parameter (PGN import / broadcast) → isUserAdded: false
- Adding a move within an existing sideline → isUserAdded: false


https://github.com/user-attachments/assets/d7925790-0999-41d7-87a0-6c4a70054d6e

I checked it on the website. In my opinion the behavior is now the same.

image

image

veloce:
Your last comment is ambiguous:

> In pgn.dart, the InlineMove widget applies italic styling whenever branch.isUserAdded is true, regardless of whether the move is currently on the mainline or a sideline. This matches the web behavior: a user-added move remains italic even if it is later moved to a sideline via promotion of another variation, making it always clear which moves were not part of the original game.

This is not what I see in the code or in the tests.

This is also contradicting the rest of the comment:

```quote
Tests
Four tests were added to node_test.dart in the addMoveAt isUserAdded group, covering:

Extending the mainline → isUserAdded: true
Creating a sideline mid-game → isUserAdded: false
No isUserAdded parameter (PGN import / broadcast) → isUserAdded: false
Adding a move within an existing sideline → isUserAdded: false
```

I believe that the boolean value of tests 2 and 3 should be `true` and not `false` if we are following the logic you mentioned above.

Moreover, I tested the web behaviour. And unless I'm missing something, the moves I have added are definitely **not** in italic. So this whole comment about web behaviour is irrelevant.

Note that the comment from @tom-anders 

> I don't think it's necessary to make moves in sidelines italic as well, as by definition they are not part of the moves played in the game. I think only mainline moves that are appended to the game's moves should be italic, to differentiate them from the moves played in the game. At least that's also the intention of the original issue I think

still makes sense to me.

If the logic is too complex for what we want to achieve here I'd rather drop this feature entirely. I don't want to add unnecessary maintenance cost for a feature for which I am not sure it has a lot of value.

MaartenD:
You are right, I apologize for the confusing summary.
The actual implementation does exactly what @tom-anders described: only moves that extend the mainline after the game ends are displayed in Italic.

Regarding the web behaviour: I want to clarify what I actually observed and what I meant. I added a move at the end of a game (extending the mainline), then created a sideline and promoted it to mainline. The move I originally added became a sideline, but it remained in italic which is also exactly what happens on the Lichess website (see screenshots). So the app behaviour does match the web behaviour in this regard.

When you want to drop the feature, that's all up to you ;-) I like to help, learn and contribute to this app.

ijm8710:
I hope this is not dropped :) it would be helpful. Totally understand if it needs time to be ironed out to get it right

veloce:
Ok I can see the moves in italic in the screenshots. I wonder why it didn't work for me when I tested, but will check again.

lib/src/model/common/node.dart
```diff
@@ -212,16 +212,19 @@ abstract class Node {
     bool prepend = false,
     bool replace = false,
     Duration? clock,
+    bool isUserAdded = false,
```
veloce, line 217: Add a section in the doc comment to explain that parameter.

lib/src/model/common/node.dart
```diff
@@ -212,16 +212,19 @@ abstract class Node {
...
-    final pos = nodeAt(path).position;
+    final parentNode = nodeAt(path);
+    final normalizedMove = normalizeMove(parentNode.position, move);
 
-    final normalizedMove = normalizeMove(pos, move);
-
-    final (newPos, newSan) = pos.makeSan(normalizedMove);
+    final (newPos, newSan) = parentNode.position.makeSan(normalizedMove);
+    final willBeMainline =
+        (prepend || replace || parentNode.children.isEmpty) && isOnMainline(path);
```
veloce, line 224: Not sure this logic is entirely tested in `node_test.dart`; I'd like to see tests for all the cases here (prepend, replace, etc.).

lib/src/model/common/node.dart
MaartenD: Added the doc comment.

lib/src/model/common/node.dart
MaartenD: added 3 tests in node_test.dart

- isUserAdded is true when prepend is true on mainline
- isUserAdded is true when replace is true on mainline
- isUserAdded is false when prepend is true but not on mainline

veloce (approved):

ijm8710:
> Made the changes after @tom-anders comment. Italic moves now only visible when entered in the mainline. This works for games in the analysis and broadcasts.
> 
> 
> 
> Enjoy!
> 
> 
> 
> https://github.com/user-attachments/assets/3d8c295f-ba3e-498f-ae9c-c24aa438c75a
> 
> 
> 
> 

Hi Maarten, where did things stand for having this with broadcasts? I thought it was a good idea to have it there as well and I think the pr did include it at one point? At least for the mainlines? Did I miss any discussion about avoiding it there specifically?

MaartenD:
> So let's do this only for games played on lichess @MaartenD (not from PGN import).

@ijm8710 this is why i removed the Italic part from the broadcasts section.

ijm8710:
> > So let's do this only for games played on lichess @MaartenD (not from PGN import).
> 
> @ijm8710 this is why i removed the Italic part from the broadcasts section.

Hey Maarten, completely understandable but tagging @HaonRekcef to confirm.

I believe Veloce had that comment in mind to avoid games brought in later such as imported and saved through the pgn import tool on web and the old app. In contrast to games played directly on lichess.

I know technically broadcasts games have pgn associated with them, but this was still completely  separate from the broadcasted ones altogether. 

HaonRekcef was in fact the first one to suggest adding this to broadcast games altogether and then Veloce tagged him for guidance on which other ones to avoid so i think he himself was deferring fully to haonrekcef on what would work best here.

@HaonRekcef Mind just confirming what I'm saying please


Edit: since haonrekcef confirmed (see thumbs up-technically I had already doubled checked with him in advance), Maarten I'm assuming you can differentiate between broadcasts and the other pgn imported games correct? And is it an easy lift to potentially bring them back just for broadcasts and did we still have the tests for those still?

MaartenD:
@ijm8710 i think this isn't to hard to achieve. I will try to look at it this week. Can someone of you be so kind and create a feature request? You may assign it to me (if that's done for Lichess Mobile).

I will keep you informed.

Issue #3011, opened 2026-04-18
Remove rating filter and side choice for unrated correspondence games


ijm8710:
For both rated and unrated correspondence lobby games, rating filter is shown but always blocked out so why show at all?

For unrated, it adds a side filter which is also greyed out so I feel that should be removed as well.

If we agree these are never able to be used, should we remove them?

PR #3032, opened 2026-04-22
Explain why the rating filter is disabled

CloudyDino:
## Summary

When a user's rating for the selected perf is provisional, the rating filter in the lobby game creator is disabled. Previously the UI just showed it greyed-out with no explanation — users were left wondering why they couldn't use it (see e.g. #3011 where the reporter asked why the filter is "always blocked out").

Now tapping the disabled rating filter opens an adaptive dialog with the existing localized string `ratingRangeIsDisabledBecauseYourRatingIsProvisional`:

> *Rating filters are locked because your rating is not stable. Playing rated games will increase stability.*

Changes are in two places, since the filter is rendered twice:

- **Real-time lobby** (`create_game_widget.dart`) — the `OutlinedButton` gets an `onPressed` that shows the dialog when unavailable; `foregroundColor` is set to the theme's disabled color to keep the greyed-out appearance.
- **Inline `PlayRatingRange`** (`common_play_widgets.dart`) used by the correspondence sheet — the disabled `ListTile` is wrapped in `IgnorePointer` + `GestureDetector(HitTestBehavior.opaque)` so the inner `NonLinearSlider` widgets don't swallow the tap in the gesture arena.

A `SnackBar` was tried first but renders behind modal bottom sheets (it anchors to the root `ScaffoldMessenger`), so a dialog route is used instead.

## Test plan

- [x] `flutter analyze` clean on both files
- [x] `dart format` clean
- [x] Manual testing

https://github.com/user-attachments/assets/2748cdad-933e-487b-83d1-9c3135767510

### DIFF
--- a/lib/src/view/play/common_play_widgets.dart
+++ b/lib/src/view/play/common_play_widgets.dart
@@ -3,6 +3,7 @@ import 'package:lichess_mobile/src/model/lobby/game_setup_preferences.dart';
 import 'package:lichess_mobile/src/model/user/user.dart';
 import 'package:lichess_mobile/src/utils/l10n_context.dart';
 import 'package:lichess_mobile/src/widgets/non_linear_slider.dart';
+import 'package:lichess_mobile/src/widgets/platform_alert_dialog.dart';
 
 class PlayRatingRange extends StatefulWidget {
   const PlayRatingRange({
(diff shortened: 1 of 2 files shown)


veloce:
Thanks for this. I've added a small improvement: an info icon next to the "rating filter" label when the button is disabled.

PR #3036, opened 2026-04-24
fix infinite loop triggered by imported game deeplinks

HaonRekcef:
Unhandled Lichess links like imported games ( f.ex. `https://lichess.org/0Q6BNlj4`), currently cause an infinite loop. When the app fails to resolve the link, it tries to open it in the system browser, which the OS immediately redirects back to the app, which will open the browser again ...

To protect against similar cases in the future, this PR adds that deeplinks can not open the in-app browser.

### DIFF
--- a/lib/src/app_links_service.dart
+++ b/lib/src/app_links_service.dart
@@ -106,7 +106,8 @@ class AppLinksService {
       return;
     }
     if (context != null && context.mounted) {
-      await handleAppLink(context, uri, animated: animated);
+      // For app deep links, we don't want to allow falling back to the browser as it might trigger an infinite loop if the app isn't properly handling the link
+      await handleAppLink(context, uri, animated: animated, allowBrowserFallback: false);
     }
   }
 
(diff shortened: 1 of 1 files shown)


lib/src/app_links_service.dart
```diff
@@ -106,7 +106,8 @@ class AppLinksService {
       return;
     }
     if (context != null && context.mounted) {
-      await handleAppLink(context, uri, animated: animated);
+      // For app deep links, we don't want to allow falling back to the browser as it might trigger an infinite loop if the app isn't properly handling the link
```
veloce, line 109: Could that also happen for other links than deep links?

veloce (approved):
Thanks for debugging this!

lib/src/app_links_service.dart
HaonRekcef: yes they could have the same problem but it is also protected against. I tested it with in app links.
Then url would get launched, converted in a deeplinkm but then the url wouldn't get launched a second time.

PR #3037, opened 2026-04-24
Ignore shared URLs in file processor

HaonRekcef:
I noticed while debugging that there was an unnecessary attempted file read when opening a deeplink url as ReceiveSharingIntent would also handle it.
`I/flutter (13335): Failed to process incoming file: PathNotFoundException: Cannot open file, path = 'https://lichess.org/iMypvkd1' (OS Error: No such file or directory, errno = 2) `

### DIFF
--- a/lib/src/app.dart
+++ b/lib/src/app.dart
@@ -206,6 +206,10 @@ class _AppState extends ConsumerState<Application> {
   Future<void> _processSharedFiles(List<SharedMediaFile> files) async {
     if (files.isEmpty) return;
     final filePath = files.first.path;
+    if (filePath.startsWith('https://') || filePath.startsWith('http://')) {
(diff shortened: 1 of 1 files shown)


lib/src/app.dart
```diff
@@ -206,6 +206,10 @@ class _AppState extends ConsumerState<Application> {
   Future<void> _processSharedFiles(List<SharedMediaFile> files) async {
     if (files.isEmpty) return;
     final filePath = files.first.path;
+    if (filePath.startsWith('https://') || filePath.startsWith('http://')) {
```
veloce: You can actually use a pattern with `startWith` and I assume `if (filePath.startsWith('http')` is enough here?

HaonRekcef: sure

veloce (approved):

PR #3046, opened 2026-04-25
Fix deep link loop for imported games with anonymous players

berenbums:
## Summary 
Imported PGNs without a `[Result]` tag and without `[White]`/`[Black]` tags fall into an infinite deep-link loop when shared into the Lichess mobile app. This PR routes those games to the analysis screen instead.

Reproducible by hitting `POST /api/import` with a PGN that lacks a result and player tags and then opening the returned `https://lichess.org/{id}`.

## Root cause 
When a PGN is imported without a `Result` tag, Lichess records the game with `status: "started"`. When `[White]`/`[Black]` tags are missing too, the players come back anonymous (`{"name": "?"}` with no `user`). 

`_tryResolveGameLink` in `app_links_service.dart` then returns `null`, because both, the AnalysisScreen (`game.finished`) and TvScreen (`game.playerOf(orientation).user`) checks fail.

`handleAppLink` falls through to `launchUrl(uri)`, which the OS routes back into the app via the verified `lichess.org/{id}` intent filter, looping indefinitely. 

## Fix
Open the game in `AnalysisScreen` whenever it's finished **or** has no real player on the orientation side. Open in `TvScreen` otherwise.

### DIFF
--- a/test/app_links_service_test.dart
+++ b/test/app_links_service_test.dart
@@ -15,6 +15,7 @@ import 'package:lichess_mobile/src/model/common/chess.dart';
 import 'package:lichess_mobile/src/model/common/game.dart';
 import 'package:lichess_mobile/src/model/common/id.dart';
 import 'package:lichess_mobile/src/model/common/speed.dart';
+import 'package:lichess_mobile/src/model/game/game.dart';
 import 'package:lichess_mobile/src/model/game/game_repository.dart';
 import 'package:lichess_mobile/src/model/game/game_status.dart';
 import 'package:lichess_mobile/src/model/game/player.dart';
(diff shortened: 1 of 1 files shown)


berenbums:
Just noticed that https://github.com/lichess-org/mobile/pull/3043 was opened and merged while I was working on my PR. You guys are fast :). This one might not be required anymore.

HaonRekcef:
@berenbums I see you added a test case that is useful so you could create a new PR with the test :)

veloce:
The PR is still useful imo, can you solve the conflict @berenbums ? Thanks!

berenbums:
Combined the changes and updated the test case. Thank you for the quick feedback!

HaonRekcef:
Hmm, I am not sure we want the proposed changes: When getting a live game link of a game between two anon users we would end in the analysis board, where there wont be any move updates. Ideally we should open it in the tvScreen.
The loop is already prevented by #3036

veloce:
I see. I didn't get that, thanks for the insight @HaonRekcef . 

So we only want the test in this PR @berenbums

berenbums:
Makes sense to me, thanks @HaonRekcef . Updated the PR so it only adds a new test case: A `https://lichess.org/{id}` link for a game with `status: "started"`, `source: "import"`, and anonymous players resolves to `AnalysisScreen`.

veloce (approved):

PR #3048, opened 2026-04-25
feat: support requesting server analysis for study chapter

tom-anders:
[studyanalysis.webm](https://github.com/user-attachments/assets/a15abc4f-adce-4533-a945-ec7ca40af1e8)

### DIFF
--- a/lib/src/model/analysis/server_analysis_service.dart
+++ b/lib/src/model/analysis/server_analysis_service.dart
@@ -16,8 +16,11 @@ import 'package:lichess_mobile/src/model/game/game_repository.dart';
 import 'package:lichess_mobile/src/model/game/game_socket_events.dart';
 import 'package:lichess_mobile/src/network/http.dart';
 import 'package:lichess_mobile/src/network/socket.dart';
+import 'package:logging/logging.dart';
 
 part 'server_analysis_service.freezed.dart';
+
+final _logger = Logger('ServerAnalysisService');
 
 @freezed
 sealed class ServerAnalysisSource with _$ServerAnalysisSource {
@@ -25,7 +28,10 @@ sealed class ServerAnalysisSource with _$ServerAnalysisSource {
 
   const factory ServerAnalysisSource.game({required GameId gameId}) = _GameServerAnalysisSource;
 
-  // TODO add study chapter source
+  const factory ServerAnalysisSource.studyChapter({
+    required StudyId studyId,
+    required StudyChapterId chapterId,
+  }) = _StudyChapterServerAnalysisSource;
 }
 
 const Duration kMaxWaitForServerAnalysis = Duration(minutes: 1);
@@ -75,6 +81,7 @@ class ServerAnalysisService {
     final uri = Uri(
       path: switch (source) {
         _GameServerAnalysisSource(:final gameId) => '/watch/$gameId/${side?.name ?? Side.white}/v6',
+        _StudyChapterServerAnalysisSource(:final studyId) => '/study/$studyId/socket/v6',
       },
     );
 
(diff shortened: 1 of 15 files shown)


ijm8710:
Obviously very excited for this.

Quick question, should the overflow options seen in a typical analyst be included here such as

- Show threat
- learn from mistakes
- Board editor
- continue from here
- Share and export 

And if so should they be in studies in general or only once analysis is requested?

lib/src/model/analysis/server_analysis_service.dart
```diff
@@ -131,6 +135,15 @@ class ServerAnalysisService {
           _cancelAnalysis();
           rethrow;
         }
+
+      case _StudyChapterServerAnalysisSource(:final chapterId):
+        _currentAnalysis.value = source;
+        _socketClient!.firstConnection
+            .timeout(const Duration(seconds: 3))
+            .onError((_, _) {})
+            .whenComplete(() {
```
veloce, line 150: What is the reason of ignoring the error here?

lib/src/model/study/study_controller.dart
```diff
@@ -150,6 +160,19 @@ class StudyController extends AsyncNotifier<StudyState>
       return illegalPositionState;
     }
 
+    // If server analysis has already been requested for this chapter,
+    // assume that any comment that has an eval is actually a lichess analysis comment.
+    // (Note that regular comments by the study author might also exist.)
+    if (analysisSummary != null) {
+      for (final node in _root.mainline) {
+        final lichessCommentIndex = (node.comments ?? []).indexWhere((c) => c.eval != null);
```
veloce, line 168: Wonder if we can do a little better here? Like only pick if eval is not null AND the text contains either `Inaccuracy | mistake | blunder`?

lib/src/network/http.dart
```diff
@@ -640,6 +640,15 @@ extension ClientExtension on Client {
     return response.body;
   }
 
+  /// Sends an HTTP GET request with the given headers to the given URL and returns a Future that completes to the response.
+  ///
+  /// Throws a [ServerException] if the response doesn't have a success status code.
+  Future<Response> getAndThrowOnError(Uri url, {Map<String, String>? headers}) async {
```
veloce: This name is good because explicit but at the same time I fear it can add confusion since all the other extension helpers (readJson, postRead, etc.) also throw on error, without mentioning it in their name.

I have another suggestion, which is not perfect but is more consistent maybe:

```dart
/// Sends an HTTP GET request with the given headers to the given URL and returns a [Future<Response>]
///
/// Throws a [ServerException] if the response doesn't have a success status code.
Future<Response> readResponse(Uri url, {Map<String, String>? headers})
```

lib/src/view/study/study_screen.dart
```diff
@@ -148,33 +149,34 @@ class _StudyScreenState extends ConsumerState<_StudyScreen> with TickerProviderS
...
   }
 
   @override
   void didUpdateWidget(covariant _StudyScreen oldWidget) {
-    // If the study has not yet loaded the opening explorer and it is now available
-    // with this chapter, add the opening tab.
-    // We don't want to remove a tab, so if the opening explorer is not available
-    // anymore, we keep the tabs as they are.
-    // In theory, studies mixing chapters with and without opening explorer should be pretty rare.
-    if (tabs.length < 2 && widget.studyState.isOpeningExplorerAvailable) {
```
veloce, line 170: Why removing this?

lib/src/model/analysis/server_analysis_service.dart
tom-anders: copy paste probaly, but I added error handling now :+1:

lib/src/model/study/study_controller.dart
tom-anders: that wouldn't work, we don't only want the mistakes, we need **all** server evals because they are used to draw the acpl graph.

In theory, I think we could cross-check with the `treeParts` in the JSON object returned by the API endpoint, which contains `"by": "lichess"` for the analysis comments, but I'm not sure if it's worth the complexity

lib/src/network/http.dart
tom-anders: good point, fixed

lib/src/view/study/study_screen.dart
tom-anders: The server analysis tab also is hidden in some chapters (interactive ones), so if we keep this code it would become even more complex. IMO with the new tab bar it's not a big deal anymore if tabs get hidden in some chapters and then reappear in others. Was there some other reasoning for this logic?

lib/src/model/analysis/server_analysis_service.dart
```diff
@@ -131,6 +135,18 @@ class ServerAnalysisService {
           _cancelAnalysis();
           rethrow;
         }
+
+      case _StudyChapterServerAnalysisSource(:final chapterId):
+        _currentAnalysis.value = source;
+        _socketClient!.firstConnection
+            .timeout(const Duration(seconds: 3))
+            .onError((err, st) {
+              debugPrint('Error connecting to analysis socket: $err\n$st');
```
veloce: I'd rather use a logger than `debugPrint`.

lib/src/model/study/study_controller.dart
veloce: Well, indeed not. Too bad there is nothing in the PNG that could help us distinguish a comment from lichess.

The whole `lichessAnalysisComments` thing is not ideal by the way. I wonder if we should simply remove it. Since the PGN we get from lichess games will only have evals and comments of the server analysis, there is no point to distinguish them I think (and for other sources we can't distinguish). This is like an extra complexity for nothing.

lib/src/view/study/study_screen.dart
veloce: I am not sure anymore.

Changing the number of tabs when we switch chapters is OK. What we don't want is dynamically changing the number of tabs (especially removing) within the same screen.

But looking at the suppressed code I doubt this is useful anymore so let's keep it removed.

tom-anders: Yes, the number of tabs will only change when we switch chapters, I don't see a situation where it will change without switching chapters

lib/src/model/analysis/server_analysis_service.dart
tom-anders: fixed, I also adapted the other `debugPrint` instances in this file

veloce (approved):

PR #3052, opened 2026-04-25
try flag in github workflow

HaonRekcef:
It removes a lot of unnecessary output lines from the CI test with goal to see the failing once faster, but the failures are not marked with ❌ so I am unsure how big of an improvement it actually is. 
See here how it looks: https://github.com/lichess-org/mobile/actions/runs/24939910760/job/73031778428


veloce:
Yeah... I'd rather see the red cross.

PR #3058, opened 2026-04-27
refactor: simpler fix for #2564 and #2750

tom-anders:
Previous fix was in #2859, but this PR fixes the root cause instead of adding a workaround.

The root cause of this problem was that we check
`_game.serverAnalysis == null` to check if the RetroController needs to trigger server analysis in case it isn't already running. When the server analysis was already triggered by the `AnalysisController`, the problem was that `archivedGameProvider` is not invalidated and still returns `null` for the server analysis field. By using the `GameRepository` directly instead, we get a fresh response from the server which correctly has a non-null `serverAnalysis`.

An alternative fix would be to add
`ref.invalidate(archivedGameProvider(gameId))` when a server analysis request is triggered in the analysis screen, however this leads to the analysis screen being rebuilt when triggering server analysis, which causes a short flicker of the screen during the reload.

To verify that everything still works, I successfully tested the following scenarios:

1) Trigger server analysis in analysis screen and immediately open
   "Learn From Your Mistakes"
   -> The loading bar correctly displays server analysis progress,
      once analysis is finished Learn from your mistakes is available.
2) Trigger server analysis in analysis screen, wait for it to finish
   and only then open "Learn From Your Mistakes"
   -> Learn from your mistakes is available immediately without any
      loading bar.
3) Trigger server analysis, wait for it to finish and then close the
   analysis screen. Open analysis again for the same game and open
  "Learn from your Mistakes".
   -> Learn from your mistakes is available immediately without any
      loading bar.

### DIFF
--- a/lib/src/model/analysis/retro_controller.dart
+++ b/lib/src/model/analysis/retro_controller.dart
@@ -21,7 +21,7 @@ import 'package:lichess_mobile/src/model/engine/evaluation_preferences.dart';
 import 'package:lichess_mobile/src/model/explorer/opening_explorer_preferences.dart';
 import 'package:lichess_mobile/src/model/explorer/opening_explorer_repository.dart';
 import 'package:lichess_mobile/src/model/game/exported_game.dart';
-import 'package:lichess_mobile/src/model/game/game_repository_providers.dart';
+import 'package:lichess_mobile/src/model/game/game_repository.dart';
 import 'package:lichess_mobile/src/network/socket.dart';
 import 'package:lichess_mobile/src/view/engine/engine_gauge.dart';
 import 'package:logging/logging.dart';
@@ -108,7 +108,10 @@ class RetroController extends AsyncNotifier<RetroState> with EngineEvaluationMix
 
     socketClient = ref.watch(socketPoolProvider).open(AnalysisController.socketUri);
 
-    _game = await ref.watch(archivedGameProvider(options.id).future);
+    // Don't use archivedGameProvider, as its value gets cached when we watch it in AnalysisController,
+    // so when a server analysis is already running, _game.serverAnalysis would still be null.
+    // Request directly from the API instead to get the latest value.
+    _game = await ref.read(gameRepositoryProvider).getGame(options.id);
 
     _root = _game.makeTree();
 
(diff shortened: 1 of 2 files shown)


lib/src/model/analysis/retro_controller.dart
```diff
@@ -137,35 +140,7 @@ class RetroController extends AsyncNotifier<RetroState> with EngineEvaluationMix
       // so we don't miss the first progress event.
       serverAnalysisService.lastAnalysisEvent.addListener(_listenToServerAnalysisEvents);
 
-      // Reuse an already available event immediately if it belongs to this game.
```
veloce, line 140: Why removing this block of code? How is it part of the fix?

lib/src/model/analysis/retro_controller.dart
```diff
@@ -108,7 +108,10 @@ class RetroController extends AsyncNotifier<RetroState> with EngineEvaluationMix
 
     socketClient = ref.watch(socketPoolProvider).open(AnalysisController.socketUri);
 
-    _game = await ref.watch(archivedGameProvider(options.id).future);
+    // Don't use archivedGameProvider, as it's value gets cached when we watch it in AnalysisController,
+    // so when a server analysis is already running, _game.serverAnalysis would still be null.
+    // Request directly from the API instead to get the latest value.
+    _game = await ref.watch(gameRepositoryProvider).getGame(options.id);
```
veloce: But if the server analysis is still running, `serverAnalysis` will be null when this is requested, isn't it?

lib/src/model/analysis/retro_controller.dart
ghost, line 442: We used to have `root: recomputeRootView ? _root.view : state.requireValue.root,` here.
I have not gone over the analysis controller files yet, but on first glance it looks like that line was removed accidentally.

lib/src/model/analysis/retro_controller.dart
tom-anders: Yes, but that's not a problem, `_game.serverAnalysis` is not used for the retro after that. The retro logic uses the lichess analysis comments, which are merged into the tree via `ServerAnalysisService.mergeOngoingAnalysis(_root, event.$2.tree);` during `_listenToServerAnalysisEvents()`

lib/src/model/analysis/retro_controller.dart
tom-anders: That was not removed in this PR, but already in this commit: https://github.com/lichess-org/mobile/commit/f8d4d6945b

lib/src/model/analysis/retro_controller.dart
tom-anders: Sorry I put this PR out of draft mode a little bit too early - In the latest version, the equivalent of this code is now simply calling `_listenToServerAnalysisEvents()` when analysis is already ongoing, this should have the same effect: https://github.com/lichess-org/mobile/pull/3058/changes#diff-c6b846a7669af2edc8231be90171e689591845d0cc1d2d1801e29a7e13bd1395R147

tom-anders:
Added two tests now:

- One for when analysis has already been requested but is still running
- One for when analysis has not been requested yet

lib/src/model/analysis/retro_controller.dart
ghost: Sorry, I knew it was a much earlier PR. I just thought I would mention it since you are editing this file already. It looked accidental to me because the other lines weren't changed.
Meaning
`_refreshCurrentNode(recomputeRootView: !isSameEvalString);` to
`_refreshCurrentNode();` 
and
`void _refreshCurrentNode({bool recomputeRootView = false}) {` to
`  void _refreshCurrentNode() {`

If it was all intended then just ignore this. Sorry.

veloce (approved):

PR #3060, opened 2026-04-28
fix: dispose TextEditingController and remove listener in MessageInput

freebeartogoodhome:
`TextEditingController` in `_MessageInputState` was initialized but never disposed.

The listener in `initState` remains active even after the widget is removed from the tree.

I also added a `mounted` check to the listener logic for safety.


veloce (approved):
LGMT, thanks!

PR #3061, opened 2026-04-28
Delay engine move until piece animation completes

RobertKuszmar:
### Problem
Visible animation stuttering when making a move against local computer on Pixel 7 Pro.

### Change
Wait for move animation to complete before calling playEngineMove.


veloce (approved):
Nice! Thanks.

PR #3063, opened 2026-04-28
fix CloudEval in engine_button

ghost:
`currentEval` is currently a `LocalEval` so currently this is always false.
`if (currentEval is CloudEval)`

### DIFF
--- a/lib/src/view/engine/engine_button.dart
+++ b/lib/src/view/engine/engine_button.dart
@@ -290,7 +290,7 @@ class _EnginePopup extends ConsumerWidget {
         engineState != EngineState.computing &&
         (work == null || work.isDeeper != true);
 
-    final currentEval = switch (engineState) {
+    final ClientEval? currentEval = switch (engineState) {
       EngineState.computing || EngineState.idle => evalStateEval,
       _ => null,
     };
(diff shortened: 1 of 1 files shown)


freebeartogoodhome:
`if (currentEval is CloudEval)`  is 100% nonfunctional right now because a LocalEval can never be a CloudEval. It essentially says `if (false)`.

But I don't know if changing currentEval to a ClientEval is enough to fix everything. I have not fully tested it yet so I will convert this to a draft.

freebeartogoodhome:
My draft PR is no good. It would still always be null or a local eval.

If everything is functioning how you like it then the answer is just delete this block:

```
    if (currentEval is CloudEval) {
      return ListTile(
        contentPadding: const EdgeInsets.only(left: 16.0),
        title: Text(context.l10n.cloudAnalysis),
        subtitle: Text(context.l10n.depthX('${currentEval!.depth}')),
        trailing: canGoDeeper
            ? IconButton(
                icon: const Icon(Icons.add_circle_outlined),
                onPressed: goDeeper,
                tooltip: context.l10n.goDeeper,
              )
            : null,
      );
    }
```

PR #3081, opened 2026-04-30
less visible viewer badge + zen mode respected while playing

### LINKED ISSUE #3071: Use more subtle badge for spectate
The spectate badge I believe is red. May make more sense to use the subtle theme badge like that is used with the puzzle streak widget when you resume it for later

HaonRekcef:
I think viewer badge was way too noticeable, surely should not be red. Also stops displaying it in zen mode. Also use a badge for the tv_screen,
Before: 
image

After:
image

fix #3071

### DIFF
--- a/lib/src/view/game/game_screen.dart
+++ b/lib/src/view/game/game_screen.dart
@@ -509,7 +509,7 @@ class _WatcherButton extends ConsumerWidget {
         label: Text('$nb'),
         backgroundColor: Theme.of(context).colorScheme.surfaceContainerHighest,
         textColor: Theme.of(context).colorScheme.onSurfaceVariant,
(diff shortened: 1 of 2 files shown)


tom-anders:
Nitpick, but in the 2nd screenshot the badge covers more of the icon than in the 1st one, I think the old version looked better in that regard

ijm8710:
> Nitpick, but in the 2nd screenshot the badge covers more of the icon than in the 1st one, I think the old version looked better in that regard 

I was literally about to comment the exact same lol. Color works better tho

HaonRekcef:
thanks for pointing it out

image

veloce (approved):
Thanks!

PR #3082, opened 2026-04-30
Add numbers to the broadcast game filters

### LINKED ISSUE #2410: Show count of total boards for a broadcast round
image

Web has this which is a helpful metric 

One idea is to stick it in the gear menu where you can filter between:
- all games
- ongoing

Instead perhaps it can become:
- all games (32)
- ongoing (2)

HaonRekcef:
closes #2410
image

Allows to see the number of all and the ongoing games.

Is only computed when the user presses on the settings cog.


veloce (approved):
Thanks!

PR #3085, opened 2026-05-01
Small cleanup of misused set literals

freebeartogoodhome:
Small cleanup. Slight improvement in readability as well.
 `=>` automatically returns the value of the expression following it.
`() => { ... }` returns a set literal which is not needed here.

Also simplifies `node is root` to false for Branch nodes because they cannot be root by definition.


veloce:
() => { ... } returns a set literal which is not needed here.

Good catch! Thanks. If it was returing a set literal it could not work. I wonder why we have not seen that before...

veloce (approved):
lgtm

PR #3087, opened 2026-05-01
Fix gestures_exclusions sending null to Android and update MediaQuery

freebeartogoodhome:
The old code in gestures_exclusions sends null entries for invalid Rect objects, but the Android decodeExclusionRects() function does not filter them out before processing.  The function will crash attempting to access properties on null.

Added some `await` as well.

Removed MediaQuery.of(context) when used for only one property of MediaQueryData to avoid unnecessary rebuilds.


freebeartogoodhome:
It might be smart to also make a change like this to the Kotlin code.
```
private fun decodeExclusionRects(inputRects: List<Map<String, Int>>): List<Rect> =
    inputRects.mapNotNull { item ->
        val left = item["left"]
        val top = item["top"]
        val right = item["right"]
        val bottom = item["bottom"]

        // If any of these are null, the whole block returns null
        if (left != null && top != null && right != null && bottom != null) {
            Rect(left, top, right, bottom)
        } else {
            // Log a warning instead
            Log.w("RectDecoder", "Skipping malformed rect: $item")
            null
        }
    }
```

veloce:
> It might be smart to also make a change like this to the Kotlin code.
> 
> ```
> private fun decodeExclusionRects(inputRects: List<Map<String, Int>>): List<Rect> =
>     inputRects.mapNotNull { item ->
>         val left = item["left"]
>         val top = item["top"]
>         val right = item["right"]
>         val bottom = item["bottom"]
> 
>         // If any of these are null, the whole block returns null
>         if (left != null && top != null && right != null && bottom != null) {
>             Rect(left, top, right, bottom)
>         } else {
>             // Log a warning instead
>             Log.w("RectDecoder", "Skipping malformed rect: $item")
>             null
>         }
>     }
> ```

Can you do it in this PR as well so I can review it as a whole? Thanks.

freebeartogoodhome:
> Can you do it in this PR as well so I can review it as a whole? Thanks.

Yes, I can do it later today or tomorrow.

freebeartogoodhome:
I accidentally merged that. I'm still terrible with git. I split the MediaQuery idea into one PR now and I'll push that right now. I'll close this and get a new PR with all of the gestures stuff in one PR. It makes sense to have them in separate PRs anyway.

PR #3088, opened 2026-05-01
Be sure to properly cancel and dispose the Retro timer

freebeartogoodhome:
Named the timer and made sure we are properly storing it, canceling it, and disposing it.

Changed the second `mistakes.toIList()` because it was redundant. It has a `toIlist()` right above that.

### DIFF
--- a/lib/src/model/analysis/retro_controller.dart
+++ b/lib/src/model/analysis/retro_controller.dart
@@ -90,6 +90,8 @@ class RetroController extends AsyncNotifier<RetroState> with EngineEvaluationMix
 
   final Completer<void> _serverAnalysisCompleter = Completer<void>();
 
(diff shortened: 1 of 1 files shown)


lib/src/model/analysis/retro_controller.dart
```diff
@@ -440,6 +443,7 @@ class RetroController extends AsyncNotifier<RetroState> with EngineEvaluationMix
   void _refreshCurrentNode({bool recomputeRootView = false}) {
     state = AsyncData(
       state.requireValue.copyWith(
+        root: recomputeRootView ? _root.view : state.requireValue.root,
```
veloce, line 446: This change is not relevant in that PR... but ok, it makes sense to have it.

veloce (approved):
Thanks!

PR #3089, opened 2026-05-02
Feature: clock variations: simple delay & Bronstein delay

Matrix9494:
## Summary

Adds support for additional chess clock modes in the standalone Clock tool:

- Simple delay
- Bronstein delay

The existing increment mode remains the default. The clock settings toolbar now includes a clock type picker, and delay-based modes display their delay value separately from normal increment.

## Details

- Simple delay waits for the configured delay before subtracting main time on each move.
- Bronstein delay subtracts time while thinking, then restores the time spent on the move up to the configured delay.

## Tests

- Added focused tests for standalone clock controller behavior.
- Added regression coverage for repeated delay application and fast Bronstein taps.
- Updated `ChessClock` tests for delayed side switching and partial tick settlement.
- Manual tested on phone + targeted Flutter tests/analyze passed

## AI assistance

This PR was developed with assistance from ChatGPT/Codex. I reviewed and tested the changes manually, and I understand the submitted code. A demo video is attached.

## Demo

A demo video is attached showing the new clock type picker and the delay/Bronstein behavior in the standalone Clock tool.

https://github.com/user-attachments/assets/e18e385f-41e2-43ec-86db-158d9410aece


ijm8710:
2 feedbacks:

Personally I feel the clock toolbar has a lot going on.

I feel it may be tidier to instead have a dropdown filter in the custom section where you can adjust the increments and choose delays if one so chooses


Would also be nice to have some of the other options that were available in the old app (see photo below)

If you prefer to keep how you already did it and await Veloce thoughts all good, but if you agree, cool as well

image

Matrix9494:
I feel we may eventually want to unify clock settings across the different entry points: lobby games, OTB games, and the standalone clock. Right now they each expose slightly different clock configuration UIs and support different sets of clock styles. Maybe the lobby game time-control UI could become the shared base, with each context enabling only the clock types it actually supports.

veloce:
Quick feedback for now: I agree there are too many icons in the toolbar. Let's find a way to have less icon please (and share a new screen recording when it is done).

Also, let's not unify all clock settings. I want the online clock setup to be completely separate from the offline/OTB clocks.

We can unify OTB and standalone clocks if we want, but let's leave lobby (and all online games) time control setup out of it.

Matrix9494:
Turn to a new PR #3122 .

PR #3097, opened 2026-05-03
Fix gestures_exclusions sending null to Android

freebeartogoodhome:
The main idea here is to prevent gestures_exclusions from sending null to Android.

The Kotlin code has also been made more defensive.

Added `await` to some lines that are dealing with a `Future`.

Changed `<Rect>[]` to `<Map<String, int>>[]` to explicitly match the Kotlin function signature and use standard serializable types.

I only stopped importing `widgets.dart` here because it was redundant. material.dart transitively imports widgets.dart. You can leave that in if it's your preference.


veloce (approved):
lgtm

PR #3100, opened 2026-05-03
Fix best move arrow not shown when engine lines set to zero

### LINKED ISSUE #2871: "Best move" not showing
### What is the problem?

On the computer analysis I'm used to seeing move recommendations, but I don't see them currently (despite what I think is the relevant setting being enabled). 

Image
Image

1. View analysis of game (screenshots are of https://lichess.org/p6mNFTB3/black for example)

### App version

0.22.7

### Device Info

Fairphone 3, LineageOS

### Additional information

I have 8.0.0 installed too and that shows it fine.

Image

MaartenD:
**Fix: Best move arrow not shown when engine lines is set to 0**

When the "Show best move arrow" option was enabled but the number of engine lines was set to 0, the best move arrow was not displayed on the board.

**Root cause**
In analysis_board.dart, the best moves were filtered using .take(enginePrefs.numEvalLines). When numEvalLines is 0, .take(0) returns an empty list, so no arrow was drawn. This was inconsistent with the UCI protocol layer in uci_protocol.dart, which already uses math.max(1, multiPv) to ensure Stockfish always evaluates at least one line.

**Fix**
Changed .take(enginePrefs.numEvalLines) to .take(math.max(1, enginePrefs.numEvalLines)) in analysis_board.dart, so that at least the best move is always passed to computeBestMoveShapes() when the engine is running, regardless of the number of displayed lines.

**Test**
A new test was added to the Engine best move arrow group in analysis_screen_test.dart, verifying that the best move arrow is displayed even when the number of engine lines is set to 0.

Fixes: #2871

### DIFF
--- a/lib/src/view/analysis/analysis_board.dart
+++ b/lib/src/view/analysis/analysis_board.dart
@@ -1,3 +1,5 @@
+import 'dart:math' as math;
+
 import 'package:chessground/chessground.dart';
 import 'package:dartchess/dartchess.dart';
 import 'package:fast_immutable_collections/fast_immutable_collections.dart';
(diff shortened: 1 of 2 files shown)


veloce (approved):
Thanks!

Issue #2024, opened 2025-08-01
When fetching games, title is off-center


ijm8710:
![image](https://github.com/user-attachments/assets/edc22a96-785e-47a0-a0d6-ea634553bbca)

PR #3101, opened 2026-05-03
fix: center time control header in active game

### LINKED ISSUE #3074: For active game, the time control header is no longer centered
It is now left oriented

I am guessing this was unintentional as this bug has resurfaced before but just making sure

It's also weirdly off-center when you seek a game match as well

sajal90:
fixes #3074 

The AppBar for the active game screen was missing centerTitle: true, causing the time control header to appear left-aligned on Android.

### DIFF
--- a/lib/src/view/game/game_screen.dart
+++ b/lib/src/view/game/game_screen.dart
@@ -209,6 +209,7 @@ class _GameScreenState extends ConsumerState<GameScreen> {
           resizeToAvoidBottomInset: false,
           appBar: AppBar(
             leading: isRealTimePlayingGame ? SocketPingRatingIcon(socketUri: socketUri) : null,
+            centerTitle: true,
             title: _StandaloneGameTitle(id: createdGameId, lastMoveAt: widget.lastMoveAt),
             actions: [
               _WatcherButton(gameId: createdGameId),


ijm8710:
Thanks, there is a plethora of inconsistency on this so not really sure what should intentionally be left aligned vs intentionally centered

But also, the main part of https://github.com/lichess-org/mobile/issues/2024 the game seek (at least on iOS) is neither centered nor left aligned. Instead it leans slightly right? I also have a few other examples listed in the comment section.

If you open one of the first 10 most recent games from home, it's kind of neither left aligned nor centered but leans left. I'm guessing that's supposed to be centered?

The tabs now (such as puzzles, learn etc) are left aligned as tab headers. I think that's intentional now?

Would help to get some clarification from Veloce about what is supposed to be where. But as of right now, it's all over the place.

sajal90:
yes, on my android device(Moto G45), the title is very odd. It's not fully left or center, but after adding centerTitle: true, it sits correctly in the middle, or that's how it seems to me. 

prev
after

ijm8710:
> yes, on my android device(Moto G45), the title is very odd. It's not fully left or center, but after adding centerTitle: true, it sits correctly in the middle, or that's how it seems to me. 
> 
> 
> 
> prev
> 
> after
> 
> 
> 
> 

Yah! I saw you fixed that one in this pr but yeah was sharing that there are a ton of screens where things are neither left aligned nor fully centered

veloce (commented):
I am not sure it is an improvement, that's why I hesitate to merge it.

The `centerTitle` is not used in other parts of the app (offline games, TV, etc). So it would not be consistent.

Also if the title is not centred it is because we show app bar action buttons on the right, so there is too little place for the title to be shown entirely if it is in the centre.

ijm8710:
> I am not sure it is an improvement, that's why I hesitate to merge it.
> 
> The `centerTitle` is not used in other parts of the app (offline games, TV, etc). So it would not be consistent.
> 
> Also if the title is not centred it is because we show app bar action buttons on the right, so there is too little place for the title to be shown entirely if it is in the centre.

Hi Veloce, I am one who prefers the centering, but assuming you generally prefer left-aligned, should we have fuller consistency throughout the app? 

I am also unsure why centering makes a spacing difference. If we need all the space, centering will cover the entire bar regardless and hence centering and left assigned would be equivalent in a situation where we need max space, no?

image
image
image
image
image
image
image
image
image
image

veloce:
You should post so many screenshots, that is unreadable. If you do that, at least please reduce their size.

ijm8710:
> You should post so many screenshots, that is unreadable. If you do that, at least please reduce their size.

Is this easier to digest?

I'm sure there will be exceptions but the lack of inconsistency between center and left alignment I think is the issue. 

Again I prefer centering and don't see why size constraints matter as  it should still take up entire bar if needed regardless of center or left aligned 

But regardless, even if you prefer left-alignment shouldn't it be more consistent throughout the app

image
image

ijm8710:
Hi @veloce was my pic collage a little easier to digest and any chance you agree some more consistency may make sense?

TBH we actually center more than we don't center. I believe game boards, analysis board and board editor are only times we don't.

PR #3104, opened 2026-05-04
Update country list to turn the same as server

SergioGlorias:
https://github.com/lichess-org/lila/blob/054637e35d60b68d1a5d78b6de3f674b5a28124c/modules/user/src/main/Flags.scala


veloce (approved):
👍

PR #3107, opened 2026-05-04
fix(move-list): move list padding in scroll view

lsaudon:
| Before | After |
| - | - |
| Screenshot_20260504-220658_Lichess - DEBUG | Screenshot_20260504-220718_Lichess - DEBUG |


veloce:
Sorry but I just don't see what is the problem here.

At least I can't reproduce any padding issue on my devices.

lsaudon:
It’s subtle, but there’s a 5-pixel margin on the left side.

| Before | After |
| - | - |
|image|image|
|image|image|

veloce (approved):
Padding inside the scroll view, thanks!

PR #3117, opened 2026-05-05
fix/FEN import ignores side to move

### LINKED ISSUE #3111: FEN import ignores side to move
### What is the problem?

When I go to board editor and then import a FEN, the default side to move remains white, even if it's black in the FEN. (Possibly also applies to castling rights, haven't checked.)

1. Go to board editor
2. Click on the pencil button
3. Paste a FEN where it's black to move
4. Click the analysis button
5. It's white to move

### App version

v0.23.5

### Device Info

Pixel 6 / Android 16

### Additional information

_No response_

MaartenD:
**Fix: Board editor FEN paste ignores side to move (and castling rights)**

**Root cause**
loadFen in BoardEditorController only called _updatePosition, which updates piece positions but leaves all other state (sideToPlay, castlingRights, enPassantSquare, etc.) unchanged. Pasting a FEN where black is to move had no effect on the side to play.

**Fix**
loadFen now distinguishes between two cases:

- Partial FEN (piece placement only, 1 field): preserves the old behavior, only pieces are updated, castling rights remain in state and are filtered by piece positions via _castlingRightsPart.
- Full FEN (2+ fields): all fields are read from the parsed FEN, sideToPlay, castlingRights, enPassantSquare, halfmoves, and fullmoves are all updated.

**Test added**
A new test Pasting FEN with black to move correctly sets side to play was added to test/view/board_editor/board_editor_screen_test.dart. It pastes a full FEN with black to move and verifies that sideToPlay is correctly set to Side.black after the paste.

**Note**
The FEN parsing logic in loadFen (for the full FEN case) is similar to the inline logic in build(). Should this be extracted into a shared helper method, or is the current duplication acceptable?

In the videos i used this FEN: r1bqkbnr/pppp1ppp/2n5/4p3/2B1P3/5N2/PPPP1PPP/RNBQK2R b KQkq - 2 3

**Bug**
https://github.com/user-attachments/assets/28160b2f-c85f-4f4b-8e19-2551a86c6771

**Fixed**
https://github.com/user-attachments/assets/320e4e4e-9a70-4b14-9ea1-b84b444c5330

Fixes: #3111

### DIFF
--- a/lib/src/model/board_editor/board_editor_controller.dart
+++ b/lib/src/model/board_editor/board_editor_controller.dart
@@ -93,7 +93,37 @@ class BoardEditorController extends Notifier<BoardEditorState> {
   }
 
   void loadFen(String fen) {
-    _updatePosition(readFen(fen).lock);
+    final fenParts = fen.trim().split(RegExp(r'\s+'));
+    final setup = Setup.parseFen(fen);
+    final pieces = readFen(fen).lock;
+
+    if (fenParts.length == 1) {
+      // Partial FEN (piece placement only): preserve existing state for side to play,
+      // castling rights, etc. Castling is filtered by piece positions in _castlingRightsPart.
+      _updatePosition(pieces);
+    } else {
+      // Full FEN: update all fields from the parsed FEN.
+      final position = Position.setupPosition(
+        state.variant.rule,
+        setup,
+        ignoreImpossibleCheck: true,
+      );
+      final castlingRights = IMap({
+        CastlingRight.whiteKing: position.castles.rookOf(Side.white, CastlingSide.king) != null,
+        CastlingRight.whiteQueen: position.castles.rookOf(Side.white, CastlingSide.queen) != null,
+        CastlingRight.blackKing: position.castles.rookOf(Side.black, CastlingSide.king) != null,
+        CastlingRight.blackQueen: position.castles.rookOf(Side.black, CastlingSide.queen) != null,
+      });
+      state = state.copyWith(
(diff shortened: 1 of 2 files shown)


lib/src/model/board_editor/board_editor_controller.dart
```diff
@@ -93,7 +93,37 @@ class BoardEditorController extends Notifier<BoardEditorState> {
...
+      // castling rights, etc. Castling is filtered by piece positions in _castlingRightsPart.
+      _updatePosition(pieces);
+    } else {
+      // Full FEN: update all fields from the parsed FEN.
+      final position = Position.setupPosition(
+        state.variant.rule,
+        setup,
+        ignoreImpossibleCheck: true,
+      );
+      final castlingRights = IMap({
```
veloce, line 111: this logic could be factored out in a private method.

test/view/board_editor/board_editor_screen_test.dart
```diff
@@ -385,6 +385,25 @@ void main() {
...
+        await tester.pumpAndSettle();
+
+        await tester.tap(find.byIcon(Icons.paste));
+        await tester.pumpAndSettle();
+
+        final container = ProviderScope.containerOf(tester.element(find.byType(BoardEditorScreen)));
+        final state = container.read(boardEditorControllerProvider(null));
+        expect(state.sideToPlay, Side.black);
+      });
+
```
veloce, line 406: Another test would be useful to test the castling rights logic.

PR #3119, opened 2026-05-05
Fix/promotion correspondence with confirm overlay issue

### LINKED ISSUE #3116: When promoting a pawn in correspondence games with move confirmation, the overlay stays open
### What is the problem?

I play quite some correspondence games. 

When I move to promote a pawn I get both the overlay to pick the piece and the confirmation request, which seems to interfere with it. 
I then pick the piece and confirm the move and the overlay stays open. See screenshot. 

Image

1. Play a correspondence game with move confirmation
2. Get a pawn to promote

### App version

0.23.11 (latest at the time)

### Device Info

Pixel 10 

### Additional information

_No response_

MaartenD:
**Root cause**
In game_controller.dart, userMove() has two sequential early-return checks:

1. When a pawn reaches the promotion square, the move arrives without a promotion role (promotion: null). The controller detects this, stores the move in promotionMove, and returns early, which causes the promotion picker to appear so the player can choose a piece.
2.  If move confirmation is enabled, the controller stores the move in moveToConfirm and returns early, showing the confirmation UI. 

When the player selects a piece in the promotion picker, onPromotionSelection calls userMove again, this time with the chosen role set on the move (e.g. promotion: Role.queen). The first check is now skipped, and the second check fires, but it only set moveToConfirm without clearing promotionMove. This left both promotionMove and moveToConfirm set simultaneously, causing both the promotion picker and the confirmation UI to be visible at the same time.

**Fix**
One addition in game_controller.dart: when setting moveToConfirm, also clear promotionMove:
``` state = AsyncValue.data(curState.copyWith(moveToConfirm: move, promotionMove: null)); ```

**Test added**
A new test in test/view/game/game_screen_test.dart:

_promotion_ with move confirmation closes promotion picker after piece selection

- Sets up a game using Variant.fromPosition with a custom FEN (6k1/4P3/8/8/8/8/8/4K3 w - - 0 1), autoQueen: AutoQueen.never so the promotion picker is shown, and submitMove: true for move confirmation.
- Plays e7–e8, then verifies promotionMove is set and moveToConfirm is null.
- Selects the queen, then verifies promotionMove is null and moveToConfirm is set.

Fixes: #3116

### DIFF
--- a/lib/src/model/game/game_controller.dart
+++ b/lib/src/model/game/game_controller.dart
@@ -185,7 +185,7 @@ class GameController extends AsyncNotifier<GameState> {
     }
 
     if (curState.shouldConfirmMove && isPremove != true) {
-      state = AsyncValue.data(curState.copyWith(moveToConfirm: move));
+      state = AsyncValue.data(curState.copyWith(moveToConfirm: move, promotionMove: null));
       return;
     }
 
--- a/test/view/game/game_screen_test.dart
+++ b/test/view/game/game_screen_test.dart
@@ -299,14 +299,11 @@ void main() {
     testWidgets('promotion with move confirmation closes promotion picker after piece selection', (
       WidgetTester tester,
     ) async {
-      // Position: white pawn on e7 ready to promote, black king on g8, white king on e1.
-      // Black king must not be on d8 or f8 (pawn attack squares).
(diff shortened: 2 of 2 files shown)


veloce (approved):
Thanks!

PR #3120, opened 2026-05-05
Feature/ECO code on openings explorer

### LINKED ISSUE #3055: ECO on openings explorer
On the old lichess app, we were able to see the ECO code of the opening in the analysis window, and I found that really useful for viewing my studies, I would appreciate this feature making its way to the new app

Image
For example here, it would be nice to have the ECO of the Italian game, so that when I'm analyzing on my phone (when I don't have access to my computer) I could have slightly better organization of my openings.

Tested on version 0.23.5, iPhone 13, iOS 26.4.2

MaartenD:
Added the ECO code in the openings explorer

https://github.com/user-attachments/assets/f4168bc9-cf2d-4019-baf2-df0b83c6f6c6

Fixes: #3055

### DIFF
--- a/lib/src/view/explorer/opening_explorer_widgets.dart
+++ b/lib/src/view/explorer/opening_explorer_widgets.dart
@@ -35,7 +35,7 @@ class OpeningNameHeader extends StatelessWidget {
             ],
             Expanded(
               child: Text(
-                opening.name,
+                opening.eco.isNotEmpty ? '${opening.eco} • ${opening.name}' : opening.name,
                 style: TextStyle(
                   color: ColorScheme.of(context).onSurface,
                   fontWeight: FontWeight.bold,


lib/src/view/explorer/opening_explorer_widgets.dart
```diff
@@ -35,7 +35,7 @@ class OpeningNameHeader extends StatelessWidget {
             ],
             Expanded(
               child: Text(
-                opening.name,
+                opening.eco.isNotEmpty ? '${opening.eco} • ${opening.name}' : opening.name,
```
veloce, line 38: I'd just remove the ` • `, so the title is shorter (it is done like that on web too).

MaartenD: You're right. I should have taken a closer look myself.

PR #3125, opened 2026-05-06
Fix: Refactor online_friends and friend_screen to fix loading issues

ghost:
**online_friends.dart:**
Returns an immediate empty list (`IListConst([])` to prevent the hang by ensuring the provider completes instantly.
The socket listener then updates the state reactively when data arrives.
Added `pause()`/`resume()` so we can handle app focus without initializing the socket again.

**friend_screen.dart:**
Replaced `await .future` with a reactive provider that watches the async values.
The screen no longer waits for both lists to finish before rendering.
Swapped `StatefulBuilder` to `ref.invalidate` to keep the UI up to date after an unfollow.

### DIFF
--- a/lib/src/view/relation/friend_screen.dart
+++ b/lib/src/view/relation/friend_screen.dart
@@ -24,15 +24,20 @@ final followingProvider = FutureProvider.autoDispose<IList<User>>((ref) {
   return ref.withClient((client) => RelationRepository(client).getFollowing());
 });
 
-final onlineAndFollowingProvider =
-    FutureProvider.autoDispose<(IList<OnlineFriend> onlineFriends, IList<User> following)>((
-      ref,
-    ) async {
-      final onlineFriends = await ref.watch(onlineFriendsProvider.future);
-      final following = await ref.watch(followingProvider.future);
+final onlineAndFollowingProvider = Provider.autoDispose<AsyncValue<(IList<OnlineFriend> onlineFriends, IList<User> following)>>((ref) {
+  final online = ref.watch(onlineFriendsProvider);
+  final following = ref.watch(followingProvider);
 
-      return (onlineFriends, following);
-    });
+  return online.when(
+    data: (onlineData) => following.when(
+      data: (followingData) => AsyncValue.data((onlineData, followingData)),
+      error: (e, st) => AsyncValue.error(e, st),
+      loading: () => const AsyncValue.loading(),
+    ),
+    error: (e, st) => AsyncValue.error(e, st),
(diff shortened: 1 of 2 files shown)


lib/src/view/relation/friend_screen.dart
```diff
@@ -64,32 +72,34 @@ class _FriendScreenState extends ConsumerState<FriendScreen> with TickerProvider
...
-          appBar: PlatformAppBar(
-            title: Text(context.l10n.friends),
-            bottom: TabBar(
-              controller: _tabController,
-              tabs: <Widget>[
-                Tab(text: context.l10n.nbFriendsOnline(value.$1.length)),
-                Tab(text: context.l10n.nbFollowing(value.$2.length)),
-              ],
-            ),
+    return onlineAndFollowing.when(
```
veloce, line 75: According to riverpod author, `when` is not the preferred way anymore. The preferred way is pattern matching so please don't change that.

lib/src/view/relation/friend_screen.dart
```diff
@@ -103,29 +113,28 @@ class OnlineFriendsWidget extends ConsumerWidget {
...
             ],
           );
         },
         error: (error, stackTrace) {
-          debugPrint(
-            'SEVERE: [PlayerScreen] could not load following online users; $error\n $stackTrace',
-          );
-          return const Center(child: Text('Could not load online friends'));
+          debugPrint('SEVERE: [OnlineFriendsWidget] load failed: $error');
+          return const SizedBox.shrink();
```
veloce, line 127: Why changing that?

lib/src/view/relation/friend_screen.dart
```diff
@@ -103,29 +113,28 @@ class OnlineFriendsWidget extends ConsumerWidget {
...
         },
         loading: () => ShimmerLoading(
           isLoading: true,
-          child: ListSection.loading(itemsNumber: 5, header: true),
+          child: ListSection.loading(itemsNumber: 3, header: true),
```
veloce, line 131: This change does not seem relevant to this PR.

veloce (changes requested):
I have not reviewed everything yet. I've only added a few comments.

There are too many changes in this PR for me to review properly.

If you want to fix loading issues, please only change code related to loading issues, and don't change the view.

If you think the view needs to be changed better create an issue to explain what is wrong, or ask a question on discord. Thanks.

lib/src/view/relation/friend_screen.dart
ghost: Ok, still learning so I didn't know that.

PR #3132, opened 2026-05-07
fix(move_list): apply piece notation font and padding scroll in stacked mode

lsaudon:
I spent a bit of time trying to refactor move_list.
I noticed that in the stacked version, the Chess font wasn’t being used.
And same bug on padding and scroll view in stacked mode

| Before | After |
| - | - |
|Capture d&#39;écran 2026-05-07 210849|Capture d&#39;écran 2026-05-07 210600|
|Capture d&#39;écran 2026-05-07 212427|image|

---

I haven't included this in the PR, but `inlineColor` and `inlineDecoration` are never used by the calling widgets.

### DIFF
--- a/lib/src/widgets/move_list.dart
+++ b/lib/src/widgets/move_list.dart
@@ -113,41 +113,39 @@ class _MoveListState extends ConsumerState<MoveList> {
             ),
           )
         : Card(
-            child: Padding(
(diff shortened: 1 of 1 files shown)


veloce (approved):
Good catch, thanks!

PR #3135, opened 2026-05-08
Refactor online_friends and friend_screen to prevent a hang

ghost:
**online_friends.dart**
Change the build to non-blocking by returning an initial empty state and updating it reactively via the socket stream

**friend_screen.dart**
Remove async/await to use ref.watch(provider) directly

### DIFF
--- a/lib/src/view/relation/friend_screen.dart
+++ b/lib/src/view/relation/friend_screen.dart
@@ -25,13 +25,18 @@ final followingProvider = FutureProvider.autoDispose<IList<User>>((ref) {
 });
 
 final onlineAndFollowingProvider =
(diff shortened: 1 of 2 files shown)


lib/src/view/relation/friend_screen.dart
```diff
@@ -25,13 +25,18 @@ final followingProvider = FutureProvider.autoDispose<IList<User>>((ref) {
 });
 
 final onlineAndFollowingProvider =
-    FutureProvider.autoDispose<(IList<OnlineFriend> onlineFriends, IList<User> following)>((
```
veloce, line 28: FutureProvider already returns an `AsyncValue` so we should keep that.

veloce: Actually I kept your work but chose to go even further in the simplification and remove the combined provider.

PR #3138, opened 2026-05-08
fix/removed site and event from broadcasts pgnlabels

### LINKED ISSUE #3136: For new pgn section for broadcast the site fields plaintext links to game rather than listing location

MaartenD:
I made the new pgnlabels on the broadcasts screen inline with the website. I removed the Event and Site label.

See video for new mobile version and screenshot with current website labels.

https://github.com/user-attachments/assets/96a4df3f-aa25-4251-9178-f6d147e24758

Broadcasts-pgnlabels-website

Fixes: #3136


ijm8710:
Hey bit confused.why would we remove event? And for site, wasn't as much about removing it but feel it's pulling wrong info. There is a location field there but for some reason it was feeding a diff Link. Is that available and if not I feel it would make more sense to request an api change rather than remove a field that web has

Issue #3139, opened 2026-05-08
Lichess TV no longer continuously autoplays in new app after rematch series ends


bashtheshell:
## Summary

I am reporting this against the **new Lichess mobile app**.

I mention the legacy app only as a comparison point because the legacy app continuously plays Lichess TV from the selected category/channel without requiring manual re-selection.

In the new mobile app, Lichess TV does not appear to continuously autoplay games from the selected TV category/channel.

Autoplay works while the same two players continue rematching each other. However, when one player stops accepting rematches and the rematch series ends, Lichess TV stops instead of moving to another live game from the same selected category/channel.

## Affected Lichess TV categories/channels

This affects regular Lichess TV categories such as:

- Bullet
- Blitz
- Rapid
- Top Rated TV

## Expected behavior

When the current game or rematch series ends, Lichess TV should automatically move to another live game from the same selected TV category/channel.

For **Top Rated TV**, it should continue selecting another eligible top-rated game instead of stopping.

## Actual behavior

When the rematch series ends, autoplay stops.

I have to manually go back and re-select the Lichess TV category/channel to continue watching.

## Legacy app comparison

I had been using the legacy app for as long as possible specifically because Lichess TV could run continuously there for hours without intervention.

I recently stopped using the legacy app after seeing the warning banner encouraging users to move to the new app. That is what prompted me to report this behavior in the new app.

I understand the new app is separate from the legacy app, so this may not be a code regression in the same implementation. However, from a user-behavior standpoint, this appears to be a parity/regression issue because the legacy app continuously played Lichess TV, while the new app stops after a rematch series ends.

### Steps to reproduce the bug

1. Open the new Lichess mobile app.
2. Go to **Lichess TV**.
3. Select a TV category/channel, such as **Bullet**, **Blitz**, **Rapid**, or **Top Rated TV**.
4. Watch the current game finish.
5. If the same two players accept a rematch, observe that the app continues to the next game in that rematch series.
6. Wait until one player does not accept another rematch and the rematch series ends.
7. Observe that Lichess TV stops instead of automatically moving to another live game from the same selected category/channel.
8. Manually go back and re-select the TV category/channel to continue watching.

### App version

0.23.13

### Device Info

iPad (9th generation), iPadOS 26.4.2

HaonRekcef:
@bashtheshell I cannot replicate, can you share a video?
I noticed that it doesn't work for userTV but for the Lichess TV I could not reproduce.

[...]

PR #3141, opened 2026-05-08
Fix Lichess TV channel autoplay after rematch series ends

Matrix9494:
## Summary

Fix Lichess TV channel autoplay after the current featured game stops rematching. #3139 

When a matching `tvSelect` event is received, reload the current TV channel via `startWatching()` instead of requiring the socket event to include the next game id and orientation.

## Testing

- `flutter test test/model/tv/tv_controller_test.dart`

### DIFF
--- a/lib/src/model/tv/tv_controller.dart
+++ b/lib/src/model/tv/tv_controller.dart
@@ -120,11 +120,6 @@ class TvController extends AsyncNotifier<TvState> {
     });
   }
 
-  Future<void> _moveToNextGame((GameId id, Side orientation) game) async {
-    final newState = await _connectWebsocket(game);
-    state = AsyncValue.data(newState);
-  }
-
   bool canGoBack() => state.mapOrNull(data: (d) => d.value.stepCursor > 0) ?? false;
 
   bool canGoForward() =>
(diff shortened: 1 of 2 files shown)


ijm8710:
Is this only an issue on iPad? I don't reproduce this on iOS? Any reason why I see it switching fine?

lib/src/model/tv/tv_controller.dart
```diff
@@ -270,8 +264,7 @@ class TvController extends AsyncNotifier<TvState> {
         final json = event.data as Map<String, dynamic>;
         final eventChannel = pick(json, 'channel').asTvChannelOrNull();
         if (eventChannel != null && eventChannel == params.channel) {
-          final data = TvSelectEvent.fromJson(json);
```
veloce, line 273: I am not sure to understand the fix. Why the server would send the id and orientation if we don't use them?

Why is it not working in the current state?

Matrix9494:
@veloce Sorry for the confusion earlier. I initially thought I had reproduced this bug on my Android device, but later realized that it was a different issue. Everything appears to run correctly on my local Android device now.

I can still reproduce the original bug on my iPad, but I currently don’t have a way to debug it there, so I may not be able to make further progress at the moment.

PR #3144, opened 2026-05-08
Add some mounted checks before setState, after using await

ghost:
I just added some mounted checks for safety after we use `await` before we use `setState`


veloce (approved):
thanks!

PR #3153, opened 2026-05-09
feat(editor): implement separate function board clearing logic

fahadahmed32:
**Summary**
This PR fixes a bug in the Board Editor where the **"Clear Board"** action would fail or cause an invalid state. The previous implementation attempted to load an empty FEN string, which triggers Position.setupPosition validation. Since a valid chess position requires kings, the empty FEN was rejected, leading to crashes or the board not updating.

**Changes**
BoardEditorController: Added a dedicated **clearBoard()** method. This method bypasses FEN parsing and directly updates the state with an empty piece map and reset metadata (castling rights, en passant, etc.).
BoardEditorScreen: Updated the "Clear Board" bottom sheet action to call the new clearBoard() method instead of loadFen(kEmptyFEN).

### DIFF
--- a/lib/src/model/board_editor/board_editor_controller.dart
+++ b/lib/src/model/board_editor/board_editor_controller.dart
@@ -108,6 +108,20 @@ class BoardEditorController extends Notifier<BoardEditorState> {
     }
   }
 
+  void clearBoard() {
+    state = state.copyWith(
+      pieces: IMap(const {}), // No pieces
+      castlingRights: IMap(const {
+        CastlingRight.whiteKing: false,
+        CastlingRight.whiteQueen: false,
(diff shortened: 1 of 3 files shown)


tom-anders:
Thanks for your contribution! It would be great if you could add a regression test to https://github.com/lichess-org/mobile/blob/main/test/view/board_editor/board_editor_screen_test.dart (i.e. a test that would fail without your commit)

fahadahmed32:
> Thanks for your contribution! It would be great if you could add a regression test to https://github.com/lichess-org/mobile/blob/main/test/view/board_editor/board_editor_screen_test.dart (i.e. a test that would fail without your commit)

added

veloce (approved):
Thank you!

PR #3156, opened 2026-05-10
Home friends widget

### LINKED ISSUE #3003: Optional 'friends last online' home widget carousel
Kind of like this. Can ping for say just the last 10 or so with a view all button to expand to full friends page. 

image

HaonRekcef:
Closes #3003
To reduce the server loads this PR works as follows
- ~~`onlineFriendsSnapshotProvider` that requests the online and playing friends without listening for updates, that is called every app start and home screen refresh~~
- ~~a following cache, that is stored in a sqllite db, because of this the server has to compute the heavy `rel/api/following` only once a day, except if the user navigates explicitly to the FriendsScreen or the PickPlayerScreen screen.~~
- As users can follow up to 400 players, and the following widget only shows up to 10, the users are sorted:
1. Playing players 
2. Online players 
3. Players from the offline cache next, ordered by a mixture of factors (Amount of user interaction with that player, recently active, recently interacted with and some randomness)

~~The interactions are stored in the sqllite db locally when the player interacts with the widget.~~

[Screencast_20260510_093744.webm](https://github.com/user-attachments/assets/c9c7ff7a-e81f-4b37-b70c-a73997641018)


ijm8710:
Super duper excited about this one!

veloce:
This is really cool!

Some UI feedback for now:
* the cards look too cramped: let's add some more vertical space and padding
* the user name could be more visible, also it may be better to show the avatar (flair or initial) in a ListTile, with the user name as title and the status as subtitle (what we already have in the UserScreen app bar)
* I like how you use the tonal variant of the button to show the hierarchy of actions
* however, the button is too big. Its size could be reduced by a third I think, it would still be very visible; it also needs a bit more vertical padding around it.
* may be worthwhile to have an icon in the button
* the button would probably be better aligned on the left than central
* we may want to use the right bottom corner to have a "Send message icon button", that could be useful too (in that case we'd use an OutlinedButton to make it less important than the FilledButton)
* if it's not too costly to send requests to the server (there is an endpoint to get profiles by sending a list of ID) to get the offline user profiles, displaying "Active 2h ago" would be better than "Offline" as subtitle (this is low priority, I'm just listing here what comes to mind to not forget it)

HaonRekcef:
image

@veloce the problem is that like this we can only display 1.5 friends on a normal phone...  


in a more compacted form without avatar and inbox (which I am anyways not sure we should display as it is not a friends list really but a following list, so it is not a reflective relation)
image

ijm8710:
> image
> 
> 
> 
> @veloce the problem is that like this we can only display 1.5 friends on a normal phone...  
> 
> 
> 
> 
> 
> in a more compacted form without avatar and inbox (which I am anyways not sure we should display as it is not a friends list really but a following list, so it is not a reflective relation)
> 
> image
> 
> 

I am team Noah on this one. I think 2.5 is much more valuable than 1.5 and he expressed very well why messaging is less vital

I still would love the badge from here on the carousel header. They use it for total friend count but I would love it for online friend count. It's an easy way to bring attention to the widget that you have friends online similar to the old app where the top right friend icon had a badge when friends were online. To me the carousel is most valuable when friends are online which is not always the case when you only follow a few people.
image

veloce:
Capture d’écran 2026-05-19 à 10 41 39

@HaonRekcef agreed, the compact form is better as we have 2.5 instead of 1.5 displayed friends.

Inbox button looked weird too I agree.

However in that screenshot I still find the button is way too big. The button is taking as much space as the whole user block. Seeing this I got the feeling that the button is trying to yell "Hiiiiit meeeee!!". The most important information here should be the user name and status, thus it should take more space and be more visible than the button. The button is already quite visible by itself being a FilledButton, so it should be small.

If that helps for this card design we could even display a bit more information about the user. For instance the website displays the ratings, the score against me, the number of games played, etc.
Note that I said *we could*, not *we have to*. I'd rather see less information than useless information here.

Last thing: the "online" not capitalised is an issue here. If Lila translations don't provide this we need to have a mobile one.

HaonRekcef:
> However in that screenshot I still find the button is way too big. The button is taking as much space as the whole user block. Seeing this I got the feeling that the button is trying to yell "Hiiiiit meeeee!!". The most important information here should be the user name and status, thus it should take more space and be more visible than the button. The button is already quite visible by itself being a FilledButton, so it should be small.

That was my intention actually, but I reduced the size:
image

 
> If that helps for this card design we could even display a bit more information about the user. For instance the website displays the ratings, the score against me, the number of games played, etc. Note that I said _we could_, not _we have to_. I'd rather see less information than useless information here.

I think here simple is better.


> Last thing: the "online" not capitalised is an issue here. If Lila translations don't provide this we need to have a mobile one.

What speaks against .capitalize()?


Also the different languages are problematic. The german translation for challenge turns it into: `Zu einer Partie herausfordern` . I guess we have to create a seperate translation key, asking the translators to be succinct.

veloce:
> What speaks against .capitalize()?

It won't work in some languages.

> Also the different languages are problematic. The german translation for challenge turns it into: Zu einer Partie herausfordern . I guess we have to create a seperate translation key, asking the translators to be succinct.

Definitely.

veloce:
I am still not satisfied with the button size 😅 

Some user have long names. The username looks better now, but the font should be a little smaller I think.

HaonRekcef:
Maybe switching to an outline Button would be better?

image

Another problem I noticed, is that if we want the challenge button, we have to ask the server devs to include a field for that, as players can disable that other playes can challenge them.

veloce:
> Another problem I noticed, is that if we want the challenge button, we have to ask the server devs to include a field for that, as players can disable that other playes can challenge them.

good catch


The button size is better now, but I preferred the filled button style to the outlined one.

HaonRekcef:
image

So I should ask them to include the flag? Or we just add a button to the profile?


> Some user have long names. The username looks better now, but the font should be a little smaller I think.

We could also make the width depend on the size of the name.

veloce:
> So I should ask them to include the flag? Or we just add a button to the profile?

That would be better imo. Also even if we decide later to just show the profile button we can also support a long press context menu to show more options, where challenge would definitely be useful.

> We could also make the width depend on the size of the name.

No, it looks better if all cards have the same width.

HaonRekcef:
Okay so challenging someone that has challenges disabled is actually not a big issue: 

image


Not decided how to best handle long usernames. Maybe we could add a flag to `UserFullNameWidget` that says if the flair and patron or the username is more important and for this widget we would prioritiue the name?
I don't want to increase the width of the card more and decreasing the font size looks bad. What do you think @veloce ?

veloce:
I don't know @HaonRekcef ; maybe we should keep it simple: just decrease the font a bit, optionally play with `letterSpacing` and show an ellipsis if there is an overflow. Some long names won't show entirely but what can we do about it?

lib/src/network/aggregator.dart
```diff
@@ -22,6 +22,7 @@ final Map<Uri, ISet<({String key, RegExp pathRegexp})>> _targetUris = {
     (key: 'challenges', pathRegexp: RegExp(r'^\/api\/challenge$')),
     (key: 'tournaments', pathRegexp: RegExp(r'^\/tournament\/featured$')),
     (key: 'inbox', pathRegexp: RegExp(r'^\/inbox\/unread-count$')),
+    (key: 'friends', pathRegexp: RegExp(r'^\/api\/mobile\/following$')),
```
veloce, line 25: Would be nice to add that in aggregator_test.dart

lib/src/model/relation/relation_repository.dart
```diff
@@ -47,3 +49,27 @@ class RelationRepository {
     await client.postRead(uri);
   }
 }
+
+final followingCarouselProvider = FutureProvider.autoDispose<IList<FollowingUser>>((ref) {
+  return ref.read(mobileFollowingProvider).getFollowing();
```
veloce: This is weird. Why a `ref.read` here? Doing that would cache forever the list, I don't get why we'd do that.

veloce:
@HaonRekcef the PR description is out of date. It mentions a sqlite cache, but I don't see any in the code. Could you update the description? Thanks!

PR #3157, opened 2026-05-10
feat: add "small board" option (again), reduce "small screen" detection threshold

### LINKED ISSUE #3093: Request to bring back small analysis board option (to help see more rows of pgn move list)
Building a use-case:

1. #3090 looks very nice and will aid discoverability. But it does take away one row from the pgn move list (which already is a bit limited, especially if you have engine moves shown up top: see photo below)
2. Some users have been vocal on how information density was better in old app. This would add more space in the analysis window
3. In the past, small board added extra padding on sides. Which would lessen the conflict users would get when they tried to move pieces on the left-file and it would inadvertently try to Backswipe
4. Idea: this part may be overkill but I like to be move list view-heavy for my own analysis board and lean towards seeing a bigger board for broadcasts. Perhaps have a toggle in the top right (where the old tab switcher is being moved from). It could function like a full screen, non-full screen switcher so that one can easily shift between full screen board and small board? Thoughts? 

image

### LINKED ISSUE #3150: Layout of game screen in portrait mode on tablets should be more flexible
The game screen in portrait mode on an iPad should be more flexible. The game board is huge and the section below (moves, analysis) is extremely narrow, almost unreadable. It should be possible to have the section on the right side of the board just like in landscape mode or on the web page. After all, the 13" screen of the iPad Pro is big enough for that.
Image

tom-anders:
- Adjusts the threshold for detecting small screens, since with the new tab bar and the player names in game analysis more vertical space is taken already
- If small screen or tablet is detected, automatically sets the `smallBoard` option to `true` by default

## Screenshots

| | Normal | Small Board |
|---|---|---|
| Phone | Screenshot_1778416232| Screenshot_1778416237 |
| Tablet (Portrait) | Screenshot_1778416272 | Screenshot_1778416278 |

Closes #3150 
Closes #3093

### DIFF
--- a/lib/src/init.dart
+++ b/lib/src/init.dart
@@ -9,13 +9,10 @@ import 'package:lichess_mobile/l10n/l10n.dart';
 import 'package:lichess_mobile/src/binding.dart';
 import 'package:lichess_mobile/src/constants.dart';
 import 'package:lichess_mobile/src/db/secure_storage.dart';
-import 'package:lichess_mobile/src/model/analysis/analysis_preferences.dart';
-import 'package:lichess_mobile/src/model/broadcast/broadcast_preferences.dart';
 import 'package:lichess_mobile/src/model/notifications/notification_service.dart';
 import 'package:lichess_mobile/src/model/notifications/notifications.dart';
 import 'package:lichess_mobile/src/model/settings/board_preferences.dart';
 import 'package:lichess_mobile/src/model/settings/preferences_storage.dart';
-import 'package:lichess_mobile/src/model/study/study_preferences.dart';
 import 'package:lichess_mobile/src/utils/chessboard.dart';
 import 'package:lichess_mobile/src/utils/color_palette.dart';
 import 'package:lichess_mobile/src/utils/screen.dart';
@@ -50,8 +47,6 @@ Future<void> initializeApp() async {
         final boardPrefs = BoardPrefs.defaults.copyWith(boardTheme: BoardTheme.system);
         await prefs.setString(PrefCategory.board.storageKey, jsonEncode(boardPrefs.toJson()));
       }
-
-      _screenSizeBasedInitialization();
 
       _logger.info('First run initialization completed.');
     }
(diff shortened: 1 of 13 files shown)


lib/src/init.dart
```diff
@@ -169,12 +177,18 @@ Future<void> _screenSizeBasedInitialization() async {
...
+  // For tablets in portrait mode using the full board size makes the bottom analysis tabs tiny,
+  // see https://github.com/lichess-org/mobile/issues/3150,
+  // so use a small board there by default as well.
+  final smallBoard = isTablet || isSmallScreen;
+
+  final analysisPrefs = AnalysisPrefs.defaults.copyWith(smallBoard: smallBoard);
   await prefs.setString(PrefCategory.analysis.storageKey, jsonEncode(analysisPrefs.toJson()));
-  final studyPrefs = StudyPrefs.defaults.copyWith(showEngineLines: !isSmallScreen);
+  final studyPrefs = StudyPrefs.defaults.copyWith(smallBoard: smallBoard);
   await prefs.setString(PrefCategory.study.storageKey, jsonEncode(studyPrefs.toJson()));
```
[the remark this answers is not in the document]
tom-anders: intentional, we don't need to hide engine lines by default anymore, as with the small board we have more vertical space

lib/src/init.dart
```diff
@@ -162,19 +157,3 @@ Future<void> androidDisplayInitialization(WidgetsBinding widgetsBinding) async {
...
-
-// Adjusts some settings for small screens based on the MediaQuery data.
-Future<void> _screenSizeBasedInitialization() async {
-  final prefs = LichessBinding.instance.sharedPreferences;
-  final mediaQueryData = MediaQueryData.fromView(
-    WidgetsBinding.instance.platformDispatcher.views.first,
-  );
-  final isSmallScreen = estimateHeightMinusBoard(mediaQueryData) < kSmallHeightMinusBoard;
-
-  final analysisPrefs = AnalysisPrefs.defaults.copyWith(showEngineLines: !isSmallScreen);
```
veloce, line 174: I am not sure we can remove this logic. The small screen detection was initially used for screen ratios where very little vertical space is left. Even with the small board logic, on such screen we may still want to hide the engine lines by default.

tom-anders: alright, I've added back the `showEngineLines` check, with a slightly lower threshold than the "small board" one, to account for the fact that the small board already increases available vertical space

veloce (approved):

Issue #3094, opened 2026-05-02
New App Store Images


cappylot:
<img width="1418" height="538" alt="Image" src="https://github.com/user-attachments/assets/8c7b2991-05fa-4f21-a946-269ea36c4481" />

I am not a professional product designer but I thought I would try my best to create some polished mockups to use for a better looking app store listing. These would all be in the correct size for appstoreconnect and could be used immediately if wanted, but I can also adjust for any wishes. Let me know if this is something that is wanted :)

<img width="1242" height="2688" alt="Image" src="https://github.com/user-attachments/assets/0ad1dea2-024f-426d-91e9-954ead7e75fc" />
<img width="1242" height="2688" alt="Image" src="https://github.com/user-attachments/assets/4286844e-fdda-4fb9-b2f3-7ac1d28bd9ee" />
<img width="1242" height="2688" alt="Image" src="https://github.com/user-attachments/assets/c4106a59-c8c5-4dd5-9bcb-31e0ff6b69bb" />
<img width="1242" height="2688" alt="Image" src="https://github.com/user-attachments/assets/4e56d831-a3f2-4e4d-96a0-6c58f5a07477" />
<img width="1242" height="2688" alt="Image" src="https://github.com/user-attachments/assets/27feea43-c94b-4b55-8c35-8fec5890a61c" />
<img width="1242" height="2688" alt="Image" src="https://github.com/user-attachments/assets/e4cd3d39-f929-4ce7-a1ce-c4e8288e5b12" />

veloce:
Thanks! They look great!

The studies shown are not ideal though. Maybe we could use the study list screen directly and sort them by most popular?

veloce:
Another suggestion: for broadcasts it would be better to use the view in a broadcast with all the boards.

cappylot:
Alright, I try to update by the end of the week.

cappylot:
### Updated studies and broadcast app store images

<img width="1242" height="2688" alt="Image" src="https://github.com/user-attachments/assets/cc75e7fd-1269-4217-9a18-452035a7bee0" />
<img width="1242" height="2688" alt="Image" src="https://github.com/user-attachments/assets/6befcdfc-846a-4c30-ba2d-1e0f7cc76c31" />

Sorry for the wait :)

veloce:
Thank you!

veloce:
[...]

PR #3164, opened 2026-05-11
Add new Images

cappylot:
Added new Appstore-listing images as discussed in #3094


veloce:
Thanks for this @cappylot ; what I meant by sources was the files from the image processing software you are using. Like the `.psd` if it is photoshop.

cappylot:
Ah, sorry. I made them in Figma. I'll check later what the possibilities are.

cappylot:
@veloce There is no such equivalent to a .psd file like in photoshop. I can however download a .fig file which is a local copy of my Figma Project which then has to be imported back into Figma for editing and would work similar enough to how .psd files in photoshop work. Should I do that instead?

veloce:
@cappylot sorry I forgot to reply. Yes sure, you can include the figma file. Thanks!

cappylot:
@veloce Please excuse my very late response. I have added the Figma file. I also left the separate images from before. Should be complete now. Let me know if you need anything else!

PR #3172, opened 2026-05-12
 fix: use spacing parameter instead of custom spacing 

tom-anders:
Came across this by accident while changing other stuff :D

### DIFF
--- a/lib/src/widgets/stat_card.dart
+++ b/lib/src/widgets/stat_card.dart
@@ -1,5 +1,4 @@
 import 'package:flutter/material.dart';
-import 'package:lichess_mobile/src/constants.dart';
 import 'package:lichess_mobile/src/styles/styles.dart';
 
 const _customOpacity = 0.6;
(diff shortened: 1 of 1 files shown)


veloce (commented):
I am not sure this is an improvement. Can you elaborate?

tom-anders:
Hmm looking at it now on the next morning I think it's not worse,but also not really better :D

Looking a bit more at the rest of the code, maybe this whole function is not needed anyway, it effectively adds spacing between the widgets, which should also be achievable via the Rows's spacing parameter

tom-anders:
Yep, the custom logic was unnecessary indeed, using `spacing: 8` looks the same in the puzzle dashboard and the user perf cards. For the storm dashboard, there's a very slight difference, but it's actually an improvement: The boxes (all-time, this month, this week, today) are the exact same size now (it's very subtle, but I measured it in inkscape):

Before:

Screenshot_1778698879_before


After:

Screenshot_1778698906_after

veloce (approved):
Thanks! That logic came from a time when the Row spacing parameter was not yet a thing.

Issue #3168, opened 2026-05-12
Opening Explorer database does not match the current variant


Matrix9494:
### What is the problem?

In the new mobile app, Opening Explorer appears to use the standard chess database even when the current game/analysis variant is non-standard, such as Crazyhouse.

### Steps to reproduce the bug

Opening Explorer should query/display data for the current variant when the selected database supports variants.

For example, in a Crazyhouse position, the Lichess/player explorer data should be Crazyhouse data, not standard chess data.

### App version

0.23.13

### Device Info

Android 14

PR #3173, opened 2026-05-13
Fix variant handling in explorer and tablebase

Matrix9494:
## Summary

Fix variant handling for opening explorer and tablebase requests.  #3168 

This PR makes explorer requests variant-aware so non-standard positions no longer fall back to standard data by accident. It also prevents the Masters database from being shown for variants where the endpoint does not support variant selection.

## Changes

- Pass the current position variant through `OpeningExplorerView` and `openingExplorerProvider`.
- Add the `variant` query parameter for Lichess and player opening explorer requests.
- Keep `fromPosition` mapped to `standard`, since it is an imported/custom standard chess position rather than a real variant.
- Block Masters DB for non-standard variants because the `/masters` endpoint has no variant parameter.
- Route tablebase requests to variant-specific endpoints:
  - `/standard` for standard, from-position, and Chess960
  - `/atomic` for Atomic
  - `/antichess` for Antichess

## Screenshot

Combined screenshot:
1. Standard opening explorer
2. Antichess opening explorer
3. Antichess with Masters DB unavailable message
4. Antichess tablebase

image

### DIFF
--- a/lib/src/view/explorer/opening_explorer_view.dart
+++ b/lib/src/view/explorer/opening_explorer_view.dart
@@ -44,6 +44,8 @@ class OpeningExplorerView extends ConsumerStatefulWidget {
 }
 
 class _OpeningExplorerState extends ConsumerState<OpeningExplorerView> {
+  // Variant is part of the key because the same FEN string can be queried
+  // against different explorer datasets.
   final Map<({String fen, Variant variant, OpeningExplorerPrefs prefs}), OpeningExplorerEntry>
   cache = {};
 
@@ -65,6 +67,8 @@ class _OpeningExplorerState extends ConsumerState<OpeningExplorerView> {
     final prefs = ref.watch(openingExplorerPreferencesProvider);
     final variant = Variant.fromRule(widget.position.rule);
 
+    // The masters endpoint has no variant parameter, so avoid silently showing
+    // standard master games for variant positions.
     if (prefs.db == OpeningDatabase.master && !_isMasterDatabaseAvailableFor(variant)) {
       return const Center(
         child: Padding(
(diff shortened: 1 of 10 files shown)


HaonRekcef:
I think it would be preferable to display the lichess database all the time for variants instead of showing that the master db is not available.

veloce:
> I think it would be preferable to display the lichess database all the time for variants instead of showing that the master db is not available.

Agreed. We don't want to be forced to go to settings to change the DB.

Matrix9494:
Agreed, thanks. I updated the PR so variant positions no longer show the “Masters database is not available” message.

Since mobile currently stores a single selected explorer database, the current behavior is now:

- Standard / from-position + Masters selected: use Masters
- Variant position + Masters selected: temporarily fall back to the Lichess database
- Lichess selected: use Lichess

This fallback does not change the saved preference; it only affects the database used for the current variant position.

Longer term, I agree the better UI would be closer to the LichessWeb, where multiple databases can be shown as separate tabs and unsupported databases can simply be hidden/disabled per variant. That is a larger UI/settings change, so I’d prefer to keep it as a follow-up PR.

lib/src/view/explorer/opening_explorer_view.dart
```diff
@@ -44,7 +44,10 @@ class OpeningExplorerView extends ConsumerStatefulWidget {
 }
 
 class _OpeningExplorerState extends ConsumerState<OpeningExplorerView> {
-  final Map<OpeningExplorerCacheKey, OpeningExplorerEntry> cache = {};
+  // Variant is part of the key because the same FEN string can be queried
+  // against different explorer datasets.
+  final Map<({String fen, Variant variant, OpeningExplorerPrefs prefs}), OpeningExplorerEntry>
```
veloce, line 49: You are now using a record instead of the `freezed` class, which is fine. But I don't see the deletion of that `OpeningExplorerCacheKey` class in the PR. I suppose it is now dead code (it should be) so it needs to be removed.

Matrix9494: Done. The unused code has been removed.

veloce (approved):
LGTM!

PR #3176, opened 2026-05-13
Display the different lines in the variation bar.

### LINKED ISSUE #3069: Improve variation navigation during game analysis on mobile
While analyzing games on mobile, navigating sub-variations is difficult.
When stepping through moves, if a position contains multiple variations, the app defaults to a single main variation without prompting the user to choose another line. This makes exploring alternative lines slow and unintuitive.

Expected Behavior:
When reaching a position with multiple continuations, the app should prompt the user to select which variation to follow, or clearly show available branches to choose from.

Current Behavior:
The app automatically follows one main variation, and accessing other sub-variations requires manually navigating the move tree.

Why This Helps:
This would make analysis faster and improve usability when working with complex positions

HaonRekcef:
closes #3069

Adds a variation bar to the game notation or broadcasts, analysis board and studies.
This is especially useful for analysis or studies with a lot of lines. 
Mirrors the web behavior, except that I put it under the notation, to make it easier to reach.
The mainline is highlighted and if the user simply presses next it will continue on the mainline.

I also added test cases.

Preview:

[Screencast_20260514_000551.webm](https://github.com/user-attachments/assets/edde0ed1-4663-4180-8c82-e37be214cd85)

### DIFF
--- a/lib/src/view/analysis/tree_view.dart
+++ b/lib/src/view/analysis/tree_view.dart
@@ -4,6 +4,7 @@ import 'package:lichess_mobile/src/model/analysis/analysis_controller.dart';
 import 'package:lichess_mobile/src/model/analysis/analysis_preferences.dart';
 import 'package:lichess_mobile/src/view/game/game_result_dialog.dart';
 import 'package:lichess_mobile/src/widgets/pgn.dart';
+import 'package:lichess_mobile/src/widgets/variations_bar.dart';
 
 class AnalysisTreeView extends ConsumerWidget {
   const AnalysisTreeView(this.options);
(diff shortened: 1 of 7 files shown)


lib/src/widgets/variations_bar.dart
```diff
@@ -0,0 +1,97 @@
...
+                      ),
+                    ),
+                    child: Material(
+                      color: isMainline
+                          ? theme.colorScheme.primaryContainer
+                          : theme.colorScheme.secondaryContainer,
+                      child: InkWell(
+                        onTap: () => onJump(currentPath + child.id),
+                        child: Container(
+                          height: 38,
```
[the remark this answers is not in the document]
HaonRekcef: I kept it a bit smaller intentionally, to not take too much space away from the notation (because the variationsBar can take 2 or even 3 lines) but maybe we should update it anyways? @veloce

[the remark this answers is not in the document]
veloce: I would like to see it on a phone screen recording too.

I think it would be better to use the min interactive dimension of material lib indeed.

3 lines seems a lot. It should be a pretty rare case to have that many variations, right? Can we not constraint the number of lines in a way?

Could we have an horizontal scroll here if needed (for edge cases)?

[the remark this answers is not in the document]
HaonRekcef: @veloce 

Recording: 

[Screencast_20260519_160950.webm](https://github.com/user-attachments/assets/d946c5e5-3992-43ac-95f0-63e8b39f69d3)

Yes I think more than  5 or 6 is very rare and that will still fit in one row. And two lines of variations should fit in pretty much all screens so I think it should be fine 

I think a horizontal scroll might not work well, as it is inside of the analysis Tabs that has a similar gesture.

lib/src/view/analysis/tree_view.dart
```diff
@@ -18,34 +19,47 @@ class AnalysisTreeView extends ConsumerWidget {
     final prefs = ref.watch(analysisPreferencesProvider);
     // enable computer analysis takes effect here only if it's a lichess game
     final enableServerAnalysis = !options.isLichessGameAnalysis || prefs.enableServerAnalysis;
+    final currentNode =
```
veloce, line 22: There is already an `analysisState.currentNode`, so why computing it again?

lib/src/view/broadcast/broadcast_game_screen.dart
```diff
@@ -353,24 +354,38 @@ class _BroadcastGameTreeView extends ConsumerWidget {
     final state = ref.watch(ctrlProvider).requireValue;
 
     final broadcastPrefs = ref.watch(broadcastPreferencesProvider);
+    final currentNode = state.root.branchesOn(state.currentPath).lastOrNull ?? state.root;
```
veloce, line 357: Idem: there is already a `currentNode` in broadcast state.

lib/src/view/study/study_tree_view.dart
```diff
@@ -25,29 +26,42 @@ class StudyTreeView extends ConsumerWidget {
...
-                  root: root,
-                  currentPath: studyState.currentPath,
-                  pgnRootComments: studyState.pgnRootComments,
-                  notifier: ref.read(studyControllerProvider(options).notifier),
-                  showTopDivider: showTopDivider,
-                  shouldShowAnnotations: studyPrefs.showAnnotations,
-                  displayMode: studyPrefs.inlineNotation
-                      ? PgnTreeDisplayMode.inlineNotation
-                      : PgnTreeDisplayMode.twoColumn,
+    final currentNode = root.branchesOn(studyState.currentPath).lastOrNull ?? root;
```
veloce, line 29: Idem

veloce:
Simulator Screenshot - iPhone 17 - 2026-05-28 at 15 46 37

In your screen recording (on android) I can see that the mainline variation is highlighted. I cannot see it on iOS though.

lib/src/view/analysis/tree_view.dart
HaonRekcef: @veloce I don't think I can use them as I need the children and analysisCurrentNode doesn't have it.

```dart
 const factory AnalysisCurrentNode({
    required Position position,
    required bool hasChild,
    required bool isRoot,
    SanMove? sanMove,
    Opening? opening,
    ClientEval? eval,
    IList<PgnComment>? lichessAnalysisComments,
    IList<PgnComment>? startingComments,
    IList<PgnComment>? comments,
    IList<int>? nags,
  }) = _AnalysisCurrentNode;
```

HaonRekcef:
> Simulator Screenshot - iPhone 17 - 2026-05-28 at 15 46 37
> 
> In your screen recording (on android) I can see that the mainline variation is highlighted. I cannot see it on iOS though.

It should be different, not sure why it is not: 

```dart
                    child: Material(
                      color: isMainline
                          ? theme.colorScheme.primaryContainer
                          : theme.colorScheme.secondaryContainer,
```

goutham80808:
Thanks @HaonRekcef for solving My primary issue in lichess mobile 👍

PR #3178, opened 2026-05-13
Explicity tell the Play Store we support various screens

### LINKED ISSUE #875: New Lichess app no longer available for Amazon Fire Max 11 tablet
### What is the problem?

I was on the beta and noticed no updates for a while. Turns out you've made a change which means the (sideloaded) Play Store now says "this tablet isn't compatible with this app". It was last updated in June. It was working fine before (I've previously reported bugs one of which was fixed).

Try and install

App installed

### What happened instead?

App not installed

### Device Info

Amazon Fire Max 11

### Operating System Version

os v8.3.2.4

### Additional information

_No response_

ghost:
This might fix #875 and I think there is no real harm in trying.

The Play Store occasionally applies implicit filtering where it hides apps from devices with unusual aspect ratios or similar reasons. I'm setting these as metadata for the Play Store's distribution engine. I know these are technically deprecated but this is an attempt to fix Google Play Store filtering issues and shouldn't change any app behavior.

We can check if @veloce or someone else with access can go to Google Play Console

-> Reach and devices > Device catalog.

Search for "Fire Max 11" or "Titan Pocket."

Check the Status and see if it says "Screen layout" or something like "Missing feature: android.hardware.telephony"

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -1,5 +1,13 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android">
     <uses-permission android:name="android.permission.INTERNET" />
+    <!-- Explicitly define screen support for niche device compatibility -->
+    <supports-screens
+        android:smallScreens="true"
+        android:normalScreens="true"
+        android:largeScreens="true"
+        android:xlargeScreens="true"
+        android:anyDensity="true"
+        android:resizeable="true" />
     <application
         android:label="Lichess"
         android:name="${applicationName}"


veloce:
Are you sure this will do anything @freebeartogoodhome ? Please link the android documentation about this.

ghost:
This was an educated guess based on what I could see about those devices (unusual screens) and from reading https://developer.android.com/guide/topics/manifest/supports-screens-element and https://developer.android.com/google/play/filters (Last updated 2025-07-21)

I am finding out that when a device has an "unusual" aspect ratio, Google Play’s backend sometimes fails to apply the standard API 4+ default rule. Instead, it hits a fallback safety check: "If the developer didn't explicitly guarantee they support non-standard dimensions, hide the app to prevent a bad user experience."

Also, if any third-party SDK mention <supports-screens> then we are possibly getting a restriction on our screen support since we have nothing in our main AndroidManifest.xml.

This was an attempt to override any filtering that Google Play Store is doing that is incorrectly hiding some apps from us. 
If you can check Google Play Console
-> Reach and devices > Device catalog.
Search for "Fire Max 11" or "Titan Pocket."
Check the Status and see if it says "Screen layout" or something like "Missing feature: android.hardware.telephony"

Then we can know if I'm way off track here. That is the only way to be 100% sure, but I felt like this was a good guess as to why based on the fact the two devices in that thread had unusual screens; and based on what I read about how the Google Play Store filters can work differently than we expect.

android/app/src/main/AndroidManifest.xml
```diff
@@ -1,5 +1,13 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android">
     <uses-permission android:name="android.permission.INTERNET" />
+    <!-- Explicitly define screen support for niche device compatibility -->
+    <supports-screens
+        android:smallScreens="true"
+        android:normalScreens="true"
+        android:largeScreens="true"
+        android:xlargeScreens="true"
+        android:anyDensity="true"
```
veloce, line 9: This attribute is `true` by default, so let's drop it.

android/app/src/main/AndroidManifest.xml
```diff
@@ -1,5 +1,13 @@
...
+        android:resizeable="true" />
```
veloce, line 10: This attribute is deprecated so let's drop it.

PR #3185, opened 2026-05-14
automatically expand the current children node

### LINKED ISSUE #2494: Expand plus button as you go through moves
Taken from discord

https://discord.com/channels/280713822073913354/807722604478988348/1445386473871245364

@tom-anders

HaonRekcef:
Automatically expand the children of the currentNode, closes #2494

The perf impact is negligible, as only the current children get looped.

[Screencast_20260514_173011.webm](https://github.com/user-attachments/assets/9438b0e2-687b-45cc-8b47-d6fedddef172)

I implemented for studies and analysis, I do not see a reason to do it for broadcasts.

### DIFF
--- a/lib/src/model/analysis/analysis_controller.dart
+++ b/lib/src/model/analysis/analysis_controller.dart
@@ -484,16 +484,21 @@ class AnalysisController extends AsyncNotifier<AnalysisState>
     }
   }
 
-  void userPrevious() {
-    _setPath(state.requireValue.currentPath.penultimate, isNavigating: true);
+  void userPrevious({bool fastSeek = false}) {
+    _setPath(
+      state.requireValue.currentPath.penultimate,
+      isNavigating: true,
+      keepCollapsed: fastSeek,
+    );
   }
 
-  void userNext() {
+  void userNext({bool fastSeek = false}) {
     final curState = state.requireValue;
     if (!curState.currentNode.hasChild) return;
     _setPath(
       curState.currentPath + _root.nodeAt(curState.currentPath).children.first.id,
       isNavigating: true,
+      keepCollapsed: fastSeek,
     );
   }
 
@@ -681,6 +686,7 @@ class AnalysisController extends AsyncNotifier<AnalysisState>
 
     /// Whether the user is navigating through the moves (as opposed to playing a move).
     bool isNavigating = false,
+    bool keepCollapsed = false,
   }) {
     _currentPath = path;
     final curState = state.requireValue;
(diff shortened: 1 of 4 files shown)


lib/src/model/analysis/analysis_controller.dart
```diff
@@ -714,7 +724,7 @@ class AnalysisController extends AsyncNotifier<AnalysisState>
     // root view is only used to display move list, so we need to
     // recompute the root view only when the nodelist length changes
     // or a variation is hidden/shown
-    final rootView = shouldForceShowVariation || shouldRecomputeRootView
+    final rootView = shouldForceShowVariation || shouldRecomputeRootView || pathWasExpanded
```
veloce, line 716: Recomputing rootView is expensive. Are you sure about the performance implications? Seems like it would impact the fast forward/rewind feature.

IMO this should be disabled if fast forward/rewind and only apply to a single jump.

veloce (commented):
I'm ok in principle for the feature and implementation.

But see comment: it should only be applied if not fast seeking the game.

Also since the asked change would complexity a bit the logic I'd like to have 2 simples tests for this: one that shows it does not expand when fast seeking, and one that shows it does for simple move jump.

Thanks!

HaonRekcef:
I found two cases where nodes are automatically collapsed:
https://github.com/lichess-org/mobile/blob/249010cfcd844a1c66644215ffb309dbd2d17afc/lib/src/model/common/node.dart#L558
and https://github.com/lichess-org/mobile/blob/249010cfcd844a1c66644215ffb309dbd2d17afc/lib/src/model/analysis/server_analysis_service.dart#L214

So I am wondering if we need the fast-forward logic:
As for the first case the collapsed nodes are not direct children of the mainline and fast-forwarding a variation with a lot of sub-variations seems very rare.

It might be a problem for the second case, as direct children of the mainline nodes are collapsed. But this is actually very niche: people requesting a server analysis and then long-pressing through it. A game that is reopened would not have them collapsed because it would use the logic from the first case. Also a normal game would be very light to compute the `rootView` for, as it barely has variations.

So the one possible problem I can see is for huge studies. After requesting a computer analysis, those lines are collapsed and then the user fast-navigates through the moves, `rootView` would need to be recomputed every time.

The only other possibility where this problem appears is when a user manually collapses a lot of variations in a huge study, but tbh this also seems very unlikely.

I think these cases are too rare to deserve their own handling logic in the code. What do you think, @veloce?

veloce:
Sorry for the late response here. I think there are other considerations other than perf @HaonRekcef .

It makes sense to me to expand automatically the node if we are jumping to moves one by one. But if you fast seek to a position it doesn't make sense. It would expand all the nodes, making the whole tree much larger, especially in case of a huge study. I would not expect to see such a behaviour as a user. We have to take the worst scenario into account, event though it is a relatively rare one.

If adding that extra logic would add a lot of complexity, I'd say let's not do it. But it does not seem to be the case, at first sight it should be easy to implement/test. What do you think?

HaonRekcef:
@veloce okay I added logic to prevent auto expanding when fast seeking

veloce (approved):

PR #3189, opened 2026-05-14
replace generic analysisboard header

HaonRekcef:
Replaces the generic "Analysis board" header to display the time control and if the game was rated. Previously this information wasn't visible in the analysis board.

image

image


ijm8710:
Hey I like this but would suggest a few adjustments

I would moreso match the format of web.

Some examples:

Symbol 5+0 • Rated • Chess960
Symbol 5+0 • Casual • Crazyhouse
Symbol 5+0 • Rated • Blitz

- I know you're showing the variant symbol, but if it's a standard game they still precede it with the time control symbol
- If it's not standard variant, the variant type overrides showing the time control name. You still have the time control number regardless

Thoughts?

veloce (approved):
lgtm

Issue #3183, opened 2026-05-14
Access to lichess forum


ijm8710:
Since lichess blog is currently opened simply as an inapp browser for now, could potentially be a quicker hit to do the same here

Matrix9494:
I can take this. Forum access can be added under the Community section. Also, I think it would make sense to add Blog there as well, instead of only exposing it through the home widgets.

PR #3193, opened 2026-05-16
Add forum and blog links to Community

Matrix9494:
#3183 add forum and blog links to Community tab. 

image

### DIFF
--- a/lib/src/view/more/more_tab_screen.dart
+++ b/lib/src/view/more/more_tab_screen.dart
@@ -171,6 +171,7 @@ class _Body extends ConsumerWidget {
                     Navigator.of(context, rootNavigator: true).push(FriendScreen.buildRoute());
                   },
                 ),
+              // Match the home blog carousel behavior by hiding community web content in kid mode.
               if (!isKidMode) ...[
                 ListTile(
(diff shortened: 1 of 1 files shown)


veloce:
I am not 100% sure we want this since it opens in a web browser. Will ask the lichess team what they think about it.

lib/src/view/more/more_tab_screen.dart
```diff
@@ -169,6 +171,25 @@ class _Body extends ConsumerWidget {
                     Navigator.of(context, rootNavigator: true).push(FriendScreen.buildRoute());
                   },
                 ),
+              // Match the home blog carousel behavior by hiding community web content in kid mode.
+              if (!isKidMode) ...[
+                ListTile(
```
veloce, line 176: This opens a web browser, so it should display the appropriate trailing icon to make it clear.

See other parts of the app where this trailing icon is shown.

veloce (commented):
Thanks for this. This is good except the missing icon.

lib/src/view/more/more_tab_screen.dart
Matrix9494: Icon has been changed. 
image

PR #3195, opened 2026-05-16
add tournament description

HaonRekcef:
Adds the tournament description to the tournament screen.

I made it expandable and hidden by default, as tournament descriptions can be long.

[Screencast_20260516_101416.webm](https://github.com/user-attachments/assets/b188c702-f886-4c1c-a358-6ec36450fa42)

Tournament links to test it:
https://lichess.org/tournament/may26wua
https://lichess.org/tournament/apr26lta

PR #3196, opened 2026-05-16
fix markdown line break for broadcast overview

HaonRekcef:
noticed while creating #3195 that in the broadcast overview tab, line breaks are not respected in the markdown:

| Before | After |
|--------|-------|
| image | image |


veloce (approved):
Thanks!

PR #3199, opened 2026-05-16
Feature/server outage screen

### LINKED ISSUE #1016: Better messaging to communicate server outage
First image is mobile web; latter 3 are app

image

image
imageimage

MaartenD:
**UPDATE: How Outage screen is now implemented June 12th 20206**
# **Summary: Outage screen (server status feature)** 

**Goal:** Distinguish between "no internet connection" (the phone has no wifi/data) and "the Lichess server is unreachable" (the phone has internet, but lichess.org is down or in maintenance), and show an appropriate message for each.

## Background
While working on this feature, I asked in the Lichess-Development Discord channel how the web client detects a server-side outage versus a regular connectivity issue. revoof explained:

> "It's a server-side decision. It only works when a frontend server is still up, but the backend isn't. Then the frontend server serves the error page — using HTTP 503 for planned maintenance, and HTTP 502 if the backend is unreachable. So if the mobile app sees one of those response codes on API calls or websocket connections, it can assume it's an issue on Lichess' side, not just a normal connection issue."

This confirmed that checking for HTTP 502/503 responses is the right, server-endorsed way to detect this — rather than relying on connectivity heuristics alone — and is the approach used for this feature.

**Two situations, two messages**
1. No internet connection (network down)
If the phone itself has no wifi/mobile data, the existing "No internet connection" message continues to work as before. Offline features (viewing saved games, playing puzzles offline, etc.) remain available — nothing changes here.

2. Server down (new outage screen)
If the phone does have internet, but the Lichess server is unreachable (outage or maintenance), we show a new, friendly outage screen with:

- The Lichess logo
- A message explaining that the server is currently unreachable
- Links to Mastodon, Bluesky and Discord, so the user can check for known outage announcements
- This screen appears on both the Home tab and the Watch tab.

## How is this detected?
We look at the responses the server sends back to requests from the app:

- A "503" response means: Lichess is in planned maintenance.
- A "502" response means: the server is unreachable (outage).

As soon as the server responds normally again, or the live connection (websocket) recovers after a real interruption, the outage screen disappears automatically.

## Pull-to-refresh
On the outage screen, the user can pull-to-refresh to manually check whether the server is back up — without causing any extra network traffic beyond what the app normally does.

## Respecting offline capabilities
This feature follows the same approach already used elsewhere in the app: whatever works offline keeps working, and anything that requires an online connection becomes non-clickable during a server outage (for example, the "Players/friends" and "Challenges" buttons in the top app bar). This avoids sending the user to a screen that wouldn't be able to load anyway, and the confusing error messages that would result from that.

## Tests
Automated tests were added/updated for this feature, including checks that:

- the outage screen appears when the server is down (on both Home and Watch),
- the "no internet" message keeps working correctly when only the network is down,
- the "Players" button in the top app bar is disabled during a server outage.

##Known limitation
This implementation does not yet cover the case where only the websocket connection is unavailable (while regular HTTP requests still succeed). In that scenario, the user currently gets no specific message about it. This is something i like to discuss further before deciding on the right approach.

For transparency: Claude helped me with the design and implementation of this feature.

## First movie when mobile data / wifi unavailable.

https://github.com/user-attachments/assets/8782ba51-7971-4fdf-925a-bf03d458db93

## Second movie how outage screen is acting. 

I tested this by stopping and starting the lila-1 service of my local Docker container. I needed to refresh i few times at the end of the video.

https://github.com/user-attachments/assets/b6fae908-6718-472e-9196-ab87fced740b

Fixes #1016


HaonRekcef:
@MaartenD There are offline features in the app, so the lichess is down indicator should be less invasive in the UI.

How do we make sure to distinguish between the server actually being down and the player simply being offline or experiencing network issues?

MaartenD:
@HaonRekcef that one i missed. I will do my research and let you know.

MaartenD:
@HaonRekcef is the Over the board game an offline feature? Are there more?

HaonRekcef:
@MaartenD there are multiple. You can disable the network on your device and see which buttons are interactable and not greyed out.

MaartenD:
@HaonRekcef what about this version? This is when in flightmode.

https://github.com/user-attachments/assets/c82b53ba-246c-45e9-b866-009ceb43df2f

~~In the video you will see that Puzzle Themes is available and when you click on it you can't select anything. I tested this on my Iphone against the production version and there the behavior is the same. In my opinion this isn't correct, but i'm not sure.~~

Will upload a version when the websocket connection isn't working later today or tomorrow. I need to figure some things out first.

MaartenD:
**Behaviour during outage **

Below what is working so far. It's still work in progress but before i continue i would like to have somen answers according my approach (see Question below).

**Video when websocket isn't available**

https://github.com/user-attachments/assets/3c9ff03c-10fc-4609-be3d-f11b99305b3b



- Home tab: outage screen is shown, the Play button remains accessible
- Puzzles, Learn, More: accessible and functional as normal (offline features work)
- Watch tab: shows the outage screen instead of "No internet connection" (offline)
- Play button: The options that require a server connection (such as Create lobby game, Challenge a friend, Correspondence and Arena tournaments) were already disabled when the network was unavailable. By replacing onlineStatusProvider with lichessOnlineProvider in the Play button components, these options are now also correctly disabled during a server outage

**A new provider in lib/src/network/lichess_online.dart combines both checks:**
```flutter
final lichessOnlineProvider = Provider.autoDispose<bool>((ref) {
  final isNetworkOnline = ref.watch(onlineStatusProvider).value ?? false;
  final isServerReachable = ref.watch(serverStatusProvider);
  return isNetworkOnline && isServerReachable;
});
```

**Currently applied to play_menu.dart, quick_game_matrix.dart and create_game_widget.dart.** There are other places in the codebase that still use onlineStatusProvider directly, these would benefit from the same treatment. 

**Question:** to fully implement this feature i would like to know if you agree with my approach. Additionally, would you prefer lichessOnlineProvider to live in connectivity.dart alongside onlineStatusProvider rather than in a separate file?

**Tests added / updated** 

- test/network/server_status_test.dart (new): two tests using fakeAsync verifying the 30-second timer fires after continuous disconnection, and that the timer is cancelled when the connection is restored within 30 seconds
- test/view/home/home_tab_screen_test.dart (updated):
outage page shown → extended to also verify the Play button (FloatingActionButton) remains visible
New test: Watch tab shows ServerOutage when offline

HaonRekcef:
Hi @MaartenD thanks for the work on this!
Correct me if I am wrong, but looking at the code, it seems the "Lichess is undergoing Maintenance" screen will still show up if the user simply loses their local internet connection (at least on the Watch tab).
Actually a local network drop is way more common than an actual server outage.
 I think the cleanest way to handle this is by introducing an enum for the connection state rather than relying on booleans:
```dart
enum ConnectionStatus {
  online,
  networkDown,
  serverDown,
} 
```
To answer your question: 
_Question: to fully implement this feature i would like to know if you agree with my approach. Additionally, would you prefer lichessOnlineProvider to live in connectivity.dart alongside onlineStatusProvider rather than in a separate file?_

I am not the final authority, but I would say it doesn't matter much as long as the code is well written and works, personally I would tend towards putting it into the same file.

MaartenD:
Hi @HaonRekcef,

Thanks for your reply and great suggestion according to the ConnectionStatus enum.  I myself was also leaning towards putting everything in the same file. I will take that route.

MaartenD:
**Reason for change**
The initial implementation used two separate booleans (onlineStatusProvider and serverStatusProvider) to determine whether the app could reach lichess. This made it impossible to distinguish between a local network drop and an actual server outage. Both situations were treated the same way. @HaonRekcef correctly pointed out that showing "Lichess is undergoing technical difficulties" when the user simply has no internet is misleading.

**New approach: ConnectionStatus enum**
Added to connectivity.dart:

```flutter
enum ConnectionStatus {
  online,       // network available and server reachable
  networkDown,  // no network connection
  serverDown,   // network available but lichess server unreachable
}
```

**Behaviour per status**
|Status | Home tab | Watch tab | Play button online options| 
|---- | ---- | ---- | ----|
|online | normal | normal | enabled|
|networkDown | normal + offline banner | "No internet connection." | disabled|
|serverDown | outage screen | outage screen | disabled|


**Tests**
- test/network/connectivity_test.dart (new) — three unit tests covering all three enum values:

1. Network available + server reachable → online
2. Network unavailable → networkDown
3. Network available + server unreachable → serverDown

- test/view/home/home_tab_screen_test.dart (updated):

1. Watch tab shows no internet message when network is down → verifies "No internet connection." text for networkDown
2. Watch tab shows outage screen when server is down → verifies ServerOutage widget for serverDown

- Final implementation of Offline scenario (no internet connection)

https://github.com/user-attachments/assets/3234b209-b7d6-433c-bdb2-05a205ce41e3

- Final implementation of Server outage scenario

https://github.com/user-attachments/assets/aab162da-dc14-4d70-9008-75e0fa0888a3

To be clear: given the scope of this PR, I used Claude as an AI assistant during development (final implementation).

veloce:
@MaartenD I have not read the comments but only the PR description (which I hope you updated based on the last code).

I have not read the code either, but based on the description I don't see how this can work. How do you distinguish a server outage from a network disconnection? Even if the socket is disconnected for more than 30s, that does not mean the lichess WS server is down.

We certainly don't want to display a message indicating that the lichess server is down if that is not the case. And I don't see how you can know that by just monitoring the WS connection.

I am pretty sure this feature cannot be implemented as is, or am I missing something?

I invite you to reach out to the lichess server devs on discord to see how this is implemented in the website.

MaartenD:
@veloce i updated the PR description and pushed my latest version of this outage screen feature. I have seen the conflicts and will check that out. Would love to hear what you think about this version.

lib/src/network/http.dart
```diff
@@ -107,6 +108,9 @@ final httpClientFactoryProvider = Provider<HttpClientFactory>((Ref ref) {
             responseCode: response.statusCode,
             responseDateTime: DateTime.now(),
           );
+          ref
```
veloce, line 111: I would not put that in the global http client factory. This can be used to create clients that target other URIs than the lichess main server.

There is already a `LichessClient`, this logic belong here.

lib/src/network/server_status.dart
```diff
@@ -0,0 +1,54 @@
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+import 'package:lichess_mobile/src/constants.dart';
+import 'package:lichess_mobile/src/network/socket.dart';
+import 'package:logging/logging.dart';
+
+final _logger = Logger('ServerStatus');
+
+final serverStatusProvider = NotifierProvider<ServerStatusNotifier, bool>(ServerStatusNotifier.new);
```
veloce, line 8: Provider name is missing.

lib/src/network/server_status.dart
```diff
@@ -0,0 +1,54 @@
...
+
+class ServerStatusNotifier extends Notifier<bool> {
```
veloce, line 10: Add doc comment to explain the purpose of this notifier.

lib/src/network/server_status.dart
```diff
@@ -0,0 +1,54 @@
...
+  @override
+  bool build() {
+    final pool = ref.watch(socketPoolProvider);
+    _lastLag = pool.averageLag.value;
+    pool.averageLag.addListener(_onLagChange);
+    ref.onDispose(() => pool.averageLag.removeListener(_onLagChange));
+    return true;
+  }
+
+  void _onLagChange() {
```
veloce, line 22: There is a flaw here, you're checking only the lag change and not whether an http request previously returned an error code.

That being said, I don't think that listening to the socket is the proper thing to do at all.

Lichess down is detected when a frontend server returns a 502/503. There is a separate websocket server that communicates with the lila instance through redis. I assume the WS server can be up when the lichess backend is down, so the connected Websocket does not mean a game can be played.

Thus we should rely only on http status code change from a lichess URI. cc @ornicar and @niklasf , do you confirm?

lib/src/network/connectivity.dart
```diff
@@ -23,6 +24,30 @@ final onlineStatusProvider = FutureProvider.autoDispose<bool>((ref) {
   return ref.watch(connectivityChangesProvider.selectAsync((status) => status.isOnline));
 }, name: 'OnlineStatusProvider');
 
+/// Represents the connection state of the app with respect to the lichess server.
+enum ConnectionStatus {
```
veloce, line 28: To remove any ambiguity, it would better be called `LichessConnectionStatus`.

lib/src/network/connectivity.dart
```diff
@@ -23,6 +24,30 @@ final onlineStatusProvider = FutureProvider.autoDispose<bool>((ref) {
...
+
+  /// The device is online but the lichess server is unreachable.
+  serverDown,
+}
+
+/// A provider that exposes the current [ConnectionStatus].
+///
+/// Use this instead of [onlineStatusProvider] when the distinction between a
+/// local network drop and a server outage matters for the UI.
+final connectionStatusProvider = Provider.autoDispose<ConnectionStatus>((ref) {
```
veloce, line 43: Rename it `lichessConnectionStatusProvider`.

lib/src/view/home/server_outage.dart
```diff
@@ -0,0 +1,74 @@
+import 'package:flutter/material.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+import 'package:lichess_mobile/src/constants.dart';
+import 'package:lichess_mobile/src/styles/styles.dart';
+import 'package:lichess_mobile/src/utils/l10n_context.dart';
+import 'package:url_launcher/url_launcher.dart';
+
+class ServerOutage extends ConsumerWidget {
```
veloce, line 8: I would name this `ServerOutageDisplay` or sth like that.

lib/src/view/home/server_outage.dart
```diff
@@ -0,0 +1,74 @@
...
+        ? 'assets/images/logo-white.png'
+        : 'assets/images/logo-black.png';
+    final mastodonLogo = brightness == Brightness.dark
+        ? 'assets/images/mastodon-logo-purple.png'
+        : 'assets/images/mastodon-logo-black.png';
+    final discordLogo = brightness == Brightness.dark
+        ? 'assets/images/discord-logo-white.png'
+        : 'assets/images/discord-logo-black.png';
+
+    return ColoredBox(
```
veloce, line 24: I don't get the purpose of that `ColoredBox`. This content is meant to be shown directly and the current scaffold already handles the background.

lib/src/view/home/server_outage.dart
```diff
@@ -0,0 +1,74 @@
...
+      color: Theme.of(context).scaffoldBackgroundColor,
+      child: LayoutBuilder(
+        builder: (context, constraints) {
+          return SingleChildScrollView(
+            physics: const AlwaysScrollableScrollPhysics(),
```
veloce, line 29: Not sure why it is needed. If the content overflows the single child scroll viewport, scrolling should work out of the box and otherwise we don't have to see the scrollbar.

lib/src/view/home/server_outage.dart
```diff
@@ -0,0 +1,74 @@
...
+              constraints: BoxConstraints(minHeight: constraints.maxHeight),
+              child: Padding(
+                padding: Styles.bodyPadding,
+                child: Center(
+                  child: Column(
+                    mainAxisAlignment: MainAxisAlignment.center,
+                    children: [
+                      Image.asset(logo, width: 150),
+                      const SizedBox(height: 16),
+                      Text(context.l10n.mobileServerOutageMessage, textAlign: TextAlign.center),
```
veloce, line 40: Let's put a larger font here, like `bodyLarge`;

lib/src/view/home/server_outage.dart
```diff
@@ -0,0 +1,74 @@
...
+                      const SizedBox(height: 16),
+                      Text(
+                        context.l10n.mobileServerOutageKeepInformed,
```
veloce, line 43: same larger font

lib/src/view/more/more_tab_screen.dart
```diff
@@ -104,7 +104,7 @@ class _Body extends ConsumerWidget {
                     ? const CupertinoListTileChevron()
                     : null,
                 title: Text(context.l10n.openingExplorer),
-                enabled: isOnline,
+                enabled: connectionStatus == ConnectionStatus.online,
```
veloce, line 107: Lichess main server may be down while the opening explorer is still available. So you should really use the regular online status provider.

lib/src/network/connectivity.dart
```diff
@@ -23,6 +24,30 @@ final onlineStatusProvider = FutureProvider.autoDispose<bool>((ref) {
...
+/// Represents the connection state of the app with respect to the lichess server.
+enum ConnectionStatus {
+  /// The device is online and the lichess server is reachable.
+  online,
+
+  /// The device has no network connection.
+  networkDown,
+
+  /// The device is online but the lichess server is unreachable.
+  serverDown,
```
veloce, line 36: There is no distinction between the server being in planned maintenance and the server being down; we should add it since the status code can tell us that.

lib/src/view/home/server_outage.dart
veloce, line 40: We should probably display a different message if the server is in maintenance mode. No need to translate it for now, especially if there is not translation already available server side.

For the maintenance mode, it would make sense if the http response would contain the datetime when this maintenance is supposed to end.

Now, this would be the ideal. But I don't know whether the website makes this distinction, and if it does not we should probably do the same.
Can you please reach out to the server team and keep me informed on the 2 questions raised here? (maintenance date for 503 and whether to show a different message).

translation/source/mobile.xml
```diff
@@ -90,5 +90,7 @@
   <string name="challengeCreated" comment="Shown as a bottom banner when another player has been challenged.">Challenge created: You will be notified when the game starts.\nYou can access it from the home tab.</string>
   <string name="previousPage" comment="Shows the previous page, e.g. in tournament standings">Previous</string>
   <string name="orImportPgnFile" comment="Button text to import a PGN file from the device">Or import a PGN file</string>
+  <string name="serverOutageMessage" comment="Shown on the home screen when the Lichess server is unreachable.">Lichess is undergoing technical difficulties. We&apos;re doing everything we can, and expect to be back up very soon.</string>
```
veloce, line 93: Have you checked that these translations are not already available server side?

If they are we should use them. If they are not, I'd rather not translate the mobile part yet. See contributing guide for the explanation why we don't translate immediately the new strings.

veloce (commented):
Thanks for your work on this and the detailed description.

I made a lot of comments, and some are not just about the code as I have more questions about this feature and how it is handled in the website.

veloce:
Closing as it is continued here: https://github.com/lichess-org/mobile/pull/3488

PR #3202, opened 2026-05-16
refactor: replace connectivity Debouncer with Throttler in ConnectivityChangesNotifier

### LINKED ISSUE #3201: ConnectivityChangesNotifier: debouncer delays reconnection by 5 seconds
## Problem

`ConnectivityChangesNotifier` uses a `Debouncer` with a 5-second delay to avoid reacting to rapid network fluctuations:

```dart
final _connectivityChangesDebouncer = Debouncer(const Duration(seconds: 5));

_connectivitySubscription = _connectivity.onConnectivityChanged.listen((result) {
  _connectivityChangesDebouncer(() => _onConnectivityChange(result));
});
```

A debouncer resets its timer on every incoming event, meaning **every** connectivity change — including a clean, single reconnection — is delayed by 5 seconds. In practice this means:

- User loses connection briefly and reconnects → the app waits 5 full seconds before resuming socket connections, syncing correspondence moves, or retrying any pending requests.
- Even a single, isolated reconnection event gets delayed, because the debouncer cannot distinguish between a burst and a single event.

## Expected behavior

The first event in a burst should be delivered immediately. Subsequent events within the same window should be suppressed to protect against rapid fluctuations (e.g. WiFi → mobile → WiFi in quick succession).

## Proposed fix

Replace the `Debouncer` with the existing `Throttler` from `lib/src/utils/rate_limit.dart`. A throttler delivers the first event immediately and suppresses duplicates within the time window — which is exactly the desired behavior here.

The `Throttler` class is already in the codebase, so no new dependencies are needed.

YairAndrade1:
Closes #3201

## Motivation

The current implementation uses a `Debouncer` with a 5-second delay to handle rapid connectivity changes. A debouncer delays every event in a burst, including the first one. This means that when the user reconnects after a brief disconnection, the app waits a full 5 seconds before reacting and resuming socket connections or syncing correspondence moves, even when the network is already stable.

## Change

Replace the `Debouncer` with the existing `Throttler` from `lib/src/utils/rate_limit.dart`. A throttler delivers the first event in a burst immediately and suppresses subsequent events within the same time window. This preserves the protection against rapid network fluctuations while eliminating the unnecessary delay on the first reconnection event.

The throttle delay is extracted into a named constant (`kConnectivityThrottleDelay`) to make it configurable in tests without relying on real clock waits.

## Behavior difference

| Scenario | Before (Debouncer) | After (Throttler) |
|---|---|---|
| Single reconnection event | Delayed 5 seconds | Delivered immediately |
| Rapid network fluctuation (WiFi → mobile → WiFi) | Delivered after 5s silence | First event delivered, rest suppressed |
| Test clock dependency | Requires real clock waits | Configurable via constant |

## Testing

Existing tests pass. No new dependencies introduced. `Throttler` already exists in `lib/src/utils/rate_limit.dart`.

---

> **Note:** This contribution was made as part of a mobile application development course. The change was identified through a micro-optimization audit of the codebase, where the behavioral difference between `Debouncer` and `Throttler` was analyzed as a real-world example of rate-limiting patterns in a production Flutter app.

### DIFF
--- a/lib/src/network/connectivity.dart
+++ b/lib/src/network/connectivity.dart
@@ -11,6 +11,8 @@ import 'package:logging/logging.dart';
 
 final _logger = Logger('Connectivity');
 
+const kConnectivityThrottleDelay = Duration(seconds: 5);
+
 /// A provider that exposes a [Connectivity] instance.
 final connectivityPluginProvider = Provider<Connectivity>((Ref _) => Connectivity());
 
@@ -39,7 +41,7 @@ class ConnectivityChangesNotifier extends AsyncNotifier<ConnectivityStatus> {
   StreamSubscription<List<ConnectivityResult>>? _connectivitySubscription;
   AppLifecycleListener? _appLifecycleListener;
 
-  final _connectivityChangesDebouncer = Debouncer(const Duration(seconds: 5));
+  final _connectivityChangesThrottler = Throttler(kConnectivityThrottleDelay);
 
   Client get _defaultClient => ref.read(defaultClientProvider);
   Connectivity get _connectivity => ref.read(connectivityPluginProvider);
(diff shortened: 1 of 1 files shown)


veloce (approved):
Good catch! Thank you.

Issue #3204, opened 2026-05-17
add support for deeplinks to a player in tournament results


FitzgeraldKrudde:
Since [#19472](https://github.com/lichess-org/lila/issues/19472) web supports deeplink to a player in tournament results.

It would be great if the App also would support this.

As a reference open this URL in web: 
https://lichess.org/tournament/spring26?player=realcyberbird

HaonRekcef:
closed by #3217

FitzgeraldKrudde:
Very nice, thanks!

PR #3217, opened 2026-05-18
Deeplink support for player stats in tournament

fahadahmed32:
### **Summary**

- Extend tournament deeplinks to support an optional ?player=<username> query parameter (e.g. https://lichess.org/tournament/spring26?player=realcyberbird)
- When the parameter is present, the tournament screen opens and immediately shows the player details bottom sheet for that user
- No behavior change for plain tournament links without the player param

[https://github.com/lichess-org/mobile/issues/3204](https://github.com/lichess-org/mobile/issues/3204)

https://github.com/user-attachments/assets/0da91fe3-3ab0-4559-948b-69e076674538

### DIFF
--- a/lib/src/app_links_service.dart
+++ b/lib/src/app_links_service.dart
@@ -157,7 +157,9 @@ class AppLinksService {
         }
       case 'tournament':
         final tournamentId = TournamentId(appLinkUri.pathSegments[1]);
-        return [TournamentScreen.buildRoute(tournamentId)];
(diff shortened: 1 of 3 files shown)


veloce (approved):
Looks good. Thank you!

ijm8710:
I "think" on web the Lila pr made it so that it scrolls to the page where the player can be found as well. Do you think we should do that as well? In this case page 2

FitzgeraldKrudde:
Very nice, thanks!

PR #3218, opened 2026-05-19
fix some leaks and minor changes

ghost:
`lib/src/view/account/edit_profile_screen.dart`
fixes a potential memory leak
Linked the `focusNode` from `Autocomplete<String>` to `_countryFocusNode`
Added `_countryFocusNode.dispose()`

`lib/src/view/message/contacts_screen.dart`
added `_controller.dispose();`
added `_focusNode.dispose();`
Used Riverpod's built-in `ref.debounce()` instead of a manual debouncer that we also have to cancel later.

`lib/src/view/user/player_screen.dart`
Ensure we really stop watching our friends.
There is a potential race condition if the user pops the `PlayerScreen` off the navigation stack and then the `(context.mounted == false)` and we don't stop watching friends.

`lib/src/view/user/search_screen.dart`
fixed typo in the variable name `_kSaveHistoryDebounceTimer`
added `_saveHistoryDebouncer.cancel`
removed a `ref.read` because it is redundant when we are already watching

`lib/src/widgets/emoji_picker/emoji_picker.dart`
added `_debouncer.cancel();`

`lib/src/widgets/platform_search_bar.dart`
Fixes a potential memory leak. When a parent widget rebuilds and passes a new `TextEditingController` the old controller still holds a reference to `_onTextChanged`. 
`didUpdateWidget` swaps the listener to the new controller.

`lib/src/view/broadcast/broadcast_game_screen.dart`
fixes a memory leak in the `BroadcastGameScreen` by replacing `TapGestureRecognizer` with `GestureDetector`
`TapGestureRecognizer` instances inside` TextSpan` must be manually disposed of to prevent memory leaks. They were never being cleaned up.
Removed `import 'package:flutter/gestures.dart';` since it's no longer used,
Made sure the broadcast pgn text still looks fine.

screenshotbroadcastpgn

### DIFF
--- a/lib/src/view/account/edit_profile_screen.dart
+++ b/lib/src/view/account/edit_profile_screen.dart
@@ -34,8 +34,8 @@ class _EditProfileScreenState extends ConsumerState<EditProfileScreen> {
     'flag': null,
     'location': null,
     'bio': null,
+    'flair': null,
     'realName': null,
-    'lastName': null,
     'fideRating': null,
     'uscfRating': null,
     'ecfRating': null,
@@ -154,9 +154,16 @@ class _EditProfileFormState extends ConsumerState<_EditProfileForm> {
 
   final FocusNode _countryFocusNode = FocusNode();
 
+  late final TextEditingController _countryController = TextEditingController(
+    text: widget.user.profile?.country != null
+        ? (countries[widget.user.profile!.country] ?? '')
+        : '',
+  );
+
   @override
   void dispose() {
     _countryFocusNode.dispose();
+    _countryController.dispose();
     super.dispose();
   }
 
@@ -315,9 +322,7 @@ class _EditProfileFormState extends ConsumerState<_EditProfileForm> {
                     const SizedBox(height: 6.0),
                     Autocomplete<String>(
                       focusNode: _countryFocusNode,
-                      initialValue: field.value != null
-                          ? TextEditingValue(text: countries[field.value]!)
-                          : null,
+                      textEditingController: _countryController,
                       optionsBuilder: (TextEditingValue value) {
                         if (value.text.isEmpty) {
                           return const Iterable<String>.empty();
--- a/lib/src/view/account/edit_profile_screen.dart
+++ b/lib/src/view/account/edit_profile_screen.dart
@@ -152,6 +152,14 @@ class _EditProfileForm extends ConsumerStatefulWidget {
 class _EditProfileFormState extends ConsumerState<_EditProfileForm> {
   Future<void>? _pendingSaveProfile;
 
+  final FocusNode _countryFocusNode = FocusNode();
+
+  @override
+  void dispose() {
+    _countryFocusNode.dispose();
+    super.dispose();
+  }
+
   @override
   Widget build(BuildContext context) {
     final String? initialLinks = widget.user.profile?.links?.map((e) => e.url).join('\r\n');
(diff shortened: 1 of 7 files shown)


lib/src/view/account/edit_profile_screen.dart
```diff
@@ -306,6 +314,7 @@ class _EditProfileFormState extends ConsumerState<_EditProfileForm> {
                     Text(context.l10n.countryRegion, style: Styles.formLabel),
                     const SizedBox(height: 6.0),
                     Autocomplete<String>(
+                      focusNode: _countryFocusNode,
```
veloce, line 324: I don't see how this could be a memory leak since the argument was not provided.

Do you assume that internally, Flutter does not dispose the focus node? While it is possible, I doubt it.

lib/src/view/broadcast/broadcast_game_screen.dart
```diff
@@ -445,13 +444,9 @@ class _PgnTagsView extends ConsumerWidget {
...
                         final url = tag.isLink ? tag.buildUrl(value) : null;
                         if (url != null) {
-                          return RichText(
-                            text: TextSpan(
-                              text: value,
-                              style: Styles.linkStyle,
-                              recognizer: TapGestureRecognizer()
-                                ..onTap = () => launchUrlString(url),
-                            ),
+                          return GestureDetector(
```
veloce, line 447: Please explain this change, I don't see why it is better.

ghost: This was because `GestureDetector` manages its own disposal and we were not disposing the `TapGestureRecognizer`.  If the clicking behavior is going to be worse then I didn't know that. The display worked fine for me and I clicked on everything fine in my testing.

lib/src/view/account/edit_profile_screen.dart
ghost: I'll look into that. That was my understanding but I will check everything out to be 100% sure.

ghost: Bad news first:
That code ran, analyzed, tested, etc. but apparently I didn't remember to go to the profile screen and attempt to change the country. 😩  Right now it will give:

```
══╡ EXCEPTION CAUGHT BY WIDGETS LIBRARY ╞═══════════════════════════════════════════════════════════
The following assertion was thrown building Autocomplete<String>(dirty):
'package:flutter/src/widgets/autocomplete.dart': Failed assertion: line 191 pos 15: '(focusNode ==
null) == (textEditingController == null)': is not true.

Either the assertion indicates an error in the framework itself, or we should provide substantially
more information in this error message to help you determine and fix the underlying cause.
In either case, please report this assertion by filing a bug on GitHub:
  https://github.com/flutter/flutter/issues/new?template=02_bug.yml
```

What I read about the old code was right though. I stress tested it and can see it leaking.
stress_test

I fixed the assertion and did a memory test with the new one. It is showing a delta of 0 now.

ghost:
I found [[RawAutocomplete] Remove assert that don't seem to be needed](https://github.com/flutter/flutter/issues/134890) from 2023 so I guess I don't need to tell them about that.

lib/src/view/account/edit_profile_screen.dart
```diff
@@ -34,8 +34,8 @@ class _EditProfileScreenState extends ConsumerState<EditProfileScreen> {
     'flag': null,
     'location': null,
     'bio': null,
+    'flair': null,
     'realName': null,
-    'lastName': null,
     'fideRating': null,
```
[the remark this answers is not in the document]
ghost: I set a flair with this build, and cleared the flair too. It worked for me.

Now I'm really, really taking a break.

veloce:
Closing because it does too many things and I'm not confident enough to merge this.

PR #3220, opened 2026-05-19
Compress and add new feds

SergioGlorias:
and remove W.png due is not a real fide (only link to RUS and BRL in Lila)


veloce:
Thanks!

PR #3223, opened 2026-05-19
Study pref study order

### LINKED ISSUE #3198: Remember study sorting preference in the mobile app
I would like the sorting preference for Studies in the mobile app to be remembered.

At the moment, the Studies list always opens with “Hot”, and I have to change the sorting manually every time I start the app. It would be much more convenient if the app kept the last selected sorting option and restored it on the next launch.

Expected behavior:

The app should save the last chosen Studies sorting option
On the next app start, the same sorting should be applied automatically

Current behavior:

The Studies section always starts with “Hot”
The sorting has to be changed manually every time.

Why this would be useful:

It improves usability
It saves time for users who always prefer a different default sorting
It makes the app feel more consistent and personalized

nirajk14:
Study order changing also saves in prefs.
Closes #3198 .

### DIFF
--- a/lib/src/view/study/study_list_screen.dart
+++ b/lib/src/view/study/study_list_screen.dart
@@ -5,6 +5,7 @@ import 'package:lichess_mobile/src/model/auth/auth_controller.dart';
 import 'package:lichess_mobile/src/model/study/study.dart';
 import 'package:lichess_mobile/src/model/study/study_filter.dart';
 import 'package:lichess_mobile/src/model/study/study_list_paginator.dart';
+import 'package:lichess_mobile/src/model/study/study_preferences.dart';
 import 'package:lichess_mobile/src/styles/lichess_icons.dart';
 import 'package:lichess_mobile/src/styles/styles.dart';
 import 'package:lichess_mobile/src/utils/l10n.dart';
@@ -37,7 +38,7 @@ class StudyListScreen extends ConsumerStatefulWidget {
 
 class _StudyListScreenState extends ConsumerState<StudyListScreen> {
   late StudyCategory category;
-  StudyListOrder order = StudyListOrder.hot;
+  late StudyListOrder order;
 
   String? search;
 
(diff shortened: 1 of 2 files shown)


lib/src/view/learn/learn_tab_screen.dart
```diff
@@ -18,11 +19,13 @@ import 'package:lichess_mobile/src/widgets/list.dart';
 import 'package:lichess_mobile/src/widgets/platform.dart';
 import 'package:material_symbols_icons/symbols.dart';
 
-final _hotStudiesProvider = FutureProvider.autoDispose<IList<StudyPageItem>>((Ref ref) {
+final _studiesProvider = FutureProvider.autoDispose<IList<StudyPageItem>>((Ref ref) {
```
veloce: While I think it is useful to persist the pref in the study list screen, here I think we should keep showing the hot studies.

lib/src/view/study/study_list_screen.dart
```diff
@@ -135,30 +137,42 @@ class _StudyListScreenState extends ConsumerState<StudyListScreen> {
...
+                },
               ),
               ContextMenuAction(
                 icon: order == StudyListOrder.popular ? Icons.check : null,
                 label: context.l10n.studyMostPopular,
-                onPressed: () => setState(() {
-                  order = StudyListOrder.popular;
-                }),
+                onPressed: () {
+                  setState(() {
```
veloce: I'd rather have a private method `_onStudyOrderChange` than duplicating several times the logic.

veloce (commented):
Sorry I still have a requested change that I missed in the previous review.

lib/src/view/study/study_list_screen.dart
nirajk14: working on it.

veloce (approved):

veloce:
Thanks!

PR #3233, opened 2026-05-22
Fix zen mode toggle not persisting across games

### LINKED ISSUE #3226: Zen mode toggle resets between games when account preference is "In Game Only"
### What is the problem?

When my account zen mode preference is set to "In Game Only", toggling zen mode off during a game doesn't carry over to the next game. Each new game starts with zen mode ON again, forcing me to turn it off every single time.

I expected that once I disable zen mode mid-game, it would either remember my preference for the next game, or at least stay off for the remainder of that game session. Instead, every new game resets to the lichess.org website state.

1. Set zen mode preference to "In Game Only" on lichess.org
2. Start a game on mobile — zen mode is ON (synced from account)
3. During the game, open settings and toggle zen mode OFF
4. Finish the game
5. Start a new game
6. Still the zen mode is turned on

### App version

0.23.14

### Device Info

Realme 9 pro plus / Android 14

### Additional information

Technical details from codebase analysis:
- The in-game zen toggle (zenModeGameSetting) is stored as a local, ephemeral override in the GameState and is not persisted or synced to the server.
- Each new game creates a fresh GameController instance, which initializes zenModeGameSetting to null, causing it to fall back to the account preference (Zen.gameAuto).
- The toggle in game settings (game_settings.dart) always appears regardless of the account preference, unlike the move confirmation and auto-queen toggles which only show when relevant.
- The toggle switch callback ignores the actual value parameter and always calls toggleZenMode(), which could cause state mismatches.
- When a game finishes, zen mode immediately deactivates for Zen.gameAuto users (even if they had it toggled on), because the isZenModeActive getter only checks for Zen.yes on finished games.

goutham80808:
Fixes #3226
When the account zen mode preference is "In Game Only" (`Zen.gameAuto`), toggling zen mode off during a game would reset on every new game. The toggle was stored as an ephemeral `zenModeGameSetting` field on `GameState`, which was discarded when the `GameController` was recreated for each new game.
## Changes
- `toggleZenMode()` now syncs to the server via `accountPreferencesProvider.setZen()`, making the server the single source of truth
- Removed the `zenModeGameSetting` field from `GameState`
- `isZenModeEnabled` simplified to a pure switch on `game.prefs.zenMode`
- `isZenModeActive` unchanged — still correctly deactivates zen for `Zen.gameAuto` users when the game finishes
## Testing
- 10 new unit tests covering `isZenModeEnabled` and `isZenModeActive` for all Zen values
- All 926 tests pass
- `flutter analyze` — no issues
- `flutter test` — passes
- `dart format` — clean

No manual edits to app_en.arb (no translations touched)
Code follows the style guide


test/model/game/game_preferences_test.dart
```diff
@@ -0,0 +1,158 @@
...
+import 'package:lichess_mobile/src/model/common/perf.dart';
+import 'package:lichess_mobile/src/model/common/speed.dart';
+import 'package:lichess_mobile/src/model/game/game.dart';
+import 'package:lichess_mobile/src/model/game/game_controller.dart';
+import 'package:lichess_mobile/src/model/game/game_status.dart';
+import 'package:lichess_mobile/src/model/game/playable_game.dart';
+import 'package:lichess_mobile/src/model/game/player.dart';
+
+void main() {
+  group('GameState.isZenModeEnabled', () {
```
veloce: This test game state (so game controller) but you put these in a game_preferences_test.dart.

This will be a little harder to setup but I'd like to see these tests in game_screen_test.dart instead. Create a new test group in that big file.

You can access the controller to make asserts as you do here. Or you can find widgets to makes assertions on the presence/absence of widgets to verify zen mode is working as intended (preferred method).

goutham80808:
Tests moved from game_preferences_test.dart into game_screen_test.dart as suggested, including a new toggleZenMode switches zen off and on test that exercises the full toggle cycle through the controller.
- Added _initialZenPref to capture the actual server preference at game load. Previously, toggling zen back on always went to Zen.gameAuto. If your account had Zen.yes, you'd get downgraded to "In Game Only" — which is technically correct in some sense, but not what the user intends to. Now it restores whatever the server said.

veloce (approved):
All good! Thank you.

PR #3235, opened 2026-05-22
small refactor of move_feedback

ghost:
`model/common/service/move_feedback`
I refactored to remove the `Ref` dependency from `MoveFeedbackService` by shifting the `boardPreferencesProvider` watch logic directly into the provider. `MoveFeedbackService` now operates as a pure class accepting a `hasHapticFeedback` boolean.


veloce:
@freebeartogoodhome here's a little reminder: please don't mix several unrelated subjects in one PR. One PR -> one purpose, one change.

Thanks!

ghost:
> @freebeartogoodhome here's a little reminder: please don't mix several unrelated subjects in one PR. One PR -> one purpose, one change.
> 
> Thanks!

Sounds good, this was way back in May so I've been a bit better on that. But I don't mind a reminder.

ghost:
I just reverted the changes to that file since I made a mistake in it. 

I made a separate PR that only adds in the stopwatch stuff.

veloce:
There is no gain here, and I think it introduces a regression: usually services are not watched but read (I know it is an anti-pattern to do `ref.read` in riverpod), so if you change the setting mid-game it would fail to refresh the provider and new setting would not be activated.

Issue #2697, opened 2026-03-01
Help


anmardragon99-collab:
### What is the problem?

Im trying to log in mobile to my account BM_anmar cars when i press authorize nothing happen and im stuck on anonymous id really like to use lichess but this is preventing me from doing to do 

### Steps to reproduce the bug

Im trying to log in mobile to my account BM_anmar

### App version

0.19.4

### Device Info

Android 15

HaonRekcef:
We had reports that this seems to be an issue with Microsoft Edge and switching to another browser has solved it for them. Which browser are you using? @anmardragon99-collab https://github.com/lichess-org/mobile/issues/1598

PR #3236, opened 2026-05-23
fix Android login issues

### LINKED ISSUE #3234: Cannot sign in after upgrade to 0.23.14.fdroid
### What is the problem?

After upgrading to 0.23.14.fdroid I got signed out and when trying to sign in again it redirects me to https://lichess.org/oauth?response_type=code&client_id=lichess_mobile&redirect_uri=org.lichess.mobile%3A%2F%2Flogin-callback&scope=web%3Amobile&code_challenge=FKk_MKVSEceRZkPYVkVsdvJV2JP_Q3wPrH0-yIXhjH8&code_challenge_method=S256&state=9an9OEqnSVfKfC9Xh6bTGA.
When I click `AUTHORIZE`, I get
```
All set! You can now close this page and return to the app.
```
But when returning to the app nothing changed and I'm still signed out. 

Upgrade and try to sign in.
I use Cromite (a Chromium based browser) as default browser.
Before the upgrade I was signed in. 

### App version

0.23.14.fdroid 

### Device Info

Moto g54 5G Android 15

### Additional information

_No response_

ghost:
Update `<queries>` to ensure the browser can talk back to the app when the user hits "Authorize."

Update `onNewIntent` to ensure the app reads the token instead of accidentally throwing it away.

Should close #3234 and #2697

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -164,5 +164,11 @@
             <action android:name="android.intent.action.PROCESS_TEXT"/>
             <data android:mimeType="text/plain"/>
         </intent>
+        <!-- Explicitly allows the app to discover and launch web browsers -->
+        <intent>
+            <action android:name="android.intent.action.VIEW" />
+            <category android:name="android.intent.category.BROWSABLE" />
+            <data android:scheme="https" />
+        </intent>
     </queries>
 </manifest>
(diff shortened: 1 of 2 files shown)


veloce (commented):
Thanks for trying to fix this.

I would like to understand what this fix does. Can you add a detailed explanation in the PR description please? There is too little information here for me to review.

I would also like to know whether you were able to reproduce the bug. If yes, with what device?
If you were able to reproduce, a screen recording showing the bug and one showing the fix would be appreciated.

Thank you!

PR #3237, opened 2026-05-23
Fix 2 notification bugs

julien4215:
- Fix wrong channel id for challenge created notifications.
  
  The wrong channel id caused the notification to be parsed incorrectly when the user taps on it. Because of that, it would bring the user to the home screen instead of the challenge screen.

-  Only display the challenge received notifications from the socket if the app is in the foregroud because fcm already shows the notification when the app is in the background.
    
    Before this fix, when the app is in the background, it might show the notification from the socket (when the socket is still alive in the background) which is later overriden by the notification from fcm resulting in a confusing experience for the user.

### DIFF
--- a/test/model/challenge/challenge_service_test.dart
+++ b/test/model/challenge/challenge_service_test.dart
@@ -134,6 +134,10 @@ void main() {
       ),
     ).thenAnswer((_) => Future.value());
 
+    // notifications from socket are only displayed if app is in foreground
+    final binding = TestWidgetsFlutterBinding.ensureInitialized();
+    binding.handleAppLifecycleStateChanged(AppLifecycleState.resumed);
+
     final container = await makeContainer(
       authUser: fakeAuthUser,
       overrides: {
@@ -222,6 +226,10 @@ void main() {
     when(
       () => notificationDisplayMock.cancel(id: any(named: 'id')),
     ).thenAnswer((_) => Future.value());
+
+    // notifications from socket are only displayed if app is in foreground
+    final binding = TestWidgetsFlutterBinding.ensureInitialized();
+    binding.handleAppLifecycleStateChanged(AppLifecycleState.resumed);
 
     final container = await makeContainer(
       authUser: fakeAuthUser,
--- a/test/model/challenge/challenge_service_test.dart
+++ b/test/model/challenge/challenge_service_test.dart
@@ -62,7 +62,7 @@ class _ShowDeclineDialogWidget extends ConsumerWidget {
 }
 
 void main() {
-  TestWidgetsFlutterBinding.ensureInitialized();
+  final binding = TestWidgetsFlutterBinding.ensureInitialized();
 
   setUpAll(() {
     registerFallbackValue(const ChallengeId(''));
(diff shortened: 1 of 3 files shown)


test/model/challenge/challenge_service_test.dart
```diff
@@ -134,6 +134,10 @@ void main() {
       ),
     ).thenAnswer((_) => Future.value());
 
+    // notifications from socket are only displayed if app is in foreground
+    final binding = TestWidgetsFlutterBinding.ensureInitialized();
+    binding.handleAppLifecycleStateChanged(AppLifecycleState.resumed);
```
veloce: I don't think this is useful. More useful would be a test that assert that the notification is not shown in background.

julien4215: The tests fail without it. I think it doesn't consider that the app is in foreground by default.

veloce: That would be very weird if the app wasn't in foreground by default during tests. The tests would not even work if that was the case.

I can't accept the PR until I have the real explanation for why the tests fail with the new code.

And also it needs a test that shows the notification is not shown in background.

julien4215: I have added a test that shows no local notification is disaplayed in background.

When I print the value of `lifecycleState` in the test, it is null. I don't find it that surprising that the app is not set to foreground during tests since it doesn't run a on an emulator or a device. By looking quickly at the Flutter source code, it seems to leave the initial value empty https://github.com/flutter/flutter/blob/924134a44c189315be2148659913dda1671cbe99/packages/flutter_test/lib/src/window.dart#L320-L328.

veloce: Thanks for the new test.

veloce:
> Before this fix, when the app is in the background, it might show the notification from the socket (when the socket is still alive in the background) which is later overriden by the notification from fcm resulting in a confusing experience for the user.

@julien4215 were you able to reproduce this?

See the copilot comment for the related change you made in challenge_service. I'm very cautious about this, because I don't think we've received complaints about this. I don't want to introduce new problems by solving non-existent ones.

veloce:
FYI I cherry-picked your commit that fixes the wrong ID. So if we do a release before this PR is merged, at least we'll have this fix.

julien4215:
> @julien4215 were you able to reproduce this?

Yes but it was with the fdroid version. However the behavior should be the same as with FCM. Can you try to reproduce with FCM? I am not familiar with how to test FCM notifications locally.

julien4215:
[notification-bug.webm](https://github.com/user-attachments/assets/0b9d0900-13f8-4ce4-bbac-fdc6cb08be42)

Here is a video of the issue with fcm.

The behavior is a little different than what I got with Unified Push. The notification doesn't get overriden. Instead there are two notifications shown for the same challenge.

julien4215:
> See the copilot comment for the related change you made in challenge_service. I'm very cautious about this, because I don't think we've received complaints about this. I don't want to introduce new problems by solving non-existent ones.

It is quite hard to tell what fcm considers as background and what lifecycle states correspond to it. Maybe there isn't even a clear correspondance. Anyway it is better to have duplicate notifications than missing a notification so I took copilot changes. If we want something cleaner, I think we will have to rethink how challenge notifications are handled (I have added a TODO so that we don't forget there is room for improvement). Until then, it should somewhat solve duplicated notifications for challenges.

PR #3244, opened 2026-05-25
fix showTenths in duration.dart

HaonRekcef:
We displayed the hundreds of the second not the tenths currently, this PR fixes this and introduces tests.


veloce (approved):
Thanks!

PR #3246, opened 2026-05-26
ci(crowdin-pull): guard crowdin pull jobs by repository owner

lsaudon:
CI is also launching on forks via the schedule.

The only approach I’ve found is to add an if statement to the job.

https://github.com/orgs/community/discussions/26409#discussioncomment-3251822


veloce (approved):
Thanks!

PR #3247, opened 2026-05-27
Fix: broadcasts result showing "½" when points scored was 0

allanjoseph98:
For context: https://github.com/lichess-org/mobile/pull/2943#issuecomment-4554151067


veloce (approved):
Thanks!

PR #3249, opened 2026-05-28
Localize speed and variant labels

Matrix9494:
## Summary

This PR localizes speed and variant labels in the mobile UI by reusing existing translation keys.

It adds mobile l10n generation for existing variant names from `variant.xml`, then uses those localized labels across game creation, challenges, offline computer setup, over-the-board setup, studies, analysis, and variant pickers.

It also switches speed labels such as Bullet, Blitz, Rapid, Classical, and Correspondence to existing localized strings where they are shown in the UI.


lib/src/model/common/chess.dart
```diff
@@ -130,6 +130,31 @@ enum Variant {
   final String label;
```
veloce: Let's remove `label` if we are going to use the translated one instead (which should then be called `label`).

lib/src/model/common/speed.dart
```diff
@@ -51,6 +52,23 @@ enum Speed {
     }
   }
 
+  String translatedLabel(AppLocalizations l10n) {
```
veloce: Same: remove existing `label` and name this one label, we don't want to use the non-translated string anymore.

lib/src/model/common/chess.dart
Matrix9494: Done locally. One detail I noticed is that the old `Variant.label` was not only used as UI text. A few places use it for PGN/export or protocol-like data, where the stable English value should be kept.

So I split the old usage into two explicit APIs:

- `label(l10n)` for localized UI text
- `pgnName` for the old stable English value

`pgnName` is effectively the previous `Variant.label`, just renamed to make its non-UI purpose explicit. I originally used `translatedLabel(l10n)` because I wanted to avoid breaking those existing non-UI usages.

Does that naming look good to you, or would you prefer another name for the old stable English label?

veloce (approved):
Thanks!

PR #3253, opened 2026-05-29
Remove unused files and exports

ghost:
Nothing is importing `utils/async_value.dart` or `widgets/adaptive_date_picker.dart` so I deleted them.

Nothing is using `export 'package:lichess_mobile/src/widgets/material_diff.dart';` from `view/game/game_player.dart` so I deleted that line.

`widgets/emoji_picker/widget.dart` was exporting 3 things, but only two of them were being used. The only file using those two was `view/account/edit_profile_screen.dart` so I just deleted `widget.dart` and had `edit_profile_screen.dart` import those two things directly.


veloce (approved):
lgtm! thanks

PR #3257, opened 2026-05-29
Fix last move highlights in offline games

Matrix9494:
## Summary

Fixes missing last move highlights/animations in offline interactive game screens.

After the recent `GameLayout` controller refactor, interactive boards now read the last move from `GameBoardParams.interactive.lastMove`, while `GameLayout.lastMove` is only used for readonly boards. Some offline interactive screens were still passing `lastMove` to `GameLayout`, so the board controller never received it.

Updated:
- offline computer games
- over-the-board games
- offline correspondence games


veloce:
I just added non-regression tests on top of it. Thanks!

PR #3263, opened 2026-05-31
Add query parameters to getOnlineBots API call

ComradeRamen:
Api returns only 100 bots by default.

### DIFF
--- a/lib/src/model/user/user_repository.dart
+++ b/lib/src/model/user/user_repository.dart
@@ -82,7 +82,7 @@ class UserRepository {
   }
 
   Future<IList<User>> getOnlineBots() {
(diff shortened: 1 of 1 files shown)


lib/src/model/user/user_repository.dart
```diff
@@ -82,7 +82,10 @@ class UserRepository {
   }
 
   Future<IList<User>> getOnlineBots() {
-    return client.readNdJsonList(Uri(path: '/api/bot/online'), mapper: User.fromServerJson);
+    return client.readNdJsonList(
+      Uri(path: '/api/bot/online', queryParameters: {'nb': '500'}),
```
veloce, line 86: If we return that many bots, we should stream the response.

There's already a method for that: https://github.com/lichess-org/mobile/blob/a3e328431bc9e845412313cf93474d19a7c2cbb0/lib/src/network/http.dart#L753

veloce (approved):

PR #3269, opened 2026-06-01
Add several ref.mounted checks for safety

ghost:
I added a bunch of `ref.mounted` checks. Hopefully these all make sense and are self-explanatory.

Still trying to be on a break, but I'm not doing a great job.


veloce:
I can see that. Hope you're not pushing yourself too hard!

ghost:
> I can see that. Hope you're not pushing yourself too hard!

Thanks! I pretty much just did a regex for this one so I didn't use my brain too much. Looks like it went well for the most part!

My headaches are down and I'm sleeping better -- not grinding my teeth in my sleep. I will probably put up some draft PRs of stuff I was working on so someone can use those or scrap them.

veloce:
Good to hear that! Just fyi I haven't yet merged this because this is not just a couple of `ref.mounted` checks, as the copilot review says, some of the checks you added throw exceptions or may change the behaviour.

I am reluctant to merge such a code when there is no proven bugs. If it were only `ref.mounted` checks after an async gap that would be different since those ones are safe.

ghost:
Got it! Should we just take out the ones in the TV channels then? The rest are simple, right?

What would be the best way to do that? I still suck at git so if I did it I would make a new PR with 6 files changed instead of 8 so it's simple and clean. Haha, just let me know what is best.

veloce:
I don't mind how you do it, if you modify this one or close it and creates a new one. Feel free to do as you are the most comfortable with.

ghost:
I decided to try it the way I wouldn't normally do it so I could see how it went. Haha! Looks good now I think.

PR #3279, opened 2026-06-05
Localize chess mode labels

Matrix9494:
## Summary

Localize chess mode, speed, variant, and TV channel labels across the mobile UI instead of displaying hardcoded English names such as `Bullet`, `King of the Hill`, or `Rapid`.

This adds centralized localization helpers for `Perf` and TV channels, then applies them to user-facing screens where those labels are shown.

## Localization coverage

- Added localized `Perf` labels:
  - standard speeds: UltraBullet, Bullet, Blitz, Rapid, Classical, Correspondence
  - variants: From Position, Chess960, Antichess, King of the Hill, Three-check, Atomic, Horde, Racing Kings, Crazyhouse
  - puzzle perf uses the existing puzzle localization
  - English keeps existing short labels where compact UI needs them

- Added localized TV channel labels:
  - Best, Blitz, Rapid, Classical, Bullet, UltraBullet
  - Bot, Computer, Horde, Racing Kings, Crazyhouse
  - Chess960, King of the Hill, Three-check, Antichess, Atomic

- Updated Watch / TV surfaces:
  - live TV channel list
  - watch tab TV card
  - TV screen title

- Updated user profile and stats surfaces:
  - perf cards
  - perf stats screen
  - user activity entries
  - leaderboard labels

- Updated game-related labels:
  - game screen title when showing a perf name
  - shared game title
  - game history perf filter labels
  - game filter selection summary
  - recent seek quick actions

- Updated Opening Explorer settings:
  - speed filter chip tooltips now use localized speed labels

- Updated tournament-related static labels:
  - tournament info line uses localized perf labels
  - tournament FAQ/help variant tables use localized variant names

## Notes

PGN metadata still uses the original English perf title, so exported/shared PGN data is not affected.

Tournament list names are not rewritten in this PR because they currently come from the server-provided tournament `fullName`. Web has a separate `tourname` translation layer for that, which would be a separate follow-up.

### DIFF
--- a/lib/src/model/common/perf.dart
+++ b/lib/src/model/common/perf.dart
@@ -1,6 +1,7 @@
 import 'package:deep_pick/deep_pick.dart';
 import 'package:fast_immutable_collections/fast_immutable_collections.dart';
 import 'package:flutter/widgets.dart';
+import 'package:lichess_mobile/l10n/l10n.dart';
 import 'package:lichess_mobile/src/model/common/chess.dart';
 import 'package:lichess_mobile/src/model/common/speed.dart';
 import 'package:lichess_mobile/src/styles/lichess_icons.dart';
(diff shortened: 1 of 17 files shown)


lib/src/model/common/perf.dart
```diff
@@ -32,6 +33,50 @@ enum Perf {
...
+      case Perf.puzzle:
+        return l10n.puzzles;
+      case Perf.storm:
+      case Perf.streak:
+        return title;
+    }
+  }
+
+  String shortLabel(AppLocalizations l10n) {
+    return l10n.localeName.startsWith('en') ? shortTitle : label(l10n);
```
veloce: Short label is meant to be short, so you can't do that as it would break UI.

We need a short label translation to support this.

lib/src/model/tv/tv_channel.dart
```diff
@@ -26,6 +27,43 @@ enum TvChannel {
   final String label;
   final IconData icon;
 
+  String localizedLabel(AppLocalizations l10n) {
```
veloce: Same remark as in the other PR: if we are to use the localised label, then remove the `label` member and this method should be named label.

lib/src/view/explorer/opening_explorer_settings.dart
```diff
@@ -49,7 +47,7 @@ class OpeningExplorerSettings extends ConsumerWidget {
                     String.fromCharCode(speed.icon.codePoint),
                     style: TextStyle(fontFamily: speed.icon.fontFamily, fontSize: 18.0),
                   ),
-                  tooltip: Perf.fromVariantAndSpeed(Variant.standard, speed).title,
+                  tooltip: speed.label(context.l10n),
```
veloce: You can't change it like that. Perf is different than speed for a reason so you should keep the perf here (which is also translated).

lib/src/view/explorer/opening_explorer_settings.dart
```diff
@@ -159,7 +157,7 @@ class OpeningExplorerSettings extends ConsumerWidget {
...
```
veloce: idem, keep perf.

lib/src/model/common/perf.dart
Matrix9494: I will fix the rest and keep this in next PR. I need to first figure out how to work with Crowdin.

veloce (approved):
LGTM!

PR #3282, opened 2026-06-06
Replace deprecated linkify package

### LINKED ISSUE #2092: Replace flutter_linkify
Package is not maintained and there is an issue with selectable text.

meenbeese:
I looked at some other packages but they were also either unmaintained or had few stars to be a reliable dependency. Since our linkify use case in this app is minimal, I decided to rewrite a small custom version without external deps.

I would appreciate any guidance since I have not written any Flutter code before.

Closes #2092

### DIFF
--- a/lib/src/app_links_service.dart
+++ b/lib/src/app_links_service.dart
@@ -29,8 +29,8 @@ import 'package:lichess_mobile/src/view/study/study_screen.dart';
 import 'package:lichess_mobile/src/view/tournament/tournament_screen.dart';
 import 'package:lichess_mobile/src/view/user/user_or_profile_screen.dart';
 import 'package:lichess_mobile/src/view/watch/tv_screen.dart';
+import 'package:lichess_mobile/src/widgets/app_link.dart';
 import 'package:lichess_mobile/src/widgets/feedback.dart';
-import 'package:linkify/linkify.dart';
 import 'package:logging/logging.dart';
 import 'package:url_launcher/url_launcher.dart';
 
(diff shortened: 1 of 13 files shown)


lib/src/app_links_service.dart
```diff
@@ -387,23 +387,21 @@ class AppLinksService {
     return route;
   }
 
-  static const kLichessLinkifiers = [UrlLinkifier(), EmailLinkifier(), UserTagLinkifier()];
```
veloce, line 390: Not a big deal, but we are not linking emails anymore with that change. Can be done later.

lib/src/view/chat/chat_screen.dart
```diff
@@ -203,13 +203,11 @@ class _MessageBubble extends ConsumerWidget {
...
                   ),
-                Linkify(
-                  onOpen: (link) async =>
-                      await ref.read(appLinksServiceProvider).onLinkifyOpen(context, link),
-                  linkifiers: AppLinksService.kLichessLinkifiers,
+                AppLinkify(
                   text: message.message,
                   style: TextStyle(color: _textColor(context)),
                   linkStyle: Styles.linkStyle,
+                  onTap: (link) => ref.read(appLinksServiceProvider).onLinkTap(context, link),
```
veloce: Since the `onTap` is always the same call to app links service, could we not simplify and handle it in the `AppLinkify` class directly?

I don't think we'll ever need to have custom logic on caller sites, so it's fine if it's less flexible.

lib/src/widgets/app_link_parser.dart
```diff
@@ -0,0 +1,52 @@
...
+  const LinkChunk({required this.text, required this.link});
+
+  final String text;
+  final AppLink link;
+}
+
+class AppLinkParser {
+  const AppLinkParser._();
+
+  static final _pattern = RegExp(r'(https?:\/\/[^\s]+)|(@[A-Za-z0-9_-]+)', caseSensitive: false);
```
veloce: We should have a better regex here. You can find many examples of url parsing regexes.

lib/src/widgets/app_link_parser.dart
```diff
@@ -0,0 +1,52 @@
...
+
+  static List<AppLinkSpan> parse(String text) {
```
veloce, line 29: This code should be unit tested.

lib/src/widgets/app_linkify.dart
```diff
@@ -0,0 +1,56 @@
+import 'dart:async';
+
+import 'package:flutter/gestures.dart';
+import 'package:flutter/material.dart';
+
+import 'package:lichess_mobile/src/widgets/app_link.dart';
+import 'package:lichess_mobile/src/widgets/app_link_parser.dart';
+
+class AppLinkify extends StatelessWidget {
```
veloce, line 7: We need widget tests for this.

veloce (commented):
Thanks for tackling this.

Can you please add tests? If we're not using an external package anymore it should be tested extensively.

Also your PR has conflicts, please solve them. And the current test suite is not passing, can you check that as well?

lib/src/view/chat/chat_screen.dart
meenbeese: Done.

lib/src/widgets/app_link_parser.dart
meenbeese: Done.

lib/src/widgets/app_link_parser.dart
meenbeese: Done.

lib/src/widgets/app_linkify.dart
meenbeese: Done.

meenbeese:
> Thanks for tackling this.
> 
> Can you please add tests? If we're not using an external package anymore it should be tested extensively.
> 
> Also your PR has conflicts, please solve them. And the current test suite is not passing, can you check that as well?

I'm not sure why some unrelated tests are failing...

PR #3286, opened 2026-06-07
better localize broadcast player tab

### LINKED ISSUE #2612: When looking at the players in a broadcast, text is cut off
See the screenshot for details. I am using German text, but this might happen in other languages too. 

image

HaonRekcef:
Uses the translation string for Standings Disclaimer, increase the max lines to 3 as languages can be verboser than english.
Also fix #2612 
| Before| After|
|-|-|
|image| image|

sorting still works: 

[Screencast_20260607_140422.webm](https://github.com/user-attachments/assets/10f50074-b2bd-4b2d-9dd4-e790ec63d5d5)


veloce (approved):
Thanks!

PR #3288, opened 2026-06-07
Broadcast Team Results

### LINKED ISSUE #2614: Broadcast Team Results
Add a tab to show the Team results that were added to lila:
https://lichess.org/broadcast/german-bundesliga-202526/round-12/SR1eddMe#team-results
https://lichess.org/broadcast/german-bundesliga-202526/round-12/SR1eddMe#team-results/SC%20Viernheim

HaonRekcef:
close #2614 

|||||
|-|-|-|-|
|image|image|image|image|

I decided to put the team results on top of the teams tab, to not overload the tab bar.

A lot of code from the player screen I could reuse and adjust. 
One problem is the table headers as they take to much space to put on one line, I put them on two. But with this approach we can not use the current localization.
added one simple test

### DIFF
--- a/lib/src/view/broadcast/broadcast_team_standings_screen.dart
+++ b/lib/src/view/broadcast/broadcast_team_standings_screen.dart
@@ -20,7 +20,7 @@ enum _SortingTypes { elo, score }
 typedef _TeamPicker<T> = T? Function(BroadcastTeamStanding team);
 
 const _kTableRowPadding = EdgeInsets.symmetric(horizontal: 8.0, vertical: 12.0);
-const _kHeaderTextStyle = TextStyle(fontWeight: FontWeight.bold, overflow: .ellipsis);
+const _kHeaderTextStyle = TextStyle(fontWeight: .bold, overflow: .ellipsis);
 
 class BroadcastTeamStandingsScreen extends ConsumerWidget {
   const BroadcastTeamStandingsScreen({super.key, required this.tournamentId});
@@ -39,8 +39,10 @@ class BroadcastTeamStandingsScreen extends ConsumerWidget {
     return PlatformScaffold(
       appBar: PlatformAppBar(title: AppBarTitleText(context.l10n.broadcastTeamResults)),
       body: switch ((standingsAsync, tournamentAsync)) {
-        (AsyncData(value: final teams), AsyncData()) when teams.isNotEmpty =>
-          BroadcastTeamStandingsList(teams: teams, tournamentId: tournamentId),
+        (AsyncData(value: final teams), AsyncData()) =>
+          teams.isNotEmpty
+              ? BroadcastTeamStandingsList(teams: teams, tournamentId: tournamentId)
+              : Center(child: Text(context.l10n.nothingToSeeHere)),
         (AsyncError(:final error), _) ||
         (_, AsyncError(:final error)) => Center(child: Text('Cannot load data: $error')),
         _ => const Center(child: CircularProgressIndicator.adaptive()),
@@ -91,6 +93,7 @@ class _BroadcastTeamStandingsListState extends ConsumerState<BroadcastTeamStandi
   ) => (team1, team2) {
     final field1 = picker(team1);
     final field2 = picker(team2);
+    if (field1 == null && field2 == null) return 0;
     if (field1 == null) return -1;
     if (field2 == null) return 1;
     return field1.compareTo(field2);
@@ -175,8 +178,9 @@ class _BroadcastTeamStandingsListState extends ConsumerState<BroadcastTeamStandi
                       Expanded(
                         child: _TableTitleCell(
                           title: Text(
-                            '${context.l10n.broadcastTeams} (avg. Elo)',
+                            '${context.l10n.broadcastTeams} (${context.l10n.averageElo})',
                             style: _kHeaderTextStyle,
+                            maxLines: 2,
                           ),
                           onTap: () => toggleSort(.elo),
                           sortIcon: currentSort == .elo ? sortIcon : null,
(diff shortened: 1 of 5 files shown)


veloce:
> One problem is the table headers as they take to much space to put on one line, I put them on two. But with this approach we can not use the current localization.

Is there no way to force the localised text to wrap?

veloce:
I like how you handled the UI. The team result button should have some horizontal padding though.

lib/src/view/broadcast/broadcast_teams_tab.dart
```diff
@@ -83,30 +85,76 @@ class BroadcastTeamsList extends ConsumerWidget {
...
+  const _TeamStandingsButton({required this.tournamentId});
+
+  final BroadcastTournamentId tournamentId;
+
+  @override
+  Widget build(BuildContext context) {
+    final colorScheme = Theme.of(context).colorScheme;
+
+    return Padding(
+      padding: const EdgeInsets.fromLTRB(4.0, 12.0, 4.0, 8.0),
```
veloce: A value of `horizontalBodyPadding` constant would be better I think for the left and right (or at least `8.0`).

lib/src/view/broadcast/broadcast_team_standings_screen.dart
```diff
@@ -0,0 +1,355 @@
...
+              childCount: teams.length,
+            ),
+          ),
+        ),
+      ],
+    );
+  }
+}
+
+class _TableTitleCell extends StatelessWidget {
```
veloce, line 248: Perhaps I'm wrong but it looks like we could reuse some of these widgets across broadcast screens where tables are used.

Only if it is really straightforward. The goal is to ensure a consistent styling across all the broadcast screens.

lib/src/view/broadcast/broadcast_team_standings_screen.dart
```diff
@@ -0,0 +1,355 @@
...
+              ),
+            ],
+          ),
+        ),
+      ),
+    );
+  }
+}
+
+class _BroadcastTeamStandingRow extends StatelessWidget {
```
veloce, line 295: Same idea here. If possible we should have a broadcast standing row widget (if not adding complexity).

Not in the scope of this PR.

lib/src/view/broadcast/broadcast_team_standings_screen.dart
```diff
@@ -0,0 +1,355 @@
...
+  ///
+  /// Null values are smaller than any other value.
+  Comparator<BroadcastTeamStanding> nullableCompare<T extends Comparable<T>>(
+    _TeamPicker<T> picker,
+  ) => (team1, team2) {
+    final field1 = picker(team1);
+    final field2 = picker(team2);
+    if (field1 == null) return -1;
+    if (field2 == null) return 1;
+    return field1.compareTo(field2);
```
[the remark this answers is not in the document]
HaonRekcef: fixed also in the other places

lib/src/view/broadcast/broadcast_team_standings_screen.dart
```diff
@@ -0,0 +1,355 @@
...
+
+    return PlatformScaffold(
+      appBar: PlatformAppBar(title: AppBarTitleText(context.l10n.broadcastTeamResults)),
+      body: switch ((standingsAsync, tournamentAsync)) {
+        (AsyncData(value: final teams), AsyncData()) when teams.isNotEmpty =>
+          BroadcastTeamStandingsList(teams: teams, tournamentId: tournamentId),
+        (AsyncError(:final error), _) ||
+        (_, AsyncError(:final error)) => Center(child: Text('Cannot load data: $error')),
+        _ => const Center(child: CircularProgressIndicator.adaptive()),
+      },
```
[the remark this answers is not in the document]
HaonRekcef: fixed

HaonRekcef:
image

image


> Is there no way to force the localised text to wrap?

It wraps but maybe not at the "correct" place in the word, but I used the translations anyways now as it is better anyways. In English it wraps correctly.

lib/src/view/broadcast/broadcast_team_standings_screen.dart
```diff
@@ -0,0 +1,364 @@
...
+  @override
+  Widget build(BuildContext context) {
+    return ListTile(
+      contentPadding: const EdgeInsetsDirectional.only(start: 16.0, end: 8.0),
+      visualDensity: .compact,
+      tileColor: index.isEven ? context.lichessTheme.rowEven : context.lichessTheme.rowOdd,
+      title: Text(team.name, maxLines: 2, overflow: .ellipsis),
+      subtitle: team.averageRating != null
+          ? Text(
+              'Ø ${team.averageRating}',
```
allanjoseph98: `Ø` is a German speaking convention. `X̄` is more common internationally but it looks worse. You could also just drop the symbol entirely. I don't think it's _too_ hard to guess the number's meaning from context. 
EDIT - The table header already has context.

ijm8710: Had also suggested μ which may not be as universally known but mathematically makes sense similar to x-bar.

HaonRekcef: makes sense, I will just removed it. The header should be enough context indeed.

lib/src/view/broadcast/broadcast_teams_tab.dart
```diff
@@ -83,30 +85,76 @@ class BroadcastTeamsList extends ConsumerWidget {
...
-            roundId: roundId,
-            tournamentSlug: tournamentSlug,
-            roundSlug: value.round.slug,
-            title: value.round.name,
-            showEvaluationGauge: showEvaluationGauges,
-            customScoring: value.round.customScoring,
-            showTeamScores: showTeamScores,
-          );
-        },
+      AsyncData(:final value) => CustomScrollView(
```
veloce, line 88: CustomScrollView doesn't handle view padding automatically, contrary to ListView; it breaks the view on iOS.

veloce, line 88: See how it's done in players tab.

PR #3300, opened 2026-06-08
Polish app bar scroll behavior

Matrix9494:
## Summary

This PR smooths out app bar color transitions and makes a few scroll-related app bar behaviors more consistent.

## Changes

- Enables animated color transitions for regular app bars, so Material scrolled-under color changes no longer feel abrupt.
- Reworks the theme settings screen to use a regular `AppBar` with a fixed board preview area.
  - The board preview now stays on the page background instead of being part of a scrolling `SliverAppBar`.
  - This keeps the preview visually stable while the settings list scrolls underneath.
- Handles the broadcast list separately:
  - The page app bar and sticky section headers use the same background color.
  - The broadcast app bar does not change color on scroll, avoiding a distracting double state change between the page app bar and section headers.
  - Section headers keep their pinned `SliverAppBar` behavior, but do not apply additional scrolled-under elevation.
 

<table>
  <tr>
    <td>
      <video src="https://github.com/user-attachments/assets/32b421d3-05ec-4653-9904-98370f825368" width="300" controls></video>
    </td>
    <td>
      <video src="https://github.com/user-attachments/assets/e466983a-ff7a-4cf3-a111-010d59350057" width="300" controls></video>
    </td>
  </tr>
</table>

### DIFF
--- a/lib/src/view/broadcast/broadcast_list_screen.dart
+++ b/lib/src/view/broadcast/broadcast_list_screen.dart
@@ -84,10 +84,14 @@ class _BroadcastListScreenState extends State<BroadcastListScreen> {
       ),
     );
 
+    final appBarBackground = ColorScheme.of(context).surfaceContainerHigh.withValues(alpha: 1);
+
     return Scaffold(
       body: _Body(filter),
       appBar: AppBar(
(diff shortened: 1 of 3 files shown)


veloce:
Looks good! Thanks. I'll need to test this on iOS too.

veloce (approved):
thanks!

PR #3306, opened 2026-06-09
Clamp displayed centipawn evaluations to +/-99 pawns

### LINKED ISSUE #3222: Clamp cp to +/- 99
As shared by niklasf in #2694 and in relation to #3216, there is some overflow for tablebase evals such as +200. Niklasf shared that these should be clamped to +/- 99

> Btw. on the web we clamp cp to +/-99. There's no tablebase scores, so these huge scores have basically no utility for humans.

> Mate scores have no such clamping and even an additional #. 

Example:

image

image

CloudyDino:
Fixes #3222.

  Tablebase / extreme engine scores currently render as values like
  `+200.0` in the analysis gauge and move list, which has no practical
  meaning for humans. Mirror the web behaviour by clamping the
  centipawn value to `[-9900, 9900]` inside `_evalString` before
  converting to pawns. Mate scores are unaffected (`#3`, `#-2`, …).

  **Changes**
  - `lib/src/model/common/eval.dart` — clamp `cp` to `±9900` in `_evalString`.
  - `test/model/common/eval_test.dart` — new unit tests covering small / negative / zero / mate / and the new clamp boundaries.

### DIFF
--- a/lib/src/model/common/eval.dart
+++ b/lib/src/model/common/eval.dart
@@ -324,9 +324,17 @@ double _rawWinningChances(num cp) {
   return 2 / (1 + math.exp(multiplier * cp)) - 1;
 }
 
+/// Maximum absolute centipawn value displayed in evaluations.
+///
+/// Values above this magnitude (e.g. tablebase scores like +200) are clamped
+/// to +/-99 pawns when rendered, since such large advantages have little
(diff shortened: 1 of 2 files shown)


veloce (approved):
Thanks for the improvement!

PR #3310, opened 2026-06-10
Fix rematch for finished correspondence games

### LINKED ISSUE #1183: Rematch offer for correspondence should not be online-dependent
Please correct me if I'm wrong, this is merely from my own observations.

It seems like the rematch-logic relies on both you and the opponent being online (both when creating and accepting). This makes sense for real-time games, but not necessarily for correspondence games IMO.

My suggestion is that rematch-offers should not be online-dependent for correspondence games, as I think it aligns better with the game style overall.

### LINKED ISSUE #2089: Rematch Doesn't Work for Correspondence Games
### What is the problem?

I play friends in correspondence games and when the game is over the dialog comes up but "Rematch" is greyed out. I don't know why.

Image

start a correspondence game with a friend
resign the game
look at the dialog

### App version

0.17.3

### Device Info

Pixel 8, Graphene OS (Android 16)

### Additional information

_No response_

twesk:
Closes #1183
Closes #2089

Rematch stayed greyed out on finished correspondence games because mobile only ever sent a socket offer, which an offline opponent never receives. This matches the web client: when the opponent is offline in a clockless game, create a persistent rematch challenge via `POST /challenge/rematch-of/{id}` instead. Same endpoint and behavior as the web UI.

Requires an update to the `lila` repo as well: https://github.com/lichess-org/lila/pull/20645

Assisted by Opus 4.8 and Fable 5


twesk:
Addressed the two actionable items in 14a56d5a2; declined the l10n/timing ones to match CLAUDE.md conventions and the existing test suit - details inline

HaonRekcef:
thanks! Can you share a video recording?

twesk:
Here are some videos I grabbed after setting up the test environment.

The Issue (Rematch button disabled when opponent leaves the board, even in a correspondence match where the 2 players are less likely to be in the board at the same time, especially after a game has concluded):

https://github.com/user-attachments/assets/f200382f-0182-449a-8a94-027bf82b84aa

The fix:

(mobile player's opponent leaves the board, shows them as offline, rematch button still works)

https://github.com/user-attachments/assets/a0ae8927-a850-4e87-b7cc-800f62a6cc94

(mobile player's opponent closes the browser, shows them as offline, rematch button still works)

https://github.com/user-attachments/assets/50a863c6-3733-40d5-b452-98d60b950abf

veloce:
Thanks for this! I'm not reviewing immediately since it requires a server change. Will do once the server change is accepted.

lib/src/model/game/game_controller.dart
```diff
@@ -458,6 +459,26 @@ class GameController extends AsyncNotifier<GameState> {
     _socketClient.send('rematch-yes', null);
   }
 
+  Future<void> challengeRematch() async {
+    final repository = ref.read(challengeRepositoryProvider);
+    await repository.rematchOfGame(gameFullId.gameId);
+    // The server doesn't return the created challenge, so fetch the outgoing
+    // challenges to find it, which allows the user to cancel it later.
+    final challenges = await repository.list();
```
veloce, line 490: It's always good to add a `if (!ref.mounted) return;` check after an asynchronous gap in a Notifier.

lib/src/model/game/game_controller.dart
```diff
@@ -458,6 +459,26 @@ class GameController extends AsyncNotifier<GameState> {
...
+    final curState = state.requireValue;
+    state = AsyncValue.data(curState.copyWith(rematchChallengeId: challenge?.id));
+  }
+
+  /// Cancels the rematch challenge previously created by [challengeRematch].
+  Future<void> cancelRematchChallenge() async {
+    final curState = state.requireValue;
+    final challengeId = curState.rematchChallengeId;
+    if (challengeId == null) return;
+    await ref.read(challengeRepositoryProvider).cancel(challengeId);
```
veloce, line 501: missing ref.mounted check

test/view/game/game_rematch_test.dart
```diff
@@ -0,0 +1,205 @@
...
+      await pumpFinishedGame(
+        tester,
+        client: client,
+        requestedPaths: requestedPaths,
+        opponentOnGame: false,
+      );
+
+      // The result dialog is shown 500ms after the game finishes, then its
+      // buttons activate after a further second.
+      await tester.pump(const Duration(milliseconds: 600));
```
veloce: Now the game end dialog delay is exported in a const in `game_body.dart`. It is private, but you could make it public and use it to make the wait duration.

veloce (commented):
Thanks for this effort, and for the detailed PR description!

lib/src/model/game/game_controller.dart
```diff
@@ -1035,6 +1056,10 @@ sealed class GameState with _$GameState {
 
     /// Game full id used to redirect to the new game of the rematch
     GameFullId? redirectGameId,
+
+    /// Id of the rematch challenge created when the opponent is offline in a
+    /// clockless game (see [GameController.challengeRematch]).
+    ChallengeId? rematchChallengeId,
```
dev-shady: I think we should change variable name to offlineRematchChallengeId (or correspondenceRematchId).Since this is specifically a fallback for clockless/offline scenarios without an active socket connection, adding the offline prefix makes the constraint self-documenting and stops it from being mistaken for a general real-time rematch tracker across the rest of the codebase.  @veloce what do you think ?

veloce: Not sure I have followed everything here. I thought it was used in clockless games anytime (even if opponent is online).

Well, in any case I agree `correspondenceRematchId` is a better name @twesk

lib/src/model/game/game_controller.dart
twesk: Good to know, addressed!

lib/src/model/game/game_controller.dart
twesk: Addressed

test/view/game/game_rematch_test.dart
twesk: Good idea, thanks.  Addressed.

lib/src/model/game/game_controller.dart
twesk: Big fan of good naming here :D Thanks. Addressed

veloce (approved):

PR #3312, opened 2026-06-10
Show a toast when a user does not accept challenges

### LINKED ISSUE #3243: Toast message for players unavailable to be challenged
Decently minor but on web, challenge is omnipresent on a profile, but if someone (ie thibault) opts out, you get a toast message that they don’t accept challenges from friends

In the app, currently we just hide the challenge button  

I prefer the alert becuase

1) I think buttons sometimes appearing and sometimes not is more confusing to a user
2) I like to hear why. Thibaults is hidden because he "doesn’t accept from non friends"

CloudyDino:
Fixes #3243.

When viewing a user profile, the **Challenge** action was previously hidden whenever the server returned `canChallenge != true`. The action then silently came and went depending on each user's challenge preference, which (as the reporter noted) is confusing — users had no way to know *why* the button was missing.

Match the web behaviour:

- Show the Challenge action whenever `canChallenge` is **non-null** (i.e. the server is informing us about the challengeability state). On the user's own profile `canChallenge` is null, so the action stays hidden as before.
- When the user has opted out (`canChallenge == false`), tapping the action now shows a snackbar using the already-translated `challengeXDoesNotAcceptChallenges` string instead of triggering the challenge flow.
- Same logic applied to the user context-menu bottom sheet.

## Changes

- `lib/src/view/user/user_screen.dart`
- `lib/src/view/user/user_context_menu.dart`
- `test/view/user/user_screen_test.dart` — three new widget tests (`canChallenge` null / true / false), including asserting that the toast appears with the localized message.

## Verification

```
flutter analyze lib/src/view/user/user_screen.dart lib/src/view/user/user_context_menu.dart test/view/user/user_screen_test.dart  # 0 issues
flutter test test/view/user/user_screen_test.dart  # 5/5 pass
```

Also verified manually on an Android emulator (Pixel 10 / API 36) against `lichess.dev`:

1. Signed in as `CD-Test-1`.
2. Set `CD-Test-2`'s "Let other players challenge you" preference to **Nobody** on the website.
3. Opened `CD-Test-2`'s profile in the app:
   - Challenge action is **visible** (the new behaviour).
4. Tapped Challenge:
   - Snackbar appears reading **"CD-Test-2 does not accept challenges."** instead of opening the challenge composer.

UI accessibility-tree dump after tap confirms the snackbar text:

```
content-desc="CD-Test-2 does not accept challenges."
```

Screenshots attached in a follow-up comment.

### DIFF
--- a/lib/src/view/user/user_context_menu.dart
+++ b/lib/src/view/user/user_context_menu.dart
@@ -13,6 +13,7 @@ import 'package:lichess_mobile/src/view/user/user_or_profile_screen.dart';
 import 'package:lichess_mobile/src/view/user/user_screen.dart';
 import 'package:lichess_mobile/src/view/watch/tv_screen.dart';
 import 'package:lichess_mobile/src/widgets/adaptive_bottom_sheet.dart';
+import 'package:lichess_mobile/src/widgets/feedback.dart';
 import 'package:lichess_mobile/src/widgets/list.dart';
 import 'package:lichess_mobile/src/widgets/user.dart';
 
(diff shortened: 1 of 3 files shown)


CloudyDino:
## Manual verification screenshots

**Profile screen** — Challenge action is now visible even when the user has opted out of challenges (previously hidden):

CD-Test-2 profile with Challenge action visible

**After tapping Challenge** — Snackbar reading `"CD-Test-2 does not accept challenges."` appears at the bottom (instead of the challenge composer opening):

Snackbar: CD-Test-2 does not accept challenges.

veloce (approved):
LGTM, thanks for the improvement!

PR #3313, opened 2026-06-10
fix isTyping displaying if different user is typing

HaonRekcef:
Currently when being in a chat `is typing` would be displayed even if a different user was typing and even worse the message from a different user would be displayed in the chat the user is viewing too. 

Added tests that fail before and pass now.
Also verified the fix on the app.


veloce (approved):
Thanks!

PR #3314, opened 2026-06-10
Feature/configurable highlight color

### LINKED ISSUE #1989: Board highlight may be barely visible when the board hue is changed
Turning any board green makes the move color practically invisible. It would be useful to be able to change the move color, even if it's just a selection between 4 colors like the move arrows.

Here I really like this board in this color with this background but I cant see when any move lands on a dark square.

Image

adharshio:
Fixes #1989
a new setting in Theme screen to change last move highlight color
 the default is yellow to match existing behavior

### DIFF
--- a/lib/src/model/settings/board_preferences.dart
+++ b/lib/src/model/settings/board_preferences.dart
@@ -121,6 +121,10 @@ class BoardPreferences extends Notifier<BoardPrefs> with PreferencesStorage<Boar
   Future<void> setShapeColor(ShapeColor shapeColor) {
     return save(state.copyWith(shapeColor: shapeColor));
   }
+  
+  Future<void> setHighlightColor(HighlightColor highlightColor) {
+  return save(state.copyWith(highlightColor: highlightColor));
+  }
 
   Future<void> adjustColors({double? brightness, double? hue}) {
     return save(state.copyWith(brightness: brightness ?? state.brightness, hue: hue ?? state.hue));
@@ -173,6 +177,8 @@ sealed class BoardPrefs with _$BoardPrefs implements Serializable {
     required DragTargetKind dragTargetKind,
     @JsonKey(defaultValue: ShapeColor.green, unknownEnumValue: ShapeColor.green)
     required ShapeColor shapeColor,
+    @JsonKey(defaultValue: HighlightColor.yellow, unknownEnumValue: HighlightColor.yellow)
+    required HighlightColor highlightColor,
     @JsonKey(defaultValue: false) required bool showBorder,
     @JsonKey(defaultValue: kBoardDefaultBrightnessFilter) required double brightness,
     @JsonKey(defaultValue: kBoardDefaultHueFilter) required double hue,
@@ -199,6 +205,7 @@ sealed class BoardPrefs with _$BoardPrefs implements Serializable {
     magnifyDraggedPiece: true,
     dragTargetKind: DragTargetKind.circle,
     shapeColor: ShapeColor.green,
+    highlightColor: HighlightColor.yellow,
     showBorder: false,
     brightness: kBoardDefaultBrightnessFilter,
     hue: kBoardDefaultHueFilter,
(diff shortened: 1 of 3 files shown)


adharshio:
Here is screenrecording of what all changes are made:

https://github.com/user-attachments/assets/4223682b-d666-4caa-aeb0-8b6e928e5975

veloce:
I'm always reluctant to add new settings... but I suppose this one is useful. Thanks the contribution.

lib/src/view/game/watcher_list_bottom_sheet.dart
veloce, line 69: This is not relevant to this PR at all, please remove this change.

lib/src/model/settings/board_preferences.dart
```diff
@@ -173,6 +179,8 @@ sealed class BoardPrefs with _$BoardPrefs implements Serializable {
     required DragTargetKind dragTargetKind,
     @JsonKey(defaultValue: ShapeColor.green, unknownEnumValue: ShapeColor.green)
     required ShapeColor shapeColor,
+    @JsonKey(defaultValue: HighlightColor.yellow, unknownEnumValue: HighlightColor.yellow)
+    required HighlightColor highlightColor,
```
veloce: The default value actually depends on the board, so let's make it nullable. It should act only as an override.

lib/src/model/settings/board_preferences.dart
```diff
@@ -207,10 +216,20 @@ sealed class BoardPrefs with _$BoardPrefs implements Serializable {
   bool get hasColorAdjustments =>
       brightness != kBoardDefaultBrightnessFilter || hue != kBoardDefaultHueFilter;
 
-  ChessboardSettings toBoardSettings(Variant variant) {
+  ChessboardSettings toBoardSettings() {
     return ChessboardSettings(
       pieceAssets: pieceSet.assets,
-      colorScheme: boardTheme.colors,
+      colorScheme: ChessboardColorScheme(
```
veloce, line 219: Let's wait for the chessground update where I added a `copyWith` method to ChessboardColorScheme.

adharshio:
@veloce made the necessary changes...would love to hear your suggestions if any

veloce:
@adharshio sorry I do not see these changes in the current diff.

adharshio:
@veloce the changes are in the force-pushed commits  (6a8eab0). Could you try refreshing the diff? Sorry for the confusion

veloce:
Sorry I don't see the changes in this commit. The commit says "format code" btw.

adharshio:
yeah i thought i had pushed the changes in the "format chore" , it was a mistake from my end , sorry for the confusion, i have committed the new changes...please review

lib/src/model/settings/board_preferences.dart
```diff
@@ -210,7 +216,19 @@ sealed class BoardPrefs with _$BoardPrefs implements Serializable {
   ChessboardSettings toBoardSettings(Variant variant) {
     return ChessboardSettings(
       pieceAssets: pieceSet.assets,
-      colorScheme: boardTheme.colors,
+      colorScheme: ChessboardColorScheme(
```
veloce, line 219: Please use `ChessboardColorcheme.copyWith` which is now available.

lib/src/model/settings/board_preferences.dart
```diff
@@ -173,6 +177,7 @@ sealed class BoardPrefs with _$BoardPrefs implements Serializable {
     required DragTargetKind dragTargetKind,
     @JsonKey(defaultValue: ShapeColor.green, unknownEnumValue: ShapeColor.green)
     required ShapeColor shapeColor,
+    HighlightColor? highlightColor,
```
veloce, line 180: This actually does not work. We need to be able to switch back to the default which is no custom color and here we can't.

Here it should be non nullable with a default value:
```dart
@JsonKey(defaultValue: HighlightColor.default, unknownEnumValue: HighlightColor.default)
required HighlightColor highlightColor
```

lib/src/model/settings/board_preferences.dart
```diff
@@ -199,6 +204,7 @@ sealed class BoardPrefs with _$BoardPrefs implements Serializable {
     magnifyDraggedPiece: true,
     dragTargetKind: DragTargetKind.circle,
     shapeColor: ShapeColor.green,
+    highlightColor: null,
```
veloce, line 207: HighlightColor.default

lib/src/model/settings/board_preferences.dart
```diff
@@ -253,6 +271,23 @@ enum ShapeColor {
   }).withAlpha(0xAA);
 }
 
+/// Colors for the last move highlight on the board.
+enum HighlightColor {
+  yellow,
```
veloce, line 276: Add a `default` associated with a dummy color (ex black) fully transparent (we won't use it)

adharshio:
hello @veloce ,i know that a sorry doesn't cut the delay for this push, still i am very much sorry for this delay,i was very much distracted and couldn't follow up with this pr...happy to hear any suggestions.

adharshio:
hey @veloce ,addressed all comments,happy to hear any suggestions

PR #3319, opened 2026-06-11
Refine analysis appbar header format for finished games

### LINKED ISSUE #3259: A couple analysisboard header suggestion f/u's
So @HaonRekcef added these recently and they're awesome

I would suggest a few adjustments to match the format of web.

Some examples:

Symbol 5+0 • Rated • Chess960
Symbol 5+0 • Casual • Crazyhouse
Symbol 5+0 • Rated • Blitz

- [ ] I know you're showing the variant symbol, but if it's a standard game they still precede it with the time control symbol
- [ ] If it's not standard variant, the variant type overrides showing the time control name. You still have the time control number regardless

CloudyDino:
Fixes #3259.

Refine the analysis screen header for a finished game so the speed/variant icon leads the title and the rated/casual segment comes before any trailing variant label. On mobile space is constrained, so — unlike the web client — the speed label is omitted for standard chess; the leading icon and the time control already convey it.

Format:
- **Standard / from-position**: `{speed icon} {clockDisplay} • {Rated|Casual}` (e.g. `5+0 • Rated`). Previously standard games showed no leading icon at all.
- **Non-standard variants** (Chess960, Crazyhouse, …): keep the variant icon and append the variant label as the trailing segment (e.g. `5+0 • Rated • Chess960`); the time control is still shown so the duration is retained.

Previously the title was `{clock} • {speed} • {Rated|Casual}` for every game, and the variant icon was always used (so standard games had no leading icon).

## Changes
- `lib/src/view/analysis/analysis_screen.dart` — branch on `isStandardVariant`, pick the speed or variant icon/label accordingly, omit the speed label for standard games, reorder the rated/casual segment, and render the title as an icon + text `Row`.
- `test/view/game/game_screen_test.dart` — strengthen the analysis-from-finished-game test (was `find.textContaining("Rated")`) to assert the exact new text `2+1 • Rated`, locking in the format.

## Verification
```
flutter analyze lib/src/view/analysis/analysis_screen.dart test/view/game/game_screen_test.dart  # 0 issues
flutter test test/view/game/game_screen_test.dart  # all pass
```

Manual verification on a Pixel 10 emulator (lichess.dev, finished standard Rapid game):

| | App-bar header |
|---|---|
| BEFORE (`main`) | `10+0 • Rapid • Rated` (no leading icon) |
| AFTER (this PR) | `10+0 • Rated` (with leading speed icon, no speed label) |

### DIFF
--- a/lib/src/view/analysis/analysis_screen.dart
+++ b/lib/src/view/analysis/analysis_screen.dart
@@ -123,17 +123,16 @@ class _AnalysisScreenState extends ConsumerState<_AnalysisScreen>
         Widget appBarTitle;
         if (value.archivedGame != null) {
           final meta = value.archivedGame!.meta;
-          final isStandardVariant =
-              value.variant == Variant.standard || value.variant == Variant.fromPosition;
+          // On mobile space is constrained, so unlike the web we omit the speed for standard chess
+          final isStandardVariant = value.variant == .standard || value.variant == .fromPosition;
           final ratedOrCasual = meta.rated ? context.l10n.rated : context.l10n.casual;
-          final trailingLabel = isStandardVariant
-              ? meta.speed.label(context.l10n)
-              : value.variant.label(context.l10n);
-          final title =
-              '${value.archivedGame!.data.clockDisplay(context.l10n)} • $ratedOrCasual • $trailingLabel';
+          final clockDisplay = value.archivedGame!.data.clockDisplay(context.l10n);
+          final title = isStandardVariant
+              ? '$clockDisplay • $ratedOrCasual'
+              : '$clockDisplay • $ratedOrCasual • ${value.variant.label(context.l10n)}';
           final icon = isStandardVariant ? meta.speed.icon : value.variant.icon;
           appBarTitle = Row(
-            mainAxisSize: MainAxisSize.min,
+            mainAxisSize: .min,
             children: [
               Icon(icon),
               const SizedBox(width: 5.0),
--- a/lib/src/view/analysis/analysis_screen.dart
+++ b/lib/src/view/analysis/analysis_screen.dart
@@ -41,6 +41,7 @@ import 'package:lichess_mobile/src/widgets/adaptive_choice_picker.dart';
 import 'package:lichess_mobile/src/widgets/bottom_bar.dart';
 import 'package:lichess_mobile/src/widgets/buttons.dart';
 import 'package:lichess_mobile/src/widgets/feedback.dart';
+import 'package:lichess_mobile/src/widgets/misc.dart';
 import 'package:lichess_mobile/src/widgets/platform_context_menu_button.dart';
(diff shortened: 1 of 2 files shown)


CloudyDino:
## Manual verification screenshots

App bar comparison on a finished 10+0 Rapid standard game (Pixel 10 emulator, lichess.dev):

| Scenario | Before | After |
| --- | --- | --- |
| Rapid | Before: 10+0 • Rapid • Rated | After: 10+0 • Rated • Rapid (with speed icon) |
| Atomic | - | image |

HaonRekcef:
Thanks!

@CloudyDino check out: https://docs.flutter.dev/ui/dot-shorthands is pretty cool imo as the code is a bit less verbose

CloudyDino:
> Thanks!
> 
> @CloudyDino check out: https://docs.flutter.dev/ui/dot-shorthands is pretty cool imo as the code is a bit less verbose

Thanks for the heads-up! I tried the shorthand but looking at the codebase the verbose form is actually the prevailing style by a wide margin (e.g. `MainAxisSize.min` ~119 / `.min` ~23, `Variant.standard` etc. ~45 / shorthand 0, `Side.white|black` ~361 / shorthand ~10). I've reverted my dot-shorthand changes here to match the dominant style so this PR doesn't introduce inconsistency. Plus it helps to use the verbose style when reading through git diffs without the help of an IDE that might have type hinting

Happy to come back and run a repo-wide migration as a separate PR if we'd prefer to push the codebase toward shorthand globally.

HaonRekcef:
It was introduced fairly recently and we have just been writing the new code mostly with dot shorthands and never really changed the old code.

I prefer shorthands because otherwise there is a lot of repeated words.

```dart
mainAxisSize: MainAxisSize.min,

```
vs
```dart
mainAxisSize: .min,
```
What do you think @veloce?

veloce:
Yes we should write in shorthand for sure. But let's not bother to change the old code.

veloce:
@CloudyDino can we deviate a little from web here?

We are on mobile, space is constrained.

IMO in `10+0 • Rated • Rapid`, the `Rapid` is unnecessary. The 10+0 part is enough and there is the icon. Let's just add the perf if it is a variant different from standard chess I'd say.

ijm8710:
> @CloudyDino can we deviate a little from web here?
> 
> 
> 
> We are on mobile, space is constrained.
> 
> 
> 
> IMO in `10+0 • Rated • Rapid`, the `Rapid` is unnecessary. The 10+0 part is enough and there is the icon. Let's just add the perf if it is a variant different from standard chess I'd say.

> Let's just add the perf if it is a variant different from standard chess I'd say.

@veloce should we do the same for the board when it shows intragame. I believe if you're playing a 960 game live for example it'll show the icon but doesn't say the variant name.

veloce (approved):

PR #3320, opened 2026-06-11
Show opening name in broadcast and study opening explorer

### LINKED ISSUE #3190: Request to add opening explorer name for broadcasts
To me this would be a more coveted request. Broadcast analysis view is almost identical to the analysis board for a regular player game, but broadcasts omit this while the regular analysis have this

CloudyDino:
Fixes #3190.

The Opening Explorer tab inside a broadcast game now shows the opening name (`OpeningNameHeader`) above the move table, matching the behaviour of regular analysis and study game analysis. Previously only an unnamed move table was rendered.

### Changes
- `BroadcastAnalysisState` gains `currentBranchOpening`, the last named opening reached on the current branch (used as a fallback when the active node is not itself an opening transposition).
- `BroadcastAnalysisController` looks up openings via the local sqflite opening DB:
  - Prefetches openings for the mainline (first 30 plies) at build time.
  - On every `_setPath` it walks the active branch to find the most recent named opening and triggers a per-node fetch for the new node.
- `_OpeningExplorerTab` now passes the resolved opening to `ExplorerView`, gated on `kOpeningAllowedVariants` so the header is hidden for variants where the explorer is not applicable.

No network calls were added; opening lookups go through the existing `OpeningService` against the bundled openings DB.

### Testing
- `flutter test test/view/broadcast/` (24 tests, all pass)
- `flutter test test/model/broadcast/` 
- Manual verification on a local lila-docker broadcast (Sicilian Taimanov position): explorer tab now displays `B47 Sicilian Defense: Taimanov Variation, Bastrikov Variation` above the move table; on `main` the same position renders no header.

| Before | After |
| --- | --- |
| BEFORE-lila | AFTER |

### DIFF
--- a/lib/src/model/analysis/analysis_controller.dart
+++ b/lib/src/model/analysis/analysis_controller.dart
@@ -12,7 +12,7 @@ import 'package:lichess_mobile/src/model/analysis/analysis_player.dart';
 import 'package:lichess_mobile/src/model/analysis/analysis_preferences.dart';
 import 'package:lichess_mobile/src/model/analysis/common_analysis_state.dart';
 import 'package:lichess_mobile/src/model/analysis/forecast.dart';
-import 'package:lichess_mobile/src/model/analysis/opening_service.dart';
+import 'package:lichess_mobile/src/model/analysis/opening_explorer_mixin.dart';
 import 'package:lichess_mobile/src/model/analysis/server_analysis_mixin.dart';
 import 'package:lichess_mobile/src/model/analysis/server_analysis_service.dart';
 import 'package:lichess_mobile/src/model/common/chess.dart';
@@ -127,7 +127,10 @@ void clearSavedStandaloneAnalysis() {
 }
 
 class AnalysisController extends AsyncNotifier<AnalysisState>
-    with EngineEvaluationMixin, ServerAnalysisMixin<AnalysisState>
+    with
+        EngineEvaluationMixin,
+        ServerAnalysisMixin<AnalysisState>,
+        OpeningExplorerMixin<AnalysisState>
     implements PgnTreeNotifier {
   AnalysisController(this.options);
 
@@ -273,34 +276,17 @@ class AnalysisController extends AsyncNotifier<AnalysisState>
           }
           if (isMainline) {
             mainlinePath = mainlinePath + branch.id;
-            if (branch.position.ply <= 30) {
-              openingFutures.add(_fetchOpening(branch.position.fen, mainlinePath));
+            final openingFuture = fetchMainlineOpening(branch, mainlinePath);
+            if (openingFuture != null) {
+              openingFutures.add(openingFuture);
             }
           }
         },
       ),
     };
 
-    // wait for the opening to be fetched to recompute the branch opening
-    Future.wait(openingFutures)
-        .then((list) {
-          bool hasOpening = false;
-          for (final updated in list) {
-            if (updated != null) {
-              hasOpening = true;
-              final (path, opening) = updated;
-              _root.updateAt(path, (node) => node.opening = opening);
-            }
-          }
-          return hasOpening;
-        })
-        .then((hasOpening) {
-          if (hasOpening) {
-            scheduleMicrotask(() {
-              _setPath(state.requireValue.currentPath);
-            });
-          }
-        });
+    // wait for the openings to be fetched to recompute the branch opening
+    applyFetchedOpenings(openingFutures);
 
     final currentPath = switch (options) {
       Standalone() when _savedStandalone != null => _savedStandalone!.path,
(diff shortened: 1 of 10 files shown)


veloce (commented):
This is essentially a code duplication from analysis controller.

And it could be added to studies too.

We could have `mixin` here (you already have examples of mixin used in analysis controllers), it should not be too hard to do. And add it to the study opening explorer at the same time.

lib/src/model/study/study_controller.dart
```diff
@@ -176,6 +188,17 @@ class StudyController extends AsyncNotifier<StudyState>
     const currentPath = UciPath.empty;
     Move? lastMove;
 
+    final openingFutures = <Future<(UciPath, FullOpening)?>>[];
```
veloce: I think we can put that in the mixin as well. I think it is achievable like that:

```dart
// opening_explorer_mixin.dart

void runBuild() {
    super.runBuild();

    // after build
    final openingFutures = <Future<(UciPath, FullOpening)?>>[];
    UciPath mainlinePath = UciPath.empty;
      for (final branch in positionTree.mainline) {
        mainlinePath = mainlinePath + branch.id;
        final openingFuture = fetchMainlineOpening(branch, mainlinePath);
        if (openingFuture == null) break;
        openingFutures.add(openingFuture);
      }

    // doesn't need to be public anymore 
     _applyFetchedOpenings(openingFutures);
}
```

So this block can be removed in broadcast controller as well, and fetching the openings in analysis controller is not necessary anymore.

lib/src/model/analysis/analysis_controller.dart
```diff
@@ -273,34 +276,17 @@ class AnalysisController extends AsyncNotifier<AnalysisState>
           }
           if (isMainline) {
             mainlinePath = mainlinePath + branch.id;
-            if (branch.position.ply <= 30) {
-              openingFutures.add(_fetchOpening(branch.position.fen, mainlinePath));
+            final openingFuture = fetchMainlineOpening(branch, mainlinePath);
```
veloce: I think we can remove this, see comment in study controller.

lib/src/model/analysis/analysis_controller.dart
```diff
@@ -273,34 +276,17 @@ class AnalysisController extends AsyncNotifier<AnalysisState>
...
-        })
-        .then((hasOpening) {
-          if (hasOpening) {
-            scheduleMicrotask(() {
-              _setPath(state.requireValue.currentPath);
-            });
-          }
-        });
+    // wait for the openings to be fetched to recompute the branch opening
+    applyFetchedOpenings(openingFutures);
```
veloce: Should be handled in the mixin

lib/src/model/broadcast/broadcast_analysis_controller.dart
```diff
@@ -130,16 +130,17 @@ class BroadcastAnalysisController extends AsyncNotifier<BroadcastAnalysisState>
       UciPath mainlinePath = UciPath.empty;
       for (final branch in _root.mainline) {
         mainlinePath = mainlinePath + branch.id;
-        if (branch.position.ply > 30) break;
-        openingFutures.add(_fetchOpening(branch.position.fen, mainlinePath));
+        final openingFuture = fetchMainlineOpening(branch, mainlinePath);
```
veloce: Should be in the mixin

lib/src/model/analysis/opening_explorer_mixin.dart
```diff
@@ -0,0 +1,142 @@
...
+    return fetchOpening(branch.position.fen, mainlinePath);
+  }
+
+  /// Awaits the [openingFutures] collected during build, stores the fetched
+  /// openings on the tree, and refreshes the current branch opening if any
+  /// opening was found.
+  ///
+  /// This is fire-and-forget: it must not block the build, since openings are a
+  /// non-critical enhancement.
+  void applyFetchedOpenings(List<Future<(UciPath, FullOpening)?>> openingFutures) {
```
veloce: Should be private in the mixin, if we fetch the opening there.

lib/src/model/analysis/opening_explorer_mixin.dart
```diff
@@ -35,12 +35,17 @@ mixin OpeningExplorerMixinState {
 /// [OpeningExplorerMixinState.currentBranchOpening] is recomputed whenever the
 /// current path changes or a new opening is fetched.
 ///
+/// The build-time mainline opening fetch is centralized here: this mixin
+/// overrides [runBuild] to schedule [initMainlineOpenings] once the initial
+/// state has been built. Controllers therefore do not need to fetch the
```
veloce: Claude is always a bit verbose in its comments, I think this one is useless, let's remove it.

CloudyDino: Sorry about that, I had claude working in the background in auto mode and I didn't notice it pushed! I'll take a look over what it wrote and fix it before I make the PR ready for review again.

veloce: The rest looked good but thanks for verifying again!

veloce (approved):
Great stuff!

PR #3324, opened 2026-06-13
Add same side over the board mode

### LINKED ISSUE #3076: Add "same side" mode for OTB/local play
Currently it's only possible to have mirrored pieces or to flip the board between moves. I think it makes sense to also have a mode for having the player view always at the bottom, so people can pass the phone back and forth and have their view look normal for their turn. 

This may require rework of the menu options too so it's clear.

OddSteak:
Closes #3076
I added a same side over the board mode which flips board after every turn to make the color to play move to the bottom, this makes it easier to pass and play for people sitting side by side the need for which was felt strongly while I wanted to play chess with my buddy sitting beside me on a long flight.

I have added tests to make sure the widgets work correctly.

the flip board option is disabled in the same side mode as the orientation is determined by the turn.

https://github.com/user-attachments/assets/31a2d022-8080-4102-9286-c10d39f162b7

### DIFF
--- a/lib/src/model/over_the_board/over_the_board_preferences.dart
+++ b/lib/src/model/over_the_board/over_the_board_preferences.dart
@@ -38,6 +38,10 @@ class OverTheBoardPreferencesNotifier extends Notifier<OverTheBoardPrefs>
     return fetch();
   }
 
+  Future<void> toggleFlipOpponentInfo() {
+    return save(state.copyWith(flipOpponentInfo: !state.flipOpponentInfo));
+  }
+
   Future<void> setMyView(OverTheBoardMyView myView) {
     return save(state.copyWith(myView: myView));
   }
@@ -77,6 +81,13 @@ enum OverTheBoardMyView {
     flipPieces => 'Flip pieces after move',
     flipBoard => 'Same side (flip board after move)',
   };
+
+  /// Whether the "flip opponent info" setting is user-configurable for this view.
+  ///
+  /// Only [symmetricPieces] exposes a toggle. The other views have fixed
+  /// behavior: [whiteBottom] always flips the opponent info, [flipBoard] never
+  /// does, and [flipPieces] flips both info bars to face whoever is to move.
+  bool get canConfigureFlipOpponentInfo => this == symmetricPieces;
 }
 
 @Freezed(fromJson: true, toJson: true)
@@ -86,12 +97,14 @@ sealed class OverTheBoardPrefs with _$OverTheBoardPrefs implements Serializable 
   static const _defaultTimeIncrement = TimeIncrement(300, 3);
 
   const factory OverTheBoardPrefs({
+    required bool flipOpponentInfo,
     @Default(OverTheBoardMyView.whiteBottom) OverTheBoardMyView myView,
     @Default(TimeControlType.clock) TimeControlType timeControlType,
     @Default(OverTheBoardPrefs._defaultTimeIncrement) TimeIncrement timeIncrement,
   }) = _OverTheBoardPrefs;
 
   static const defaults = OverTheBoardPrefs(
+    flipOpponentInfo: true,
     myView: OverTheBoardMyView.whiteBottom,
     timeControlType: TimeControlType.clock,
     timeIncrement: _defaultTimeIncrement,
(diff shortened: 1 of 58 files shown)


ijm8710:
Nice pr!

One potential comment. The flipped opponent info is not only for the main mode. It could also be applied to symmetric piece mode too.

Hence, I think it may make more sense to have settings reworked.

First a 3 option chooser between main mode, symmetric mode and the mode you just added.

And then below that, a toggle to have "flip opponent info on/off"

I think this toggle would be disabled from toggling off (hence always on) for the new mode you're adding because there would never be a reason to not have it on. Sorta like how it has a shaded out toggle ability [here](https://github.com/user-attachments/assets/d33e5cf0-749a-462f-84b3-254e330a69a5) for play against the computer.

Thoughts?

OddSteak:
@ijm8710 

that's a good catch! didn't think about that honestly

to be clear there are four modes now, here's how I think it should work, let me know if it makes sense -

1. white at the bottom - this is the old default mode, I think the toggle shouldn't be visible in this mode and the opponent's info should always be flipped because this mode only makes sense if the players are sitting opposite each other
2. symmetric pieces - toggle works as expected
3. flip pieces after move - there should be no toggle, both the players' info should be facing towards the player whose turn it is
4. flip board after move - no option to flip and both the players' info on the same side

I have implemented this, I reckon this is the most intuitive way to go, let me know what you think

ijm8710:
Makes sense but part of me wonders if we have too many options. I definitely agree with adding this new mode but I don't see the point in symmetrical and wonder if Amy others are worth sunsetting as well but that's a question for Veloce once he reviews the pr

veloce:
@OddSteak can you please remove the translations you added before I review this further. As per contributing guide we don't write translation on first intent. Let's use hard coded English string for now.

Also can you please solve the conflicts. Thanks.

OddSteak:
> @OddSteak can you please remove the translations you added before I review this further. As per contributing guide we don't write translation on first intent. Let's use hard coded English string for now.
> 
> Also can you please solve the conflicts. Thanks.

done, sorry for the confusion

lib/src/model/over_the_board/over_the_board_preferences.dart
```diff
@@ -69,22 +69,43 @@ enum TimeControlType {
   }
 }
 
+enum OverTheBoardMyView {
+  whiteBottom,
```
tom-anders, line 73:
```suggestion
  whiteAtTheBottom,
```

would make this a bit more readable when this pops up in other parts of the code IMO

lib/src/model/over_the_board/over_the_board_preferences.dart
```diff
@@ -69,22 +69,43 @@ enum TimeControlType {
...
 sealed class OverTheBoardPrefs with _$OverTheBoardPrefs implements Serializable {
   const OverTheBoardPrefs._();
 
   static const _defaultTimeIncrement = TimeIncrement(300, 3);
 
   const factory OverTheBoardPrefs({
-    required bool flipPiecesAfterMove,
-    required bool symmetricPieces,
+    required bool flipOpponentInfo,
+    @Default(OverTheBoardMyView.whiteBottom) OverTheBoardMyView myView,
```
tom-anders, line 101:
```suggestion
    @Default(OverTheBoardMyView.whiteBottom, unknownEnumValue: OverTheBoardMyView.whiteBottom) OverTheBoardMyView myView,
```

IIRC this is needed to avoid problems when the enum's fields are changed in future versions

tom-anders (commented):
Nice work! Only two small comments from my side.

ijm8710:
Can we move "flip pieces after move" to 2nd since it's mostly a derivative of the 1st option "white at the bottom"


I would suggest just removing symmetric. Seems pointless with the availability of the other options. If players wants to sit across from each other they already have the choice to have the pieces each face them or flip after each turn.

shirshov:
Frankly, I find the proposed redesign a bit confusing. How about we do something like this instead? (Copy needs some work.) The four choices are mutually exclusive, the selected one is highlighted. And I agree with @ijm8710 that symmetrical pieces option can be removed.

Screenshot 2026-08-26 at 9 03 26 am

PR #3327, opened 2026-06-14
Fix 2 bugs in evaluation_service, with tests

ghost:
The old code will fail on the two new tests.

1. **`EngineEvaluationNotifier`:** Fixed the immediate, visible bug where the UI got stuck displaying old positions. In the old implementation's `_listener()`, if the service updated its state to a position or game that didn't match the current screen's filters, the notifier simply ignored the update. It left the UI stuck displaying the last known valid evaluation from the previous game.
2. **`_emitEval()`:** Fixed the silent data corruption race condition where late engine frames hijacked new positions.
3. **`_onThrottleExpired()`:** Added defensive insurance ensuring that when a user says "quit," the evaluation pipeline is dead silent, with zero ghost frames leaking out.
4. **Cached Listenable Instance:** Stored `_listenable` as a class property rather than a local variable in `build()`. This eliminates the need to call `ref.read()` inside the async `_listener` callback, avoiding potential edge-case lifecycle crashes or redundant provider lookups during provider disposal.
5. I like using `if (!ref.mounted) return;` better than wrapping everything in a mounted block and indenting it.[ Like the examples here.](https://dart.dev/tools/linter-rules/use_build_context_synchronously) But if you prefer it the other way I won't change it any more files I edit.

### DIFF
--- a/lib/src/model/engine/evaluation_service.dart
+++ b/lib/src/model/engine/evaluation_service.dart
@@ -483,6 +483,8 @@ class EvaluationService {
     final pending = _pendingEvalResult;
     if (pending != null) {
       _pendingEvalResult = null;
+      // Drop results if they are flagged to be discarded
+      if (_discardEvalResults) return;
       _emitEval(pending);
       // Start new throttle window for trailing emission
       _evalThrottleTimer = Timer(kEngineEvalEmissionThrottleDelay, _onThrottleExpired);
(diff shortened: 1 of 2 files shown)


veloce (approved):
Thanks! There was clearly a regression concerning the evals emitted for previous position. It's good to have a test now.

PR #3329, opened 2026-06-14
Guard async callbacks with ref.mounted to prevent use-after-dispose

CloudyDino:
## What

Several fire-and-forget async callbacks in Riverpod notifiers can run after their notifier has been disposed and then access `state`/`ref`, throwing an unhandled `Cannot use the Ref of ... after it has been disposed`. This happens when the user leaves a screen while a non-cancellable async operation (socket `firstConnection`, an opening fetch, a post-game-data fetch, or a delayed/timer callback) is still in flight.

This adds the project's existing `if (!ref.mounted) return;` guard to the callbacks that lacked it:

- **`analysis_controller`** — socket `firstConnection.whenComplete` (→ `requestEval`), the opening-fetch result microtask (→ `_setPath`), and `_fetchOpening().then` (→ `_updateOpening`)
- **`study_controller` / `retro_controller`** — socket `firstConnection.then` (→ `requestEval`)
- **`game_controller`** — the post-game-data `.then` (→ `state`) and the end-of-game dong-sound `Timer` (→ `ref.read(soundServiceProvider)`)
- **`storm_controller`** — the move-feedback `Future.delayed` (→ `ref.read` + `state`)

Timers and stream subscriptions that are already cancelled in `onDispose` cannot fire after disposal and are left unchanged.

## Why

This matches the guard convention already used elsewhere in the codebase (e.g. `evaluation_mixin.dart`, `broadcast_analysis_controller.dart`, `online_friends.dart`) and aligns with Riverpod's documented `Ref.mounted` usage.

## Testing

- `flutter analyze` clean, `dart format` clean.
- Relevant widget tests pass (analysis, retro, storm, study, board editor, opening).
- Reproduced the exception live on an Android emulator (board editor → analyze → leave quickly), then verified the guard eliminates it: with the device offline the analysis socket `firstConnection` stays pending until its 3s timeout, so `whenComplete` fires after the controller is disposed on every cycle — 0 `Cannot use the Ref` / `Unhandled Exception` across the runs after the fix.

### DIFF
--- a/lib/src/model/analysis/analysis_controller.dart
+++ b/lib/src/model/analysis/analysis_controller.dart
@@ -297,6 +297,7 @@ class AnalysisController extends AsyncNotifier<AnalysisState>
         .then((hasOpening) {
           if (hasOpening) {
             scheduleMicrotask(() {
+              if (!ref.mounted) return;
               _setPath(state.requireValue.currentPath);
             });
           }
@@ -379,6 +380,7 @@ class AnalysisController extends AsyncNotifier<AnalysisState>
         .timeout(const Duration(seconds: 3))
         .onError((_, _) {})
         .whenComplete(() {
+          if (!ref.mounted) return;
           if (state.requireValue.isEngineAvailable(evaluationPrefs)) {
             requestEval();
           } else if (options case ActiveCorrespondenceGame(:final gameFullId)) {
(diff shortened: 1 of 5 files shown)


veloce:
@CloudyDino I assume you forgot to mark it as ready for review?

CloudyDino:
> @CloudyDino I assume you forgot to mark it as ready for review?

While I was working on other PRs, Claude found this issue and tried to fix it. I don't fully understand the issue yet so I didn't mark it for review, but if everything looks good to you, feel free to review!

veloce (approved):

Issue #2988, opened 2026-04-15
Tracker for deep links remaining  to be supported


ijm8710:
Tom agreed would help to have a tracker for this.

- [x] Profiles https://lichess.org/@/kupggstest

- [ ] link to overall Broadcast (round-specific links work tho which is weird) -https://lichess.org/broadcast/fide-candidates-2026-women/xj4qM8Nw
- [x] Watch TV https://lichess.org/tv/bullet
- [x] Watch friend https://lichess.org/@/Ceres_D/tv
- [ ] Reusable challenge links (one time links work now tho) https://lichess.org/?user=kupggstest&variant=standard&gameMode=casual&time=unlimited#friend

tom-anders:
>       [ ]  link to overall Broadcast (round-specific links work tho which is weird) -https://lichess.org/broadcast/fide-candidates-2026-women/xj4qM8Nw

how can I even obtain such a link? on the website it also goes to the first round immediately

ijm8710:
> 
> >       [ ]  link to overall Broadcast (round-specific links work tho which is weird) -https://lichess.org/broadcast/fide-candidates-2026-women/xj4qM8Nw
> 
> how can I even obtain such a link? on the website it also goes to the first round immediately
> 
> 

Top link here https://lichess.org/broadcast/european-individual-chess-championship-2026--boards-1-66/JWIgkNkS


Seems it goes to round 1 when broadcast is completed and current round if active.
Not sure if that makes sense but the focal point is that it doesn't deep link regardless

![image](https://github.com/user-attachments/assets/8bc60016-4344-4562-9e7c-32a190bc3a4c)

PR #3330, opened 2026-06-14
Add board editor deep link support

CloudyDino:
part of https://github.com/lichess-org/mobile/issues/2988

## What

Adds deep link support for the board editor, matching lichess.org URLs:

- `lichess.org/editor` → opens the editor at the default starting position
- `lichess.org/editor/<fen>?color=<side>` → opens the editor at the given position and orientation

For example, `https://lichess.org/editor/rnbqkbnr/pppppppp/8/8/4P3/8/PPPP1PPP/RNBQKBNR_w_KQkq_-_0_1?color=black`.

## How

- **`app_links_service.dart`** — new `case 'editor'` in `resolveAppLinkUri`. A FEN's rank separators (`/`) and metadata spaces are encoded in the lichess URL as path separators and `_` respectively, so the handler rejoins `pathSegments[1..]` with `/` and restores the spaces to reconstruct the FEN. `?color=black` sets the orientation (anything else → white).
- **`board_editor_controller.dart`** — adds an `initialOrientation` field to `BoardEditorControllerParams`; `build()` uses it (defaulting to white).
- **`AndroidManifest.xml`** — registers `/editor` and `/editor/.*` App Links path patterns.

### Edge cases
- **Empty FEN** (bare `/editor`, or a trailing-slash `/editor/`) → default starting position, no error.
- **Invalid FEN** (e.g. `/editor/not-a-valid-fen`) → opens the editor at the default position **and** shows an `Invalid FEN` snackbar, rather than failing silently or refusing to navigate.

### Also fixes a pre-existing bug
A link that **cold-starts** the app was handled twice — `getInitialLink()` and the first event of `uriLinkStream` both delivered it. This was masked for navigation by `_pushDeepLinkRoute`'s same-screen dedup, but the new invalid-FEN snackbar (snackbars queue rather than dedup) surfaced it as a duplicate. Fixed by following [app_links' documented pattern](https://pub.dev/packages/app_links#getting-started) of subscribing to `uriLinkStream` alone (it already emits the cold-start link as its first event), instead of also calling `getInitialLink()`.

## Server-side dependency

App Links/universal-link verification for `/editor` lives in the `.well-known` files served from `lichess-org/lichess-org.github.io`. Companion PR: **lichess-org/lichess-org.github.io#6** (adds `/editor` to `apple-app-site-association` and `assetlinks.json`). Until that merges, `/editor` links open the app on Android ≤14 only; iOS and Android 15+ fall back to the browser.

## Testing

- Unit tests in `app_links_service_test.dart` cover: bare `/editor`, trailing slash, FEN reconstruction, `?color=black`, and invalid FEN.
- Verified on an Android emulator (API 36) by firing each link: correct position, orientation, the invalid-FEN snackbar, and — on cold start — the editor opening with a single (no longer duplicated) snackbar.

### Demo

https://github.com/user-attachments/assets/071253b9-140a-4124-abe4-54338d871fa8

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -123,6 +123,10 @@
                 <data android:pathPattern="/broadcast/.*/.*/........" />
                 <data android:pathPattern="/broadcast/.*/.*/......../........" />
 
+                <!-- Board editor: bare /editor and /editor/<fen> (FEN ranks contain '/'). -->
+                <data android:pathPattern="/editor" />
+                <data android:pathPattern="/editor/.*" />
+
                 <!--
                   Game or challenge link: any 8-char path (e.g. /abcd1234).
                   pathPattern cannot exclude known static 8-char routes (e.g. /analysis,
--- a/lib/src/app_links_service.dart
+++ b/lib/src/app_links_service.dart
@@ -58,28 +58,33 @@ class AppLinksService {
   StreamSubscription<Uri>? _linkSubscription;
 
   Future<void> start() async {
-    // Handle the link that cold-started the app (if any) after the first frame
-    // so the navigator is ready. Push without animation — the user launched the
-    // app via this link so the target screen should just be there.
-    WidgetsBinding.instance.addPostFrameCallback((_) async {
-      try {
-        final initialUri = await _appLinks.getInitialLink();
-        if (initialUri != null) {
-          await _handleUri(initialUri, animated: false);
-        }
-      } catch (e, st) {
-        _logger.severe('Error handling initial app link: $e\n$st');
+    // app_links' uriLinkStream emits the link that cold-started the app as its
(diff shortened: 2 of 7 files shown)


veloce:
Thanks for the PR!

About that:

> A link that cold-starts the app was handled twice — getInitialLink() and the first event of uriLinkStream both delivered it. This was masked for navigation by _pushDeepLinkRoute's same-screen dedup, but the new invalid-FEN snackbar (snackbars queue rather than dedup) surfaced it as a duplicate. Fixed by following [app_links' documented pattern](https://pub.dev/packages/app_links#getting-started) of subscribing to uriLinkStream alone (it already emits the cold-start link as its first event), instead of also calling getInitialLink().

Since it was a non-detected bug, is it possible to have a non-regression test? thanks.

veloce (approved):
Thanks!

PR #3331, opened 2026-06-14
fix bug where the check indicator was not showing up when spectating a game by adding an optional position parameter to ReadOnlyBoardParams

### LINKED ISSUE #2128: Check indicator not shown when watching other players' games
The red check indicator is not shown when watching other players' games. This should probably be shown for clarity.

In the example below, white is clearly in check, but the red indicator is not visible.

image

OddSteak:
Closes #2128 

I have tested it by spectating a real game

image

### DIFF
--- a/lib/src/model/game/game_board_params.dart
+++ b/lib/src/model/game/game_board_params.dart
@@ -14,6 +14,7 @@ sealed class GameBoardParams with _$GameBoardParams {
     required Variant variant,
     required Pockets? pockets,
     Move? lastMove,
+    Position? position,
   }) = ReadonlyBoardParams;
 
   const factory GameBoardParams.interactive({
--- a/lib/src/model/game/game_board_params.dart
+++ b/lib/src/model/game/game_board_params.dart
@@ -10,11 +10,9 @@ sealed class GameBoardParams with _$GameBoardParams {
   const GameBoardParams._();
 
   const factory GameBoardParams.readonly({
-    required String fen,
     required Variant variant,
-    required Pockets? pockets,
+    required Position position,
     Move? lastMove,
-    Position? position,
   }) = ReadonlyBoardParams;
 
   const factory GameBoardParams.interactive({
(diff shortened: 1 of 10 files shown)


lib/src/model/game/game_board_params.dart
```diff
@@ -14,6 +14,7 @@ sealed class GameBoardParams with _$GameBoardParams {
     required Variant variant,
     required Pockets? pockets,
     Move? lastMove,
+    Position? position,
```
veloce: Let's simplify this by removing the `fen` and `pockets` parameters and keep the (non nullable) `position`.

veloce (commented):
I have an improvement suggestion, after that we're good. Thanks!

lib/src/model/game/game_board_params.dart
OddSteak: done! 

I had to add an EmptyBoardParams type for the emptyBoard constant since an empty board is not a valid dartchess position

lib/src/view/game/game_loading_board.dart
```diff
@@ -450,27 +450,37 @@ class StandaloneGameLoadingContent extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
+    final loadingFen = position?.fen;
+    Position? loadingPosition;
+    if (loadingFen != null) {
+      try {
+        loadingPosition = Chess.fromSetup(Setup.parseFen(loadingFen));
```
veloce: Why trying to parse again the fen to obtain a position? We already have the `position`.

lib/src/view/game/game_loading_board.dart
```diff
@@ -450,27 +450,37 @@ class StandaloneGameLoadingContent extends StatelessWidget {
...
-              // This leads to a "duplicate key" error, since chessground would try to highlight the same square twice.
-              // The dropped role does not matter, since we just use it for the square highlight.
-              NormalMove(:final from, :final to) when from == to => DropMove(
-                to: to,
-                role: Role.pawn,
-              ),
-              _ => lastMove,
-            },
-          ),
+          boardParams: loadingPosition == null
```
veloce, line 472: Using the already available `position` should be enough here. See previous comment.

lib/src/view/game/game_loading_board.dart
OddSteak: the position variable here is of type LoadingPosition which is loaded from the api (I think?). It only has the fen string so it needs to be parsed

veloce: Oh my bad. I couldn't see the type so I got confused.

How about renaming `LoadingPosition` to `LoadingParams` to avoid confusion then?

OddSteak: done, I assume you meant the variable name, not the type

lib/src/view/game/game_loading_board.dart
```diff
@@ -443,14 +443,14 @@ class _GameParamsDisplay extends StatelessWidget {
 }
 
 class StandaloneGameLoadingContent extends StatelessWidget {
-  const StandaloneGameLoadingContent({this.position, this.userActionsBar, super.key});
+  const StandaloneGameLoadingContent({this.loadingParam, this.userActionsBar, super.key});
 
-  final LoadingPosition? position;
+  final LoadingPosition? loadingParam;
```
veloce: Can you also change the name of the type? Thanks.

OddSteak: done, sorry for the confusion

lib/src/view/game/game_loading_board.dart
```diff
@@ -443,34 +443,44 @@ class _GameParamsDisplay extends StatelessWidget {
...
   final Widget? userActionsBar;
 
   @override
   Widget build(BuildContext context) {
-    final lastMove = position?.lastMove;
+    final loadingFen = loadingParam?.fen;
+    Position? loadingPosition;
+    if (loadingFen != null) {
+      try {
+        loadingPosition = Chess.fromSetup(Setup.parseFen(loadingFen));
```
veloce: I just realised this can fail if the fen is from another variant.

We actually need the variant in the `LoadingParam`, and then do
```dart
loadingPosition = Position.setupPosition(variant.rule, Setup.parseFen(initialFen), ignoreImpossibleCheck: true);
```

OddSteak: fixed

OddSteak: the CI errors are coming from the last merged pr in main, unrelated to my changes.

veloce: going to rebase your branch to main to run CI again and if it's good I'll merge it; thanks again

veloce (approved):
Thanks!

PR #3332, opened 2026-06-15
Fix crash in perf stats columns when a perf has zero games

### LINKED ISSUE #1513: Issue with null columns for perf stats
image

Username for reference in top left of image

CloudyDino:
## Problem

Fixes #1513.

On the perf stats screen, when a perf has a rating but **zero counted games** (`stat.count.all == 0`), the win/draw/loss percentage columns render as empty/grey boxes (error widgets).

The cause is a divide-by-zero in `_PercentageValueWidget`: the percentage is computed as `((numerator / denominator) * 100).round()`, and with `denominator == 0`, `0 / 0` is `NaN`. In Dart `NaN.round()` throws `Unsupported operation: Infinity or NaN toInt`, so each of the seven percentage cards (wins, draws, losses, rated, tournament, berserk, disconnections) throws during build and is replaced by an `ErrorWidget` (a grey box in release builds).

## Fix

Guard the division in `_getPercentageString` so it falls back to `0%` when the denominator is `0`.

## Test

Added a regression test that loads the screen for a perf with zero games and asserts no exception / no `ErrorWidget` is rendered. Verified the new test fails on the previous code (7 `NaN` exceptions) and passes with the fix.

### DIFF
--- a/lib/src/view/user/perf_stats_screen.dart
+++ b/lib/src/view/user/perf_stats_screen.dart
@@ -525,8 +525,8 @@ class _PercentageValueWidget extends StatelessWidget {
   const _PercentageValueWidget(this.value, this.denominator, {this.color, this.isShaded = false});
 
   String _getPercentageString(num numerator, num denominator) {
-    if (denominator == 0) return '0%';
-    return '${((numerator / denominator) * 100).round()}%';
+    final fraction = denominator == 0 ? 0 : numerator / denominator;
+    return NumberFormat.percentPattern().format(fraction);
   }
 
   @override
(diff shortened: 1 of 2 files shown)


lib/src/view/user/perf_stats_screen.dart
```diff
@@ -525,6 +525,7 @@ class _PercentageValueWidget extends StatelessWidget {
   const _PercentageValueWidget(this.value, this.denominator, {this.color, this.isShaded = false});
 
   String _getPercentageString(num numerator, num denominator) {
+    if (denominator == 0) return '0%';
```
veloce: Thanks for the fix. Now I realise the number is not correctly localised. Can you fix that at the same time by using a `NumberFormat`?

CloudyDino: Updated!

veloce (approved):

PR #3339, opened 2026-06-15
Add clock delay option to the over-the-board clock tool

CloudyDino:
## Summary

Adds a **clock delay** (simple/US delay) option to the over-the-board clock tool. Each side can be given a delay that holds its main clock for that long before it starts counting down — so a move played within the delay window costs no time.

Delay is only offered in the **local clock tool**; lichess online play does not support it, so the time-control modal gates the delay slider behind a `showDelay` flag and online flows (lobby/challenge) are unaffected.

## What's included

- **Per-side delay**, configurable from each player's tune sheet, and **for both sides at once** via the shared time-control modal's Custom section (new "Delay in seconds" slider).
- A static **`d{n}` chip** shown before the game starts, indicating the configured delay.
- Delay **behaviour**: the active side's main clock is held for the delay before ticking; **pause/resume carries over** the remaining delay rather than restarting or discarding it.
- `TimeIncrement` gains a named `delay` field; `ChessClock` gains a `pause()` (split from `stop()`) that returns the remaining delay for carry-over.

## Incidental cleanup (separate commits)

- `fix(clock)`: clip the time-control bottom sheets to their rounded corners (scrolled content was bleeding past the corners).
- `refactor`: extract a shared `NonLinearSliderTile` widget to de-duplicate the time-control sliders.

## Testing

- Unit tests: `ChessClock` holding the main clock for the start delay, not flagging during the delay, and `pause()` returning the remaining delay; `ClockToolController` applying delay across start/onTap/resume/reset and per side.
- Widget tests: the time-control modal delay slider (shown only when enabled, OK carries the delay) and the clock-tool `d{n}` chip.
- `flutter analyze` and `dart format` clean.

## Demo

https://github.com/user-attachments/assets/48f6f380-adb8-46fc-8233-2aeb699dfa36

### DIFF
--- a/lib/src/model/clock/chess_clock.dart
+++ b/lib/src/model/clock/chess_clock.dart
@@ -116,8 +116,6 @@ class ChessClock {
   ///
   /// Trying to start an already running clock on the same side is a no-op.
   ///
-  /// The [delay] parameter can be used to add a delay before the clock starts counting down. This is useful for lag compensation.
-  ///
   /// Returns the think time of the active side before switching or `null` if the clock is not running.
   Duration? startSide(Side side, {Duration? delay}) {
     if (isRunning && _activeSide == side) {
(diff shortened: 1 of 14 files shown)


veloce:
Have you checked this one before? https://github.com/lichess-org/mobile/pull/3122

looks like it's a similar work

CloudyDino:
I didn't see that PR, but it's exactly what I needed!

PR #3352, opened 2026-06-18
update lichessFlagSrc to use .webp format for flag assets

SergioGlorias:
require: lichess-org/lila#20697


SergioGlorias:
Already merge in lila

PR #3354, opened 2026-06-18
Sort language picker by native name

### LINKED ISSUE #3341: Sort language options alphabetically
Currently language options in Lichess are sorted by an internal two letter ID, like "en", "es", "ru", etc. This simplifies the sorting of the languages, but for the human user it is confusing and hard to navigate. Different languages written in different scripts get mixed together, like latin, cyrillic, arabic, etc. Some users may find it hard to find their native language in the list. The current way of doing things is not ideal. The better way is to order the language options by their language name in Unicode. It automatically sorts alphabetically and also groups languages written in a similar script together. It is easy to implement, just sort by the language name itself, like "English", "Français", "Русский", "한국어", etc.

Note: this already has been implemented for the lichess website here: https://github.com/lichess-org/lila/pull/20660
It would make sense to port this change to the mobile app as well.

Image

CloudyDino:
## What

Sorts the language selector (Settings → Language) by each locale's **native display name** using Unicode ordering, instead of the current ordering by two-letter language code.

## Why

Closes #3341.

Ordering by language code interleaves different scripts (Latin, Cyrillic, Arabic, CJK), making languages hard to find. Sorting by native name groups similar scripts together and is far easier to scan. This mirrors the website's behavior from lichess-org/lila#20660, where the language list is ordered by native name.

## How

- Added `localesSortedByLocalizedName(...)` next to `localeToLocalizedName` in `lib/src/utils/l10n.dart`, which sorts locales by their localized name with a plain `String.compareTo` (Unicode/code-unit order — no ICU collator needed, matching the issue's request).
- The Android language picker in `settings_screen.dart` now feeds the sorted list into `showChoicePicker`. (iOS opens the system locale settings, so there's no in-app list to sort there.)

## Tests

Added unit tests in `test/utils/l10n_test.dart` covering: native-name ordering across scripts, that the input list is not mutated, and that all supported locales come back in sorted order.

[language ordering.webm](https://github.com/user-attachments/assets/16fd170d-2c3b-4252-9eab-c42a6eb88cb8)

### DIFF
--- a/lib/src/utils/l10n.dart
+++ b/lib/src/utils/l10n.dart
@@ -191,3 +191,11 @@ String localeToLocalizedName(Locale locale) => switch (locale) {
   Locale(languageCode: 'zu') => 'isiZulu',
   _ => locale.toString(),
 };
+
+/// Returns the given [locales] sorted alphabetically by their localized
+/// (native) name, following Unicode ordering.
+///
+/// This groups locales written in the same script together, instead of the
+/// default ordering by language code.
+List<Locale> localesSortedByLocalizedName(Iterable<Locale> locales) =>
+    [...locales]..sort((a, b) => localeToLocalizedName(a).compareTo(localeToLocalizedName(b)));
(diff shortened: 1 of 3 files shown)


veloce (approved):
Thanks!

PR #3355, opened 2026-06-19
Open patron page in external browser

### LINKED ISSUE #3345: Thank you message after donating
### What is the problem?

I donate monthly, being thankful for the work you do. Now every time after I open the app, I get the same thank you message that I have to click away. Your thanks is understood and appreciated, I'd just like to play chess. Username Koeiluh17. 

1. Open app
2. Thank you message appears every time I open the app
3. I have to 'go back' to get to the homepage to start playing. 

### App version

0.24.6

### Device Info

Samsung S25, android 8.0 version 16

### Additional information

_No response_

sawirricardo:
## Summary
- Add a shared patron page launcher that requests the external application launch mode.
- Use it from the Android donation entry points so the donation thank-you page is not kept on top of the app task.

Fixes #3345

## Testing
- `flutter analyze lib/src/utils/launch.dart lib/src/view/home/home_tab_screen.dart lib/src/view/more/more_tab_screen.dart lib/src/view/account/account_menu.dart test/utils/launch_test.dart`
- `flutter test test/utils/launch_test.dart`

### DIFF
--- a/lib/src/utils/launch.dart
+++ b/lib/src/utils/launch.dart
@@ -1,0 +1,13 @@
+import 'package:url_launcher/url_launcher.dart';
+
+const _patronUri = 'https://lichess.org/patron';
+
+typedef UrlLauncher = Future<bool> Function(Uri uri, LaunchMode mode);
+
+Future<bool> launchPatronPage({UrlLauncher? launcher}) {
+  return (launcher ?? _launchUrl)(Uri.parse(_patronUri), LaunchMode.externalApplication);
+}
+
+Future<bool> _launchUrl(Uri uri, LaunchMode mode) {
+  return launchUrl(uri, mode: mode);
+}
--- a/lib/src/view/account/account_menu.dart
+++ b/lib/src/view/account/account_menu.dart
@@ -14,6 +14,7 @@ import 'package:lichess_mobile/src/styles/lichess_icons.dart';
 import 'package:lichess_mobile/src/styles/styles.dart';
 import 'package:lichess_mobile/src/utils/http_network_image.dart';
 import 'package:lichess_mobile/src/utils/l10n_context.dart';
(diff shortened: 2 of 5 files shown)


lib/src/utils/launch.dart
veloce, line 1: `launch.dart` is not a very clear, better name it `launch_url.dart` so we know what it is about.

lib/src/utils/launch.dart
```diff
@@ -0,0 +1,13 @@
+import 'package:url_launcher/url_launcher.dart';
+
+const _patronUri = 'https://lichess.org/patron';
```
veloce, line 3: You could also declare a final and parse it once for all:

```suggestion
final _patronUri = Uri.parse('https://lichess.org/patron');
```

test/utils/launch_test.dart
veloce, line 1: Also rename it to launch_uri_test.dart

veloce:
@sawirricardo if you were able to reproduce this behaviour, I'd like to see a screenshot/screen recording of it. Thank you!

lib/src/view/account/account_menu.dart
```diff
@@ -237,7 +238,7 @@ class _AccountMenuScreenState extends ConsumerState<AccountMenuScreen> with Widg
                   ),
                   title: Text(context.l10n.patronDonate),
                   enabled: isOnline,
-                  onTap: () => launchUrl(Uri.parse('https://lichess.org/patron')),
+                  onTap: () => launchPatronPage(),
```
dev-shady, line 241: Since launchPatronPage has only optional parameters, I think we can remove closure wrapper ( () => ) . Avoiding unnecessary object instantiations on every widget build pass.

onTap: launchPatronPage

We can do the same in other invocations as well.

PR #3369, opened 2026-06-22
add a try catch for premoves

### LINKED ISSUE #3124: Premove bug
### What is the problem?

When I premove and then the opponent makes a move that invalidates the premove the premove is made two moves later. 

I’ve lost multiple games this week because of this. I premove (say bishop takes bishop after they take my pawn) but instead of taking the pawn they make another move. I play to respond to that move but then on my next move it applies the premove from the move before. 

My premove should only be valid for one move. After they move it shouldn’t be applied anymore. 

Wrote above 

### App version

0.23.5

### Device Info

iPhone / iOS 26

### Additional information

_No response_

ghost:
`moveToConfirm` was not being reset when an opponent made a move that invalidated our local premove. I added a check at the end of `case 'move' || 'drop':`

Closes #3124

### DIFF
--- a/lib/src/model/game/game_controller.dart
+++ b/lib/src/model/game/game_controller.dart
@@ -758,6 +758,14 @@ class GameController extends AsyncNotifier<GameState> with ChatMixin<GameState> 
           ref.read(ongoingGamesProvider.notifier).updateGame(gameFullId, newState.game);
         }
 
+        if (newState.moveToConfirm != null) {
+          try {
+            newState.game.lastPosition.makeSan(newState.moveToConfirm!);
+          } catch (_) {
+            newState = newState.copyWith(moveToConfirm: null);
+          }
+        }
+
         state = AsyncValue.data(newState);
 
       // End game event


veloce:
@freebeartogoodhome thanks for contributing by fixing bugs.

Now, each time you want to fix a bug that involves online game play, please add a test that proves that:
- the bug was here before
- the new code is actually fixing the bug described in the test scenario

veloce:
Even better: along the test, a screen recording proving the bug before, and another proving the fix after.

lib/src/model/game/game_controller.dart
```diff
@@ -758,6 +758,14 @@ class GameController extends AsyncNotifier<GameState> with ChatMixin<GameState>
           ref.read(ongoingGamesProvider.notifier).updateGame(gameFullId, newState.game);
         }
 
+        if (newState.moveToConfirm != null) {
+          try {
+            newState.game.lastPosition.makeSan(newState.moveToConfirm!);
+          } catch (_) {
+            newState = newState.copyWith(moveToConfirm: null);
```
veloce, line 765: moveToConfirm is already set to null line 707 so I believe this is unnecessary.

veloce:
Closing since this bug was already fixed (but the issue not closed).

lib/src/model/game/game_controller.dart
ghost: Okay, I see now it was in `tryExecutePremove`

Sorry about this one.

PR #3375, opened 2026-06-23
Use refresh icon for New Streak button in puzzle streak

Zinkelburger:
When using puzzle streak, I was really confused by the "Play" icon. I kept clicking it when I didn't mean to. I feel like a refresh icon would be less ambiguous that I am going to start a new session.

I've tested the change in Android Studio

### DIFF
--- a/lib/src/view/puzzle/streak_screen.dart
+++ b/lib/src/view/puzzle/streak_screen.dart
@@ -555,7 +555,7 @@ class _BottomBar extends ConsumerWidget {
                 : null,
             highlighted: true,
             label: context.l10n.puzzleNewStreak,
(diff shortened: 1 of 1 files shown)


veloce (approved):
Makes sense.

ijm8710:
> Makes sense.

It's minor but Not sure I agree. There should be no confusion; you're not accidentally ending the streak. The only time the play button shows is once you already messed up and ended the streak.

So you're pressing play to begin a new attempt with the play button for a new game consistent with web. There's no risk of messing up an active streak.

If anything maybe the trash can or resign flag that you see in puzzle storm. 

But I still think play is the correct icon since starting a new game is what the action is that is being solicited

PR #3377, opened 2026-06-23
chore: update Flutter 3.44.3 & deps, use private named parameters

lsaudon:
- dart sdk to 3.12.2 & fix `Private named parameters` when linter find issues
- flutter to 3.44.3
- flutter_riverpod: 3.3.1, the regression occurs starting with version 3.3.2
- cronet_http to 1.9.0
- firebase_messaging to 16.4.1

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -5,9 +5,9 @@ publish_to: "none"
 version: 0.25.1+002501 # See README.md for details about versioning
 
 environment:
-  sdk: ^3.11.5
+  sdk: ^3.12.2
   # this should be updated with every new flutter stable release
-  flutter: 3.44.2
+  flutter: 3.44.3
 
 dependencies:
   app_links: ^7.1.2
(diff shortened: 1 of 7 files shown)


pubspec.yaml
```diff
@@ -43,7 +43,7 @@ dependencies:
     sdk: flutter
   flutter_markdown_plus: ^1.0.7
   flutter_native_splash: ^2.4.8
-  flutter_riverpod: 3.2.1 # current 3.3.2 version has a breaking change
+  flutter_riverpod: 3.3.1 # 3.3.2 version has a breaking change
```
veloce, line 46: I don't see a stable 3.3.1 version published on pub.dev. How can this work?

lsaudon: I see here https://pub.dev/packages/flutter_riverpod/versions/3.3.1

Maybe error with cache ?

veloce: Got it, it's flutter_riverpod, not riverpod.

PR #3393, opened 2026-06-25
Puzzle streak: offline & previous puzzles

Zinkelburger:
I was trying to do puzzle streak on the train, and I realized it didn't have an offline option. I also kept wanting to view puzzles I got correct, but wasn't 100% sure of the solution.

Architectural:
1. Fetch 30 puzzles at a time with the new "apiMany"
2. Save them for offline use

Live puzzle:
1. Add previous button
2. Once we've failed the streak, change the weird refresh icon to be a "skip" icon, and label it "New Streak" to eliminate user confusion.

On past puzzles:
1. Remove unlabelled "<" ">" arrows. If the user wants to analyze, they can go to analysis.
2. Replace "<" ">" buttons with "previous" / "next" buttons

[Screen_recording_20260902_160358_gh.webm](https://github.com/user-attachments/assets/cc8375fb-7796-4bef-9385-b3ac71bc6cda)

---
AI Generated description:
## The commits

Five commits, best read in order.

**puzzle: fetch several puzzles by id in one request**

`PuzzleRepository.fetchMany` on top of `/api/puzzle/many`, plus `PuzzleStorage.cachedPuzzleIds` and `saveAll` so a caller can ask only for what isn't on disk and write the answer in one transaction. The endpoint drops ids it can't serve, so the response can be shorter than the request. Nothing calls it yet at this point in the series.

**puzzle: load a puzzle from the cache or the server in one place**

`PuzzleService.loadPuzzle` reads a puzzle from SQLite and falls back to the server on a miss; `cachePuzzle` writes one and logs a failed write instead of throwing, because the cache is an optimisation, not the source of truth. `puzzleProvider` and `solve()` go through these instead of open-coding the same thing.

**network: keep puzzle fetches out of the retry, and name the failures worth retrying**

`/api/puzzle/many` joins the batch endpoints in the "don't retry a 429" list: a prefetch that just got rate limited should stop, not double the burst. `isPermanentFailure` says whether a request is worth trying again later. A 4xx will fail the same way next time, except 429 and 401, which pass once the limit resets or the session is renewed. The streak uses it to decide what to drop and what to leave queued.

**streak: play offline from a local look-ahead buffer**

The controller reads puzzles from the cache first, and keeps 30 puzzles cached ahead of the current one with a single `/api/puzzle/many` request, refilling once fewer than 10 remain. Each refill starts where the cached range ends so the windows don't overlap. Refills are best-effort: a failed one backs off for a few advances, a 4xx stands the refill down for the rest of the run, and play continues one puzzle at a time from `/api/puzzle/{id}`, which isn't rate limited.

A solve whose next puzzle isn't available yet is saved as `advancePending` and counts towards the score from that moment, so cutting the run short or failing later doesn't replay a puzzle you already solved. `PuzzleStreak.score` is the one place that answers "how long is this run", for the tab badge, the counter on screen and the posted run. On reconnect the controller tries the advance once more; if that fails too, the bar offers a retry. Connectivity only reports online once lichess itself answers, so there's no captive portal to retry through. If the server refuses the next puzzle outright, the run ends with the solve counted and the score posted, the same way a saved run in that state is ended on the next visit.

A run that ends offline queues its score in the preferences. `StreakScoreSync` posts it on reconnect or on a cold start that is already online, so it doesn't sit there until the streak screen is opened again. Only the best pending run is kept (the server keeps the best score anyway), and a score the server rejects for good is dropped rather than re-posted forever.

The puzzle tab lets you open a saved streak offline, and the new streak button stays tappable offline so it can say why nothing happens instead of being greyed out.

**streak: review the solved puzzles of the run**

Previous/Next in the bottom bar walk the run puzzle by puzzle, and the feedback tile says which one is on the board ("Solved puzzle 2 of 4"). A reviewed puzzle gets its own `puzzleControllerProvider` key, so the live puzzle keeps its state and stays the only thing driving the streak. It carries no user id, otherwise each load would refresh the glicko through the batch endpoint.

Reviewing replays the puzzle: a right move is answered, a wrong one taken back. That needed a distinct `PuzzleMode.replay`. On the streak screen a puzzle that's over (won, lost or reviewed) keeps checking moves against the solution, whereas on the puzzle screen view mode is a free analysis board. Everything that used to care about "view mode, but on a streak" now just switches on the mode. The replay goes through the existing play branch of `onUserMove` rather than a second copy of the solution checking; only the error sound and the take-back differ.

Back from a review returns to the live puzzle, not to the puzzle tab. Starting a new streak from a review drops the review. The bar keeps one previous/next pair in every state, greyed out at the ends, so it doesn't change shape.

## Things worth knowing

- The run is saved to disk as soon as it starts, since it has to be there to resume offline. That means backing out of the first puzzle without playing it brings the same puzzle back next time.
- The controller tests poll with short real delays instead of `FakeAsync`, because they go through a real sqflite database.
- The old move-by-move Previous/Next after a loss is gone; the arrows now walk the run. The solution is still in the move list and on the analysis board.

## Tests

Widget tests on the streak screen for review, new streak from a review, a solve stranded offline, and a retry after a failed load while online. Controller tests for the cache-first loading, prefetch and refill windows, back-off and rate limiting, reconnect, the pending score queue, and a run whose puzzle the server refuses.

AI tools used: Claude Opus 5, Fable 5 and Fable 5.1.

### DIFF
--- a/lib/src/app.dart
+++ b/lib/src/app.dart
@@ -22,6 +22,7 @@ import 'package:lichess_mobile/src/model/correspondence/correspondence_service.d
 import 'package:lichess_mobile/src/model/log/app_log_service.dart';
 import 'package:lichess_mobile/src/model/message/message_service.dart';
 import 'package:lichess_mobile/src/model/notifications/notification_service.dart';
+import 'package:lichess_mobile/src/model/puzzle/streak_score_sync.dart';
 import 'package:lichess_mobile/src/model/settings/board_preferences.dart';
 import 'package:lichess_mobile/src/model/settings/general_preferences.dart';
 import 'package:lichess_mobile/src/model/study/study_preferences.dart';
@@ -182,6 +183,11 @@ class _AppState extends ConsumerState<Application> {
     ref.listenManual(connectivityChangesProvider, (prev, current) async {
       final prevWasOffline = prev?.value?.isOnline == false;
       final currentIsOnline = current.value?.isOnline == true;
+
+      // Post a streak score queued while offline, on reconnect or on the first online check.
+      if (currentIsOnline && (prevWasOffline || !_firstTimeOnlineCheck)) {
+        ref.read(streakScoreSyncProvider).flushCurrentUser().ignore();
+      }
 
       // Play registered moves whenever the app comes back online.
       if (prevWasOffline && currentIsOnline) {
--- a/lib/src/model/puzzle/puzzle_controller.dart
+++ b/lib/src/model/puzzle/puzzle_controller.dart
@@ -86,14 +86,10 @@ class PuzzleController extends Notifier<PuzzleState> {
     // update puzzles that are remaining in replay
     _replayRemaining = context.replayRemaining;
 
-    // play first move after 1 second
-    _firstMoveTimer = Timer(const Duration(seconds: 1), () {
-      _setPath(state.initialPath, firstMove: true);
-    });
-
     final initialPath = UciPath.fromId(_gameTree.children.first.id);
+    final initialNode = _gameTree.branchAt(initialPath);
 
-    return PuzzleState(
+    final loaded = PuzzleState(
       puzzle: context.puzzle,
       glicko: context.glicko,
       mode: PuzzleMode.load,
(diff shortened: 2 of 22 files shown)


veloce:
Sorry if I did not react to this earlier. I am not sure we want this because of the added complexity, but I'll have a look.

PR #3395, opened 2026-06-26
fix UserTV advancing to next game

HaonRekcef:
UserTV was not auto connecting to the next game automatically. There were two problems:
- The Queryparameter usertv? was not sent to the server, so the server did not know that it should send a `resync` event, when the user starts a new game.
- `resync/rematchTaken` reconnected to the same game instead of fetching the new game

Added a test that fails without both fixes.

Video: 

[Screencast_20260626_101113.webm](https://github.com/user-attachments/assets/54c6560f-0451-4bb6-8a6c-bf196ec89f52)


HaonRekcef:
To test the current and fixed behavior I recommend to open the LichessTV for Ultrabullet and then go to one of the players and then watch the games.

veloce:
Thanks!

PR #3403, opened 2026-06-28
Fix msg display

vinayak-vikram:
Fix text padding in each message bubble and timestamp display on the messages/inbox thing.
Before/after images attached.
Simulator Screenshot - iPhone 16 Pro - 2026-06-28 at 14 20 57
Simulator Screenshot - iPhone 16 Pro - 2026-06-28 at 14 37 05

### DIFF
--- a/lib/src/view/message/conversation_screen.dart
+++ b/lib/src/view/message/conversation_screen.dart
@@ -365,10 +365,11 @@ class _MessageBubble extends ConsumerWidget {
 
     return ChatBubbleContextMenu(
       message: message.text,
-      child: Align(
-        alignment: isMe ? Alignment.centerRight : Alignment.centerLeft,
-        child: ConstrainedBox(
-          constraints: BoxConstraints(maxWidth: MediaQuery.sizeOf(context).width * 0.85),
+      child: FractionallySizedBox(
+        alignment: isMe ? .centerRight : .centerLeft,
+        widthFactor: 0.85,
+        child: Align(
+          alignment: isMe ? .centerRight : .centerLeft,
           child: Container(
             margin: EdgeInsets.only(
               bottom: !isInGroup || isFirstInGroup ? 8 : 2,
--- a/lib/src/view/message/conversation_screen.dart
+++ b/lib/src/view/message/conversation_screen.dart
@@ -364,59 +364,61 @@ class _MessageBubble extends ConsumerWidget {
 
     return ChatBubbleContextMenu(
       message: message.text,
-      child: FractionallySizedBox(
+      child: Align(
         alignment: isMe ? Alignment.centerRight : Alignment.centerLeft,
-        widthFactor: 0.85,
-        child: Container(
-          margin: EdgeInsets.only(
-            bottom: !isInGroup || isFirstInGroup ? 8 : 2,
-            top: !isInGroup || isLastInGroup ? 8 : 2,
-            left: 8,
-            right: 8,
-          ),
-          padding: const EdgeInsets.symmetric(vertical: 10, horizontal: 14),
-          decoration: BoxDecoration(
-            color: _bubbleColor(context),
-            borderRadius: BorderRadius.only(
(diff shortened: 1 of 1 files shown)


lib/src/view/message/conversation_screen.dart
```diff
@@ -364,66 +365,100 @@ class _MessageBubble extends ConsumerWidget {
...
-                  ? _bubbleRadius
-                  : isInGroup && !isFirstInGroup
-                  ? _inGroupRadius
-                  : _bubbleRadius,
-              bottomRight: isMe
-                  ? isInGroup && !isFirstInGroup
-                        ? _inGroupRadius
-                        : _bubbleRadius
-                  : _bubbleRadius,
+        child: ConstrainedBox(
```
julien4215: I would keep the `FractionallySizedBox` widget and put the `Align` inside it. There is no need to use `ConstrainedBox`.

In `Align` widget you can use the dot shorthand notation with `.centerLeft` instead of `Alignment.centerLeft`.

(If we want to do it with constraints, the constraints can be put directly in the `Container` widget instead of using `ConstrainedBox`. To get the width, I think I would use `LayoutBuilder` for safety even though here the width should match the one from `MediaQuery`. If `MediaQuery` is kept, `widthOf` should be used instead of `sizeOf`.)

lib/src/view/message/conversation_screen.dart
```diff
@@ -364,66 +365,100 @@ class _MessageBubble extends ConsumerWidget {
...
+  Widget build(BuildContext context) {
+    final timeStyle = TextStyle(fontSize: 11, color: textColor.withValues(alpha: 0.6));
+    final spacer = WidgetSpan(
+      alignment: PlaceholderAlignment.baseline,
+      baseline: TextBaseline.alphabetic,
+      child: Opacity(
+        opacity: 0,
+        child: Padding(
+          padding: const EdgeInsets.only(left: 8),
+          child: Text(time, style: timeStyle),
```
julien4215, line 442: Why does `time` appear twice? Once here and once in the `Positioned` widget.

vinayak-vikram: for display reasons, the invisible one is just there to force the positioned one to be in the right spot (if the invisible one fits on the same line as the text, the positioned one overlaps, and if the invisible one is on the next line the positioned one is forced to the end of the line in order to render properly)

julien4215: I see. It avoids the text message overlapping with the text time.

lib/src/view/message/conversation_screen.dart
```diff
@@ -365,65 +366,100 @@ class _MessageBubble extends ConsumerWidget {
...
+      linkify(text, linkifiers: AppLinksService.kLichessLinkifiers),
+      style: TextStyle(color: textColor),
+      linkStyle: Styles.linkStyle,
+      onOpen: onLinkOpen,
+    );
+
+    return Stack(
+      children: [
+        RichText(text: TextSpan(children: [linkSpan, spacer])),
+        Positioned(right: 0, bottom: 0, child: Text(time, style: timeStyle)),
```
veloce, line 457: What if the message text goes full length to the right? I assume the time will overlap with the text.

veloce (commented):
It's rather an improvement than a fix (more compact view). See my comment about overlapping text.

lib/src/view/message/conversation_screen.dart
vinayak-vikram: the issue is that if the text and the timestamp were in a row (in multiline messages) the message text would have extra padding on the right because of the space allocated for the timestamp... is there another way to do this? i thought this would be better than that. there could also be some sort of dynamic layout determination i guess, but i was trying to avoid that.

veloce: I agree it looks better like that

veloce (approved):

PR #3405, opened 2026-06-29
Save clock tool's last used time preference

### LINKED ISSUE #3394: [Feature Requests] clock tool: remember last time selection
We are using the clock tool of the lichess mobile app a lot in our local pub tournaments. 
It would be useful for users that use the clock often, If the last time control was stored. It is frustrating having to re-configure the time control especially when using a custom time control. 

Current behavior: Default is 10 + 0
Better behavior: last selected time Control is selected. 

I think that the clock tool is a mobile only Feature, so i hope that i am in the correct place here.

dev-shady:
Closes #3394 

1. Created a new clock tool preferences file as there wasn't already.
2. Earlier 10sec was the fixed value used so kept same as the default config value.
3.  The UI is rendered by the state driven by ClockToolController. hence let it read the default value from preferences in build().
4. updateOptions and updateOptionsCustom of ClockToolController are the only two methods that update the clock tool state, so updated them to write to the prefs as well.

### Clock Preferences:

* Asymmetrical Tracking: Refactored ClockToolPrefs to store topTimeIncrement and bottomTimeIncrement separately. The tool now remembers separate settings for each clock across app sessions. 
* Helper Method: Added getStartTime(ClockSide) to ClockOptions to simplify and avoid repetitive logic within the controller.

### Emergency Threshold Bug Fix:

* Problem: Previously, a single shared _emergencyThreshold was used for both clocks. In games with different starting times, the low-time sound would trigger at incorrect intervals.

* Fix: Implemented a map-based _emergencyThresholds to track threshold separately for the two clocks. onClockEmergency now correctly references the active player's specific threshold.

### Clean-up & Dead Code Removal:

* updateDuration Removal: Identified and removed the updateDuration method. The function was unused in the codebase and contained a logical error where side increments were swapped. ClockSide bottom using topIncrement and ClockSide top using bottomIncrement which is wrong.

### Impact:

* Users can now play handicap games with remembered settings for both sides.
* Low-time warnings are now accurate for both players in asymmetrical games.

### Tested

Clocks separate default time tracking:
* Used central settings button to change both clocks time. The default time persists for both.
* Used top or bottom aligned settings button to change individual clocks time. The default time persists.
https://github.com/user-attachments/assets/35428792-f86b-4247-85f5-a33b188b2438

Emergency threshold bug (No Sound at Emergency threshold (10% of the total time)) :
https://github.com/user-attachments/assets/7120efc9-3bab-4a0c-a925-eac386693ed6

Emergency threshold bug fixed:
https://github.com/user-attachments/assets/3d8dbad4-f2f1-423e-a131-b8cecf502dc7

### DIFF
--- a/lib/src/model/clock/clock_tool_controller.dart
+++ b/lib/src/model/clock/clock_tool_controller.dart
@@ -33,7 +33,10 @@ class ClockToolController extends Notifier<ClockState> {
     ClockSide.top: false,
     ClockSide.bottom: false,
   };
-  late Duration _emergencyThreshold;
+  final Map<ClockSide, Duration> _emergencyThresholds = {
+    ClockSide.top: Duration.zero,
+    ClockSide.bottom: Duration.zero,
+  };
   DateTime? _delayStartedAt;
   Duration? _delayDuration;
   Duration? _pausedDelayRemaining;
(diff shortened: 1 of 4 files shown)


dev-shady:
@veloce need your review and inputs.

veloce:
@dev-shady not sure I saw that in the video:

> Meaning you can change one clock to be say 10 mins and other clock to be 5 mins and play but right now, only the last updated value will be used as the default value.

Anyway, I think that we should keep both clocks separated in case they have different time setup.

> We can keep separate defaults for the two clocks as well but need confirmation on the UX before doing it.

Why confirmation on the UX? the UX should not change. Instead of saving one TimeIncrement for both clocks, save two for each clock. There is no need for an UX or UI change.

dev-shady:
> @dev-shady not sure I saw that in the video:
> 
> > Meaning you can change one clock to be say 10 mins and other clock to be 5 mins and play but right now, only the last updated value will be used as the default value.
> 
> Anyway, I think that we should keep both clocks separated in case they have different time setup.
> 
> > We can keep separate defaults for the two clocks as well but need confirmation on the UX before doing it.
> 
> Why confirmation on the UX? the UX should not change. Instead of saving one TimeIncrement for both clocks, save two for each clock. There is no need for an UX or UI change.

@veloce Ah, sorry for the confusion. By UX , I didn't mean UI change but the user experience for two clock with separate defaults or only 1 default for both clocks.
Anyway, understood. I will add one more TimeIncrement in prefs to save two defaults separately. 
Please review code changes as well, if there is any other changes required. I will make them as well.

veloce:
> @veloce Ah, sorry for the confusion. By UX , I didn't mean UI change but the user experience for two clock with separate defaults or only 1 default for both clocks.

Understood. Actually not sure what is the best here. I feel like saving the current actual setup is better and less confusing than saving same time for both clocks if having different clock times.

dev-shady:
@veloce the changes have been made to support asymmetric clocks. Also fixed another bug related to emergency threshold. Please review the changes.

PR #3406, opened 2026-06-29
fix: improve accessibility semantics in CreateGameWidget

### LINKED ISSUE #3328: Accessibility VoiceOver: "Time control" and "Variant" labels are not grouped with their inputs on the CreateGame screen
## Objective
Hey! I'm learning the iOS Accessibility, so tested the Lichess iOS app a bit.
A one small bug I've found (and can elaborate on the fix if it's welcome) is the one described in this issue.
I know that there's a [whole checklist](https://github.com/lichess-org/mobile/issues/467) related to the accessibility, so please tell me if the current issue is a duplicate.

I'm attaching the AI-generated bug report which I proofread. I haven't tested the `Semantics`/`MergeSemantics` proposed fix yet, but judging from Flutter documentation and the current project's code it sounds reasonable.

## Environment

- Platform: iOS, VoiceOver enabled
- Screen: the "new game" panel on the home tab (Play / create game)
- Affected build: current App Store release

## Problem

The create-game panel shows four settings in a 2×2 grid:

- Row 1: **Time control** / **Variant**
- Row 2: **Game mode** / **Rating filter**

With VoiceOver on, the first row is announced incorrectly. The two labels are
read first, then the two values, so each label is separated from its input:

**Heard**: "Time Control", "Variant", "10+0", "Standard", "Game mode", "Rated"

**Expected**: "Time control, 10+0", "Variant, Standard", "Game mode, Rated", "Rating filter, …"

"Game mode" / "Rating filter" sound correct, but only by accident (see root
cause). 

Image

## Where

`lib/src/view/play/create_game_widget.dart` — `CreateGameWidget`.

## Root cause

Each field is a `Column` containing a `Text(label)` **sibling** above an
`OutlinedButton`:

```dart
Column(
  children: [
    Text(context.l10n.timeControl, style: labelStyle), // label node
    OutlinedButton.icon(... label: Text('10+0') ...),   // input node
  ],
)
```

A `Column` does not merge semantics and the label is a sibling of the button
(not a child), so Flutter emits two independent accessibility nodes per field.
There is no `Semantics`/`MergeSemantics` wrapper in the file, so VoiceOver falls
back to **geometric traversal order** (top-to-bottom, left-to-right): on row 1
the two labels share a baseline and the two inputs share a lower baseline, so
both labels are read before both values.

"Game mode" / "Rating filter" *appear* fine only incidentally — the
Rating-filter label is wrapped in an extra `Row` for the info icon
(`Icons.info_outline`), which shifts the baselines out of alignment so the sort
happens to interleave label + value. This is pixel-geometry luck, not real
grouping, and would also break under a different text scale.

## Proposed fix

Wrap each label + input pair in a single semantic container so each field is
announced as one element, with an explicit combined label. Apply this to all
four `Column`s in `CreateGameWidget` (not just the broken row), so the behavior
is uniform rather than dependent on layout:

```dart
Semantics(
  container: true,
  button: true,
  label: '${context.l10n.timeControl}: ${playPrefs.timeIncrement.display}',
  excludeSemantics: true,
  child: Column(
    crossAxisAlignment: CrossAxisAlignment.stretch,
    mainAxisSize: MainAxisSize.min,
    children: [
      Text(context.l10n.timeControl, style: labelStyle),
      OutlinedButton.icon(/* unchanged */),
    ],
  ),
)
```

- `container: true` — merges the label + input into one accessibility node.
- `label: '<label>: <value>'` — explicit combined announcement
  (e.g. "Time control: 10+0").
- `excludeSemantics: true` — suppresses the children's individual nodes so
  VoiceOver does not re-read the label and value separately.
- `button: true` — each field opens a picker/modal, so it should be announced
  as a button.

The same wrapper is applied to Variant, Game mode, and Rating filter, building
the label from each field's current value. For Rating filter the value string is
the displayed range (or the disabled "-500 / +500" placeholder).

## How this aligns with the existing architecture

This is not a new pattern — it reuses the convention already established in the
codebase:

- `lib/src/widgets/settings.dart:36` (`SettingsListTile`) wraps a label + value
  control in exactly this construct:
  `Semantics(container: true, button: true, label: '$settingsLabel: $settingsValue', excludeSemantics: true, child: ...)`.
  The proposed fix mirrors it directly.
- `Semantics(container: true, ...)` is also used in
  `lib/src/widgets/buttons.dart:64`, `lib/src/widgets/bottom_bar.dart:136`,
  `lib/src/view/user/perf_stats_screen.dart:295`, and
  `lib/src/view/study/study_gamebook.dart:159`. `MergeSemantics` is used in
  `lib/src/widgets/buttons.dart` and `lib/src/widgets/misc.dart`.

madhaviGit:
Problem
With VoiceOver enabled, the create game panel announces labels and values separately due to geometric traversal order:

❌ "Time Control" → "Variant" → "10+0" → "Standard"

Each label is completely separated from its corresponding input, which is confusing for blind users.
Solution
Wrapped each label+button pair in a Semantics widget so VoiceOver and TalkBack announce each field as a single grouped element with its current value:

✅ "Time control: 10+0" → "Variant: Standard" → "Game mode: Casual" → "Rating filter: -500 / +500"

This follows the existing pattern already established in SettingsListTile (lib/src/widgets/settings.dart:36).
Fixes #3328

### DIFF
--- a/lib/src/view/play/create_game_widget.dart
+++ b/lib/src/view/play/create_game_widget.dart
@@ -40,73 +40,89 @@ class CreateGameWidget extends ConsumerWidget {
           mainAxisAlignment: MainAxisAlignment.spaceBetween,
           children: [
             Expanded(
-              child: Column(
-                crossAxisAlignment: CrossAxisAlignment.stretch,
-                mainAxisSize: MainAxisSize.min,
-                children: [
-                  Text(context.l10n.timeControl, style: labelStyle),
-                  OutlinedButton.icon(
-                    style: OutlinedButton.styleFrom(
-                      side: BorderSide(color: Theme.of(context).dividerColor),
+              child: Semantics(
+                container: true,
+                button: true,
+                label: '${context.l10n.timeControl}: ${playPrefs.timeIncrement.display}',
+                excludeSemantics: true,
+                child: Column(
+                  crossAxisAlignment: CrossAxisAlignment.stretch,
+                  mainAxisSize: MainAxisSize.min,
+                  children: [
+                    Text(context.l10n.timeControl, style: labelStyle),
+                    OutlinedButton.icon(
+                      style: OutlinedButton.styleFrom(
+                        side: BorderSide(color: Theme.of(context).dividerColor),
+                      ),
+                      icon: Icon(playPrefs.timeIncrement.speed.icon),
+                      label: Text(
+                        playPrefs.timeIncrement.display,
+                        style: const TextStyle(letterSpacing: 2.0),
+                      ),
+                      onPressed: () {
+                        final double screenHeight = MediaQuery.sizeOf(context).height;
+                        showModalBottomSheet<void>(
+                          context: context,
+                          isScrollControlled: true,
+                          constraints: BoxConstraints(
+                            maxHeight: screenHeight - (screenHeight / 10),
+                          ),
+                          builder: (BuildContext context) {
+                            return TimeControlModal(
+                              timeIncrement: playPrefs.timeIncrement,
+                              onSelected: (choice) {
+                                ref
+                                    .read(gameSetupPreferencesProvider.notifier)
+                                    .setTimeIncrement(choice);
+                              },
+                            );
+                          },
+                        );
+                      },
                     ),
-                    icon: Icon(playPrefs.timeIncrement.speed.icon),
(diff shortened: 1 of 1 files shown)


veloce:
Thanks for this!

PR #3412, opened 2026-06-30
display arena tournament that game was/is played in

HaonRekcef:
Displays the tournament name and links to it in various places.

Preview:

| | | | 
|-|-|-|
| image | image |image |
| image |image |

Note that there is bug with the tournament sockets. Currently when going to a tournament -> going to a game -> going to the tournament again but by pushing a new screen again not by going back.
This can now happen for example from the tv screen or the archived games. Notice that it doesn't manage to connect to the socket. This issue exists already but this PR will make it happen more likely to occur, before maybe it could only happen from deeplinks.

Video:

[Screencast_20260630_170149.webm](https://github.com/user-attachments/assets/21ce019e-738b-49b1-8b89-ecf9da90d3ea)

I think this issue should probably be investigated before this PR is merged as it is quite likely behavior. I wont be able to investigate it in the next month though as I will be on holidays, if not done until then I could do it.


veloce:
I assume the mentioned issue is #2521 

I converted the PR to a draft, because you're right this issue should be fixed before the PR is merged.

HaonRekcef:
Not exactly the same, but possibly related.

veloce (approved):

veloce:
Code lgtm. Will investigate the tournament socket bug.

PR #3396, opened 2026-06-26
Add preview flavor

julien4215:
Introduce a preview product flavor configured to use the website `preview.test.example.com` and a separate FCM configuration from production to allow easier testing.

PR #3415, opened 2026-07-02
Add preview flavor

julien4215:
Reopening of #3396


julien4215:
> @julien4215 sorry I had to revert because this broke iOS. Can you reopen it please?
> 
> I'll have to make it work on iOS before merging, but can't do it now. Leave it as a draft for now, so we know it is not ready.

I didn't realize it would break iOS. I assume it is because there is a default flavor set in the `pubspec.yaml`  that must be used by each platform. This is annoying because we either need to also add flavors to iOS or set no default flavor but then the usual `flutter run` command will have to become `flutter run --flavor prod` on android.

Related Flutter issue https://github.com/flutter/flutter/issues/176820

julien4215:
@veloce I think renaming the Xcode scheme to `prod` should solve the build issue on iOS. If that is preferable, we can rename the `prod` flavor to `Runner` instead.

Or we can remove the `default-flavor` from the `pubspec.yaml` but each flutter command that builds the app on android will need to specify a flavor, e.g. `flutter run --flavor prod`.

What do you think would be the best approach?

veloce:
> @veloce I think renaming the Xcode scheme to `prod` should solve the build issue on iOS. If that is preferable, we can rename the `prod` flavor to `Runner` instead.
> 
> Or we can remove the `default-flavor` from the `pubspec.yaml` but each flutter command that builds the app on android will need to specify a flavor, e.g. `flutter run --flavor prod`.
> 
> What do you think would be the best approach?

Not sure yet, I have to investigate. Ideally android and iOS should behave the same way so the work you did for android needs to be done for iOS too.

veloce:
I rebased the branch fyi @julien4215 ; will add support for iOS from here.

PR #3424, opened 2026-07-08
chore(logging): pass exceptions as args to Logger

lsaudon:
- I passed the error and the stack trace as arguments to comply with the `logging` contract. (`(Object? message, [Object? error, StackTrace? stackTrace])`).
- I replaced the `;` with `:` in the log messages.

I think there might be a regression.
On the App Logs page, the logs that displayed the error and/or stack trace will no longer be visible on that page.
Perhaps add a pop-up with the log details when clicked.

### DIFF
--- a/lib/src/app_links_service.dart
+++ b/lib/src/app_links_service.dart
@@ -85,7 +85,7 @@ class AppLinksService {
     try {
       await _handleUri(uri, animated: animated);
     } catch (e, st) {
-      _logger.severe('Error handling app link: $e\n$st');
+      _logger.severe('Error handling app link:', e, st);
     }
   }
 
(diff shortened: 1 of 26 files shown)


veloce:
Ok I guess this makes sense. But I don't want the regression.

You should first modify the `AppLogService` so that if the `record.error` is not null, you concatenate it in the `debugPrint`.

Then in the App logs page you could do the same, or as you suggest, a tap could provide the detail (but no pop-up, I'd suggest an ExpansionTile instead).

lsaudon:
Should I include the stackTrace in all logs that are inside a `catch` block?

veloce:
> Should I include the stackTrace in all logs that are inside a `catch` block?

If you wish to.

veloce (approved):

Issue #1788, opened 2025-05-22
Add Time in Clock Tool While in a Game


landry314:
Just like you can give time to an opponent in an online game, it would be nice to give time to an opponent in an irl game using the clock. This could be symmetric or asymmetric. Sometimes both players just wish they were playing a longer game during the game and want to add a minute to each person's clock. Would be a cool feature.

PR #3437, opened 2026-07-12
Add time in clock tool while in a game

Giuseppe499:
Add a feature to add time to the clock tool while in a game.
If the clock is running and the game is paused, the settings menu can be used to add time.
If the clock is reset, the settings menu can be used as usual to set the time and increment.

See the attached video with a usage example:

https://github.com/user-attachments/assets/b96250a5-e236-416c-b789-f709a81c758b

It solves issue #1788


ghost:
If you run `dart format lib/src/` it will pass the tests.
Also the linter wants you to change the double quotes to single.
```
   info - Unnecessary use of double quotes. Try using single quotes unless the string contains single quotes -
          lib\src\view\clock\clock_tool_settings_modal.dart:186:40 - prefer_single_quotes
   info - Unnecessary use of double quotes. Try using single quotes unless the string contains single quotes -
          lib\src\view\clock\clock_tool_settings_modal.dart:186:48 - prefer_single_quotes
```

Giuseppe499:
Fixed now. Thanks @freebeartogoodhome

veloce:
@Giuseppe499 I can't see the video. Please upload a GIF or any format that is supported in GitHub. Thanks!

Giuseppe499:
@veloce here it's a GIF instead of the mp4:
output1

I would also like to add that I opened a new PR (#3460) that would also close #1788.
The new PR reuses the existing per-player settings UI. Therefore, it might be preferable to this one.

PR #3438, opened 2026-07-12
feat: show game info in app bar

ghost:
Tapping the app bar of a game you're watching will show the game info. You can switch back and forth with a tap. 

I didn't add the icon or play around with the font size. I'm sure people might have thoughts on that but I don't like UI stuff so I just didn't play around with it a ton. I would rather introduce it and then let people give feedback on it. But whatever is best.

gameInfo

### DIFF
--- a/lib/src/view/watch/tv_screen.dart
+++ b/lib/src/view/watch/tv_screen.dart
@@ -7,6 +7,7 @@ import 'package:lichess_mobile/src/model/account/account_repository.dart';
 import 'package:lichess_mobile/src/model/auth/auth_controller.dart';
 import 'package:lichess_mobile/src/model/chat/chat.dart';
 import 'package:lichess_mobile/src/model/common/id.dart';
+import 'package:lichess_mobile/src/model/common/time_increment.dart';
 import 'package:lichess_mobile/src/model/game/game_board_params.dart';
 import 'package:lichess_mobile/src/model/tv/tv_channel.dart';
 import 'package:lichess_mobile/src/model/tv/tv_controller.dart';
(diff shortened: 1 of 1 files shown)


veloce:
Wrong screenshot: it does not show what the PR says it does.

ghost:
> Wrong screenshot: it does not show what the PR says it does.

It is at the very top where it says 3+0 Rated Blitz. I can make a video of me tapping and switching between them if you would like.

ghost:
Here is a gif that makes it a lot clearer.
gameinfo

veloce:
Ok now I see. Sorry I don't really like it.

I don't think that info is necessary in the first place: usually bullet TV will be 1+0 and usually you can guess the time control from the clocks.

ghost:
Oh, I miss it from the Lichess website. It's one thing I don't like about watching from the app in bed. There is no easy way to see if a game is rated and it stays set to however you like it when you click it. I just clicked it a bunch of times to show how it worked.

Oops, meant to add this image from the website.
timecontrol

ijm8710:
> Ok now I see. Sorry I don't really like it.
> 
> 
> 
> I don't think that info is necessary in the first place: usually bullet TV will be 1+0 and usually you can guess the time control from the clocks.
> 
> 
> 
> 

Veloce, the issue is more noticeable when your watching a player. So it would say Veloce TV with no hint at the time control at all.

I get your dislike of the switching animation for this pr, but I still feel showing 1+0 is more important than showing Veloce TV. Plus you already know who you are spectating becuase its the user on the lower half of the board and their username is already present there

Would you be open to a pr to just replace out "Veloce TV" with the typical 1+0 etc game info? Can still potentially keep a TV symbol in the header (if there is room) to reflect that it's a watch game.

ghost:
> I get your dislike of the switching animation for this pr,

I think I messed up there with my gif showing it switch back and forth. When you click once it will stay set in the new state until you ever click again. If you start viewing another user, or switch to another TV screen it will default back to the "Whatever TV" but if you stay on the same user it will stay as game info or "User TV" until you choose to change it.

veloce:
ok @freebeartogoodhome , for a *User TV* you can add the time control as a subtitle in the app bar (see for instance player profile we have connection status as subtitle).

PR #3441, opened 2026-07-13
refactor(chat): remove inputText in ChatState for reduce rebuild

lsaudon:
On v0.25.3, every time I type a letter, the keyboard disappears.
This prevents the watch from being triggered in the parent.
There is no processing of `inputText` in the provider.
We can remove it to avoid duplicating the truth between the TextController and the Provider.

I also replace `authUser == null` by `isLoggedIn` from `isLoggedInProvider`

### DIFF
--- a/lib/src/view/chat/chat_screen.dart
+++ b/lib/src/view/chat/chat_screen.dart
@@ -287,16 +287,6 @@ class _ChatBottomBarState extends ConsumerState<_ChatBottomBar> {
   final _textController = TextEditingController();
 
   @override
-  void initState() {
-    super.initState();
-    final draft = ref.read(chatProvider(widget.options)).asData?.value?.inputText ?? '';
-    _textController.text = draft;
-    _textController.addListener(() {
-      ref.read(chatNotifierProvider(widget.options)).setInputText(_textController.text);
-    });
-  }
-
-  @override
   void dispose() {
     _textController.dispose();
     super.dispose();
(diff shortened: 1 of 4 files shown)


ghost:
Some stuff was recently committed for this. I would compare it to that if you get a chance. Yours might be better but I haven't looked.

https://github.com/lichess-org/mobile/commits/main/

lsaudon:
Indeed, this commit c09e3db49e91e08daa3c2d382229eb6d7c752e61 already solves the problem.
My fix simply reduces the number of page rebuilds. Currently, the page is rebuilt every time someone writes something.

ghost:
Got it! My brain can only do so much and it's maxed out for a few days. :)

veloce:
> Indeed, this commit [c09e3db](https://github.com/lichess-org/mobile/commit/c09e3db49e91e08daa3c2d382229eb6d7c752e61) already solves the problem. My fix simply reduces the number of page rebuilds. Currently, the page is rebuilt every time someone writes something.

Then it's better if the title of the PR reflects what it actually does @lsaudon ; gonna change it

lib/src/view/chat/chat_screen.dart
```diff
@@ -286,16 +286,6 @@ class _ChatBottomBar extends ConsumerStatefulWidget {
 class _ChatBottomBarState extends ConsumerState<_ChatBottomBar> {
   final _textController = TextEditingController();
 
-  @override
-  void initState() {
-    super.initState();
-    final draft = ref.read(chatProvider(widget.options)).asData?.value?.inputText ?? '';
```
veloce, line 292: Not sure but removing this may introduce a regression: there was an issue where the chat input (draft text not yet sent) was lost when temporarily putting the app into background.

Also since the draft text was saved in the mixin state, if you closed the chat window during a game with a text already entered it would save it for if you open the chat window again. Ideally we'd want to keep that.

Ideally this fix should take that into account. So the `inputText` should be renamed to `draftText` in the state and it should be saved only when the app is leaving foreground or the chat window is closed.

We would also need tests to assert this is working properly.

Thank you!

lsaudon: Actually, I hadn't thought of that. My pull request is therefore completely useless.
Do you want me to create a testWidget to check if the text is preserved when the user leaves and returns to the page?

veloce: I would not say it is useless because we don't really want to rebuild the screen on each key stroke. It could be slow and lead to a poor user experience.

If you can modify the PR to keep both it would be nice. So yes, I'd start with a widget test that checks this spec is respected (also when app goes to background).

PR #3447, opened 2026-07-14
Add Broadcast Widget For Android OS

r4khul:
## Description

This PR implements the native, theme-aware home screen widget for Lichess broadcasts on Android.

*   Implemented `BroadcastWidgetProvider.kt` to handle configuration, update loops, and async JSON fetching from the Lichess API.
*   Updated `app.dart` to persist the active Lichess host to shared preferences, enabling correct API targeting on Android.
*   Designed adaptive XML layouts supporting live status tags, rounded thumbnails, and theme-aware styling.
*   Added localization resources in `strings.xml` (to introduce translations if needed in future) and integrated system-locale datetime formatting for broadcast start times.

## Test

This change has been physically tested on the following Android devices:
*   Realme C67 5G (Mobile Phone)
*   Realme Pad 2 (Tablet)
### Widget Picker

| Light Mode | Dark Mode |
|------------|-----------|
| image | image |

### Widget Preview

| Light Mode | Dark Mode |
|------------|-----------|
| Widget Preview - Light | Widget Preview - Dark |

### Mobile Demo

https://github.com/user-attachments/assets/c4737ecc-d685-4627-aa82-467d6f192533

### Tablet Demo

https://github.com/user-attachments/assets/f216dfcc-8e91-41de-ac9d-4da59886a4b3


r4khul:
PTAL @veloce

veloce:
Awesome @r4khul ! Will review asap.

android/app/src/main/kotlin/org/lichess/mobileV2/widgets/BroadcastWidgetProvider.kt
```diff
@@ -0,0 +1,304 @@
...
+                val availableHeight = if (isPortrait && maxHeight > 0) maxHeight else minHeight
+
+                val numRows = if (availableHeight == 0) 3 else ((availableHeight - 44) / 60).coerceIn(1, 4)
+                val showThumbnails = minWidth == 0 || minWidth >= 200
+                
+                val displayItems = broadcasts.take(numRows)
+                
+                displayItems.forEachIndexed { index, item ->
+                    val itemViews = RemoteViews(context.packageName, R.layout.widget_broadcast_item)
+                    setupItemViews(context, itemViews, item, index, showThumbnails, lichessHost, appWidgetId)
```
veloce: This will sequentially fetch the thumbnails. I know there aren't many to fetch but in case of bad network it would be better to get them concurrently (and that's what we do on iOS).

android/app/src/main/kotlin/org/lichess/mobileV2/widgets/BroadcastWidgetProvider.kt
```diff
@@ -0,0 +1,304 @@
...
+            try {
+                return connection.inputStream.use { stream ->
+                    val jsonString = stream.bufferedReader().use { it.readText() }
+                    val active = JSONObject(jsonString).getJSONArray("active")
+                    
+                    (0 until active.length()).mapNotNull { i ->
+                        val item = active.getJSONObject(i)
+                        val tour = item.getJSONObject("tour")
+                        
+                        if (tour.optInt("tier", 0) < 4) return@mapNotNull null
```
veloce: I don't think this is necessary: according to the API docs, all broadcasts are already sorted and the active ones are all on the first page.

I know iOS has the top and all filters to change that, but I think I'll remove it so both platforms will be equal.

android/app/src/main/res/xml/widget_broadcast_info.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
+<appwidget-provider xmlns:android="http://schemas.android.com/apk/res/android"
+    android:description="@string/widget_broadcast_description"
+    android:minWidth="250dp"
+    android:minHeight="110dp"
+    android:minResizeWidth="150dp"
+    android:minResizeHeight="110dp"
+    android:maxResizeHeight="290dp"
+    android:updatePeriodMillis="1800000"
```
veloce: Let's match here iOS refresh rate which is 1h.

android/app/src/main/res/layout/widget_broadcast_item.xml
```diff
@@ -0,0 +1,68 @@
...
+            android:layout_marginTop="4dp"
+            android:gravity="center_vertical"
+            android:orientation="horizontal">
+
+            <FrameLayout
+                android:id="@+id/item_live_indicator"
+                android:layout_width="6dp"
+                android:layout_height="6dp"
+                android:layout_marginEnd="4dp"
+                android:background="#FF3B30"
```
veloce: This should be also in colors.xml. Also it doesn't appear to be the standard material color red. Let's use that one instead since we're on android.

veloce (commented):
Nice job, thank you!

The internationalisation of the widget seems to be missing, but it can be done in another PR.

r4khul:
thanks for the review @veloce! addressed changes:

* concurrent thumbnail downloads using a thread pool before binding `remoteviews` (matches ios `withtaskgroup`)
* removed the client-side tier filter 
* moved the live indicator color to `colors.xml` using material red 500 (`#f44336`)
* updated `updateperiodmillis` to `3600000` (1 hour)

android/app/src/main/kotlin/org/lichess/mobileV2/widgets/BroadcastWidgetProvider.kt
```diff
@@ -0,0 +1,323 @@
...
+            try {
+                return connection.inputStream.use { stream ->
+                    val jsonString = stream.bufferedReader().use { it.readText() }
+                    val active = JSONObject(jsonString).getJSONArray("active")
+                    
+                    (0 until active.length()).map { i ->
+                        val item = active.getJSONObject(i)
+                        val tour = item.getJSONObject("tour")
+                        val round = item.getJSONObject("round")
+                        val id = item.optJSONObject("roundToLink")?.optString("id") ?: round.getString("id")
```
veloce: `optString` returns `""` so fallback would not work.

```suggestion
  val id = item.optJSONObject("roundToLink")?.optString("id")?.takeIf { it.isNotBlank() }
      ?: round.getString("id")
```

android/app/src/main/kotlin/org/lichess/mobileV2/widgets/BroadcastWidgetProvider.kt
```diff
@@ -0,0 +1,323 @@
...
+            requestCode,
+            intent,
+            PendingIntent.FLAG_UPDATE_CURRENT or PendingIntent.FLAG_IMMUTABLE
+        )
+        itemViews.setOnClickPendingIntent(R.id.item_container, pendingIntent)
+    }
+
+    private fun setFallbackThumbnail(itemViews: RemoteViews) {
+        itemViews.setImageViewResource(R.id.item_thumbnail, R.drawable.ic_stat_lichess_notification)
+        itemViews.setViewPadding(R.id.item_thumbnail, 12, 12, 12, 12)
```
veloce: Need to take pixel density into account here (as done above):

```
val pad = (12 * context.resources.displayMetrics.density).toInt()
```

android/app/src/main/kotlin/org/lichess/mobileV2/widgets/BroadcastWidgetProvider.kt
```diff
@@ -0,0 +1,323 @@
...
+            context,
+            requestCode,
+            intent,
+            PendingIntent.FLAG_UPDATE_CURRENT or PendingIntent.FLAG_IMMUTABLE
+        )
+        itemViews.setOnClickPendingIntent(R.id.item_container, pendingIntent)
+    }
+
+    private fun setFallbackThumbnail(itemViews: RemoteViews) {
+        itemViews.setImageViewResource(R.id.item_thumbnail, R.drawable.ic_stat_lichess_notification)
```
veloce, line 210: No tint override here, so the thumbnail would appear as white on light grey (#E6E2DC) which would be barely visible on a light theme.

veloce (approved):

veloce:
Thanks @r4khul !

It's been integrated into main with a rebase.

Issue #3442, opened 2026-07-14
Feature Request: Support navigating back to the first move in the Puzzle mode


AminiamEva:
While playing puzzles, I can only navigate back for only one move while on the lichess website users are able to navigate back many moves to review the full game. I would appreciate if full navigation on mobile devices are added.

Device: iphone
Version: 0.25.3

nleonh:
Hello,
I agree! I'll try to take care of it.

AminiamEva:
Thank you very much！

veloce:
There is already the "See the game button" that allow to rewind the game moves.

AminiamEva:
> There is already the "See the game button" that allow to rewind the game moves.

Yes I checked it on my device and there's a view game button after completing the puzzle.

I'd appreciate it if rewinding back before completing the puzzle was added because often we need to review the game and try to see both sides' attacking/defencing ideas and strategic plans to better solve the puzzle, which is a feature yet not supported.

PR #3453, opened 2026-07-16
Feature: Support navigating back to the first move in the Puzzle mode (issue #3442)

nleonh:
Hello,
I modified the ``PuzzleController`` class so that you can navigate back to the first move while solving puzzles, just like on lichess.org (issue #3442). This also applies for analysing the puzzle afterwards.
I've tested it on my phone and it works perfectly fine (-> screen recording)

``flutter analyze`` doesn't find any problems.

EDIT: third commit fixes tests

Best regards

https://github.com/user-attachments/assets/366b814d-6438-48d4-9701-be16039a4b9e

### DIFF
--- a/lib/src/model/puzzle/puzzle_controller.dart
+++ b/lib/src/model/puzzle/puzzle_controller.dart
@@ -365,17 +365,14 @@ class PuzzleController extends Notifier<PuzzleState> {
   }
 
   String makePgn() {
-    final initPosition = _gameTree.nodeAt(state.initialPath).position;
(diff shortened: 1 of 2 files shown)


veloce:
Hi, thanks for contributing but I don't think I want this. It is not because it is on website that we need to have it here. Puzzle UI is a bit different on mobile.

veloce:
Closing it as not planned. See https://github.com/lichess-org/mobile/issues/3442

Issue #215, opened 2023-06-20
Download a new puzzle batch if rating evolves too much


veloce:
To ensure puzzle difficulty is always adapted.

We should save the rating at the time the batch is downloaded and then compare from time to time the current rating of the user with that saved rating. If the rating diff goes beyond a certain threshold, a new puzzle batch should be downloaded.

braden-godley:
I totally agree, this is a frustrating experience as someone who mostly plays in mobile. I am currently rated 1825, but I am consistently being given puzzles rated 1550-1600. Theoretically once the batch runs out it will be resolved, but the experience of only getting puzzles that are rated 200 below does kind of suck.

ijm8710:
I def had the same thoughts when my puzzle rating was more chaotic/not as matured when I first joined the site...

I feel a way to manually dump the batch and resync would be helpful as well.

PR #3454, opened 2026-07-16
Make offline puzzle count configurable

### LINKED ISSUE #926: [Feature] prompt offline puzzle download, or increase number of offline puzzles downloaded
I often play through all 50 default downloaded puzzles while I am without mobile data. The app is rather sluggish with refilling this puzzles, and in many cases (ie, airport layover, intermittent network service, etc) I can't coax it to download new puzzles before I lose internet again. As far as I know there is no UI to prompt Lichess to download more puzzles, and workarounds like restarting the app or my phone have proved inconsistent. Is there some button I am missing in the app? Alternatively, is there some way to increase the number of puzzles downloaded by default?

The latter seems like it would be easier to add, although it might have a bad effect on the accuracy of puzzle ratings.

wernerhp:
Closes #926

## What

Make the number of offline puzzles configurable, and fill the offline queue reliably when the setting changes.

- Adds an **Offline puzzles** setting on the puzzle screen (100 / 150 / 200 / 300), default and minimum **100**.
- Fills the offline queue up to the configured count as a **one-time action** triggered on the setting change and on difficulty change, rather than trickling puzzles in via the solve path.

## How

The offline queue can exceed the server's 50-per-request batch cap, so it is filled with several sequential requests. Following review, the fill is kept out of the solve/sync loop entirely:

- **Sync path reverted** to the original single-request logic, unchanged. Offline solves still reconcile in one shot.
- **`PuzzleQueueFiller`** is a stable `NotifierProvider` that fills the queue to the configured count with sequential, single-flight, **select-only** requests. It re-reads storage each pass, dedups by id, never submits solves, and is safe offline. Being a stable provider (not a per-call service instance), its single-flight guard actually holds.
- The settings tile shows a spinner while a fill is running.

This removes the concurrency hazards raised in review: the inert `_isFilling` guard (a fresh `PuzzleService` was built per call) and the within-pass stale-solve race are both gone, because no `solved` list is ever in flight during a fill.

## Tests

- `PuzzleQueueFiller` covered in isolation: fill-to-count over multiple requests, deficit-only fetch, no-op when full, stop when the server is exhausted, offline-safe, select-only (never POSTs solves), and single-flight.
- Puzzle screen test asserts the setting change triggers the fill.
- Removed the obsolete background-fill service tests.

### DIFF
--- a/lib/src/model/puzzle/puzzle_preferences.dart
+++ b/lib/src/model/puzzle/puzzle_preferences.dart
@@ -34,20 +34,20 @@ class PuzzlePreferences extends Notifier<PuzzlePrefs> with SessionPreferencesSto
     return clamped == prefs.nbOfflinePuzzles ? prefs : prefs.copyWith(nbOfflinePuzzles: clamped);
   }
 
-  Future<void> setDifficulty(PuzzleDifficulty difficulty) async {
-    save(state.copyWith(difficulty: difficulty));
+  Future<void> setDifficulty(PuzzleDifficulty difficulty) {
+    return save(state.copyWith(difficulty: difficulty));
   }
 
-  Future<void> setAutoNext(bool autoNext) async {
-    save(state.copyWith(autoNext: autoNext));
+  Future<void> setAutoNext(bool autoNext) {
+    return save(state.copyWith(autoNext: autoNext));
   }
 
-  Future<void> setRated(bool rated) async {
-    save(state.copyWith(rated: rated));
+  Future<void> setRated(bool rated) {
+    return save(state.copyWith(rated: rated));
   }
 
-  Future<void> setNbOfflinePuzzles(int nbOfflinePuzzles) async {
-    save(
+  Future<void> setNbOfflinePuzzles(int nbOfflinePuzzles) {
+    return save(
       state.copyWith(
         nbOfflinePuzzles: nbOfflinePuzzles.clamp(kMinOfflinePuzzles, kMaxOfflinePuzzles),
       ),
--- a/lib/src/model/puzzle/puzzle_preferences.dart
+++ b/lib/src/model/puzzle/puzzle_preferences.dart
@@ -40,7 +40,24 @@ class PuzzlePreferences extends Notifier<PuzzlePrefs> with SessionPreferencesSto
   Future<void> setRated(bool rated) async {
     save(state.copyWith(rated: rated));
   }
+
+  Future<void> setNbOfflinePuzzles(int nbOfflinePuzzles) async {
+    save(
+      state.copyWith(
+        nbOfflinePuzzles: nbOfflinePuzzles.clamp(kMinOfflinePuzzles, kMaxOfflinePuzzles),
+      ),
+    );
+  }
 }
+
+/// Minimum number of puzzles kept in the offline queue.
+const kMinOfflinePuzzles = 50;
+
+/// Maximum number of puzzles kept in the offline queue.
+const kMaxOfflinePuzzles = 500;
+
+/// Available choices for the size of the offline puzzle queue.
+const kOfflinePuzzlesChoices = [50, 100, 150, 200, 300, 500];
 
 @Freezed(fromJson: true, toJson: true)
 sealed class PuzzlePrefs with _$PuzzlePrefs implements Serializable {
@@ -56,10 +73,19 @@ sealed class PuzzlePrefs with _$PuzzlePrefs implements Serializable {
     /// If `true`, the puzzle will be rated for logged in users.
     /// Defaults to `true`.
     @Default(true) bool rated,
+
+    /// Number of puzzles to keep downloaded in the offline queue.
+    /// Defaults to `50`.
+    @Default(50) int nbOfflinePuzzles,
   }) = _PuzzlePrefs;
 
-  factory PuzzlePrefs.defaults({UserId? id}) =>
-      PuzzlePrefs(id: id, difficulty: PuzzleDifficulty.normal, autoNext: false, rated: true);
+  factory PuzzlePrefs.defaults({UserId? id}) => PuzzlePrefs(
+    id: id,
+    difficulty: PuzzleDifficulty.normal,
+    autoNext: false,
+    rated: true,
+    nbOfflinePuzzles: 50,
+  );
 
   factory PuzzlePrefs.fromJson(Map<String, dynamic> json) => _$PuzzlePrefsFromJson(json);
 }
--- a/lib/src/model/puzzle/puzzle_preferences.dart
+++ b/lib/src/model/puzzle/puzzle_preferences.dart
@@ -26,7 +26,12 @@ class PuzzlePreferences extends Notifier<PuzzlePrefs> with SessionPreferencesSto
 
   @override
   PuzzlePrefs build() {
-    return fetch();
+    final prefs = fetch();
+    // Clamp the offline queue length loaded from storage: corrupted or
+    // manually edited prefs could otherwise bypass the bounds enforced by
+    // [setNbOfflinePuzzles] and trigger very large downloads.
+    final clamped = prefs.nbOfflinePuzzles.clamp(kMinOfflinePuzzles, kMaxOfflinePuzzles);
+    return clamped == prefs.nbOfflinePuzzles ? prefs : prefs.copyWith(nbOfflinePuzzles: clamped);
   }
 
   Future<void> setDifficulty(PuzzleDifficulty difficulty) async {
--- a/lib/src/model/puzzle/puzzle_preferences.dart
+++ b/lib/src/model/puzzle/puzzle_preferences.dart
@@ -59,10 +59,10 @@ class PuzzlePreferences extends Notifier<PuzzlePrefs> with SessionPreferencesSto
 const kMinOfflinePuzzles = 100;
 
 /// Maximum number of puzzles kept in the offline queue.
-const kMaxOfflinePuzzles = 500;
+const kMaxOfflinePuzzles = 300;
 
 /// Available choices for the size of the offline puzzle queue.
-const kOfflinePuzzlesChoices = [100, 150, 200, 300, 500];
+const kOfflinePuzzlesChoices = [100, 150, 200, 250, 300];
 
 @Freezed(fromJson: true, toJson: true)
 sealed class PuzzlePrefs with _$PuzzlePrefs implements Serializable {
--- a/lib/src/model/puzzle/puzzle_preferences.dart
+++ b/lib/src/model/puzzle/puzzle_preferences.dart
@@ -56,13 +56,13 @@ class PuzzlePreferences extends Notifier<PuzzlePrefs> with SessionPreferencesSto
 }
 
 /// Minimum number of puzzles kept in the offline queue.
-const kMinOfflinePuzzles = 50;
+const kMinOfflinePuzzles = 100;
 
 /// Maximum number of puzzles kept in the offline queue.
 const kMaxOfflinePuzzles = 500;
 
 /// Available choices for the size of the offline puzzle queue.
-const kOfflinePuzzlesChoices = [50, 100, 150, 200, 300, 500];
+const kOfflinePuzzlesChoices = [100, 150, 200, 300, 500];
 
 @Freezed(fromJson: true, toJson: true)
 sealed class PuzzlePrefs with _$PuzzlePrefs implements Serializable {
@@ -80,8 +80,8 @@ sealed class PuzzlePrefs with _$PuzzlePrefs implements Serializable {
     @Default(true) bool rated,
 
     /// Number of puzzles to keep downloaded in the offline queue.
-    /// Defaults to `50`.
-    @Default(50) int nbOfflinePuzzles,
+    /// Defaults to `100`.
+    @Default(100) int nbOfflinePuzzles,
   }) = _PuzzlePrefs;
 
   factory PuzzlePrefs.defaults({UserId? id}) => PuzzlePrefs(
@@ -89,7 +89,7 @@ sealed class PuzzlePrefs with _$PuzzlePrefs implements Serializable {
     difficulty: PuzzleDifficulty.normal,
     autoNext: false,
     rated: true,
-    nbOfflinePuzzles: 50,
+    nbOfflinePuzzles: 100,
   );
 
   factory PuzzlePrefs.fromJson(Map<String, dynamic> json) => _$PuzzlePrefsFromJson(json);
--- a/lib/src/model/puzzle/puzzle_preferences.dart
+++ b/lib/src/model/puzzle/puzzle_preferences.dart
@@ -1,7 +1,9 @@
 import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:freezed_annotation/freezed_annotation.dart';
 import 'package:lichess_mobile/src/model/common/id.dart';
+import 'package:lichess_mobile/src/model/puzzle/puzzle_angle.dart';
 import 'package:lichess_mobile/src/model/puzzle/puzzle_difficulty.dart';
+import 'package:lichess_mobile/src/model/puzzle/puzzle_theme.dart';
 import 'package:lichess_mobile/src/model/settings/preferences_storage.dart';
 import 'package:lichess_mobile/src/model/user/user.dart';
 
@@ -63,6 +65,17 @@ const kMaxOfflinePuzzles = 300;
 
 /// Available choices for the size of the offline puzzle queue.
 const kOfflinePuzzlesChoices = [100, 150, 200, 250, 300];
+
+/// Whether the configurable offline-queue size ([PuzzlePrefs.nbOfflinePuzzles])
+/// applies to [angle]. Where it doesn't, the queue stays at [kMinOfflinePuzzles].
+bool isConfigurableOfflineQueueAngle(PuzzleAngle angle) =>
+    angle == const PuzzleTheme(PuzzleThemeKey.mix);
+
+/// The offline-queue length for [angle]: the configurable [nbOfflinePuzzles]
+/// where [isConfigurableOfflineQueueAngle] allows it, otherwise the fixed
+/// [kMinOfflinePuzzles].
+int offlineQueueLengthForAngle(PuzzleAngle angle, int nbOfflinePuzzles) =>
+    isConfigurableOfflineQueueAngle(angle) ? nbOfflinePuzzles : kMinOfflinePuzzles;
 
 @Freezed(fromJson: true, toJson: true)
 sealed class PuzzlePrefs with _$PuzzlePrefs implements Serializable {
--- a/lib/src/model/puzzle/puzzle_queue_filler.dart
+++ b/lib/src/model/puzzle/puzzle_queue_filler.dart
@@ -39,14 +39,20 @@ class PuzzleQueueFiller extends Notifier<bool> {
   /// This is a one-time, select-only action triggered when the offline puzzle
   /// count or the difficulty changes. It never submits solves, so it can't race
   /// with the solve/sync path over the stored `solved` list. Re-reads storage
-  /// on every pass and merges by puzzle id so a solve made while the fill runs
-  /// is never clobbered. Guarded by [state] so overlapping calls can't
-  /// double-fetch. Never throws: network and storage failures stop the fill.
+  /// *after* each network request and merges by puzzle id, so a solve made
+  /// while a request is in flight is never clobbered. Guarded by [state] so
+  /// overlapping calls can't double-fetch. No-op for anonymous users: the
+  /// configurable queue length is a logged-in-only feature. Never throws:
+  /// network and storage failures stop the fill.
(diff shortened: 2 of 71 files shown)


lib/src/model/puzzle/puzzle_service.dart
```diff
@@ -18,12 +18,16 @@ import 'package:result_extensions/result_extensions.dart';
...
+/// queue length is user-configurable via [PuzzlePrefs.nbOfflinePuzzles].
 const kPuzzleLocalQueueLength = 50;
 
 /// A provider for [PuzzleService].
 final puzzleServiceProvider = FutureProvider<PuzzleService>((Ref ref) {
-  return ref.read(puzzleServiceFactoryProvider)(queueLength: kPuzzleLocalQueueLength);
+  final nbOfflinePuzzles = ref.watch(
+    puzzlePreferencesProvider.select((prefs) => prefs.nbOfflinePuzzles),
+  );
+  return ref.read(puzzleServiceFactoryProvider)(queueLength: nbOfflinePuzzles);
```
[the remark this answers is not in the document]
wernerhp: Intentional. The sync only tops up so we don't discard already-downloaded puzzles a user may still want to solve offline. Lowering the setting stops future refills beyond the new value; the surplus clears as puzzles are solved. Trimming on decrease would throw away offline content, which is the opposite of what this setting is for.

lib/src/model/puzzle/puzzle_service.dart
```diff
@@ -18,12 +18,16 @@ import 'package:result_extensions/result_extensions.dart';
 
 part 'puzzle_service.freezed.dart';
 
-/// Size of puzzle local cache
+/// Default size of puzzle local cache. Used as a fallback; the effective
+/// queue length is user-configurable via [PuzzlePrefs.nbOfflinePuzzles].
 const kPuzzleLocalQueueLength = 50;
```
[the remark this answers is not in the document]
wernerhp: Done in 6cb8c4c. Removed the constant and its stale comment; all call sites read the preference.

lib/src/model/puzzle/puzzle_preferences.dart
```diff
@@ -40,8 +40,25 @@ class PuzzlePreferences extends Notifier<PuzzlePrefs> with SessionPreferencesSto
   Future<void> setRated(bool rated) async {
     save(state.copyWith(rated: rated));
   }
+
+  Future<void> setNbOfflinePuzzles(int nbOfflinePuzzles) async {
+    save(
+      state.copyWith(
+        nbOfflinePuzzles: nbOfflinePuzzles.clamp(kMinOfflinePuzzles, kMaxOfflinePuzzles),
+      ),
```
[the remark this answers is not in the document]
wernerhp: Done in 6cb8c4c. `build()` now clamps `nbOfflinePuzzles` to 50-500 on load, so corrupted or manually edited prefs can't bypass the bounds.

lib/src/view/puzzle/puzzle_screen.dart
```diff
@@ -973,6 +976,23 @@ class _PuzzleSettingsBottomSheet extends ConsumerWidget {
                 ref.read(puzzlePreferencesProvider.notifier).setAutoNext(value);
               },
             ),
+            SettingsListTile(
+              settingsLabel: Text(context.l10n.mobileNbOfflinePuzzles),
+              settingsValue: nbOfflinePuzzles.toString(),
+              onTap: () {
+                showChoicePicker(
```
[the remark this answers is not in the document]
wernerhp: Done in 6cb8c4c. Added a widget test that opens puzzle settings, asserts the tile shows the default 50, selects 200 via the picker, and asserts the tile updates. Passes on both platform variants.

translation/source/mobile.xml
```diff
@@ -25,6 +25,7 @@
   <string name="notFollowingAnyUser" comment="This is shown in the list that contains user followed by the user. Displayed if the user does not follow any users.">You are not following any users.</string>
   <string name="okButton" comment="This is used in varius confirmations dialogs. Translation should be short.">OK</string>
   <string name="overTheBoard" comment="Used in the main play menu to launch the Over the board game mode.">Over the board</string>
+  <string name="nbOfflinePuzzles" comment="Label for the puzzle setting that controls how many puzzles are downloaded and kept available for offline play.">Offline puzzles</string>
```
veloce: Normally the recommended way for a new feature is to hardcode in English first, and then translate when the feature is mature enough.
I'll pass on this one, let's keep it.

lib/src/model/puzzle/puzzle_preferences.dart
```diff
@@ -40,8 +45,25 @@ class PuzzlePreferences extends Notifier<PuzzlePrefs> with SessionPreferencesSto
...
+    );
+  }
 }
 
+/// Minimum number of puzzles kept in the offline queue.
+const kMinOfflinePuzzles = 50;
+
+/// Maximum number of puzzles kept in the offline queue.
+const kMaxOfflinePuzzles = 500;
```
veloce: I'd like to lower a bit the max value here, maybe 300, to be on a conservative side. But I wanted the input of the server team here, so let's leave it like that for now. I might just lower it after.

veloce (commented):
Thanks! the logic is fine. Just waiting a bit to see what number of max puzzle we'll decide to keep.

lib/src/model/puzzle/puzzle_preferences.dart
wernerhp: Thanks for the review.

I had a look at the server code and found that [`apiBatchSelect`](https://github.com/lichess-org/lila/blob/d1337281337dd99d43094702d72e8b73767a1357/app/controllers/Puzzle.scala#L375) clamps every request with [`nb.atMost(50)`](https://github.com/lichess-org/lila/blob/d1337281337dd99d43094702d72e8b73767a1357/app/controllers/Puzzle.scala#L395), so the endpoint caps at 50 puzzles per sync regardless of the requested `nb`. The max setting shouldn't be a problem for the server either way.

This did make me look through the solution again, and I found a gap in the offline pre-load. The queue tops up by deficit (≤50 per sync) and syncs after every online solve, so a user who sets a high limit and goes offline won't have the full set cached (it fills gradually while playing online not up front). Bumping the limit needs a quick one-off loop to fill the queue to capacity client-side, otherwise the higher cap doesn't deliver its purpose before a flight.  Since the default (50) is unchanged, bumping to the max (500) means ~10 server calls up front, the same calls that would otherwise trickle in one-per-solve as you play, until capacity is filled. Once full, each solve fetches just 1 new puzzle, exactly as it does today.

Happy to wait for the server team's feedback.

veloce: Well, I thought it was not an issue because you usually play online and on some (less frequent) occasions you want to have offline puzzles. So it needs only 10 puzzles to fill the queue for 500 puzzles.

The problem is only because of the settings, people will assume that changing the settings will populate the queue right away. So yes, that is added complexity but we need to fill that queue when bumping the limit. Just don't do it concurrently but sequentially (we don't want to trigger rate limits server side). I guess with a loading indicator to show we're actually downloading the puzzles.

A middle ground would be no settings and raise the limit, but I don't want to do that because I fear this could lead to more issues regarding puzzles ratings and people tend to complain about that (too much puzzle in the cache increases the changes the ratings are off).

wernerhp: I've already updated it so the queue will populate right away.  The cache will fill sequentially.  No rate limit concerns.  It syncs it in the background, so users can start playing puzzles immediately.

If people already complain about "ratings are off", that sounds like an existing server-side issue that we can look into.  I think the issue is wrt how the puzzles difficulties are selected for the sync.  I imagine player elo typically remains within a range (e.g. +/-100), so a bell distribution of puzzles for the range would help and increasing from 50 to 500 would then also increase the pool of puzzles for the range.  For an even distribution, if we currently have 50 puzzles, that's 2 puzzles for every 5 elo in the range. Bumping to 500 gives us 50 puzzles per 5 elo range, so it should take longer before it feels like the rating is off.  A bell distribution would give even more puzzles in the middle range, so I think that might fix that issue.  Will post a separate MR to fix it, since it's not directly related to this.

wernerhp: Scratch the separate distribution MR I floated. I went through the server selection code and there's nothing to fix there: it already matches the player's rating band and samples from it, so puzzles stay well-centred. A bigger pool pulls from that same band, so it's just more in-range puzzles, not a spread across ratings. The cache stays as well-rated as today's 50, it just lasts longer offline.

veloce: Yes, the rating issue is for people who do not have a stable rating yet. But this not the most frequent case (by far I'd say), so let's not bother with it for now.

lib/src/model/puzzle/puzzle_preferences.dart
```diff
@@ -40,8 +45,25 @@ class PuzzlePreferences extends Notifier<PuzzlePrefs> with SessionPreferencesSto
...
+    save(
+      state.copyWith(
+        nbOfflinePuzzles: nbOfflinePuzzles.clamp(kMinOfflinePuzzles, kMaxOfflinePuzzles),
+      ),
+    );
+  }
 }
 
+/// Minimum number of puzzles kept in the offline queue.
+const kMinOfflinePuzzles = 50;
```
veloce: We could increase the min nb of puzzles to 100 I think.

lib/src/model/puzzle/puzzle_service.dart
```diff
@@ -162,48 +184,121 @@ class PuzzleService {
...
+      unsolved: IList([...unsolved, ...value.puzzles]),
+    );
+    await batchStorage.save(userId: userId, angle: angle, data: newBatch);
+
+    // The server caps each batch at 50, so a single fetch already covers any
+    // deficit up to that cap. Only spin up a background fill when the remaining
+    // deficit still exceeds one batch (i.e. the user raised the offline limit).
+    if (value.puzzles.isNotEmpty && queueLength - newBatch.unsolved.length >= serverBatchCap) {
+      unawaited(_backgroundFill(userId, angle, difficulty));
     }
```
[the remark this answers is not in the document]
wernerhp: Fixed in `8f9c5a0`.

1. Background fill now triggers when the foreground batch comes back cap-sized (`value.puzzles.length >= serverBatchCap`) and we're still under `queueLength`, so the `queueLength=100, unsolved=10` case tops up past the first 50 instead of stalling. Gated on cap-sized rather than any remaining deficit to avoid a redundant request on every sub-cap solve.
2. The `unawaited` fill now has a `catchError` that logs, so a storage I/O failure can't become an unhandled async error.

lib/src/model/puzzle/puzzle_service.dart
```diff
@@ -162,48 +184,121 @@ class PuzzleService {
...
+    try {
+      // Tracks the stored queue length across passes. If a save doesn't grow
+      // the queue (e.g. the write silently no-ops), stop instead of looping
+      // forever re-requesting the same deficit.
+      int lastLength = -1;
+      while (true) {
+        final current = await batchStorage.fetch(userId: userId, angle: angle);
+        final unsolved = current?.unsolved ?? IList(const []);
+        if (unsolved.length <= lastLength) break;
+        lastLength = unsolved.length;
```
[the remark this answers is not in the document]
wernerhp: Fixed in `8f9c5a0`. Guard now breaks on `== lastLength` instead of `<=`, so a solve made mid-fill (which lowers the count and widens the deficit) no longer stops the loop early. Exhaustion is still handled by the empty-response and dedup-stall breaks.

lib/src/model/puzzle/puzzle_service.dart
```diff
@@ -162,48 +184,129 @@ class PuzzleService {
 
     final deficit = max(0, queueLength - unsolved.length);
 
-    if (deficit > 0 || solved.isNotEmpty) {
```
veloce, line 165: Seems there is a bug with the background refill mechanism (found by Claude Opus):


> _backgroundFill reads the batch from storage, then awaits the network, then saves the snapshot it read before the request. solve() mutates the same row in the meantime — so a puzzle
  solved during the fetch gets restored to unsolved and its PuzzleSolution dropped from solved. The next sync sees solved: [], calls selectBatch instead of solveBatch, and the solve is
  never reported: no rating update, and the puzzle is served again. Silent.

  Fix: re-read storage after the network call and merge into that, rather than into the pre-request snapshot.

Another bug found by opus:

> _isFilling doesn't prevent concurrent fills. It's an instance field, but PuzzleServiceFactory.call() builds a fresh PuzzleService per invocation and PuzzleController._service is a
  getter, so every access is a new instance with its own flag. Two fills can race on the same row with separate stale snapshots.

Another issue that I found: when the user change the settings to increase queue length, the refill is not triggered since it is only part of the solve logic (so nothing happens unless a puzzle is loaded). No feedback when changing the settings, and nothing happens, the user might just think the puzzles are downloaded, then exit the screen. And next time they go offline to solve puzzles the queue is not filled.

That, combined to the fact that the increase complexity of the new solving mechanism is more error prone, makes me think we should do it differently:

- the queue refill should be a one time action separate from the solving mechanism already in place (so please rollback to the existing one)
- this one time action should be triggered when the settings change (and I guess when puzzle difficulty change, since it invalidates the queue)
- since it is a one time action is is easier to provide feedback (a loading spinner shown when settings change )
- the benefits are a simpler logic, which can (and should) be tested on its own
- since the filling is still an asynchronous background task, the 2 issues mentioned above should be addressed (and with tests proving that)

veloce: I might add that the queue refill when settings change is an edge case (it is rather unlikely that someone change the settings without playing puzzles after since nowadays being online is more common that being offline, and the refill mechanism already in place should be largely enough).
But this edge case should be addressed nevertheless I think, and if it is a separate action that can be properly tested it mitigates the added complexity (which I general I try to avoid).

wernerhp: Agreed.

1. Roll the sync path back to the original single-request logic, unchanged.
2. Move refill into a one-time action triggered on settings and difficulty change: sequential, cancellable, single-flight, with a spinner.
3. Fill action tested in isolation.

Will raise the min to 100 as recommend.

ijm8710:
@veloce may it also make sense to incorporate some logic for #215 as well

I could see it possibly being consisting of to 2 logics:

- an automated flushing of the current puzzle batch if one's puzzle rating goes beyond a certain threshold
- a manual dump of current puzzle batch 

Or would you prefer to keep that separate?

wernerhp:
> @veloce may it also make sense to incorporate some logic for #215 as well
> 
> I could see it possibly being consisting of to 2 logics:
> 
> * an automated flushing of the current puzzle batch if one's puzzle rating goes beyond a certain threshold
> * a manual dump of current puzzle batch
> 
> Or would you prefer to keep that separate?

Good idea, but I'd keep #215 separate. This PR is about the queue size and a reliable fill; #215 needs a stored "rating at download" snapshot and a drift comparison, which is a different data change. The refactor here actually sets it up nicely though. The new one-time queue filler is the natural hook for both an automatic re-fill on rating drift and a manual "refresh puzzles" action, so #215 can build on top once this lands.

ijm8710:
Does down-switching from say 100 puzzles to 50 maintain the same first 50

And how about if you're back at 50 preloaded, once you finish that batch would puzzles 51-100 mirror what was fetched in the batch of 100 before downsizing?

Or does bouncing between sizes fetch a new original batch of 50?

veloce:
> @veloce may it also make sense to incorporate some logic for #215 as well
> 
> I could see it possibly being consisting of to 2 logics:
> 
> * an automated flushing of the current puzzle batch if one's puzzle rating goes beyond a certain threshold
> * a manual dump of current puzzle batch
> 
> Or would you prefer to keep that separate?

we could do that yes, but in another PR.

test/model/puzzle/puzzle_queue_filler_test.dart
```diff
@@ -0,0 +1,241 @@
...
+      final first = notifier.fill(userId: null, queueLengthOverride: 2);
+      final second = notifier.fill(userId: null);
+      await Future.wait([first, second]);
+
+      // only the first fill ran: exactly 2 requests to reach a queue of 2
+      expect(nbReq, equals(2));
+      final data = await storage.fetch(userId: null);
+      expect(data?.unsolved.length, equals(2));
+    });
+  });
```
veloce, line 381: Would be nice to have a test ensuring solving a puzzle when a background fill is currently in process doesn't  mess with the queue (the bug already mentioned before).

lib/src/model/puzzle/puzzle_preferences.dart
```diff
@@ -40,8 +45,25 @@ class PuzzlePreferences extends Notifier<PuzzlePrefs> with SessionPreferencesSto
...
+      ),
+    );
+  }
 }
 
+/// Minimum number of puzzles kept in the offline queue.
+const kMinOfflinePuzzles = 100;
+
+/// Maximum number of puzzles kept in the offline queue.
+const kMaxOfflinePuzzles = 500;
```
veloce: As far as I understand server code, `apiBatchSolve` rate limit allows solving 400 per hour.

For that reason let's change the max number of puzzle to 300 for now. I still haven't got a response from the server team on that matter, but 300 seems a reasonable number and we could still change it later.

veloce (commented):
This is good progress, thank you!

There's one more change needed on top of the 2 code comments: we need to disable the feature for anon players.

The reason is that server rate limit for fetching puzzles is much higher for anon players, so we'll only authorise logged in users to increase the queue length.

There are many place in the code to change, so that's why I kept this request as a general comment.
We need also tests to ensure that: widget tests to show that settings is not shown for anons and service test to show that background fill is a no-op when the user is not logged in.

test/model/puzzle/puzzle_queue_filler_test.dart
veloce: @wernerhp apparently the lost-update bug is still there with the new refill logic, so I confirm it should be addressed and tested.

veloce (commented):
Thanks for the update.

There is one more issue we should guard against: increase the queue when logged in -> log out -> puzzle prefs should still return the default size queue.

Since puzzle prefs inherits from SessionPreferencesStorage this is already covered, but a test ensuring that prefs always read the default size for anon would be welcome.

lib/src/model/puzzle/puzzle_queue_filler.dart
```diff
@@ -0,0 +1,109 @@
...
+      final batchStorage = await ref.read(puzzleBatchStorageProvider.future);
+
+      // Tracks the stored queue length across passes. Stop when a pass fails to
+      // grow the queue (server exhausted or a silent no-op write), not when it
+      // shrinks: a solve made mid-fill lowers the count and legitimately widens
+      // the deficit, so that pass should still top up.
+      int lastLength = -1;
+      while (true) {
+        final current = await batchStorage.fetch(userId: userId, angle: angle);
+        final unsolved = current?.unsolved ?? IList(const []);
```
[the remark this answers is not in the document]
wernerhp: Fixed in `479f9a5`. `fetch()` and the merge `fetch()`/`save()` are now wrapped in try/catch that logs and stops the fill, so a `FormatException` on corrupt prefs or an sqflite error can't surface as an unhandled async error to the fire-and-forget caller. Added a regression test asserting `fill()` completes when the storage read throws and the single-flight guard is released.

lib/src/model/puzzle/puzzle_preferences.dart
```diff
@@ -26,7 +26,12 @@ class PuzzlePreferences extends Notifier<PuzzlePrefs> with SessionPreferencesSto
 
   @override
   PuzzlePrefs build() {
-    return fetch();
+    final prefs = fetch();
+    // Clamp the offline queue length loaded from storage: corrupted or
+    // manually edited prefs could otherwise bypass the bounds enforced by
+    // [setNbOfflinePuzzles] and trigger very large downloads.
+    final clamped = prefs.nbOfflinePuzzles.clamp(kMinOfflinePuzzles, kMaxOfflinePuzzles);
```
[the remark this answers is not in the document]
wernerhp: `int.clamp(int, int)` returns `int` on this SDK (the `num.clamp` override), so `clamped` is inferred as `int` and `copyWith(nbOfflinePuzzles:)` type-checks. `flutter analyze` passes with no cast; adding `.toInt()` would be flagged as an unnecessary no-op.

veloce:
Thanks @wernerhp ! Now I'm going to test this thoroughly on a device (puzzles are an important feature) and probably make some changes if I find any issues.

veloce:
A summary of what I added to the PR:
- only the mix angle gets to have the nb puzzle settings (to avoid problems, I don't want to have the possibility of users having 300 x N angle puzzles, at least for now)
- fix some concurrency issues that prevented the refill to work properly
- throttle queue refill
- handle rate limits on solveBatch (this is the only part not tested, since it's hard to test)
- cosmetic changes

Looks good to merge now. Thanks again for your work!

PR #3460, opened 2026-07-18
Edit clock tool to allow time and increment change to one player while paused

Giuseppe499:
Allow users to modify the time and increment settings while the clock is paused.
Previously, the per-player settings UI became unavailable when the clock started.

New behaviour:
output2

Old behaviour:
output3

Solves issue #1788  as time can now be edited during a game

### DIFF
--- a/lib/src/model/clock/clock_tool_controller.dart
+++ b/lib/src/model/clock/clock_tool_controller.dart
@@ -141,35 +141,49 @@ class ClockToolController extends Notifier<ClockState> {
   }
 
   void updateOptionsCustom(TimeIncrement clock, ClockSide player) {
+    Duration topTime = state.options.topTime;
+    Duration bottomTime = state.options.bottomTime;
+    Duration topIncrement = state.options.topIncrement;
+    Duration bottomIncrement = state.options.bottomIncrement;
+    if (player == ClockSide.top) {
+      topIncrement = Duration(seconds: clock.increment);
+      // Update stored preferences only if the clock hasn't started yet
+      if (state.activeSide == null) {
+        topTime = Duration(seconds: clock.time);
+        ref.read(clockToolPreferencesProvider.notifier).setTopTimeIncrement(clock);
+      }
+      _clock.setTimes(
+        blackTime: Duration(seconds: clock.time),
+        whiteTime: state.bottomTime.value,
+      );
+    } else {
+      bottomIncrement = Duration(seconds: clock.increment);
+      // Update stored preferences only if the clock hasn't started yet
+      if (state.activeSide == null) {
+        bottomTime = Duration(seconds: clock.time);
+        ref.read(clockToolPreferencesProvider.notifier).setBottomTimeIncrement(clock);
+      }
+      _clock.setTimes(
+        blackTime: state.topTime.value,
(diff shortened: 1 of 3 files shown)


veloce:
Thanks for this @Giuseppe499 

My feedback based on your screen recording:

I would only allow the per clock time change when a game is started. So the global cog should stay disabled, but the per clock tune icons would appear when the clock is paused mid-game.

Important note: when time is added using that per-clock button, it should **not** modify the saved settings of the clock, just add time to the current game.

Giuseppe499:
Hi @veloce. Thanks for the feedback!

I need some clarifications:

> I would only allow the per clock time change when a game is started. So the global cog should stay disabled, but the per clock tune icons would appear when the clock is paused mid-game.

Do you mean that the per-clock tune icons shouldn't be visible before the clock is started, or just that the global cog should be disabled when the clock is paused after it is started?
I think that the first one would be a regression. In the current app, the per-clock tune icons are enabled only when the clock isn't started, and they are used to have different settings for each player and to store the settings for subsequent sessions.

> Important note: when time is added using that per-clock button, it should **not** modify the saved settings of the clock, just add time to the current game.

The current code should already do this. Per-clock preferences are stored only if the clock is not started yet.
If the per-clock button is used with the clock started and paused, it just updates the time and increment, but those settings are not saved for future sessions.
For reference, see this [snippet](https://github.com/lichess-org/mobile/pull/3460/changes/f05c695f818de397394c2c09a75544d8c8f55953#diff-8920b0001fc9840b1c3bb7ffd033b0900b3bd13a73bdf83d56b3cba2294de03fR150-R154).

veloce:
> Do you mean that the per-clock tune icons shouldn't be visible before the clock is started, or just that the global cog should be disabled when the clock is paused after it is started?

No, the existing behaviour (before your PR) should remain unchanged. What I meant is that the global cog should be disabled when playing (the actual behaviour too), but the per-clock tune (which are currently not shown when pausing the clock) now are show when pausing the clock (during game), to allow changing time in-game.

Hope it is more clear now :)

Giuseppe499:
> What I meant is that the global cog should be disabled when playing (the actual behaviour too)

As you can see from the changed files, this PR does not change the global cog behaviour.

ijm8710:
> > What I meant is that the global cog should be disabled when playing (the actual behaviour too)
> 
> 
> 
> As you can see from the changed files, this PR does not change the global cog behaviour.

I could be wrong but I think what Veloce is suggesting is to actually grey out the button so that it can't be actioned in this scenario

Giuseppe499:
> I could be wrong but I think what Veloce is suggesting is to actually grey out the button so that it can't be actioned in this scenario

I thought that too, but he said that the existing behaviour (before my PR) should remain unchanged.
If he wants that, I can easily do another commit with this feature.
@veloce, in that case, let me know

veloce:
My bad: indeed the cog is available when the clock is paused, and changing the setting just restart the clock. Let's keep that behaviour indeed @Giuseppe499

Giuseppe499:
@veloce Is there any update on this? If you need me to make some changes to get this merged, let me know

veloce (approved):
Tested on iOS simulator: works as intended.

PR #3461, opened 2026-07-19
Add animated app bar gradients to remaining settings menus

Matrix9494:
## Description

This updates the remaining settings menu screens that were still using plain `AppBar`s without the sliding gradient animation.

Covered screens:

- Background theme choice
- Board style choice
- Piece set choice
- Sound settings
- Chess engine settings

`PlatformAppBar`-based screens were already covered, and `ThemeSettingsScreen` already had `animateColor: true`, so this patch only fills the missing direct `AppBar` usages in the settings/menu flow.

### DIFF
--- a/lib/src/view/settings/background_theme_choice_screen.dart
+++ b/lib/src/view/settings/background_theme_choice_screen.dart
@@ -32,7 +32,7 @@ class BackgroundChoiceScreen extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
     return Scaffold(
-      appBar: AppBar(title: Text(context.l10n.background)),
+      appBar: AppBar(title: Text(context.l10n.background), animateColor: true),
       body: _Body(),
     );
   }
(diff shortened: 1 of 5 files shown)


lib/src/view/settings/background_theme_choice_screen.dart
```diff
@@ -32,7 +32,7 @@ class BackgroundChoiceScreen extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
     return Scaffold(
-      appBar: AppBar(title: Text(context.l10n.background)),
+      appBar: AppBar(title: Text(context.l10n.background), animateColor: true),
```
veloce: I'd rather have PlatformScaffold and PlatformAppBar instead for consistency.

veloce (commented):
Thanks. Have you verified those screens to ensure it is ok visually?

lib/src/view/settings/background_theme_choice_screen.dart
Matrix9494: Have fixed.

Matrix9494:
> Thanks. Have you verified those screens to ensure it is ok visually?

Yes the animation has been added to the rest of setting menus.

PR #3475, opened 2026-07-21
feat: Add same side mode for OTB/local play (#3076)

assassinaj602:
Implements feature #3076: Add 'same side' mode for OTB/local play. Both players can now view the board and clocks from the same perspective.


tom-anders:
Closing in favor of #3324

PR #3464, opened 2026-07-19
feat: add game phase accuracies to server analysis summary

tom-anders:
Screenshot_1784462222

PR #1499, opened 2025-02-27
Moved colors to lichess_color and added color to game analysis summary.

ItsDang:
For #815 Point 3, Coloring in the Computer Analysis Summary page. 

AnalysisSummaryColor

PR #3477, opened 2026-07-22
Moved colors to lichess_color and added color to game analysis summary.

tom-anders:
Screenshot_1784747176

Based on https://github.com/lichess-org/mobile/pull/1499

As discussed in https://github.com/lichess-org/mobile/pull/3464 we only show the colors if no custom background is used for now.

### DIFF
--- a/lib/src/styles/lichess_colors.dart
+++ b/lib/src/styles/lichess_colors.dart
@@ -87,4 +87,8 @@ class LichessColors {
   static const cyan = Color(0xFF56B4E9);
   static const blue = Color(0xFF0072B2);
   static const purple = Color(0xFF8572ff);
+
+  static const inaccuracy = cyan;
+  static const mistake = Color(0xFFe69f00);
+  static const blunder = Color(0xFFdf5353);
 }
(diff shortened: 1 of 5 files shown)


veloce:
@tom-anders thank. But there's one thing here: changing the inaccuracies color will disrupt the other colors for the analysis from PGN (where we have more annotations possible and where the color system was chosen to be consistent with the possible annotations.

For that reason I'd like to keep the old colors. I think people can make the mental switch that app can show different colors for inaccuracies. We've never received complaints that the color were different from the website (or not that I can remember of).

tom-anders:
@veloce Maybe I'm missing something (I based this on https://github.com/lichess-org/mobile/pull/1499, so the original code wasn't from me), but this does not change any of the colors, does it?

Previously in `pgn.dart`:

```
const inaccuracyColor = LichessColors.cyan;
const mistakeColor = Color(0xFFe69f00);
const blunderColor = Color(0xFFdf5353);
```

And now in `lichess_colors.dart`:

```
  static const inaccuracy = cyan;
  static const mistake = Color(0xFFe69f00);
  static const blunder = Color(0xFFdf5353);
```

So this just moved the constants and renamed named them slightly, but the colors are still the same

veloce:
Oh, my bad then. The color was already cyan.

veloce (approved):

PR #3492, opened 2026-07-26
fix: use existing l10n strings for game history display mode options (#3479)

### LINKED ISSUE #3479: Decide on strings for game history view options
There was a question on Crowdin as to where the strings `displayModeDetailed` and `displayModeCompact` appear. These were introduced by @tom-anders in https://github.com/lichess-org/mobile/pull/2582.

They were meant to be used in the game history view:
Screenshot

The values were hardcoded at some point, instead of calling the translatable variables:

https://github.com/lichess-org/mobile/blob/75ef1140479befe897faaef7d8306d191a20b918/lib/src/view/user/game_history_screen.dart#L101-L120

These strings have both added the word "view", when compared with the two translatable strings.

We need to choose which way we'll go with this menu:
1. Either to match the strings (so including the word "view"), and introduce new translation keys for them
2. Use the translatable text we have already.

assassinaj602:
Replace hardcoded 'Detailed view' and 'Compact view' strings with the existing mobileDisplayModeDetailed and mobileDisplayModeCompact translation keys, making the game history view options translatable.

Fixes #3479

### DIFF
--- a/lib/src/view/user/game_history_screen.dart
+++ b/lib/src/view/user/game_history_screen.dart
@@ -100,7 +100,7 @@ class GameHistoryScreen extends ConsumerWidget {
       actions: [
         ContextMenuAction(
           icon: Icons.ballot_outlined,
-          label: 'Detailed view',
+          label: context.l10n.mobileDisplayModeDetailed,
           onPressed: () {
             ref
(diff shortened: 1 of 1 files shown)


veloce (approved):
Good catch!

PR #3496, opened 2026-07-26
Show the number of puzzles left in the queue when offline

### LINKED ISSUE #499: Add number of remaining puzzles when offline

jadddon:
Closes #499

The puzzle tab previews already compute how many unsolved puzzles are cached for each angle, but the count is discarded --  `savedBatchesProvider` returns `IList<(PuzzleAngle, int)>` and `_MaterialTabBody` maps it down to `.$1`. Offline, that number is the only signal a user has about how much training is left before the queue runs dry.

Per @veloce's guidance in the issue thread, this adds a provider dedicated to this purpose rather than reusing `savedThemeBatchesProvider`. `savedBatchCountProvider` reads a single angle's batch via `PuzzleBatchStorage.fetch` -- deriving it from `savedBatchesProvider` wouldn't work, since `fetchAll` skips the mix angle, which is the entry most users see first.

`PuzzleAnglePreview` renders the count in its existing bottom row, gated on `isDeviceOnlineProvider`: online the queue is continuously refilled, so the number is noise.

The string is hardcoded English per `docs/internationalisation.md` .

2 tests added to the existing `tactical training preview` group (offline shows the count, online doesn't). `flutter analyze` is clean and the full suite passes.

### DIFF
--- a/lib/src/model/puzzle/puzzle_providers.dart
+++ b/lib/src/model/puzzle/puzzle_providers.dart
@@ -89,6 +89,20 @@ final savedBatchesProvider = FutureProvider.autoDispose<IList<(PuzzleAngle, int)
   return storage.fetchAll(userId: authUser?.user.id);
 }, name: 'SavedBatchesProvider');
 
+/// Fetches the number of unsolved puzzles saved locally for the given [PuzzleAngle].
+///
+/// Unlike [savedBatchesProvider], which is backed by [PuzzleBatchStorage.fetchAll] and skips the
+/// mix angle, this queries the batch of a single angle, so it also reports the mix count.
+final savedBatchCountProvider = FutureProvider.autoDispose.family<int, PuzzleAngle>((
+  Ref ref,
+  PuzzleAngle angle,
+) async {
+  final authUser = ref.watch(authControllerProvider);
+  final storage = await ref.watch(puzzleBatchStorageProvider.future);
+  final batch = await storage.fetch(userId: authUser?.user.id, angle: angle);
+  return batch?.unsolved.length ?? 0;
+}, name: 'SavedBatchCountProvider');
+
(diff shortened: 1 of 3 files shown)


veloce:
Can you post a screenshot or a screen recording of the feature please? Thanks.

jadddon:
Sure here are some screenshots from an iPhone simulator, device offline (Wi-Fi off), full queue of 50 cached puzzles.

**Offline, puzzle tab**



The count sits next to "Black to play" on the angle preview. Puzzle Streak and Puzzle Storm are greyed out because they need a connection — existing behaviour, but it's the giveaway that the device really is offline here.

**Solving one from the offline queue**



**Back on the tab, count has gone down**



50 → 49 after solving one, so it tracks the queue as it drains.

ijm8710:
Veloce, tiny point. Seeing 35 puzzles left may imply to a user who's not as versed with lichess that puzzles may be capped 

Would it make sense to have wording more similar to "35 puzzles in current batch"

Or "35 puzzles left offline"

Or "35 downloaded"

"35 queued" (I may like this the best)

Or something?

Not sure best wording but something that better implies it's just remaining in current sequence

jadddon:
That's great feedback, thanks @ijm8710. Will defer to @veloce or whoever else feels strongly about the new wording.

veloce:
Thanks!

Merged in https://github.com/lichess-org/mobile/commit/5720e6093d438c87d21cf392f92e07038a968171

veloce:
Oops, and I forgot to mention: I solved the ambiguity issue, by just displaying the number. It's then self explanatory I think, because it is shown only when offline, and the number decreases. Like that it also saves a translation.

jadddon:
Sounds good, thanks @veloce!

PR #3497, opened 2026-07-26
fix(tournament): keep "Open tournaments" header visible when featured feed is empty (#3489)

### LINKED ISSUE #3489: Home screen hides “Open tournaments” when featured feed is empty despite active arenas
### What is the problem?

The home screen can hide the entire **Open tournaments** section even while Arena tournaments are actively running.

`FeaturedTournamentsWidget` currently returns `SizedBox.shrink()` when the featured response contains no tournament supported by the app:

```dart
if (value.where((t) => t.isSupportedInApp).isEmpty) {
  return const SizedBox.shrink();
}
```

However, the tappable **Open tournaments** header that navigates to `TournamentListScreen` is inside that same widget. As a result, an empty or fully filtered featured feed removes the navigation entry itself.

The home-screen widget and native Arena list use different data sources:

- Featured preview: `/tournament/featured`
- Full Arena list: `/api/tournament`

This means `/tournament/featured` can be empty or filtered to zero supported entries while `/api/tournament.started` still contains active arenas.

**Expected behavior:** The Arena tournaments navigation entry remains visible even when there are no featured tournament cards.

**Actual behavior:** The entire section, including the navigation header, disappears.

1. Open the iOS app while Arena tournaments are actively running.
2. Reach a state where `/tournament/featured` returns no supported entries.
3. View the app home screen.
4. Observe that **Open tournaments** is hidden even though active tournaments exist.

### App version

0.25.5

### Device Info

iPhone / iOS 27.0

### Additional information

Relevant files:

- `lib/src/view/tournament/tournament_list_screen.dart`
- `lib/src/model/tournament/tournament_providers.dart`
- `lib/src/model/tournament/tournament_repository.dart`

Suggested fix: always render a navigation row or header that opens `TournamentListScreen`. Only hide the featured tournament cards when the supported featured list is empty.

GalaxyPhoenix716:
Fixes #3489

### Description
Currently, `FeaturedTournamentsWidget` returns `SizedBox.shrink()` whenever `value.where((t) => t.isSupportedInApp).isEmpty`. Because the tappable **"Open tournaments"** header (which navigates to `TournamentListScreen`) is rendered inside `FeaturedTournamentsWidget`, filtering out all items removes the navigation header entirely — even if active Arena tournaments are running on `/api/tournament`.

### Fix
Updated `FeaturedTournamentsWidget` to only return `SizedBox.shrink()` when the featured response is truly empty (`value.isEmpty`, e.g. when offline or when no tournaments are returned). If data is returned, the `"Open tournaments"` header remains visible for users to navigate to the full tournament list, while unsupported featured tournament cards continue to be filtered out.

### Changes Made
- **`lib/src/view/tournament/tournament_list_screen.dart`**: Changed empty check in `FeaturedTournamentsWidget` from `value.where((t) => t.isSupportedInApp).isEmpty` to `value.isEmpty`.
- **`test/view/tournament/tournament_list_screen_test.dart`**: Added unit widget tests covering both the empty state (hidden) and populated state (header visible).

### DIFF
--- a/lib/src/view/tournament/tournament_list_screen.dart
+++ b/lib/src/view/tournament/tournament_list_screen.dart
@@ -123,9 +123,6 @@ class FeaturedTournamentsWidget extends ConsumerWidget {
   Widget build(BuildContext context, WidgetRef ref) {
     switch (featured) {
       case AsyncData(:final value):
-        if (value.isEmpty) {
-          return const SizedBox.shrink();
-        }
         return ListSection(
           hasLeading: true,
           header: Text(context.l10n.openTournaments),
--- a/lib/src/view/tournament/tournament_list_screen.dart
+++ b/lib/src/view/tournament/tournament_list_screen.dart
@@ -123,7 +123,7 @@ class FeaturedTournamentsWidget extends ConsumerWidget {
   Widget build(BuildContext context, WidgetRef ref) {
     switch (featured) {
       case AsyncData(:final value):
-        if (value.where((t) => t.isSupportedInApp).isEmpty) {
+        if (value.isEmpty) {
           return const SizedBox.shrink();
         }
         return ListSection(
--- a/test/view/tournament/tournament_list_screen_test.dart
+++ b/test/view/tournament/tournament_list_screen_test.dart
@@ -56,16 +56,6 @@ void main() {
   });
 
   group('FeaturedTournamentsWidget', () {
-    testWidgets('hides entirely when the featured list is truly empty', (tester) async {
-      final app = await makeTestProviderScopeApp(
-        tester,
-        home: const FeaturedTournamentsWidget(featured: AsyncValue.data(IListConst([]))),
-      );
-      await tester.pumpWidget(app);
-      // No header should appear when there is no data (e.g. offline).
-      expect(find.text('Open tournaments'), findsNothing);
-    });
-
     testWidgets(
       'shows "Open tournaments" header when featured list has app-supported tournaments',
       (tester) async {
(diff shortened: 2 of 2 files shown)


lib/src/view/tournament/tournament_list_screen.dart
```diff
@@ -123,7 +123,7 @@ class FeaturedTournamentsWidget extends ConsumerWidget {
   Widget build(BuildContext context, WidgetRef ref) {
     switch (featured) {
       case AsyncData(:final value):
-        if (value.where((t) => t.isSupportedInApp).isEmpty) {
```
veloce, line 126: This is not supposed to filter anything anymore. This was needed at the time some variants where not supported in the app.

If you really found items filtered in that list, please post your findings here.

GalaxyPhoenix716: Fair point. I kept the value.isEmpty guard as a safety check for edge cases (like offline or empty tournament list), but actually offline returns AsyncError, not AsyncData([]). . I'll remove the if entirely so the header is always visible.

GalaxyPhoenix716:
@veloce kindly review

veloce:
@GalaxyPhoenix716 I still don't understand the fix.

You removed the filtering but the filtering should be kept. In the event (very unlikely but still) the server add a new variant, this filtering ensures that the app is not displaying an unsupported variant.

The app now supports all the variants, so this filtering should not filter anything.

That being said there may be an issue that make an empty widget to display. But it is not the one you mentioned.

GalaxyPhoenix716:
> @GalaxyPhoenix716 I still don't understand the fix.
> 
> You removed the filtering but the filtering should be kept. In the event (very unlikely but still) the server add a new variant, this filtering ensures that the app is not displaying an unsupported variant.
> 
> The app now supports all the variants, so this filtering should not filter anything.
> 
> That being said there may be an issue that make an empty widget to display. But it is not the one you mentioned.

Thanks for clarifying @veloce!

That makes total sense. That coditional statement ensures we don't render an empty section container if the server ever returns only unsupported variants in the future.

Since all variants are currently supported by the app anyway, I'll go ahead and close this PR. Thanks for taking the time to review!

PR #3498, opened 2026-07-26
fix(game): display PGN player name fallback for imported games (#3485)

### LINKED ISSUE #3485: Imported games don't show player names
In cases where a game has been imported and bookmarked, the app shows the game as anonymous - since no LightUser is returned. However, there is often PGN tag information on the players, so we could use that as a fallback option for labelling.

GalaxyPhoenix716:
Fixes #3485

### Description
For imported/bookmarked PGN games, no `LightUser` object is present because the players are not linked to Lichess user accounts. However, the player names from the PGN header tags are parsed into `Player.name`. 
Previously, `UserFullNameWidget` did not accept a `name` parameter and computed `displayName` as `user?.name ?? (aiLevel != null ? ... : l10n.anonymous)`, causing imported games to always display "Anonymous".

### Changes Made
- **`lib/src/widgets/user.dart`**: Added `name` property to `UserFullNameWidget` (and constructor `.player`), and updated `displayName` to check `user?.name ?? name ?? ...`.
- **Call sites (`game_list_tile.dart`, `game_list_detail_tile.dart`, `analysis_screen.dart`)**: Passed `name: opponent.name` / `name: player.name` to `UserFullNameWidget.player`.
- **`test/widgets/user_test.dart`**: Added widget tests for `UserFullNameWidget.player` fallback behavior.

### DIFF
--- a/lib/src/view/analysis/analysis_screen.dart
+++ b/lib/src/view/analysis/analysis_screen.dart
@@ -418,6 +418,7 @@ class _PlayerWidget extends StatelessWidget {
             Expanded(
               child: UserFullNameWidget.player(
                 user: player.user,
+                name: player.name,
                 rating: player.rating,
                 provisional: player.provisional,
                 aiLevel: player.aiLevel,
--- a/lib/src/view/game/game_list_detail_tile.dart
+++ b/lib/src/view/game/game_list_detail_tile.dart
@@ -57,6 +57,7 @@ class GameListDetailTile extends ConsumerWidget {
           builder: (context) => GameContextMenu(
             opponentTitle: UserFullNameWidget.player(
               user: opponent.user,
+              name: opponent.name,
               aiLevel: opponent.aiLevel,
               rating: opponent.rating,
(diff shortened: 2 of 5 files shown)


veloce:
I have a question: what is the behaviour of the lichess website for imported games? Can you send me a link to an imported game pls? thanks.

GalaxyPhoenix716:
> I have a question: what is the behaviour of the lichess website for imported games? Can you send me a link to an imported game pls? thanks.

https://lichess.org/35Md2cui

lib/src/view/analysis/analysis_screen.dart
```diff
@@ -418,6 +418,7 @@ class _PlayerWidget extends StatelessWidget {
             Expanded(
               child: UserFullNameWidget.player(
                 user: player.user,
+                name: player.name,
```
veloce, line 421: I checked the app, the player names are already shown in the analysis screen, so we don't need that change.

Issue #3421, opened 2026-07-06
In messages, if a link is included in brackets the closing bracket appears as part of the link


neilgd:
### What is the problem?

If a message contains a link, eg:

```
Please visit your challenges page (https:/lichessladders.com/challenges)
```

Then the closing ) appears part of the link in blue, so the link directs to https:/lichessladders.com/challenges) 

### Steps to reproduce the bug

Write a message and include a link in ()

### App version

0.24.26

### Device Info

iOS

ishaan-jindal:
the fix for this landed in #3504

PR #3504, opened 2026-07-29
Replace stalled flutter_linkify package

### LINKED ISSUE #2092: Replace flutter_linkify

ishaan-jindal:
Resolves #2092 

I wrote a custom `rich_link_text.dart` since, I couldn't find any other trust worthy and maintained packages.

Drop in replacement.
Removes external dependency over `flutter_linkify` and `linkify` packages

I did see a relavent PR #3282 for this issue, but it seemed stale.

Closes #3282

### DIFF
--- a/lib/src/widgets/rich_link_text.dart
+++ b/lib/src/widgets/rich_link_text.dart
@@ -1,0 +1,332 @@
+import 'package:flutter/gestures.dart';
+import 'package:flutter/material.dart';
+
+typedef LinkCallback = void Function(LinkableElement link);
+
+abstract class LinkifyElement {
+  final String text;
+  final String originText;
+
+  LinkifyElement(this.text, [String? originText]) : originText = originText ?? text;
+
+  @override
+  String toString() {
+    return '$runtimeType: "$text"';
+  }
+}
+
+class TextElement extends LinkifyElement {
+  TextElement(super.text);
+}
+
+class LinkableElement extends LinkifyElement {
+  final String url;
+
+  LinkableElement(String? text, this.url, [String? originText]) : super(text ?? url, originText);
+}
+
+class UrlElement extends LinkableElement {
+  UrlElement(String url, [String? text, String? originText]) : super(text, url, originText);
+}
+
+class EmailElement extends LinkableElement {
+  final String emailAddress;
+
+  EmailElement(this.emailAddress) : super(emailAddress, 'mailto:$emailAddress');
+}
+
+class UserTagElement extends LinkableElement {
+  final String userTag;
+
+  UserTagElement(this.userTag) : super(userTag, userTag);
+}
+
+class LinkifyOptions {
+  final bool humanize;
+  final bool defaultToHttps;
+  final bool excludeLastPeriod;
+
+  const LinkifyOptions({
+    this.humanize = true,
+    this.defaultToHttps = false,
+    this.excludeLastPeriod = true,
+  });
+}
+
+abstract class Linkifier {
+  const Linkifier();
+
+  List<LinkifyElement> parse(List<LinkifyElement> elements, LinkifyOptions options);
+}
+
+final _urlRegex = RegExp(r'^(.*?)(https?:\/\/\S+|www\.\S+)', caseSensitive: false, dotAll: true);
+
+final _protocolRegex = RegExp(r'^https?:\/\/', caseSensitive: false);
+
+class UrlLinkifier extends Linkifier {
+  const UrlLinkifier();
+
+  @override
+  List<LinkifyElement> parse(List<LinkifyElement> elements, LinkifyOptions options) {
+    final result = <LinkifyElement>[];
+
+    for (final element in elements) {
+      if (element is! TextElement) {
+        result.add(element);
+        continue;
+      }
+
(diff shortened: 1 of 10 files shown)


ishaan-jindal:
@veloce please take a look

lib/src/widgets/rich_link_text.dart
```diff
@@ -0,0 +1,331 @@
...
+    this.onOpen,
+    this.options = const LinkifyOptions(),
+    this.style,
+    this.linkStyle,
+    this.textAlign = TextAlign.start,
+    this.textDirection,
+    this.maxLines,
+    this.overflow = TextOverflow.clip,
+    this.textScaler = TextScaler.noScaling,
+  });
```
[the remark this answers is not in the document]
ishaan-jindal: alright

lib/src/widgets/rich_link_text.dart
```diff
@@ -0,0 +1,331 @@
...
+      if (match.group(2)?.isNotEmpty ?? false) {
+        var matchedUrl = match.group(2)!;
+        var originText = matchedUrl;
+        var trailing = '';
+
+        if (options.excludeLastPeriod && matchedUrl.endsWith('.')) {
+          matchedUrl = matchedUrl.substring(0, matchedUrl.length - 1);
+          originText = originText.substring(0, originText.length - 1);
+          trailing = '.$remaining';
+        } else {
```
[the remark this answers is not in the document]
ishaan-jindal: set `remaining = '';` on line 100. prevents the post-URL text from being parsed twice

ishaan-jindal:
PTAL @veloce

ishaan-jindal:
@veloce please take a look

lib/src/widgets/rich_link_text.dart
```diff
@@ -0,0 +1,338 @@
...
+      );
+    } else {
+      children.add(TextSpan(text: element.text, style: style));
+    }
+  }
+
+  return TextSpan(children: children);
+}
+
+class Linkify extends StatelessWidget {
```
veloce: I'd call it `RichLinkText` so it matches the file name (and it's a better name).

Also I'd put it at the top of the file (kind of a convention in this project).

lib/src/widgets/rich_link_text.dart
```diff
@@ -0,0 +1,338 @@
...
+      if (match.group(2)?.isNotEmpty ?? false) {
+        var matchedUrl = match.group(2)!;
+        var originText = matchedUrl;
+        var trailing = '';
+
+        if (options.excludeLastPeriod && matchedUrl.endsWith('.')) {
+          matchedUrl = matchedUrl.substring(0, matchedUrl.length - 1);
+          originText = originText.substring(0, originText.length - 1);
+          trailing = '.$remaining';
+          remaining = '';
```
veloce: `remaining` is cleared before the recursive parse, so everything after a URL that ends with a period is emitted as one plain text run and never linkified.

We should fix that and have a test case to ensure we can linkify several urls that end with a period.

lib/src/widgets/rich_link_text.dart
```diff
@@ -0,0 +1,338 @@
...
+  const Linkifier();
+
+  List<LinkifyElement> parse(List<LinkifyElement> elements, LinkifyOptions options);
+}
+
+final _urlRegex = RegExp(r'^(.*?)(https?:\/\/\S+|www\.\S+)', caseSensitive: false, dotAll: true);
+
+final _protocolRegex = RegExp(r'^https?:\/\/', caseSensitive: false);
+
+final _trailingPunctRegex = RegExp(r'[.,;:!?)]$');
```
veloce, line 167: In this regex, trailing ) is stripped unconditionally, truncating Wikipedia-style URLs.

Ex: `http://en.wikipedia.org/wiki/Saw_(film)` should not be stripped, but `(https://example.com)` should.

See also: https://github.com/lichess-org/mobile/pull/3566/changes

lib/src/widgets/rich_link_text.dart
```diff
@@ -0,0 +1,338 @@
...
+  EmailElement(this.emailAddress) : super(emailAddress, 'mailto:$emailAddress');
+}
+
+class UserTagElement extends LinkableElement {
+  final String userTag;
+
+  UserTagElement(this.userTag) : super(userTag, userTag);
+}
+
+class LinkifyOptions {
```
veloce: This was part of linkify package but I don't think we want to have options. Let's remove them and keep the sane defaults since we are not using the options anywhere in the code afair.

veloce (commented):
Thanks for the work @ishaan-jindal , here's the review

lib/src/widgets/rich_link_text.dart
ishaan-jindal: I added the balancing logic into the parser itself. #3566 is probably redundant now.
And hence closes #3421

veloce:
Currently testing it live @ishaan-jindal ; don't worry about the copilot commit that made the tests fail, I'll revert it. I just have to be afk now, will resume later.

PR #3510, opened 2026-07-30
Added icons to SignIn, Donate and About Lichess 

Anurajx:
Added icons to SignIn, Donate and About Lichess on home screen and made appropriate changes to tests for it

### Before and After
Screenshot_20260730_200750
Screenshot_20260730_201924


Anurajx:
@veloce

superuser-does:
Would it be possible to use the wings icon instead please on the Donate button?

superuser-does:
I would also say as feedback that [it's sometimes better to go without an icon](https://tonsky.me/blog/tahoe-icons/) and I would suggest taking the icon off of <kbd>About Lichess...</kbd>. Personally I think that shouldn't even be a button, but that can be saved for another PR.

Icons can be kept on <kbd>Sign in</kbd> to match with the user icon circle on the top-right, and on <kbd>Donate</kbd> to match the website and as an implicit explanation when people see the wings on player profiles.

Anurajx:
thanks for the feedback and blog great read. I have made listed changes here's how it looks now 

Screenshot_20260731_192336

superuser-does:
Thanks, looks great to my eyes.

PR #3515, opened 2026-08-01
add abort training l10n

aleksandr4842:


veloce:
Before adding new translations, please read the internationalisation guide. Translations are managed centrally and these files should not be edited by hand.

PR #3532, opened 2026-08-03
feat: allow creating new study chapters

tom-anders:
[create-chapter.webm](https://github.com/user-attachments/assets/3afe9700-5418-4a17-9cec-330e4baaae22)

### DIFF
--- a/lib/src/view/study/create_study_chapter_bottom_sheet.dart
+++ b/lib/src/view/study/create_study_chapter_bottom_sheet.dart
@@ -1,0 +1,323 @@
+import 'dart:convert';
+
+import 'package:dartchess/dartchess.dart';
+import 'package:fast_immutable_collections/fast_immutable_collections.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter/services.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+import 'package:lichess_mobile/src/model/common/chess.dart';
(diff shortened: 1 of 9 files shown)


lib/src/view/study/create_study_chapter_bottom_sheet.dart
```diff
@@ -0,0 +1,316 @@
...
+import 'package:lichess_mobile/src/view/more/import_pgn_screen.dart';
+import 'package:lichess_mobile/src/widgets/adaptive_bottom_sheet.dart';
+import 'package:lichess_mobile/src/widgets/adaptive_choice_picker.dart';
+import 'package:lichess_mobile/src/widgets/board_preview.dart';
+import 'package:lichess_mobile/src/widgets/feedback.dart';
+import 'package:lichess_mobile/src/widgets/list.dart';
+
+sealed class CreateStudyChapterParams {}
+
+class CreateChapterOfExistingStudy extends CreateStudyChapterParams {
```
tom-anders, line 23: Just to prevent confusion: There's only one class actually extending `CreateStudyChapterParams` right now, but in the future there will also be `CreateChapterOfNewStudy` (for the "create study" flow)

veloce:
Tested this live on lichess.org.

Changes made:
- Use a `Card` instead of a `ListSection`
- Added error handling
- Fixed a bug where the keyboard would overlap with the bottom sheet.

PR #3544, opened 2026-08-07
add nextChapterButton in non gamebook studies

### LINKED ISSUE #3525: Missing next chapter button as result of adding hamburger menu to studies
See before and after 

image

image

HaonRekcef:
fix #3525 

image


HaonRekcef:
Test were failing as pumpAndSettle never settled, as the bottom bar was blinking with the empty pgn, is just added two moves now, so it doesnt.

veloce:
Ok, despite the comment in #3525 it might be useful for some studies so let's add it.

PR #3548, opened 2026-08-07
Auto-paste FEN in board editor while allowing edit on error

jmr:
Previously, tapping the FEN edit button opened a read-only dialog that automatically pasted from the clipboard and immediately closed the dialog regardless of whether the FEN was valid or invalid. If the pasted FEN had a typo or was invalid, the user was shown an error snackbar without any opportunity to edit or correct the string.

Make the text field interactive and prepopulate it with the current position's FEN. On open, auto-read the clipboard: if it contains a valid FEN, automatically load the position and close the dialog to preserve 1-touch convenience. If the FEN is invalid or unparseable, keep the dialog open with the string focused for editing and display inline error feedback.

### DIFF
--- a/lib/src/view/board_editor/board_editor_screen.dart
+++ b/lib/src/view/board_editor/board_editor_screen.dart
@@ -58,6 +58,7 @@ class BoardEditorScreen extends ConsumerWidget {
             onPressed: () => showDialog<void>(
               context: context,
               builder: (_) => _FenDialog(
+                initialFen: boardEditorState.fen,
                 onFenLoaded: (fen) =>
                     ref.read(boardEditorControllerProvider(params).notifier).loadFen(fen),
               ),
@@ -460,8 +461,9 @@ class _BottomBar extends ConsumerWidget {
 }
 
 class _FenDialog extends StatefulWidget {
-  const _FenDialog({required this.onFenLoaded});
+  const _FenDialog({required this.initialFen, required this.onFenLoaded});
 
+  final String initialFen;
   final void Function(String fen) onFenLoaded;
 
   @override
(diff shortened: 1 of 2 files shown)


ijm8710:
My only question is why would it ever need to be modified. Correct me if I'm wrong but you're virtually never typing in the fen. It's grabbed from somewhere else and should be good. If it's not good your probably need to re-grab it not edit it

jmr:
I take screenshots/pictures of boards/books and give them to LLMs to transcribe to FEN. When it's wrong, it's very annoying since I just get "invalid FEN" with no info. With the dialog box, I could scan the FEN and try to correct it myself. I'll also make a PR to provide more detail in the error.

ijm8710:
Eh personally I hate the idea of using LLMS to transcribe. They're nowhere close to ready to do it competently. You're better off using chessvision. I still don't think fen construction should be a guessing game composed like a rough draft

HaonRekcef:
I don't see the point of allowing FEN edits for IllegalSetups, seems way more complicated then allowing IllegalSetups to load the board editor and then being able to adjust the position on the editor.

For malformed FENs this could make sense though.

jmr:
> seems way more complicated then allowing IllegalSetups to load the board editor and then being able to adjust the position on the editor.

Good idea. I now understand there are separate illegal FEN and illegal position errors. I'll make a PR to treat a FEN with an illegal position the same way as if you had clicked the pieces there. (Editable, but no analysis possible.)

> I still don't think fen construction should be a guessing game composed like a rough draf

The user experience is exactly the same for correct FEN and improved if incorrect, so this seems like an improvement to me.

jmr:
> I'll make a PR to treat a FEN with an illegal position the same way as if you had clicked the pieces there.

https://github.com/lichess-org/mobile/pull/3687

PR #3550, opened 2026-08-08
Provide specific localized error messages for FEN paste

jmr:
When a user pastes an invalid FEN string into the board editor FEN dialog, the app displays a generic "Invalid FEN" SnackBar message. Users receive no feedback explaining whether the failure was caused by malformed FEN syntax or illegal board positions.

Displaying generic error messages makes diagnosing invalid FEN strings difficult. Catching strongly-typed exceptions from package:dartchess allows providing specific, actionable feedback.

Catch PositionSetupException and FenException from package:dartchess using typed 'on ... catch' clauses, inspecting their underlying .cause enums (IllegalSetupCause and IllegalFenCause) to map each error condition to a dedicated localized string in mobile.xml.


HaonRekcef:
Code looks good to me, thanks!

I'm wondering if we should change when the error snackbar displays. Instead of showing it right when a user copies a FEN into the editor, maybe we can only show it when they try to navigate away to the analysis board or challenge from position?

I remember a user telling about that they scan positions with AI and then try to import the FEN, where they needed an easier way to adjust minor position mistplacements after pasting. To support that, my idea is that we show a snackbar for IllegalFenCause, but allow IllegalSetupCause and just improve the feedback when they actually try to navigate away from the editor. (currently we just disable the button which is bad UX)

What do you think?

edit: ah it was you in a PR I see it now, was only checking the issues :facepalm:

veloce:
While this is a good improvement, I think it should be a joint effort with the website and not a mobile-only translation. cc @superuser-does

jmr:
> I think it should be a joint effort with the website and not a mobile-only translation.

Do you have a pointer to what I should copy to share translations with the website? What translations in the mobile app do this currently?

superuser-does:
Agreed, translation-wise you can get the strings in by following this workflow: https://github.com/lichess-org/lila/wiki/Tips-for-UI%2C-translations%2C-insights%2C-swiss%2C-accessibility

You can see  an example of _just_ strings being added here: https://github.com/lichess-org/lila/pull/21011

So you could perhaps do a PR on Lila that doesn't change thet logic but just adds the strings, with the expectation that the validation logic and error messaging will be added in time.

It's worth checking on the [Lichess Discord's](https://discord.gg/lichess) `#lichess-development` channel if lila developers would accept that i.e. adding the strings to lila without the logic. Unless of course you do the logic on that side as well. :)

PR #3554, opened 2026-08-10
Fix puzzle history stale solved/incorrect state

### LINKED ISSUE #3549: Puzzle history list has stale "solved"/"incorrect" state
### What is the problem?

Reported by email: "Hi,
The new iPad app has gotten really fabulous. Thank you. I wouldn't call this a big bug but just trying to help. You can see in the attached screenshot that the last puzzle I tried shows that I got it incorrect. I did on the first try. I then figured out my mistake, and solved it correctly. I returned to the main puzzle page and it still shows incorrect. I go to the homepage and return to the puzzle page and it still shows incorrect. The only way I've been able to get it to update is to kill the app and restart it. Like I said not a huge bug, but I like trying to be helpful. I don't know what I would do without Lichess!
Thanks"

1. Go to the puzzle history list 
2. Click a puzzle that says incorrect
3. Solve the puzzle 
4. Go back to the puzzle history list and see that it still says incorrect
5. Restart the app and the same puzzle now shows solved

### App version

0.26.2

### Device Info

Pixel 7

### Additional information

_No response_

Mohamadmourad:
Summary
 - Fixes #3549
 - After solving a puzzle, puzzleRecentActivityProvider is now invalidated in PuzzleController._onFailOrWin(), so the
  puzzle history list fetches fresh data from the server instead of showing stale cached results.

 Root cause

 puzzleActivityProvider watches puzzleRecentActivityProvider (which fetches from /api/puzzle/activity) and caches the
 result for 5 minutes. After solving a puzzle, neither provider was invalidated, so navigating back to the history
 screen kept showing the old solved/incorrect state until the app was restarted.

 Test plan
 - Added unit test verifying that invalidating puzzleRecentActivityProvider causes puzzleActivityProvider to re-fetch
 from the server


test/model/puzzle/puzzle_activity_test.dart
```diff
@@ -0,0 +1,55 @@
...
+import 'package:lichess_mobile/src/network/http.dart';
+
+import '../../test_container.dart';
+import '../../test_helpers.dart';
+import '../auth/fake_auth_storage.dart';
+import 'mock_server_responses.dart';
+
+void main() {
+  // triggers puzzleActivityProvider to rebuild with fresh data.
+  test('invalidating puzzleRecentActivityProvider re-fetches puzzle history', () async {
```
veloce: This test is not really useful I think. It is clear that invalidating the provider would re-fetch the history.

A more useful test would be a widget test:
- go to puzzle history
- do a puzzle
- go again to puzzle history: puzzle should be there

But not mandatory for this fix.

veloce (approved):
lgtm!

PR #3556, opened 2026-08-11
change RichText to Text.rich() to scale with system font size

HaonRekcef:
I changed all the RichText to Text.rich() to scale  with the system font size, it is suggested here:
https://api.flutter.dev/flutter/widgets/RichText-class.html . The problem was reported in the google play reviews.

> Consider using the [Text](https://api.flutter.dev/flutter/widgets/Text-class.html) widget to integrate with the [DefaultTextStyle](https://api.flutter.dev/flutter/widgets/DefaultTextStyle-class.html) automatically. When all the text uses the same style, the default constructor is less verbose. The [Text.rich](https://api.flutter.dev/flutter/widgets/Text/Text.rich.html) constructor allows you to style multiple spans with the default text style while still allowing specified styles per span.

Pictures of all the occurrences changed with the maximum android font size for dramatic effect:

| RichText | Text.rich() |
| ----| ---|
|image|image |
|image | image|
| image| image|
|image |image|

### DIFF
--- a/lib/src/view/broadcast/broadcast_game_screen.dart
+++ b/lib/src/view/broadcast/broadcast_game_screen.dart
@@ -437,8 +437,8 @@ class _PgnTagsView extends ConsumerWidget {
                         final value = pgnHeaders[tag.tagName]!;
                         final url = tag.isLink ? tag.buildUrl(value) : null;
                         if (url != null) {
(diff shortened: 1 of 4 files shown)


veloce (approved):
Good catch. Thanks!

Issue #3541, opened 2026-08-06
Feature request: AMOLED board pattern


r0landd:
e.g. #000000 and #202020

superuser-does:
There have been these requests for the website too. We have never implemented them officially because they obscure the black pieces in many piece sets, including the default. These sets don't have an outline so they'd basically "blend in" with black squares.

I think this would be problematic to officially support when so many piece sets would look so poor on it, so I am not sure if it would be desirable, mainly because of the pure-black squares (LEDs turned off completely on an AMOLED displays).

PR #3557, opened 2026-08-11
amoled background

HaonRekcef:
adds a 4th background option, Amoled black. It turns a lot of surfaces to pitch black (RGB 0,0,0)
Video:

[Screencast_20260811_102647.webm](https://github.com/user-attachments/assets/de07ea5f-1d37-48e7-a770-fb8e77a30c8b)

The advantage is that amoled screens physically turn off pixels to display true black (RGB 0,0,0), which reduces battery consumption. Also I personally like the look of a pure black background.

Note that I did not make all surfaces pitch black, so that for example the search bar is still highlighted in a different color.

Inspired by #3541

### DIFF
--- a/lib/src/model/settings/brightness.dart
+++ b/lib/src/model/settings/brightness.dart
@@ -20,13 +20,10 @@ class CurrentBrightness extends Notifier<Brightness> {
       }
     };
 
-    switch (themeMode) {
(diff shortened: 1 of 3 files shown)


veloce (approved):
LGTM

PR #3558, opened 2026-08-11
fix game layout on foldables

### LINKED ISSUE #3145: Material notation hidden behind chessboard on Z Fold 5
### What is the problem?

When I play chess on bigger screen on samsung z fold 5 the material notation is hidden behind chessboard. I think it is related to scaling chessboard. When it is scaled out on such aspect ratio player is not able to see all elements of gameplay. Scaling should be fitted to size that will not cover the rest elements.

Image

Play game on device with similar aspect ratio like z fold 5

### App version

0.23.13

### Device Info

Samsung Z Fold 5 / Android 16 one UI 8.0

### Additional information

_No response_

HaonRekcef:
Fixes the overflow for z-fold devices, as simple as possible.
Fixes #3145 

after:
image

before:
image

### DIFF
--- a/lib/src/widgets/game_layout.dart
+++ b/lib/src/widgets/game_layout.dart
@@ -496,6 +496,7 @@ class _GameLayoutState extends ConsumerState<GameLayout> {
           );
         } else {
           final defaultBoardSize = constraints.biggest.shortestSide;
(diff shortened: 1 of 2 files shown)


veloce (approved):
LGTM. We can still improve later but I like the simplicity of this fix.

PR #3567, opened 2026-08-14
PR template

HaonRekcef:
Add an PR template, largely copied from lila


veloce (commented):
Nice, thanks! The only missing thing is a mention about screenshots / screen recording if applicable. Will add it.

PR #3570, opened 2026-08-15
Ignore own chat messages in unread counts

### LINKED ISSUE #3418: annoying game chat badge unread increment
When:

- I offer a draw or a rematch
- I say "Good game, well played" automatically

Then the game chat badge counter should not increment.

In general, messages of a user should not increment unreads of that user

Related: https://github.com/lichess-org/mobile/issues/1855

HaonRekcef:
fix #3418 

Don't increase the badge if the message was sent by the own account (f.ex. with the gg wp setting) or if it is the own draw offer or draw decline message by us.

I am not sure if we should also filter out `rematch sent ` messages, as this is a bit disturbing too, the problem is that it is the same message no matter which color sent the request.

Tested manually, video:

[Screencast_20260815_170847.webm](https://github.com/user-attachments/assets/30e48e70-2480-4503-8ad0-d69751c6940c)

 - AI tools used:
 Copilot to help generate the tests and autocomplete

### DIFF
--- a/lib/src/model/chat/chat_mixin.dart
+++ b/lib/src/model/chat/chat_mixin.dart
@@ -1,6 +1,7 @@
 import 'dart:async';
 import 'dart:math' as math;
 
+import 'package:dartchess/dartchess.dart';
 import 'package:fast_immutable_collections/fast_immutable_collections.dart';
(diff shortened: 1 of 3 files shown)


veloce (approved):
LGTM. Thanks!

PR #3571, opened 2026-08-15
add showThreat +Boardeditor and continue from here to Broadcasts and studies

### LINKED ISSUE #2942: Adding some additional analysis board options for the Broadcast boards
on web, broadcast allows for 

- “show Threat” 
- “board Editor” 
- “continue from here”

Do you think we should have a way to access those in the app similar to how we can for a normal analysis board? "Continue from here" is the one I'm most interested in.

HaonRekcef:
closes #2942

Adds  showThreat, Boardeditor and continue from here to broadcasts and studies (if Computer Analysis is allowed), I extracted it in a separate file as it is reused in 3 places now:

Screenshots:
image
image

AI-Tools used:
Copilot to refactor into a shared file and autocomplete.

### DIFF
--- a/lib/src/view/analysis/analysis_actions.dart
+++ b/lib/src/view/analysis/analysis_actions.dart
@@ -1,0 +1,38 @@
+import 'package:dartchess/dartchess.dart';
+import 'package:flutter/widgets.dart';
+import 'package:lichess_mobile/src/model/common/chess.dart';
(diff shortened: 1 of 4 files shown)


ijm8710:
Super nitpicky; but for consistency, we always have flip
Board after settings when it's in the menu. And members kinds goes with chat room.

Are you down to swap the order of flip Board and members?

veloce (approved):
Looks good! thanks.

PR #3574, opened 2026-08-16
Fix broadcast error when analysis summary has no accuracy

Assios:
Loading a broadcast game without accuracy / acpl crashes with a `PickException`. This can happen when a game only has one ply, like this one: https://lichess.org/broadcast/kanizsa-open-2026-group-a-top-boards/round-1/PohT1PTB/kBZCI9JC

Screenshot 2026-08-16 at 12 58 58

This video shows the crash: https://github.com/user-attachments/assets/4542eab2-a336-4914-a0e1-1b47ab3f5875

Verified and tested the fix in the iOS simulator.


veloce (approved):
Good catch. Thanks!

Issue #1258, opened 2024-12-11
Multiple premoves 


Boubou78000:
Add the very useful option to make several premoves (that play in the order)

This feature already exists on chess com, and the only small complexity is to handle the new pieces’ positions when doing multiple premives.

HaonRekcef:
This is something that would need to be on the main website first. It has been discussed there a lot: https://github.com/lichess-org/lila/issues/3890
This issue can be closed.

tom-anders:
agreed, won't be added to the app unless the website gets it first

PR #3575, opened 2026-08-16
Add multiple premoves

AlexZander85:
## Summary

Add an opt-in multiple-premove mode to live games while preserving the existing single-premove behavior for existing users.

This depends on lichess-org/flutter-chessground#104, which adds the bounded premove queue to the board controller.

## User-facing behavior

Board settings now expose three premove modes:

- **Multiple** — queue up to 10 premoves
- **One** — existing single-premove behavior
- **Disabled** — disables premoves

Existing users remain on **One** unless they explicitly select **Multiple**.

## Implementation

The live-game `_PlayableGameBoard` already owns its `ChessboardController`, so the selected premove mode is applied there rather than in generic board widgets used by analysis or puzzles.

After each authoritative opponent update, only the current queue head is validated against the real game position. If legal, exactly one premove is consumed and submitted through the existing move path. If the head is illegal, the dependent queue is cleared. A promotion requiring user input also clears the queued sequence and follows the existing promotion UI.

The server protocol, websocket transport, authentication, matchmaking, and move submission protocol are unchanged. The server still receives each executed premove as a normal move.

## Compatibility / migration

The existing `premoves` preference is retained. A new `multiplePremoves` flag makes the migration backward-compatible:

- existing enabled premoves → **One**
- existing disabled premoves → **Disabled**
- **Multiple** is only enabled explicitly

## Dependency

While lichess-org/flutter-chessground#104 is under review, this Draft temporarily pins `chessground` to the tested fork commit `cb95459a4e5771bfbc7e7db2c477f6a319ab81bc`.

Before this PR is marked ready for review, that temporary pin will be replaced with the official package/version once the core change is merged/released.

## Tests

Tests cover:

- default/legacy preference migration
- persistence of Multiple / One / Disabled
- consuming exactly one legal queued head
- clearing a queue whose head is no longer legal against the authoritative position

The clean validation branch passed dependency installation, model code generation, formatting, `flutter analyze`, and the full `flutter test` suite in GitHub Actions.

I also manually tested the feature on a physical Android device against lichess.org using the forked application. The device/OS blocks internal screen capture during a live game, so I was not able to attach an on-device recording yet; I can provide an externally recorded demonstration if useful.

## AI disclosure

OpenAI ChatGPT (GPT-5.6 Sol) was used to inspect the earlier prototype, isolate the upstream-sized change, compare the Flutter and web approaches, identify edge cases, and assist with implementation and test preparation. I have personally reviewed the submitted diff, understand the implementation, and manually tested the feature on the forked Android application.

### DIFF
--- a/lib/src/model/settings/board_preferences.dart
+++ b/lib/src/model/settings/board_preferences.dart
@@ -13,6 +13,7 @@ part 'board_preferences.g.dart';
 
 const kBoardDefaultBrightnessFilter = 1.0;
 const kBoardDefaultHueFilter = 0.0;
+const kMultiplePremoveLimit = 10;
 
 final boardPreferencesProvider = NotifierProvider<BoardPreferences, BoardPrefs>(
   BoardPreferences.new,
@@ -44,6 +45,17 @@ class BoardPreferences extends Notifier<BoardPrefs> with PreferencesStorage<Boar
     await save(state.copyWith(boardTheme: boardTheme));
   }
 
+  Future<void> setPremoveMode(PremoveMode mode) async {
+    await save(
+      state.copyWith(
+        premoves: mode != PremoveMode.disabled,
+        multiplePremoves: mode == PremoveMode.multiple,
+      ),
+    );
+  }
+
+  /// Compatibility helper for callers that still expose the old on/off switch.
+  /// New UI should use [setPremoveMode].
   Future<void> togglePremoves() async {
     await save(state.copyWith(premoves: !state.premoves));
   }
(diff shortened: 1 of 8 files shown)


ijm8710:
So I'd really really like to see this but if this is not supported natively on web, do you see any chance it gets acceptance on mobile

Again, I'd love it and especially if it's opt-in I don't see the harm but something like this, I have hesitance that thibault would be cool with divergent philosophies from web to app, especially when I infer that the lack of premoves on web seemed intentional.

AlexZander85:
> So I'd really really like to see this but if this is not supported natively on web, do you see any chance it gets acceptance on mobile
> 
> Again, I'd love it and especially if it's opt-in I don't see the harm but something like this, I have hesitance that thibault would be cool with divergent philosophies from web to app, especially when I infer that the lack of premoves on web seemed intentional.

That's a fair concern, and after looking through the earlier discussions I think you're right that this should not be treated as a mobile-only divergence.

I found lichess-org/mobile#1258, where the conclusion was explicitly that multiple premoves should not be added to the app unless the website gets them first, as well as the older discussion in lichess-org/lila#3890.

I opened this as a Draft partly to make the implementation concrete and to demonstrate the downstream use of lichess-org/flutter-chessground#104, rather than to argue that mobile should independently adopt a different gameplay philosophy.

I think the right next step is to explore the same behavior on web first (or in parallel), keeping the current single-premove behavior as the default and making multiple premoves strictly opt-in.

If the web maintainers are not interested in revisiting that decision, then I would not expect the mobile PR to proceed independently either. But if they are open to reconsidering it, I'd be happy to work on the web implementation as well.

veloce:
@AlexZander85 thanks for contributing but I'm afraid I won't accept this. You should have asked before starting it if we'd want such a feature.

Reasons for not accepting is:
- web client does not support this either
- reasons mentioned in the lila issue above
- complexity and a maintenance cost that we are not ready to accept now

PR #3579, opened 2026-08-17
add pinned comment to broadcast boards tab

### LINKED ISSUE #3130: Tiebreak banner for broadcasts
image

HaonRekcef:
closes #3130 as https://github.com/lichess-org/lila/commit/cf0c9f3e45415135f3e2261b4ef8c9b2cf5631c1 was deployed now.

Preview:

image

image

Remarks:
- I tried to make sure the safe area works on iOS too, but I cannot test it.
- It takes vertical space in the boards tab, which I do not like that much, but I think in the overview tab it would not be seen. I think pinned comments are mostly used for tiebreaks regulations or if there are lichess server problems so it might be useful to show prominently in the boards tab.


veloce:
Perhaps as a future improvement we could add a close button to enable people to dismiss it. But it can be done in another PR.

PR #3581, opened 2026-08-18
improve 960 support in editor

### LINKED ISSUE #1179: Castling is possible even if the king the misplaced
Latest lichess beta on iOS
Btw this is an edited board but castling in that position is still impossible

https://reddit.com/r/AnarchyChess/comments/1gyicui/guys_whats_this_move_called/

https://imgur.com/a/2hkn8jA

HaonRekcef:
closes #1179 

Makes 960 its own variant in the editor, this helps makes the editor less confusing for people entering a standard chess position, where the rooks/kings are on the backrank but not on their initial square. Currently when entering a position with the Rook, lets say on g1, the default was that kingside was still possible which can mess with the analysis later.

Things implemented/fixed:

- In all chess Variants except 960: if Kings + Rooks not on initial square the choice chips for castling are not interactable
- In chess960: if Kings + Rooks not on backrank the choice chips for castling are not interactable
- In variants where castling is not possible, do not add castling rights to the fen, leave them empty instead
- Only enable load position in standard chess as for other chess variants the opening and endgame positions are not relevant
- Open a 960Analysisboard if the variant is 960 (just a visual change to add the 960 icon in the analysis header)
- Adds a 960 randomizer and position loader to the board editor
- Added a bunch of tests

Video:

[960 Editor](https://github.com/user-attachments/assets/70c83996-29aa-44ea-bb9b-451c8ff53984)

AI tools:
Copilot, mostly for test generation and autocomplete

### DIFF
--- a/lib/src/view/board_editor/board_editor_screen.dart
+++ b/lib/src/view/board_editor/board_editor_screen.dart
@@ -1,3 +1,4 @@
+import 'dart:math' as math;
 import 'package:chessground/chessground.dart';
 import 'package:cupertino_ui/cupertino_ui.dart';
 import 'package:dartchess/dartchess.dart';
@@ -8,8 +9,10 @@ import 'package:lichess_mobile/src/model/analysis/analysis_controller.dart';
 import 'package:lichess_mobile/src/model/auth/auth_controller.dart';
 import 'package:lichess_mobile/src/model/board_editor/board_editor_controller.dart';
 import 'package:lichess_mobile/src/model/common/chess.dart';
+import 'package:lichess_mobile/src/model/common/chess960.dart';
 import 'package:lichess_mobile/src/model/common/id.dart';
 import 'package:lichess_mobile/src/model/settings/board_preferences.dart';
+import 'package:lichess_mobile/src/styles/lichess_icons.dart';
 import 'package:lichess_mobile/src/styles/styles.dart';
 import 'package:lichess_mobile/src/utils/l10n_context.dart';
 import 'package:lichess_mobile/src/utils/navigation.dart';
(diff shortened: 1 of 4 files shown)


lib/src/view/board_editor/board_editor_screen.dart
```diff
@@ -514,3 +527,72 @@ class _FenDialogState extends State<_FenDialog> {
...
+  void _generateRandom() {
+    // Chess960 IDs range from 0 to 959.
+    final randomId = math.Random().nextInt(960);
+    _controller.text = randomId.toString();
+  }
+
+  void _loadPosition() {
+    final id = int.tryParse(_controller.text);
+    if (id == null || id < 0 || id > 959) {
+      showSnackBar(context, 'Please enter a number between 0 and 959', type: .error);
```
veloce: We can keep that as a safeguard, but the best way to handle this is with a custom [TextInputFormatter](https://api.flutter.dev/flutter/services/TextInputFormatter-class.html), to provide as-you-type validation of the allowed range.

veloce (commented):
Thanks!

The UX is good, but the suggested refinement would make it even better.

lib/src/view/board_editor/board_editor_screen.dart
HaonRekcef: Not sure this will improve UX, it will just not allow the user to type a number higher than 959 if he tries it will seem as if the device did not register the key press

veloce: what I want is not something that prevents you to type a number higher than, it's something that shows the input error: "number cannot exceed 959" (input in red, error in red below if we want to be consistent with other forms in the app). 
Basic client side form validation, either after submitting or before (as-you-type before is better though) I don't really care but I think it can be achieved in flutter.

HaonRekcef:
I changed the feedback from after submitting to as you type feedback:

[Screencast_20260819_081950.webm](https://github.com/user-attachments/assets/3aaf544a-5007-4c2d-8683-4297eeaeacea)

veloce (approved):
Thanks!

PR #3582, opened 2026-08-18
Broadcasts: Filter boards by team

allanjoseph98:
https://github.com/user-attachments/assets/a679a234-ac20-4c54-ad59-b9a1ad83ad63

* Add a team filter that filters boards to only the selected `team`. When filtering by team, boards and eval bars are oriented with **team on the bottom** instead of white on the bottom. Clicking on any game will load the game with that POV.
* This is an additive filter. It can be combined with the existing `ongoing` filter.
* Add a warning message when there are no boards to show due to filters.
* Load the `team` translation for use in broadcasts.

Misc:
* Add missing FIDE federations and sort federations alphabetically

AI use disclosure:
Copilot chat was used in conjunction with Claude Haiku and GPT 5.6 to
Add the tests (reviewed and modified by me)
Modify the team picker to re-use the existing bottom sheet modal instead of drawing over it.


lib/src/view/broadcast/broadcast_round_screen.dart
```diff
@@ -566,66 +584,142 @@ class _BroadcastSettingsBottomSheet extends ConsumerStatefulWidget {
...
+                                  ? widget.allGamesCount
+                                  : widget.ongoingGamesCount;
+                              return Text('$label ($count)');
+                            },
+                            onSelected: (value, selected) {
+                              setState(() => filter = value);
+                              widget.onGameFilterChange.call(value, selectedTeam);
+                            },
+                          ),
+                          if (widget.teams != null)
```
HaonRekcef, line 757: This currently also shows the Chip for non Team tournaments as `broadcastAllTeams` is part of the list

test/view/broadcast/broadcast_round_screen_test.dart
```diff
@@ -194,6 +196,111 @@ void main() {
...
+      // Load the tournament
+      await tester.pump();
+
+      // Load the round
+      await tester.pump();
+
+      expect(find.byType(BoardThumbnail), findsNWidgets(2));
+
+      await tester.tap(find.byType(ContextMenuIconButton));
+      await tester.pumpAndSettle();
```
HaonRekcef: In the rest of the codebase tester.pump() is generally preferred over tester.pumpAndSettle() if there are no animations, though IMO it doesn't make a difference really

lib/src/view/broadcast/broadcast_round_screen.dart
```diff
@@ -231,16 +234,26 @@ class _BroadcastRoundScreenState extends ConsumerState<BroadcastRoundScreen>
                     final games = asyncRound.value.games.values;
                     final allCount = games.length;
                     final ongoingCount = games.where((g) => g.isOngoing).length;
+                    final teams = games
+                        .expand((g) => g.players.values.map((p) => p.player.team))
+                        .toISet()
+                        .nonNulls
```
HaonRekcef, line 268: Reordering should be better here, first filtering out the Nulls and then `.toISet`, not that it makes a big difference though

lib/src/view/broadcast/broadcast_boards_tab.dart
```diff
@@ -54,26 +58,48 @@ class BroadcastBoardsTab extends ConsumerWidget {
...
+    String? teamFilter,
+  ) {
+    final ongoingFiltered = showOnlyOngoingGames
+        ? games.values.where((game) => game.isOngoing).toIList()
+        : games.values.toIList();
+    return teamFilter == null
+        ? ongoingFiltered
+        : ongoingFiltered
+              .where(
+                (game) => Side.values.map((s) => game.players[s]?.player.team).contains(teamFilter),
```
HaonRekcef:
```dart
(game) => game.players.values.any((p) => p.player.team == teamFilter)
```
would be a bit more readable

HaonRekcef (commented):
Thanks, I reviewed the code and found only one thing that really needs fixing, rest are small nitpicks

HaonRekcef:
I like the idea a lot, currently it is very hidden which is a bit sad but not sure how best to improve that.
One thing to add could be to navigiate to `BroadcastBoardsTab`  with the filter from the team results match history instead of navigating to the teams tab.

ijm8710:
Super minor but I slightly prefer the wording "All teams" as on web more than "Team" for when it is unfiltered 

Great addition otherwise

veloce:
> I like the idea a lot, currently it is very hidden which is a bit sad but not sure how best to improve that.
> One thing to add could be to navigiate to `BroadcastBoardsTab` with the filter from the team results match history instead of navigating to the teams tab.

Not sure to understand exactly but also not sure about that.

To make the filters more discoverable I think the easiest option is to move them to a menu entry instead of putting them in the settings. So a new menu entry "Filter games" would display a bottom sheet containing the filters. And the settings bottom sheet only has the "Show evaluation".

HaonRekcef:
image

I meant that clicking a match from the match history here would lead to the boards tab with the team that is currently viewed applied as a filter.

veloce:
> I meant that clicking a match from the match history here would lead to the boards tab with the team that is currently viewed applied as a filter.

Instead of going to the team. Yes that is a good idea, it's very logical to show this instead of the team page.

We should do this and also the change I suggested above.

veloce (approved):
Code is LGMT.

Tested live with a team broadcast and a non-team broadcast. Everything look and work fine!

PR #3583, opened 2026-08-18
Add separate app icon for light and dark mode for iOS.

cappylot:
Previously, the mobile app only had a dark app icon, meaning it would stay dark even if the user changes their app icons to light mode. This change introduces a second app icon with inverted colors for light mode.


veloce:
This is for the iOS app icons in Home Screen right? Not for the store?

I'd like to see screenshots please, for light mode, dark mode and the transparency mode. Thanks.

PR #3605, opened 2026-08-24
Play move sound instantly in bullet

HaonRekcef:
Before the sounds in bullet where not played instantly if played with tap to move or for opponents move.

### DIFF
--- a/lib/src/model/game/game_controller.dart
+++ b/lib/src/model/game/game_controller.dart
@@ -610,14 +610,10 @@ class GameController extends AsyncNotifier<GameState> with ChatMixin<GameState> 
 
   /// Move feedback while playing
   void _playMoveFeedback(SanMove sanMove, {bool skipAnimationDelay = false}) {
-    final speed = state.value?.game.meta.speed;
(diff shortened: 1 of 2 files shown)


lib/src/model/game/game_controller.dart
```diff
@@ -610,7 +610,12 @@ class GameController extends AsyncNotifier<GameState> with ChatMixin<GameState>
 
   /// Move feedback while playing
   void _playMoveFeedback(SanMove sanMove, {bool skipAnimationDelay = false}) {
-    final animationDuration = ref.read(boardPreferencesProvider).pieceAnimationDuration;
+    final speed = state.value?.game.meta.speed;
+    final isBullet = speed == Speed.bullet || speed == Speed.ultraBullet;
+
+    final animationDuration = isBullet
```
veloce: There was a discrepancy because of the piece animation duration is overridden for bullet. To avoid such issue in the future we could also add a new getter in `GameState` so this logic is in a single place.

veloce (approved):

PR #3612, opened 2026-08-26
Fix blank checkbox in home widgets settings

philyawj:
The "Home widgets" settings screen can show a checkbox with nothing next to it — here between Quick pairing and Blog:

Home widgets settings screen showing a checkbox with no content next to it

**Cause:** in edit mode, `_EditableWidget` always renders a checkbox beside the widget's own content — the content is the row's only label. When that content renders as `SizedBox.shrink()`, the checkbox is left floating. In the screenshot that's the featured tournaments widget: `FeaturedTournamentsWidget` shrinks when no featured tournament is supported by the app, but its `shouldShow` is only `hasServerContent`. The friends carousel has the same problem when the user follows no one.

**Fix:** gate `shouldShow` for those two widgets on actually having content. Loading and error states are unchanged (shimmer / error text still show). Added widget tests covering both states of the tournaments case.

I couldn't find an existing issue or open PR about this. Feel free to push changes to this branch if you'd like it done differently.

AI disclosure: this fix was made with Claude Code (Claude Fable 5), targeting this isolated problem; I reviewed the change.

### DIFF
--- a/lib/src/view/home/home_tab_screen.dart
+++ b/lib/src/view/home/home_tab_screen.dart
@@ -183,6 +183,17 @@ class _HomeScreenState extends ConsumerState<HomeTabScreen> {
                 ? ref.watch(followingCarouselProvider)
                 : const AsyncValue.data(IListConst<FollowingUser>([]));
 
+            // Widgets whose content can be empty should not show a checkbox in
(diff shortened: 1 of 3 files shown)


veloce (approved):
Looks good! thanks.

PR #3616, opened 2026-08-27
study: add player names and clocks from PGN dataa

tom-anders:
Like analysis of imported PGNs already does, this adds player names and their clocks to the header/footer of the board in studies.

I also did some refactoring to share most of the code between AnalysisScreen and StudyScreen for this.

Screenshot_1787861815


veloce (commented):
lgtm

PR #3620, opened 2026-08-30
OTB: add "players side-by-side" board arrangements (#3076)

### LINKED ISSUE #3076: Add "same side" mode for OTB/local play

shirshov:
Closes #3076. Replaces the flip/symmetric piece toggles with a `BoardArrangement` enum that supports four layouts:

- Face-to-face with opponent's pieces upside down
- Face-to-face with pieces flipping to face the current player
- Side-by-side with white always at the bottom
- Side-by-side with the board rotating to face the current player

The symmetrical pieces option removed.

## Menu
new_otb_setup

### 1. Opponent's pieces are upside down
https://github.com/user-attachments/assets/180f7043-a5b8-4448-be1f-ad1d9ca0ddf8

### 2. Pieces flip to face current player
https://github.com/user-attachments/assets/70a62a73-a768-44d7-84cf-5b785a551f3c

### 3. White stays at the bottom
https://github.com/user-attachments/assets/6440b02d-0d12-4a20-b89b-fc1dcc6a53de

### 4. Board rotates to face current player
https://github.com/user-attachments/assets/552edda9-4b97-44a8-b7fd-dcf2215bb4b0

---
95% of code was written manually, I've used DeepSeek v4 Pro 0813 model via Kilo extension for VS, and Kilo gateway for code review and plain text generation (doc comments and commit message)


ijm8710:
I appreciated the pr @OddSteak put in but I also agree this menu works much better

May be helpful to include a quick video demo of all the modes just to showcase the modes work as advertised

One thing I'm having trouble wrapping my head around is what is use case for same side white on bottom. For the same side option, I can only envision the board flip being the one that makes sense but perhaps I'm missing something

shirshov:
I've updated the PR description with videos.

I'd say the "white at the bottom" is for those who want to minimise board transitions. I'd probably use that one even when playing black.

Maybe call that option "Board doesn't change"? Seriously though, I'm not 100% happy with the copy on this screen, appreciate suggestions.

ijm8710:
Hmmm admittedly I don't get it, haha...def feels odd since you never truly play from the far end of the board otherwise

But it's your pr and I def like the menu otherwise!

Perhaps swap the two options around? I feel the 2nd option is more practical than the first? Maybe for the "face to face" options as well?

If you prefer to
Keep as is and you disagree though, all good

Thanks for the videos :)

shirshov:
I actually don't have a strong opinion. I think I prefer "upside down" or "whites at the bottom" arrangements, whether playing white or black, hence they're at the top in their categories. But I never used this feature, so I depend on other people to tell me what to do.

shirshov:
I've changed this PR to draft, as there's a conflict with a new Blindfold mode: not all choices in board arrangement make sense when Blindfold is selected. Does anyone have a good understanding of how all the choices on this screen should work together?

conflict_with_blindfold

PR #3627, opened 2026-08-31
dont doublecheck the the legality of the move parseSan already did

HaonRekcef:
No need to check again if the move is legal.
`ParseSan` already checked and only returns a move if it is legal.


veloce (approved):
good catch!

PR #3617, opened 2026-08-27
Copy line pgn and add to study

Zinkelburger:
Please read what follows, then replace it with your pull request message.

# Rules about AI-assisted code contributions

AI-generated code can be accepted under the following conditions:

- Carefully review and understand all the code you submit, and be able to explain it if asked.
- Provide proof of manual testing of the changes, with screenshots or ideally a video.
- Include the AI tools you used in the pull-request message.

For more information, see the [AI-assisted code contribution policy](https://github.com/lichess-org/mobile/blob/main/AI_POLICY.md).

# General guidelines for pull requests

- Make sure to check the [existing pull requests](https://github.com/lichess-org/mobile/pulls) to avoid duplicated work.
- Explain why the change is needed and what problem it solves.
- If your change touches the UI in any way, please provide a screen recording
- Link to any relevant issues or discussions.
- Prefer small, focused pull requests that only change one thing at a time.
- Mark the pull request as a draft if you have not run the code.
- Only mark the pull request as ready when you have confirmed that it works as intended - be on the lookout for edge cases.
- If you're unsure about something, or want to ask if the change is desired before doing the work, ask us in the programming channels on https://discord.gg/lichess

PR #3630, opened 2026-09-01
analysis: copy the main line or variation PGN from the move menu

Zinkelburger:
This feature was in my closed PR https://github.com/lichess-org/mobile/pull/3617, where the study functionality had overlap with another PR. But the "Copy Mainline" part is still unique.

---
Why this PR:
There was no way to get the moves out of the app's analysis feature. The website has the ability to copy a mainline/variation PGN, but the app does not. I really wanted this feature, behavior should be the same as on the website

---

Long-pressing a move now offers "Copy main line PGN" on a main line move and "Copy variation PGN" on a variation, matching the website's wording and its semantics:

  - the moves leading to the pressed move are exported as a single line, ignoring the sidelines that branch off them;
  - the line then continues with the first child of each node until it ends;
  - sidelines branching off after that point are kept only when copying a variation.

A Variant or FEN header is added when the line does not start from the standard initial position, so the result can always be replayed.

The tree walk lives in Node.makeLinePgn and is unit tested; the three screens with a move tree (analysis, broadcast and study) reach it through a new PgnTreeNotifier.makeLinePgn.

[Screen_recording_20260901_113707_small.webm](https://github.com/user-attachments/assets/58476652-9ea9-4d26-ba63-a109c2730eed)

I used Claude Opus & Fable 5

### DIFF
--- a/lib/src/model/common/node.dart
+++ b/lib/src/model/common/node.dart
@@ -353,6 +353,54 @@ abstract class Node {
 
     return pgnGame.makePgn();
   }
+
+  /// Exports the line going through [path] as a PGN string.
+  ///
+  /// Mirrors the website's "Copy main line PGN" and "Copy variation PGN": the moves up to [path]
+  /// are exported as a single line, dropping the sidelines branching off them, then the line
+  /// follows the first child of each node to the end. Sidelines after [path] are kept only when
+  /// [includeVariations] is true.
+  ///
+  /// Comments and annotations are dropped. A `Variant` or `FEN` tag is added when the line does not
+  /// start from the standard initial position. Returns an empty string when the line has no move.
+  String makeLinePgn(UciPath path, {required Variant variant, required bool includeVariations}) {
+    final pgnRoot = PgnNode<PgnNodeData>();
+
+    PgnNode<PgnNodeData> pgnNode = pgnRoot;
+    Node node = this;
+    for (final branch in branchesOn(path)) {
+      final pgnChild = PgnChildNode(PgnNodeData(san: branch.sanMove.san));
+      pgnNode.children.add(pgnChild);
+      pgnNode = pgnChild;
+      node = branch;
+    }
+
+    final List<({Node from, PgnNode<PgnNodeData> to})> stack = [(from: node, to: pgnNode)];
+    while (stack.isNotEmpty) {
+      final frame = stack.removeLast();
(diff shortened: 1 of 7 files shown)


lib/src/model/common/node.dart
```diff
@@ -353,6 +353,54 @@ abstract class Node {
...
+    final pgn = PgnGame(
+      headers: {
+        if (variant != Variant.standard) 'Variant': variant.pgnName,
+        if (position.fen != kInitialFEN) 'FEN': position.fen,
+      },
+      moves: pgnRoot,
+      comments: [],
+    ).makePgn();
+
+    // Strip the trailing result token and newline: this goes to the clipboard, not to a file.
```
tom-anders: The `Node` class shouldn't know what's done with the generated PGN string, so this shouldn't be done here. Can you move this to the `onPressed` handler in the `pgn.dart`?

lib/src/model/common/node.dart
```diff
@@ -353,6 +353,54 @@ abstract class Node {
...
+      final frame = stack.removeLast();
+      final children = includeVariations ? frame.from.children : frame.from.children.take(1);
+      for (final childFrom in children) {
+        final childTo = PgnChildNode(PgnNodeData(san: childFrom.sanMove.san));
+        frame.to.children.add(childTo);
+        stack.add((from: childFrom, to: childTo));
+      }
+    }
+
+    // Nothing to paste: don't return bare headers.
```
tom-anders: This shouldn't refer to a `paste` operation, see my other comment

test/model/common/node_test.dart
```diff
@@ -536,6 +536,62 @@ void main() {
         expect(root.mainline.last.sanMove.move, move);
       });
     });
+
+    group('makeLinePgn', () {
```
tom-anders, line 540: Can you also add a widget test to analysis_screen_test.dart so that we have test coverage for the UI as well?

tom-anders (changes requested):
Thanks! veloce is currently busy with engine features/bugfixes, so I left a couple of review comments for you until he has time to look at this himself

test/model/common/node_test.dart
Zinkelburger: Added "Copy Line PGN" group in `analysis_screen_test.dart`

veloce (approved):
Thanks! Looks good.

PR #3636, opened 2026-09-04
perf(db): use COUNT(*) in GameStorage.count

slick-daddy:
Replace full table query + Dart list.length with SELECT COUNT(*). Avoids loading all row data (including large JSON blobs) into memory just to count rows. O(n) allocations -> O(1).


veloce (approved):
Thanks!

PR #3638, opened 2026-09-04
perf(http): cache User-Agent static parts in LichessClient

slick-daddy:
Before, every time the app made a request it rebuilt the whole User-Agent from scratch, it fetched the app version, device info and other details again and again, even though those never change while the app is running. It also re-checked the login in a way that made it update more often than needed.
Now the app saves the parts that never change the first time it starts up,  the version, the device description, and the ID it uses for the session. After that, each request only needs to add the current user’s ID. The code that shows the User-Agent on screen was also changed so it only updates when the actual user changes, not every time the login token refreshes.
The end result is the same text being sent to the server.

### DIFF
--- a/lib/src/network/http.dart
+++ b/lib/src/network/http.dart
@@ -197,13 +197,13 @@ Duration _defaultDelay(int retryCount) =>
     const Duration(milliseconds: 900) * math.pow(1.5, retryCount);
 
 final userAgentProvider = Provider<String>((Ref ref) {
-  final authUser = ref.watch(authControllerProvider);
+  final user = ref.watch(authControllerProvider.select((value) => value?.user));
 
   return makeUserAgent(
     ref.read(preloadedDataProvider).requireValue.packageInfo,
     ref.read(preloadedDataProvider).requireValue.deviceInfo,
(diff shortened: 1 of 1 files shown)


lib/src/network/http.dart
```diff
@@ -428,12 +443,9 @@ class LichessClient implements Client {
...
     }
-    request.headers['User-Agent'] = makeUserAgent(
-      _ref.read(preloadedDataProvider).requireValue.packageInfo,
-      _ref.read(preloadedDataProvider).requireValue.deviceInfo,
-      _ref.read(preloadedDataProvider).requireValue.sri,
-      authUser?.user,
-    );
+    final userId = authUser?.user.id;
+    request.headers['User-Agent'] =
+        'Lichess Mobile/$_cachedVersion as:${userId ?? 'anon'} sri:$_cachedSri$_cachedOsPart';
```
veloce: I think we should still use `makeUserAgent` here. The fact it is not deleted in this PR means it is used elsewhere and we should use this central method of creating the UA.

veloce (approved):
lgtm

PR #3639, opened 2026-09-04
perf: puzzle screen select

slick-daddy:
The puzzle screen was rebuilding its entire layout every time any part of the puzzle state changed, because the main body widget was watching the whole state object with nineteen fields. I changed it so the parent only watches the specific pieces it needs for the board, move list and puzzle ID, and left the slices for the hint arrow and view mode already isolated, then moved the feedback banner that needs the full state into its own small consumers so it rebuilds on its own. Now the board, the move tree and the bottom bar update only when their own data changes, and the rest of the screen no longer rebuilds for unrelated updates like the rating calculation.

### DIFF
--- a/lib/src/view/puzzle/puzzle_screen.dart
+++ b/lib/src/view/puzzle/puzzle_screen.dart
@@ -437,14 +437,10 @@ class _BodyState extends ConsumerState<_Body> {
   Widget build(BuildContext context) {
     final boardPreferences = ref.watch(boardPreferencesProvider);
     final ctrlProvider = puzzleControllerProvider(widget.initialPuzzleContext);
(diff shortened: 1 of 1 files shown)


veloce:
Have you noticed performance issues in this screen?

While optimisation is needed in perf critical screens (like online game), puzzle is not one of these. Listening to the whole state has the advantage of being more robust. Riverpod doc warn about this: https://riverpod.dev/docs/how_to/select

slick-daddy:
> Using `select` slightly slows down individual read operations and increases the complexity of your code by a tiny bit. It may not be worth using it if those "other properties" rarely change.

This change is for puzzles as you said, so let's keep the robust implementation. It doesn't worth it apparently.

PR #3640, opened 2026-09-04
perf(board): cache piece assets and isolate BoardThumbnail repaints

slick-daddy:
The board thumbnails were repainting more than needed whenever the surrounding list scrolled, so I wrapped the static chessboard in a repaint boundary to keep its painting isolated. The piece image loader was also clearing and reloading the same set every time it was called, which could cause a brief flash, so I added a small cache that remembers the last piece set and skips the work when the same one is requested again, only updating the cache after a successful load so a failed attempt can still be retried.

### DIFF
--- a/lib/src/utils/chessboard.dart
+++ b/lib/src/utils/chessboard.dart
@@ -13,10 +13,16 @@ Set<Square>? atomicExplosionSquares(Position positionBefore, Move? move) {
   return squareSet.isEmpty ? null : squareSet.squares.toSet();
 }
 
+PieceSet? _lastCachedPieceSet;
+
 /// Preload piece images from the specified [PieceSet] into Chessground's image cache.
 ///
-/// This method clears the cache before loading the images.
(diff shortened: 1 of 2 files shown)


veloce:
Thanks! Can you also post here the devtools diagnostics (the "is useful" when you inspect the widget and performance overlay) to verify this is an actual gain in perf?

slick-daddy:
I tried to benchmark this on my phone but couldn't do it. Best way to see board thumbnails is seeing a player's game history. That page has many thumbnails but I couldn't login to my account. There was only a login page for devs and it is not there to function as you know. I can't think of anything else to benchmark this.

If you think numbers are pretty important, you can close this. The code is logically correct but it is up to you.

veloce:
You can also use a local Lila server (see Lila repo wiki) and the DB seed to populate this screen. Otherwise I can benchmark this myself but there will be delay.

PR #3647, opened 2026-09-06
feat(puzzle): offline vault with MB/count/all goal

connorhart99:
Draft. Adds DB v6 row-per-puzzle vault plus MB/count/all prefs that map to one row goal. Keeps small blob path. File import and sync of vault solves still to do. No flutter analyze run here (no SDK in env); needs CI + review. Base: commit f098b74.


veloce:
I know this is still a draft, but I'd like to have a clear explanation of what this PR does, and what problem it tries to solve.

It looks like you are changing the puzzle DB format to solve perf issues. Do you really see perf issues for a start? Can you share a screen recording and/or a benchmark? We may still reject it if the added complexity is deemed not worth it.

PR #3652, opened 2026-09-09
Prevent accessing opening explorer in more tab for anon users

vidya-gamer:
This is a very simple change to prevent the app from building the **Opening explorer** page when the user is not logged in. When an anon user goes to the **Opening explorer** page their only real option is to back out of the page. Let's just give them a quick snackbar instead. When an anon user clicks **Opening explorer** they get a  snackbar  that uses the current existing message.

Screenshot_1

No AI used.


veloce (approved):
Good catch, thanks!

PR #3663, opened 2026-09-12
HttpNetworkImageWidget watching a client that never changes

slick-daddy:
I noticed `HttpNetworkImageWidget` was watching the HTTP client instead of just reading it. The client never really changes, so every avatar and image in the app sat subscribed to a provider for no reason. If that provider ever fired, all of them would rebuild at once and throw away their caches.
This switches it to `ref.read`. It also forwards the height param, which the widget accepted but silently dropped until now.


veloce (approved):
In that case it is fine to change, but there is no problem using `ref.watch` instead of `ref.read` even if the provider never changes (riverpod even recommends using watch all the time).

PR #3665, opened 2026-09-12
Stop building the whole Watch tab on first paint

slick-daddy:
Opening the Watch tab built every section up front: the full broadcast carousel, the TV list, and the streamer list. Each carousel card fires image decode plus isolate color work, so the tab did all of that before showing the first frame.

First, the tab body moves from ListView to CustomScrollView with one SliverToBoxAdapter per section, so each section lays out lazily. The pixels are identical and nothing else changes. Second, the carousel builds only the first page plus two spare cards synchronously and appends the rest after the first frame. The scroll offset is still 0 at that point, so appending cannot jump the scroll position.

### DIFF
--- a/lib/src/view/broadcast/broadcast_carousel.dart
+++ b/lib/src/view/broadcast/broadcast_carousel.dart
@@ -70,10 +70,20 @@ class BroadcastCarousel extends StatefulWidget {
 class _BroadcastCarouselState extends State<BroadcastCarousel> {
   final _controller = CarouselController();
 
+  /// Offscreen cards each fire image decode + isolate color work on insert,
+  /// so only the first page is built synchronously. The rest is appended
+  /// after the first frame, while the scroll offset is still 0.
+  bool _showAll = false;
+
   @override
   void initState() {
     super.initState();
     watchTabInteraction.addListener(_onTabInteraction);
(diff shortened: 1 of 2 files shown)


veloce:
Thanks. For this kind of change I'd like to have a screen recording (both on iOS and Android if possible).

slick-daddy:
> Thanks. For this kind of change I'd like to have a screen recording (both on iOS and Android if possible).

I understand and it makes sense. But I really can't do it. I don't own any iOS device and I don't know how to set up emulators for Android. 

I can wait for your testing and keep the PR open. It is not a problem for me at all. Or you can decide to close it, I don't mind this either.

veloce:
> I don't know how to set up emulators for Android.

That you can't build on iOS I can understand, but if you want to contribute to this project, other than very trivial changes, you should setup a real dev env with the android emulator. Read the documentation of this project to know how to do that.

slick-daddy:
Before (loads all carousel at once):
[before.webm](https://github.com/user-attachments/assets/6016e762-10a8-45b6-a986-d6ba2795ef89)

After (loads lazily):

https://github.com/user-attachments/assets/2b4f46f3-4f7e-40b0-9ae6-d2f0d65183f0


I tried my best for the recording and I think the change is subtle but visible. You can tell me if the videos are not good enough and I will record again.

PR #3667, opened 2026-09-12
Index game history paging on (userId, lastModified)

slick-daddy:
Every game-history page load ran `WHERE userId = ? ORDER BY lastModified DESC LIMIT 20` against the `game `table, whose only index is the `gameId` primary key. I confirmed with `EXPLAIN QUERY PLAN` that each page does a full table scan plus a temp b-tree sort, on a table holding 90 days of games.

This adds `idx_game_user_lastModified ON game(userId, lastModified DESC)` as a v6 migration, wired into both `onCreate` and `onUpgrade`. The same plan check now shows an index search with no sort step.

Two commits: the migration, plus a test asserting the index exists and covers the paging query.


veloce:
Note: this is only for the offline game history.

PR #3668, opened 2026-09-12
Upgrade lint and SDK floor

slick-daddy:
The repo declared `lint: ^2.8.0` and `sdk: ^3.12.2` but already resolved lint
2.14.0, whose 15 new strict rules fired 3118 infos repo-wide. The tree was also
formatted with a pre-3.13 formatter, which CI's format gate rejects. Four commits:

1. Lint floor `^2.8.0` → `^2.14.0`. No rule change, the lockfile already pinned it.
2. SDK floor `^3.12.2` → `^3.13.0`, which 2.14.0 requires. The pinned Flutter 3.47.3
   ships Dart 3.13.3, so nothing else moves.
3. Format sync with the 3.13 formatter (77 files, mostly import-group blank lines).
   Generated files excluded.
4. `dart fix --apply` migration to primary constructors and the new strict lints
   (464 files), plus 5 hand repairs where the autofix was wrong: it deleted 4 params
   that downstream code uses, and added `const` where asserts aren't
   constant-evaluable. Those spots carry short comments explaining why.
   
5 hand repaired files are:
   

1. lib/src/tab_scaffold.dart — restored deleted inactiveColor/iconSize params
2. lib/src/widgets/platform_context_menu_button.dart — restored deleted isDefaultAction/isDestructiveAction params
3. lib/src/widgets/non_linear_slider.dart — dropped invalid const, documented ignore
4. lib/src/widgets/pgn.dart — dropped invalid const, documented ignore
5. lib/src/widgets/platform.dart — preferredSize field converted to getter


veloce (approved):
Thanks.

PR #3661, opened 2026-09-12
fix missing l10n

zhugeprajna:

PR #3678, opened 2026-09-15
Bulk string update

brollin:
This PR swaps a number of CrowdIn strings with new strings with updated copy. These new strings were introduced via a series of PRs in lila: https://github.com/lichess-org/lila/pull/21008 https://github.com/lichess-org/lila/pull/21009 https://github.com/lichess-org/lila/pull/21011 and we want to swap over here first, then we will swap over in lila and finally delete the old strings.

Importantly, I don't have a Dart development environment set up, so I'd appreciate if someone could double check my work.

> AI-generated code can be accepted under the following conditions:
> 
> - [x] Carefully review and understand all the code you submit, and be able to explain it if asked.
> - [ ]  Provide proof of manual testing of the changes, with screenshots or ideally a video. -> Currently untested!
> - [x] Include the AI tools you used in the pull-request message. -> Cursor, Grok 4.6.


brollin:
cc @superuser-does

lib/src/model/puzzle/puzzle_theme.dart
```diff
@@ -201,7 +201,7 @@ enum PuzzleThemeKey(final IconData icon) {
       case PuzzleThemeKey.enPassant:
         return PuzzleThemeL10n(
           name: 'En passant',
```
veloce: looks like this one is missing, would be good to translate it at the same time

veloce (commented):
Code is good, I just found a missing translation while reviewing.

lib/src/model/puzzle/puzzle_theme.dart
veloce: Never mind that comment it has been fixed here: https://github.com/lichess-org/mobile/pull/3661

brollin: Thanks for the review!

PR #3680, opened 2026-09-15
Fix stale rating range on New Opponent matchmaking (#3679)

### LINKED ISSUE #3679: Bug: "New Opponent" uses stale rating range after a rating change
### What is the problem?

When finishing a rated game and clicking the "New Opponent" button from the post-game screen, the matchmaking system uses the rating from before the previous game to calculate the custom rating range, rather than the newly updated rating.

The bug can be produced with any rating but for clear explanation have taken 1350

Have a current rating of 1350.

Create a custom game (Standard, Rated, Rapid 10+0) with the rating range set to -0 to +500. The matchmaking range correctly shows as [1350 - 1850].

Play and win the game. Your rating updates (e.g., increases to 1357).

On the post-game screen, click the "New Opponent" button.

Observe the matchmaking screen that appears.

### App version

0.26.4

### Device Info

Pixel 10a / Android 17

### Additional information

Expected behavior:
The new matchmaking search should recalculate the range based on the updated rating of 1357. The new range should be [1357 - 1857].

Actual behavior:
The new matchmaking search uses the stale, pre-game rating of 1350. The range incorrectly remains [1350 - 1850]

deepeshgodara:
Fixes #3679

### Summary of Changes
When a user completed a game with rating adjustments, clicking "New Opponent" used the pre-game rating to compute the custom rating range seek instead of the updated rating.

This PR fixes the issue by:
- **`GameSeek.newOpponentFromGame`**: Computes `ratingRange` using the updated post-game rating (`rating + ratingDiff`) or account perfs when `customRatingDelta` is configured.
- **`GameController`**: Invalidates `accountProvider` upon receiving `endData` for rated games so the latest user rating is refreshed.
- **`GameScreen`**: Passes `account` into `newOpponentFromGame` and pushes a replacement route with the updated seek.

### Testing
- Added unit tests in `test/model/lobby/game_seek_test.dart` verifying `ratingRange` calculation after rating changes.
- Added widget tests in `test/view/game/new_opponent_test.dart` verifying the "New Opponent" seek range and HTTP request payload.
- Verified with `flutter test` (all passed) and `flutter analyze` (0 issues).

### AI Assistance Disclosure
Antigravity IDE (Gemini) was used to assist with code navigation, test authoring, and implementation. All changes have been reviewed, tested, and verified.


veloce:
You left the PR template in the description, please remove it.

deepeshgodara:
Hey @veloce, Have updated the description.
Thanks for the review.

lib/src/model/game/game_controller.dart
```diff
@@ -875,6 +876,10 @@ class GameController(final GameFullId gameFullId)
           ),
         );
 
+        if (curState.game.meta.rated) {
```
BhuvaneshN09, line 879: does invalidating the provider guarantee the subsequent read gets the updated account, or could .value still contain the old account? it seems to verify the game rating diff path but not the account-perf fallback path..

BhuvaneshN09 (changes requested):
can anyone clarify why ratingDelta is still set for lobby games when ratingRange has been calculated? It expects ratingDelta to be null when we have a concrete rating range, but the ratingDelta: assignment below appears to always set it for lobby games.

lib/src/model/game/game_controller.dart
deepeshgodara: Thanks for the review, @BhuvaneshN09!

To clarify both points:

### 1. Regarding `ratingDelta` when `ratingRange` is calculated
In the previous implementation on `main`, `ratingDelta` was indeed unconditionally assigned:
```dart
ratingDelta: game.source == GameSource.lobby ? setup.customRatingDelta : null,
```

In this PR, that was replaced with:
```dart
ratingDelta: ratingDelta,
```
Where `ratingDelta` is initialized to `null`. When `currentRating` is available (i.e., `isProvisional != true && currentRating != null`), `ratingRange` is computed and `ratingDelta` remains `null`. It is only assigned to `setup.customRatingDelta` in the fallback branch `else if (currentRating == null)`. 

All unit tests in `test/model/lobby/game_seek_test.dart` explicitly verify `expect(seek.ratingDelta, isNull)` whenever `ratingRange` is present.

---

### 2. Regarding `ref.invalidate(accountProvider)` vs the `ratingDiff` path
- **Synchronous and Immediate:** The primary path used to compute `ratingRange` in `GameSeek.newOpponentFromGame` is `currentRating = me.rating! + (me.ratingDiff ?? 0)`. The `endData` socket event from the game server provides the exact `ratingDiff` synchronously when the game concludes, which means `newOpponentFromGame` does not need to wait for an asynchronous HTTP profile fetch to know the new rating.
- **Why `ref.invalidate(accountProvider)`:** Invalidating `accountProvider` triggers a background re-fetch so that the rest of the application (e.g. side drawer, user profile screen) reflects the updated rating rather than displaying stale numbers.
- **The `account` fallback:** Passing `account` into `newOpponentFromGame` acts purely as a fallback in the rare event that `me.rating` is absent on the game object. This fallback path is also covered by unit tests in `test/model/lobby/game_seek_test.dart` (`falls back to account perfs when player rating in game is null`).

PR #3692, opened 2026-09-17
support duplicate variations in game tree

HaonRekcef:
PGNs containing duplicate variations with the same move (e.g. 1. e4 (1. e4)) caused identical UciCharPair IDs across sibling branches. This led to duplicate GlobalKey crashes in the Flutter notation view, tapping duplicate sidelines jumping to the first child, VariationsBar and board moves routing to the wrong branch.

I saw a game in the olympiad that had this problem yesterday, and because of it the notation was not rendered.
I had opening files before that had duplicate like this too, that were a bit uncomfortable to handle because of it.

This PR gives all sibling branches unique UciPaths:

- For duplicate sibling moves UciCharPair.disambiguated encodes the duplicate index into the Unicode Private Use Area (0xE000..0xF8FF) 
- UCI Pairs keep the same lenthg

AI usage:

Idea how to do it was my own, used gemini for the implementation and test generation and it told me about the Private Use Area. Tested and verified manually by my own.

Video after:

[Screencast_20260917_155423.webm](https://github.com/user-attachments/assets/0cc6ac6f-0af6-40ee-b07d-7201fc24d3ff)

PGN to test it with:

```
[Event "?"]
[Site "?"]
[Date "????.??.??"]
[Variant "Standard"]
[Round "?"]
[White "Enter New Game"]
[Black "?"]
[Result "*"]
[WhiteElo "?"]
[BlackElo "?"]
[PlyCount "5"]
[GameId "2355227095752779"]

1. e4 ( 1. e4 d6 2. d4 ) ( 1. e4 ) ( 1. e4 c6 ) 1... c6 ( 1... c6 2. d4 d5 3. e5 ( 3. e5 ) ( 3. exd5 ) ) ( 1... d5 ) ( 1... d5 ) 2. Nc3 *

```


lib/src/model/common/node.dart
```diff
@@ -34,11 +34,27 @@ abstract class Node({
...
+      node.id = UciCharPair.disambiguated(baseId, count);
+    }
+    children.add(node);
+  }
 
-  /// Prepends a child to this node.
-  void prependChild(Branch node) => children.insert(0, node);
+  /// Prepends a child to this node, ensuring its [id] is unique among siblings.
+  void prependChild(Branch node) {
+    children.insert(0, node);
```
veloce, line 49: Inserting the node before disambiguating is a little strange, but since this is a synchronous method there is no problem in doing that I think.

lib/src/model/common/node.dart
```diff
@@ -709,6 +734,7 @@ sealed class const ViewBranch._() extends ViewNode with _$ViewBranch {
     required Position position,
     Opening? opening,
     required IList<ViewBranch> children,
+    UciCharPair? customId,
```
veloce, line 737: I find the name `customId` strange here, and not reflecting its purpose (disambiguating). In branch it is just `id`, why not the same here?

PR #3701, opened 2026-09-19
fix custom background selection

HaonRekcef:
currently the custom background selection is not working, I fixed the typing to make it work.

AI generated testcase.

video of it working again: 

[Screencast_20260919_184146.webm](https://github.com/user-attachments/assets/c34eb47e-dbc9-487f-907c-366a02c523b3)


veloce:
I tested this and it worked on iOS simulator. I can't reproduce any issue on a real iOS device too. Android only issue then? For a type issue that is strange.

HaonRekcef:
@veloce that is because on android a `RadioGroup` is used as a choice picker, while on iOS a `CupertinoActionSheet`

PR #3702, opened 2026-09-19
don't throw unnecessary Exceptions in search screen

HaonRekcef:
Return an empty list instead of throwing `Exception('Cancelled')` when the provider is disposed to prevent false positivs in the app logs.


lib/src/model/user/user_repository_providers.dart
```diff
@@ -59,10 +59,14 @@ final autoCompleteUserProvider = FutureProvider.autoDispose.family<IList<LightUs
   ref.onDispose(() => didDispose = true);
   await Future<void>.delayed(_kAutoCompleteDebounceTimer);
   if (didDispose) {
-    throw Exception('Cancelled');
+    return const IListConst([]);
```
veloce, line 62: Another way of doing that is to catch the provider exception, but I guess it's also fine like that.

veloce (approved):

